PR #2, opened 2018-12-21
Update README and build instructions on Mac

ajavadia:
### Summary
The current README is a bit of a boilerplate, so I'm updating it to the general style of the Qiskit project.

In addition, I struggled with building from source on Mac. Eventually figured it out with helpful feedback from @atilag. I'm updating the build instructions so that it may be useful to others.

### DIFF
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -62,9 +62,9 @@ The only required requisite is that the toolchain needs to support C++14.
 
 **Mac**
 
-On Mac we have various options depending on the compiler we wanted to use.
-If we want to use Apple's clang compiler, we need to install an extra library for
-supporting OpenMP: libomp. CMake build system will warn you otherwise.
+On Mac we have various options depending on the compiler we want to use.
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTING.rst
```diff
@@ -134,15 +143,23 @@ This is setting the CMake variable ``STATIC_LINKING`` to value ``True`` so CMake
 library, and is passing ``-j8`` flag to the underlaying build system, which in this case is Makefile, telling it that we want to
 build in parallel, using 8 processes.
 
+*N.B. on MacOS:*, you may need to turn off static linking and specify your platform name, e.g.:
+
+.. code::
+
+   python ./setup.py bdist_wheel --plat-name macosx-10.9-x86_64 -- -DSTATIC_LINKING=False -- -j8
```
atilag: This is actually the default, so there's no need for specifying STATIC_LINKING=False

ajavadia: ok i removed that.

PR #5, opened 2018-12-23
Edits to contributing

jaygambetta:
Following the instructions, there was a dash missing. 

A prompt was not on one of the install lines and since requirements were inline I missed it the first time.  

Personally, I would rethink your headings as I don’t know why the using cmake files needs a higher level than the build. 

In Terra, we rearranged to be the following headings. 

Contributing, Installing Code, Documentation, and Development cycle. 

But it builds and works for me now :-).


atilag:
Yep, liked that rearranged. We'll change it

PR #13, opened 2018-12-31
Stabilizer QasmSimulator method

chriseclectic:
### Summary

* Adds a Clifford stabilizer simulation method to the QasmSimulator.
* Reworks the QasmController to make it easier to switch between different simulation implementations.

Depends on #31 

### Details and comments

The stabilizer simulator is automatically used for any `QasmSimulator` simulations where the circuit is all Clifford gates, and the noise model is only clifford and reset errors. In other cases the usual statevector simulator will be used.

The automatic selection can be overriden by the user using the QasmSimulator backend options:

```
opts = {"method": "statevector}  # for statevector simulation
opts = {"method": "stabilizer"}  # for stabilizer simulation
```

If one tries to force a stabilizer simulation on an invalid circuit or noise model an invalid operation exception will be returned.


chriseclectic:
Yep, these are things we need to discuss which is why I've left it as WIP. Currently its just a port of the old Clifford simulator backend updated to work with the new Aer framework.

The idea to have the additional methods in the base class is because eventually this `Clifford` (or `CliffordTableau`) class will be directly exposed to be used independently of the qiskit backend for things like randomized benchmarking. Initially there will be a Python Clifford / CliffordTableau class for Ignis/Terra which should have the same interface and at some point that should be replaced with a cython wrapper to this C++ class.

It remains to be decided how best to organise the classes / methods (and also the Clifford table convention) which is some discussions I'm going to be having very soon with Sergey/Jay etc.

chriseclectic:
@ajavadia @atilag this is ready for review

PR #15, opened 2019-01-01
Noise transformation module

gadial:
### Summary
This pull request adds a **noise transformation** module to Qiskit-Aer. The module is fully working, and many unit tests were added as well.

### Details and comments

The basic use case we envision is the following: The user wants to run a Clifford simulator on a suitable circuit, using some noise model e.g. amplitude damping. It is not possible to simulate this noise on a Clifford simulator, but we can **transform** this noise into a Clifford noise (e.g. a noise which only uses Clifford operators) which is "close" to the original noise.


chriseclectic:
Hi @gadial I can't see the changes because its showing a lot of commits from master that are already merged. Could you rebase against master to fix that?

gadial:
Reopening as another pull request.

PR #19, opened 2019-01-02
Fix non OpenMP build, add qobj method to AerJob

### LINKED ISSUE #10: In the previous aer simulator we had a job.qobj
### What is the expected behaviour?
In the previous simulator, we had a job.qobj. I know @ajavadia wanted this for the IBMQ. I am not sure we need this for simulators but I'm putting it here as I think we need to either add job.qobj here or remove job.qobj from basic. I don't mind which way we go but i would like the two simulators to have the same methods with aer having more not less.

### LINKED ISSUE #16: Cannot build Aer without OpenMP
### Informations

- **Qiskit AER version**: 0.1
- **Operating system**: OSX

One cannot build the Aer simulator without OpenMP (on vanilla osx, for example) because the variable `'num_threads_circuit'` is only defined inside and omp ifdef section, but used outside.

chriseclectic:
### Summary

Fixes #16 and fixes #10

### Details and comments

* Adds `qobj` method to `AerJob` class that returns the submitted/run qobj.
* Fixes the missing definition for `num_threads_circuit` outside the `#ifdef _OPENMP` block so that single threaded execution works when building the simulators without OpenMP.

### DIFF
--- a/qiskit/providers/aer/aerjob.py
+++ b/qiskit/providers/aer/aerjob.py
@@ -9,7 +9,6 @@
 
 """This module implements the job class used for AerBackend objects."""
 
-import warnings
 from concurrent import futures
 import logging
 import sys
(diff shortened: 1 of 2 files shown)


nonhermitian:
Yes, this builds using Apple clang

PR #22, opened 2019-01-02
Use cmake_args instead of cmd line call to build on OSX

nonhermitian:
### Summary
This is a simple update so that you no longer need to pass `--plat-name macosx-10.9-x86_64` with `python setup.py bdist_wheel` in order to build on OSX.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from skbuild import setup
 from setuptools import find_packages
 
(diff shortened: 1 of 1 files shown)


chriseclectic:
This looks good to me, I'd like to be able to build without the compiler flag on default XCode!

Can you check it @atilag:  I recall you had some edge case that caused problems before, but I think that was because we were targeting 10.7 and `-stdlib=libc++` instead of 10.9 which always uses the C++14 compatible lib.

atilag:
The problem was more related to scikit-build, than CMake. @nonhermitian What's the name of the resulting output file? If I recall correctly, then name was incorrect.

nonhermitian:
@atilag is correct, the output file name is not adjusted properly, and still says `macosx_10_6_x86_64`.  This is because `10_6` is hard coded unless the plat arg is passed.  Besides being a bit silly, the default build folders all say `10_7` which is the version that Python is built against.

PR #34, opened 2019-01-22
add reset_error to noise module

chriseclectic:
### Summary

Adds `reset_error` function to the standard errors noise module.

### Details and comments

This allows a user to build a single-qubit reset error without having to deal with making a QobInstruction input.

### DIFF
--- a/test/terra/test_noise_model.py
+++ b/test/terra/test_noise_model.py
@@ -17,6 +17,7 @@ from qiskit.providers.aer.backends import QasmSimulator
 from qiskit.providers.aer.noise import NoiseModel
 from qiskit.providers.aer.noise.errors.quantum_error import QuantumError
 from qiskit.providers.aer.noise.errors.standard_errors import pauli_error
+from qiskit.providers.aer.noise.errors.standard_errors import reset_error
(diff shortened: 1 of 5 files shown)


chriseclectic:
Failing test is due to a terra issue: https://github.com/Qiskit/qiskit-terra/issues/1709

test/terra/test_noise_model.py
```diff
@@ -517,6 +518,73 @@ def test_amplitude_damping_error(self):
...
+        error = reset_error(1)
+        noise_model = NoiseModel()
+        noise_model.add_all_qubit_quantum_error(error, ['id', 'x'])
+        # Execute
+        target = {'0x0': shots}
+        qobj = compile([circuit], backend, shots=shots,
+                       basis_gates=noise_model.basis_gates)
+        result = backend.run(qobj, noise_model=noise_model).result()
+        self.is_completed(result)
+        self.compare_counts(result, [circuit], [target], delta=0)
```
atilag: Maybe as a TODO, but a test should always expect any kind of `assert` idiom, so instead of having a `self.compare_counts()` I'd rather have, for example, a `self.assertCountsEqual()`

Issue #1, opened 2018-12-21
Error with OpenMP


jaygambetta:
### Informations

- **Qiskit AER version**:
qiskit 0.1
- **Operating system**:
MacOS 

### What is the current behavior?
Running the tutorials gives the following error

OMP: Error #15: Initializing libiomp5.dylib, but found libomp.a already initialized.
OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://www.intel.com/software/products/support/.

### What is the expected behavior?
No error

atilag:
Not really, this happens because we link with libomp, but Numpy which is our only dependency is usually linked against libiomp (Intel implementation). Both implementations are incompatible, so when the runtime linker load both there could be conflicts. Paul told me that this is something Conda distribution has already fixed in the latest version, so doing a: `conda update` will fix this issue.

jaygambetta:
I will try and confirm either way tonight

chriseclectic:
I had this problem and two fixes I found were to either installing numpy/scipy with pip instead of conda, or if you use conda to install the `nomkl` package (`conda install nomkl`).

jaygambetta:
I still can’t get this working with pip.   I have tried going back to numbly 1.14 as well.  I was able to build it following the source instructions and not see this crash.

t-imamichi:
I notice that `qasm_simulator` and `unitary_simulator` are OK; but, `statevector_simulator` occurs the OMP errror. I checked it with the following code.

```
from qiskit import execute, Aer, QuantumCircuit, QuantumRegister

q = QuantumRegister(1)
qc = QuantumCircuit(q)
qc.h(q[0])
#backend = Aer.get_backend('qasm_simulator')  # OK
#backend = Aer.get_backend('unitary_simulator')  # OK
backend = Aer.get_backend('statevector_simulator')  # NG
job_sim = execute(qc, backend)
```

nonhermitian:
[...]

PR #42, opened 2019-02-01
[WIP] Fix for OpenMP library clashing on MacOS

atilag:
### Summary
Fixes Issue #1
 

### Details and comments
On Mac, OpenMP implementation is integrated through an external library: `libomp.dylib` (and some others similar like libiomp5 from Intel). This library has a constraint: It cannot be loaded twice in memory, basically because it will imply a double initialization of it's runtime, and this is potentially dangerous. We do need this library to operate, but other depedencies we have, require them as well, which is a problem because of the way Python loads extensions (and we, are an extension too).
Basically, with this fix, we have provided Aer the ability to dynamcally load libomp in every possible scenario.


atilag:
I'm closing this because I've messed things up while squashing/rebasing some commits and the history is dirty

PR #46, opened 2019-02-07
Issue 1 - Fix OpenMP crash on MacOS

### LINKED ISSUE #1: Error with OpenMP

atilag:
### Summary
Fixes #1 

### Details and comments
On Mac, OpenMP implementation is integrated through an external library: `libomp.dylib` (and some others similar like libiomp5 from Intel). This library has a constraint: It cannot be loaded twice in memory, basically because it will imply a double initialization of it's runtime, and this is potentially dangerous. We do need this library to operate, but other dependencies we have, require them as well, which is a problem because of the way Python loads extensions (and we are a Python extension too).
Basically, with this fix, we have provided Aer the ability to dynamically load libomp in every possible scenario without incurring in the double initialization problem.


chriseclectic:
@atilag Do you know whats going on with the tests? This PR is resulting in several thousand lines of build warnings/errors in Travis and all the tests erroring out with

```
ModuleNotFoundError: No module named 'qiskit.providers.aer.backends.qasm_controller_wrapper'
```

Locally I built with XCode and GCC8 on MacOS and the `aer_provider` tutorial worked both in a pip installed Numpy env, and a conda installed MLK numpy env, but the tests are all failing locally with the same error as Travis (but without all the compiler warnings).

atilag:
Yeah, I know what`s going on. Fix is comming...

atilag:
Everything seems to be working now :) , isn't it @chriseclectic ?

chriseclectic:
Thanks @atilag, you deserve a beer or 6 after that!

PR #50, opened 2019-02-08
Refactor thread management

hhorii:
### Summary

Simplify logic to determine parallelism. 

### Details and comments

- [x] Rename local variables with the same name to parameters (`max_parallel_threads_`, `max_parallel_experiments_`, `max_parallel_shots_`).
- [x] Provide one virtual method (`set_parallelization`) to determine parallelism.
- [x] Write test cases

### DIFF
--- a/src/base/controller.hpp
+++ b/src/base/controller.hpp
@@ -240,7 +240,7 @@ void Controller::set_parallelization(Qobj& qobj) {
   #endif
 
   // Set max_parallel_experiments_
-  max_parallel_experiments_ = (max_parallel_experiments_ < 1)?
+  parallel_experiments_ = (max_parallel_experiments_ < 1)?
       std::min<int>({ (int) qobj.circuits.size(), max_parallel_threads_}):
       std::min<int>({ (int) qobj.circuits.size(), max_parallel_threads_, max_parallel_experiments_});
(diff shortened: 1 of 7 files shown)


src/base/controller.hpp
```diff
@@ -168,18 +167,22 @@ class Controller {
...
+  int max_parallel_threads_;
 
+  // Parameters for parallelization management in configuration
+  int max_parallel_experiments_;
+  int max_parallel_shots_;
+
+  // Parameters for parallelization management for experiments
+  int parallel_experiments_;
+  int parallel_shots_;
+  int parallel_gates_;
```
chriseclectic: I think we should change the variable name `parallel_gates_` as it makes it sound like the gates are done in parallel which isn't the case. We could use something else like `parallel_linalg_`, `parallel_state_update_` instead.

hhorii:
Currently, parallelization is determined once for a qobj. This means that all the circuits in a qobj have similar characteristics (numbers of qubits and shots, noise/ideal). 

If many users packed various types of circuits into a qobj, parallelization for shots and state_update should be determines for each experiment in future releases.

hhorii:
@atilag Thanks. I added tests.

PR #51, opened 2019-02-08
Add Simulation method based on Stabilizer Rank Decompositions

### LINKED ISSUE #47: Clifford+T Simulator Qasm Simulator method

Adding a simulation method for the `QasmSimulator` based on the techinques in [Bravyi & Gosset (2016)](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.116.250501) for simulating Clifford + T circuits with low T gate depth.

**Note** This issue is to track the work already underway by @padraic-padraic

padraic-padraic:
### Summary

- Add a simulator based on decompositions into stabilizer states, as described in [arXiv:1808.00128](https://arxiv.org/abs/1808.00128)
- Includes support for simulating Clifford + T/CCZ/Toff/u1 gate circuits
- Falls back to a stabilizer simulator based on the 'CH' representation for Clifford circuits

### Details and comments

Resolves #47.

This is a work in progress to add a simulator for circuits with a small number of non-Clifford gates, but with support for up to 63 qubits. 

Currently, it exists as a standalone c++ simulator and controller. To integrate it with the QASMController we still need

- [x] A heuristic for switching to the CH backend when `method == 'automatic'`.
- [x] To update the tests for the 0.1.1 release
- [x] A way to check the memory requirements in `qiskit.providers.aer.QasmSimulator.validate` and `qiskit.providers.aer.CHSimulator.validate`.
- [x] Documentation updates
- [ ] CHANGELOG updates


atilag:
Thanks for the contribution @padraic-padraic , this is great!

Remember to add some tests to your TODO list ;)
I'll start reviewing soon so we can start fixing minor things while you are still working in the changes

padraic-padraic:
Thanks for handling the review!

The tests have been added and updated in my most recent commit; one caveat with this simulation method at the is its performance on strongly peaked output distributions, which is why the last build failed on the Grover case. Bumping up the number of shots leads to a reproducable pass on my end, which I'll push once I've added the QasmController integration.

src/base/state.hpp
```diff
@@ -188,7 +188,7 @@ class State {
   inline const state_t &qreg() const {return qreg_;}
   inline const auto &creg() const {return creg_;}
 
-protected:
+// protected:
```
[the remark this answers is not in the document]
padraic-padraic: Yep, exactly; I didn't realise that was commited, changed it back last night (still need to push).

qiskit/providers/aer/aerprovider.py
```diff
@@ -26,7 +27,8 @@ def __init__(self, *args, **kwargs):
         # Populate the list of Aer simulator providers.
         self._backends = [QasmSimulator(provider=self),
                           StatevectorSimulator(provider=self),
-                          UnitarySimulator(provider=self)]
+                          UnitarySimulator(provider=self),
+                          CHSimulator(provider=self)]
```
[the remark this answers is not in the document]
padraic-padraic: Yeah, I think that would work. Most of the CHController is a shadow of the QasmController API anyway.

The one addition I made to the API is the required memory validation. We could add something like this to the QasmController? Or else I port it to Python, but either way we'd need to update the _validate method of QasmSimulator to account for the different footprints of the CH/Statevector methods.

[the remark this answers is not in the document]
chriseclectic: Adding the memory validation to the QasmController sounds good. We've been meaning to do something like this for awhile, since we should really be doing it for the statevector simulator as well (stabilizer too though in that case its memory footprint is comparatively tiny).

Currently the memory is checked in python before sending the qobj so it can fail-fast, but if we handle it correctly in the C++ we could remove that part from the python backend (or just do a very basic check there).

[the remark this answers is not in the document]
padraic-padraic: For the moment I'll leave the simple check in Python, and set it to auto-switch to ch in the event that the number of qubits is too much for the statevector method (as long as it's not too high for the ch method). I guess the thing to do would be to make use of the `required_memory_mb` method of state as part of an initialisation step in QasmController?

padraic-padraic:
One question I have is for calling `compile` with the `QasmSimulator` class; as it stands, `QasmSimulator` doesn't use the `configuration` kwarg to the `AerBackend` constructor, instead just passing a `BackendConfiguration` instance initialsied with `QasmSimulator.DEFAULT_CONFIGURATION`, which has properties like `n_qubits` and `basis_gates` configured for the statevector simulator. Is that going to potentially cause problems when it comes to compiling to a `Qobj` etc? Because `ch` currently doesn't implement `u2` or `u3` gates.

This also has relevance for when we auto-switch to `ch`, so I'll add a quick check to `QasmSimulator._validate` to catch any illegal instructions before auto-selecting `ch`.

chriseclectic:
@padraic-padraic yes we may need to rethink that on the Terra side, but for now the default basis gates includes the most general basis gates for the simulator (ie the statevector method). This means by default if it has gates like `h, x, y` etc they won't be converted to `u2, u3` gates, however if a circuit already has u2 or u3 gates in it then it can't be run on the CH or stabilizer simulator. We add multiple default configuration dicts for each of the simulation methods to make it more explicit what each method supports. Any ideas for this @atilag @ajavadia?

The general gateset shouldn't be a problem for the automatic method as it will check if gates in the circuit are supported on the different methods and choose one. If a user tried to force running on the stabilizer or CH simulator on a circuit with `u3` gates (eg `"method": "ch"`) the simulator (or python validation before simulation) will raise an exception that the circuit has unsupported gates. You can see how this behave with the current stabilizer simulator.

padraic-padraic:
Just a note on this; I've got most of the benchmarks done to implement a heuristic swticher for CH/QASM, and I've added memory validation in the QasmController as part of validate_state (which takes into account experiment parallelisation. To avoid cross-platofrm headaches, I used `psutil` to pass available memory in as part of the Qobj config). I also fixed a bug in the standalone simulator where `QasmController::set_config` was never being called due to the qobj being run through `BaseController::execute`. 

However, I've been running into some odd intermittent crashes when calling the CH simulator from within Python. I'm just working on debugging those, hopefully should have a sense of what's going on by tomorrow!

padraic-padraic:
Hi both,

This is pretty much done from my end! Here's what's in the latest change

1. Memory validation: `Base::Controller::validate_required_memory` will, if the QObj has `available_memory` in its config, check the requirements of the circuit. This is a separate method from `validate_state` as it requires a member variable. If `validate_state` weren't static (which could be the case, as currently it's only called from inside `run_circuit_helper`), then this function could be folded in to that.

2. The AerBackend will use `qiskit._util.local_hardware_info` to pass the `available_memory` to the `QasmController` if it's not specified by the user.

3. The `QasmController` will automatically switch to the CH method only if the circuit is too large to tackle with the `statevector` method. I think this makes most sense. As CH is an approximate simulation method, and also requires some configuration in the case of strongly peaked output distributions, I think in most cases its best that using it is a conscious choice on the part of the user.

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -160,17 +196,36 @@ def _validate(self, qobj, backend_options, noise_model):
...
-                                   'for "{}" (method=statevector) '.format(self.name()) +
-                                   'with {} GB system memory.'.format(system_memory))
+                    err_string = ('Number of qubits ({}) is greater than '
+                                 'maximum ({}) for "{}" (method=statevector) '
+                                 'with {} GB system memory')
+                    err_string = err_string.format(n_qubits, max_qubits,
+                                                   self.name(), system_memory)
+                    if method != "automatic":
+                        raise AerError(err_string + '.')
+                    else:
```
atilag: There's no need for this `else` here... raising will terminate the function.
We can decrease the indentation then

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -160,17 +196,36 @@ def _validate(self, qobj, backend_options, noise_model):
...
+                        if n_qubits > 63:
+                            raise AerError(err_string + ', and has too many ' +
+                                           'qubits to fall back to the ' + 
+                                           'CH simulator.')
+                        elif not ch_supported:
```
atilag: No need for `elif`,  a simple `if` will be enough as the above `raise` will terminate the function

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -160,17 +196,36 @@ def _validate(self, qobj, backend_options, noise_model):
...
+                            raise Aererror(err_string + ', and contains ' +
+                                           'instructions not supported by ' + 
+                                           'the CH simulator.')
+                        else:
```
atilag: And we can remove this `else` as well for the same reasons exposed before

src/base/controller.hpp
```diff
@@ -265,6 +278,18 @@ void Controller::set_parallelization(Qobj& qobj) {
...
 
   parallel_state_update_ = std::max<int>({1, max_parallel_threads_/(parallel_experiments_*parallel_shots_)});
+
+  if(state_available_memory_mb_ > 0)
+  {
+    if (parallel_experiments_ > 1)
+    {
+      state_available_memory_mb_ /= parallel_experiments_;
+    }
+    else if (parallel_shots_ > 1)
```
atilag: What if `parallel_experiments_` are greater than 1 and `parallel_shots_` are greater than 1 as well?
We will never enter in the second conditional: `else if( parallel_shots_ > 1)` so I guess that the intention is to prioritize parallel experiments over parallel shots, isn't it?.
I could think that in this situation, we may want to divide twice: one by parallel shots, and the resulting available memory needs to be divided again by parallel experiments.

src/base/controller.hpp
```diff
@@ -301,6 +328,34 @@ bool Controller::validate_state(const state_t &state,
   throw std::runtime_error(msg.str());
 }
 
+template <class state_t>
+bool Controller::validate_memory_requirements(state_t &state,
```
atilag: One of the reasons to calculate real `available memory` in the simulator, is that we can know at runtime whether we have enough memory to run the experiments or not, because we have all the information at this stage. 
Anyway, we have another PR holding (#61) where there's a mechanism to  query available memory in the system, but seems like we want this PR merged before that one, so let's keep this code as is, and refactor it after PR #61 is merged.

src/base/controller.hpp
```diff
@@ -301,6 +328,34 @@ bool Controller::validate_state(const state_t &state,
...
+                                  const Circuit &circ,
+                                  bool throw_except) const
+{
+  bool sufficient_memory = (state_available_memory_mb_ == 0); //No memory info in qobj
```
atilag: What we can do is simplifying this function:
```c++
...
if (state_available_memory_mb_ == 0){
    return true;
}

auto required_mb = state.required_memory_mb(circ.num_qubits, circ.ops);
if(state_available_memory_mb_ < required_mb){
    if(thow_except){
        std::string name;
        JSON::get_value(name, "name", circ.header);
        throw std::runtime_error("AER::Base::Controller: State " + state.name() +
             " has insufficient memory to run the circuit " + name);
    }
    return false;
}
return true;
...
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+#endif
+
+namespace CHSimulator {
+  
+using complex_t = std::complex<double>;
+using uint_t = uint_fast64_t;
+using int_t = int_fast64_t;
+using chstabilizer_t = StabilizerState;
+
+double t_angle = M_PI/4.;
```
atilag: this is a constant, right?
```c++
const double T_ANGLE = M_PI/4.;
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+double tdg_angle = -1.*M_PI/4.;
```
atilag: Another constant, so:
```c++
const double TDG_ANGLE = -1. * M_PI/4.;
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+
+const U1Sample t_sample(t_angle);
+const U1Sample tdg_sample(tdg_angle);
+
+uint_t zero = 0ULL;
```
atilag: A constant, so:
```c++
const uint_t ZERO = 0LL;
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+
+class Runner
+{
+private:
+  uint_t n_qubits;
+  uint_t chi;
+  uint_t three_qubit_gate_count; //Needed for normalising states
+  std::vector<chstabilizer_t> states;
+  std::vector<complex_t> coefficients;
+  uint_t n_omp_threads;
```
atilag: `num_threads` is a more descriptive name for this variable.

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  uint_t omp_threshold;
+
+  bool accept;
+  complex_t old_ampsum;
+  uint_t x_string;
+  uint_t last_proposal;
+
+  std::unordered_map<double, U1Sample> z_rotations;
+
+  void InitMetropolis(AER::RngEngine &rng);
```
atilag: We always use  `lower_snake_case` for naming functions and variables, and `UpperCamelCase` for classes.
so this should be: 
```c++
void init_metropolis( ... )
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  void MetropolisStep(AER::RngEngine &rng);
```
atilag: Same as above:
```c++
void metropolis_step()
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  {
+    return (n_qubits == 0 || chi == 0);
+  }
+
+  uint_t get_chi() const;
+  uint_t get_n_qubits() const;
+  bool check_omp_threshold();
+
+  //Convert each state to a json object and return it.
+  std::vector<std::string> serialised_decomposition() const;
```
atilag: functions performs actions, so a better name for this functions could be:
```c++
std::vector<std::string> serialize_decomposition() const;
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  //Methods for non-clifford gates, including parameters for the Sampling
+  void apply_t(uint_t qubit, double r, int rank);
+  void apply_tdag(uint_t qubit, double r, int rank);
+  void apply_u1(uint_t qubit, complex_t lambda, double r, int rank);
+  void apply_ccx(uint_t control_1, uint_t control_2, uint_t target, uint_t branch, int rank);
+  void apply_ccz(uint_t control_1, uint_t control_2, uint_t target, uint_t branch, int rank);
+  //Measure a Pauli projector on each term in the decomposition and update their coefficients
+  // omega.
+  void apply_pauli_projector(const std::vector<pauli_t> &generators);
+  inline void apply_pauli_projector(const std::vector<pauli_t> &generators, uint_t rank);
```
atilag: There's no need for specifying `inline` in header files

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  //Routine for Norm Estimation, thin wrapper for the CHSimulator method that uses AER::RngEngine
+  //to set up the estimation routine.
+  double NormEstimation(uint_t n_samples, AER::RngEngine &rng);
```
atilag: `NormEstimation` -> `norm_estimation`

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  double NormEstimation(uint_t n_samples, std::vector<pauli_t> generators, AER::RngEngine &rng);
```
atilag: `NormEstimation` -> `norm_estimation`

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+
+  //Metropolis Estimation for sampling from the output distribution
+  uint_t MetropolisEstimation(uint_t n_steps, AER::RngEngine &rng);
```
atilag: `MetropolisEstimation` -> `metropolis_estimation`

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  std::vector<uint_t> MetropolisEstimation(uint_t n_steps, uint_t n_shots, AER::RngEngine &rng);
```
atilag: `MetropolisEstimation` -> `metropolis_estimation`

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  //Efficient Sampler for the output distribution of a stabilizer state
+  uint_t StabilizerSampler(AER::RngEngine &rng);
```
atilag: `StabilizerSampler` -> `stabilizer_sampler`

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  std::vector<uint_t> StabilizerSampler(uint_t n_shots, AER::RngEngine &rng);
```
atilag: `StabilizerSampler` -> `stabilizer_sampler`

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  //Utilities for the state-vector snapshot.
+  complex_t amplitude(uint_t x_measure);
```
atilag: This can be `const`:
```c++

File: src/simulators/ch/ch_runner.hpp [hunk 24]
```diff
@@ -0,0 +1,661 @@
...
+//=========================================================================
+
+Runner::Runner(uint_t num_qubits)
+{
+  initialize(num_qubits);
+}
+
+void Runner::initialize(uint_t num_qubits)
+{
+  if (states.size() > 0)
```
atilag: I think we can safely remove the conditionals and just call:
```c++
states.clear();
coefficients.clear(); 
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  for(uint_t i=1; i<chi; i++)
+  {
+    states.push_back(base_sate);
+    coefficients.push_back(coeff);
+  }
+}
+
+void Runner::initialize_omp(uint_t n_threads, uint_t threshold_rank)
+{
+  if(n_threads == 0)
```
atilag: There's a ternary operator in C++ that is very handy for this simple situations:
```c++
n_omp_threads = (n_threads == 0 ? 1 : n_threads);
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  {
+    throw std::runtime_error(std::string("CHSimulator::Runner was initialized without") + 
+                             std::string("being properly cleared since the last ") +
+                             std::string("experiment."));
+  }
+  chstabilizer_t base_sate(states[0]);
+  complex_t coeff(coefficients[0]);
+  for(uint_t i=1; i<chi; i++)
+  {
+    states.push_back(base_sate);
```
atilag: hmmm is this right? Do we want to fill states vector with the same `base_state` instance?

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+//-------------------------------------------------------------------------
+//Measurement
+//-------------------------------------------------------------------------
+
+double Runner::NormEstimation(uint_t n_samples, AER::RngEngine &rng)
+{
+  std::vector<uint_t> adiag_1(n_samples, 0ULL);
+  std::vector<uint_t> adiag_2(n_samples, 0ULL);
+  std::vector< std::vector<uint_t> > a(n_samples, std::vector<uint_t>(n_qubits, 0ULL));
+  #pragma omp parallel for if(n_omp_threads > 1)
```
atilag: I think we can use OMP `collapse(3)` clause here:
```c++
 #pragma omp for collapse(3) if(n_omp_threads > 1)
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+
+const U1Sample t_sample(t_angle);
+const U1Sample tdg_sample(tdg_angle);
+
+uint_t zero = 0ULL;
+
+class Runner
+{
+private:
+  uint_t n_qubits;
```
atilag: Please, suffix all class variables with an ending underscore:
```c++
uint_t n_qubits_;
uint_t chi_;
...
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+    shots[i] = x_string;
+  }
+  return shots;
+}
+
+void Runner::InitMetropolis(AER::RngEngine &rng)
+{
+  accept = 0;
+  //Random initial x_string from RngEngine
+  uint_t max = (1ULL<<n_qubits) - 1;
```
atilag: I'm not a big fun of this type we have created: `uint_t` which is really a 64 bits unsigned integer... not your fault though, we have been using it all over the code base.

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+#include "gates.hpp"
+
+namespace AER{
+namespace CH {
+
+using chpauli_t = CHSimulator::pauli_t;
+using chstate_t = CHSimulator::Runner;
+using Gates = CHSimulator::Gates;
+
+uint_t zero = 0ULL;
```
atilag: `const uint_t ZERO = 0ULL;`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+uint_t toff_branch_max = 7ULL;
```
atilag: `const uint_t TOFF_BRANCH_MAX = 7ULL;`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+  const static stringmap_t<Snapshots> snapshotset_;
+
+  //-----------------------------------------------------------------------
+  //Parameters and methods specific to the Stabilizer Rank Decomposition
+  //-----------------------------------------------------------------------
+
+  //Allowed error in the stabilizer rank decomposition.
+  //The required number of states scales as \delta^{-2}
+  //for allowed error \delta
+  double approximation_error = 0.05;
```
atilag: Private class variables should be suffixed with underscore: `approximation_error_ = 0.05`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+
+  uint_t norm_estimation_samples = 100;
```
atilag: Sames as above:  `norm_estimation_samples_`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+  // How long the metropolis algorithm runs before
+  // we consider it to be well mixed and sample form the
+  // output distribution
+  uint_t metropolis_mixing_steps = 7000;
```
atilag: `metropolis_mixing_steps_`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+
+  //Minimum number of states before we try to parallelise
+  uint_t omp_threshold_rank = 100;
```
atilag: `omp_threshold_rank_`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+
+  double snapshot_chop_threshold = 1e-10;
```
atilag: `snapshot_chop_threshold_`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+
+  double probabilities_snapshot_samples = 3000.;
```
atilag: `probabilities_snapshot_samples_`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+
+  // Compute the required stabilizer rank of the circuit
+  uint_t compute_chi(const std::vector<Operations::Op> &ops);
+  // Add the given operation to the extent
+  inline void compute_extent(const Operations::Op &op, double &xi);
```
atilag: No need for `inline` as this is implicit on header files

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+
+  //Minimum number of states before we try to parallelise
+  uint_t omp_threshold_rank = 100;
+
+  double snapshot_chop_threshold = 1e-10;
+
+  double probabilities_snapshot_samples = 3000.;
+
+  // Compute the required stabilizer rank of the circuit
+  uint_t compute_chi(const std::vector<Operations::Op> &ops);
```
atilag: This can be `const`, right?: 
```c++
uint_t compute_chi(const std::vector<Operations::Op> &ops) const;
```

src/simulators/ch/ch_state.hpp
atilag: And this can be `const` as well:
```c++
void compute_extent(const Operations::Op &op, double &xi) const;
```

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+  // Add the given operation to the extent
+  inline void compute_extent(const Operations::Op &op, double &xi);
+
+  //Compute the required chi, and count the number of three qubit gates
+  virtual std::pair<uint_t, uint_t> decomposition_parameters(const std::vector<Operations::Op> &ops);
```
atilag: I'd remove `virtual` keyword here, if we are not planning to subclass this class.

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+  
+  //Check if this is a stabilizer circuit, for the locaiton of the first non-CLifford gate
+  std::pair<bool, size_t> check_stabilizer_opt(const std::vector<Operations::Op> &ops);
+
+  //Check if we can use the sample_measure optimisation
+  bool check_measurement_opt(const std::vector<Operations::Op> &ops);
```
atilag: `const`

src/simulators/ch/ch_state.hpp
atilag: I think this can be `const` as well

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+  // Compute the required stabilizer rank of the circuit
+  uint_t compute_chi(const std::vector<Operations::Op> &ops);
+  // Add the given operation to the extent
+  inline void compute_extent(const Operations::Op &op, double &xi);
+
+  //Compute the required chi, and count the number of three qubit gates
+  virtual std::pair<uint_t, uint_t> decomposition_parameters(const std::vector<Operations::Op> &ops);
+  
+  //Check if this is a stabilizer circuit, for the locaiton of the first non-CLifford gate
+  std::pair<bool, size_t> check_stabilizer_opt(const std::vector<Operations::Op> &ops);
```
atilag: `const`

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+  //Set the truncation threshold for the probabilities snapshot.
+  JSON::get_value(snapshot_chop_threshold, "chop_threshold", config);
+  //Set the number of samples for the probabilities snapshot
+  JSON::get_value(probabilities_snapshot_samples, "probabilities_snapshot_samples", config);
+}
+
+std::pair<uint_t, uint_t> State::decomposition_parameters(const std::vector<Operations::Op> &ops)
+{
+  double xi=1.;
+  uint_t three_qubit_gate_count = 0;
```
atilag: Do we need a 64 bit integer to have this counter? I'd promote it to: `u32_t`.

src/simulators/ch/ch_state.hpp
```diff
@@ -0,0 +1,787 @@
...
+  if (xi >1)
+  {
+    chi = std::ceil(xi/std::pow(approximation_error, 2));
+  }
+  return std::pair<uint_t, uint_t>({chi, three_qubit_gate_count});
+}
+
+std::pair<bool, size_t> State::check_stabilizer_opt(const std::vector<Operations::Op> &ops)
+{
+  for(auto op = ops.cbegin(); op != ops.cend(); op++)
```
atilag: This is a good candidate for the new `for( : )` you used before:
```c++
for(auto& op : ops){
...
}
```

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+}
+
+void Runner::initialize_decomposition(uint_t n_states)
+{
+  chi = n_states;
+  states.reserve(chi);
+  coefficients.reserve(chi);
+  if(states.size() > 1 || coefficients.size() > 1)
+  {
+    throw std::runtime_error(std::string("CHSimulator::Runner was initialized without") +
```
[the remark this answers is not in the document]
padraic-padraic: Ahh okay; I was using + operators to join the component strings and getting compiler errors. That makes sense.

src/simulators/ch/ch_runner.hpp
```diff
@@ -0,0 +1,661 @@
...
+  n_qubits = num_qubits;
+  chi = 1;
+  n_omp_threads = 1;
+  states = std::vector<chstabilizer_t>(1, chstabilizer_t(num_qubits));
+  coefficients.push_back(complex_t(1.,0.));
+}
+
+void Runner::initialize_decomposition(uint_t n_states)
+{
+  chi = n_states;
```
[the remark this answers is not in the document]
padraic-padraic: `chi` (which I'll rename `chi_`) is the member variable. I used `n_states` as the argument to avoid name duplication, which will be fixed by the underscore convention.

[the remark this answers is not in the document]
padraic-padraic: I'll probably switch from `chi_` (which is lifted from the paper) to `n_states_` for clarity.

src/simulators/ch/ch_runner.hpp
padraic-padraic: We do yeah. This simulator only branches at non-Clifford gates. If the circuit starts with some `m` Clifford gates, then we basically run a clifford simulator for those n gates, and then setup a decomposition by taking `chi` copies of this initial state and doing the usual sampling method. This saves us having to repeateadly apply these first `m` gates to all the states in the decomposition.

src/base/controller.hpp
padraic-padraic: This was based on line 229 earlier in the file, as part of `Controller::set_config`
```C++
if (max_parallel_eperiments_ > 1)
{
  max_parallel_shots_ = 1;
}
```
So at any time only one of parallel experients/shots will be greater than 1.

src/simulators/ch/ch_state.hpp
padraic-padraic: I left this as is because make use of pointer arithmetic later in the function (returning `op - op.cbegin()`)

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -152,6 +185,9 @@ def _validate(self, qobj, backend_options, noise_model):
                                'result data will not contain counts.', name)
             # Check if Clifford circuit or if measure opts missing
             no_measure = True
+            ch_supported = False
```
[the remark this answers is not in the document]
padraic-padraic: I went for 
```python
ch_supported = method in ['ch', 'automatic']
```
to keep the line short!

padraic-padraic:
Hey Chris,

I’ve resolved that conflict now! Some kind of odd white space issue is all. 

I’m happy for this to be merged now. In a future update I’d like to add u2/u3/U decompositions but for now this is useable as is.

PR #54, opened 2019-02-11
Fixing capitalization in word BibTeX

cbjuan:
### Summary
According to http://www.bibtex.org/, the right capitalization for "BibTeX" is this, instead of "BibTex". This PR fixes it.


atilag:
Thanks Juan!!

PR #57, opened 2019-02-13
Refactor Unitary simulator

chriseclectic:
### Summary

The main purpose of this is to refactor the `UnitaryMatrix` to be a subclass of the `QubitVector` so that it can inherit all the optimisations from that class, and to reduce the amount of duplicated code.

This is done by storing the unitary as a column-vectorized matrix, then all the standard gate operations acting on the vector are equivalent to left matrix multiplication without any further modification.

### Details and comments

Other changes:

* Added code documentation and cleaned up naming and arguments of the lambda helper functions in QubitVector class
* Some renaming of C++ classes eg: `QubitVector::State` to `Statevector::State`, `QubitUnitary` to `UnitaryMatrix`.
* Some other minor changes of QubitVector method names due to issues that arrised in subclassing.


chriseclectic:
@hhorii I couldn't add you as reviewer, but can you have a look at this?

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
-
-template <class statevector_t>
-void QubitVector<statevector_t>::initialize(const statevector_t &statevec, const size_t num_states) {
-  if (num_states_ != num_states) {
-    std::stringstream ss;
-    ss << "QubitVector<statevector_t>::initialize input vector is incorrect length (";
-    ss << num_states_ << "!=" << num_states << ")";
-    throw std::runtime_error(ss.str());
+  for (int_t k = 0; k < end; ++k) {
+    data_[k] = 0.0;
```
atilag: And here, I'd put a `std::fill(data_, data_+END, 0.0);`, and remove the OpenMP directive on top, as this is going to be optimized to some sort of `memset` operation, which is pretty fast.

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
 
   // Free any currently assigned memory
-  if (statevector_)
-    free(statevector_);
+  if (data_)
+    free(data_);
 
   if (checkpoint_) {
     free(checkpoint_);
     checkpoint_ = 0;
```
atilag: oh, if this is a pointer we MUST use `nullptr` and never use anything else :)

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
   }
 
   // Allocate memory for new vector
-  statevector_ = reinterpret_cast<complex_t*>(malloc(sizeof(complex_t) * num_states_));
+  data_ = reinterpret_cast<complex_t*>(malloc(sizeof(complex_t) * size_));
```
atilag: ok, I think this is coming from one of the Hiroshi's optimizations... but can't really remember why we are using a plain `malloc` instead of a ` new`, which is the idiomatic way of allocating heap memory in C++

atilag: @hhorii can we change this  `malloc` into a `new`?  Is there any known performance degradation with this change?

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
-template <class statevector_t>
-void QubitVector<statevector_t>::checkpoint() {
+template <class data_t>
+void QubitVector<data_t>::checkpoint() {
   if (!checkpoint_)
-    checkpoint_ = reinterpret_cast<complex_t*>(malloc(sizeof(complex_t) * num_states_));
+    checkpoint_ = reinterpret_cast<complex_t*>(malloc(sizeof(complex_t) * size_));
 
-  const int_t end = num_states_;    // end for k loop
+  const int_t end = size_;    // end for k loop
```
atilag: `end` -> `END`

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
 #pragma omp parallel for if (num_qubits_ > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
   for (int_t k = 0; k < end; ++k)
-    checkpoint_[k] = statevector_[k];
+    checkpoint_[k] = data_[k];
```
atilag: That would probably needs to be benchmarked... but I'd remove the OpenMP directive and put a `std::copy` instead of this loop.

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
-      const complex_t z = statevector_[k] * std::conj(checkpoint_[k]);
-      z_re += std::real(z);
-      z_im += std::imag(z);
-    }
-  } // end omp parallel
-  return complex_t(z_re, z_im);
+  const int_t end = size_;    // end for k loop
+#pragma omp parallel for if (num_qubits_ > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
+  for (int_t k = 0; k < end; ++k)
+    data_[k] = checkpoint_[k];
```
atilag: Same as above: remove #pragma and change the loop to `std::copy`

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
+
+  if (!keep) {
+    free(checkpoint_);
+    checkpoint_ = 0;
```
atilag: `0` -> `nullptr`

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
+template <class data_t>
+void QubitVector<data_t>::initialize_from_vector(const cvector_t &statevec) {
+  if (size_ != statevec.size()) {
+    std::stringstream ss;
+    ss << "QubitVector::initialize input vector is incorrect length (";
+    ss << size_ << "!=" << statevec.size() << ")";
+    throw std::runtime_error(ss.str());
+  }
+
+  const int_t end = size_;    // end for k loop
```
atilag: `end` -> `END`

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
 
-  const int_t end = num_states_;    // end for k loop
 #pragma omp parallel for if (num_qubits_ > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
   for (int_t k = 0; k < end; ++k)
-    statevector_[k] = checkpoint_[k];
+    data_[k] = statevec[k];
```
atilag: `std::copy()` ?

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
+template <class data_t>
+void QubitVector<data_t>::initialize_from_data(const data_t &statevec, const size_t num_states) {
+  if (size_ != num_states) {
+    std::stringstream ss;
+    ss << "QubitVector::initialize input vector is incorrect length (";
+    ss << size_ << "!=" << num_states << ")";
+    throw std::runtime_error(ss.str());
   }
+
+  const int_t end = size_;    // end for k loop
```
atilag: `end` -> `END`

src/simulators/statevector/qubitvector.hpp
```diff
@@ -516,128 +684,120 @@ cvector_t QubitVector<statevector_t>::vector() const {
...
+
+#pragma omp parallel for if (num_qubits_ > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
+  for (int_t k = 0; k < end; ++k)
+    data_[k] = statevec[k];
```
atilag: `std::copy()` ?

src/simulators/statevector/qubitvector.hpp
```diff
@@ -845,160 +1230,600 @@ complex_t QubitVector<statevector_t>::apply_reduction_lambda(const std::vector<u
...
+                                                const cvector_t &mat) {
+  // Special low N cases using faster static indexing
+  switch (qubits.size()) {
+  case 1:
+    apply_diagonal_matrix(areg_t<1>({{qubits[0]}}), mat);
+    break;
+  case 2:
+    apply_diagonal_matrix<2>(areg_t<2>({{qubits[0], qubits[1]}}), mat);
+    break;
+    case 3: {
```
atilag: From 3 to 16, we can pass-through here as well

src/simulators/statevector/qubitvector.hpp
```diff
@@ -845,160 +1230,600 @@ complex_t QubitVector<statevector_t>::apply_reduction_lambda(const std::vector<u
...
+                                       const cvector_t &mat) {
+  // Special low N cases using faster static indexing
+  switch (qubits.size()) {
+  case 1:
+    apply_matrix(areg_t<1>({{qubits[0]}}), mat);
+    break;
+  case 2:
+    apply_matrix(areg_t<2>({{qubits[0], qubits[1]}}), mat);
+    break;
+  case 3: {
```
atilag: We can compact the code a lot by passing-through from case 3 to case 16, like:
```c++
auto qubits_size = qubits.size();
switch(qubits_size) {
...
case 3:
case 4:
case 5:
...
case 16:
    areg_t<qubits_size> qubits_arr;
    std::copy_n(qubits.begin(), qubits_size, qubits_arr.begin());
    apply_matrix(qubits_arr, mat);
    break;
}
```

src/simulators/statevector/qubitvector.hpp
```diff
@@ -369,30 +537,30 @@ class QubitVector {
 // JSON Serialization
 //------------------------------------------------------------------------------
 
-template <class statevector_t>
-inline void to_json(json_t &js, const QubitVector<statevector_t> &qv) {
+template <class data_t>
+inline void to_json(json_t &js, const QubitVector<data_t> &qv) {
```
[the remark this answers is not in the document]
chriseclectic: In this case (and the ostream one) the inline stops a pedantic compiler warning on gcc

src/simulators/statevector/qubitvector.hpp
chriseclectic: These were done this way so `data_` could be replaced with any structure that has an operator access method and doesn't necessarily work with the STL

src/simulators/statevector/qubitvector.hpp
chriseclectic: I think Hiroshi tested performance of the different ways and this was the fastest

src/simulators/statevector/qubitvector.hpp
```diff
@@ -845,160 +1230,600 @@ complex_t QubitVector<statevector_t>::apply_reduction_lambda(const std::vector<u
...
+                    const cvector_t &_mat)->void {
+    const uint_t dim = 1ULL << N;
+    std::array<complex_t, dim> cache;
+    for (size_t i = 0; i < dim; i++) {
+      const auto ii = inds[i];
+      cache[i] = data_[ii];
+      data_[ii] = 0.;
+    }
+    // update state vector
+    for (size_t i = 0; i < dim; i++)
```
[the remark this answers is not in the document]
chriseclectic: This is an inner loop of an already parallelised lambda function

src/simulators/statevector/qubitvector.hpp
chriseclectic: I don't think this will compile because for the std::array methods the size needs to be known at compile time, hence why these big switch statements are needed.

src/simulators/unitary/unitarymatrix.hpp
```diff
@@ -0,0 +1,232 @@
...
+  explicit UnitaryMatrix(size_t num_qubits);
+  UnitaryMatrix(const UnitaryMatrix& obj) = delete;
+  UnitaryMatrix &operator=(const UnitaryMatrix& obj) = delete;
+
+  //-----------------------------------------------------------------------
+  // Utility functions
+  //-----------------------------------------------------------------------
+
+  // Set the size of the vector in terms of qubit number
+  inline void set_num_qubits(size_t num_qubits);
```
atilag: no `inline` :)

src/simulators/unitary/unitarymatrix.hpp
```diff
@@ -0,0 +1,232 @@
...
+  // Type aliases
+  using BaseVector = QubitVector<data_t>;
+
+  //-----------------------------------------------------------------------
+  // Constructors and Destructor
+  //-----------------------------------------------------------------------
+
+  UnitaryMatrix() : UnitaryMatrix(0) {};
+  explicit UnitaryMatrix(size_t num_qubits);
+  UnitaryMatrix(const UnitaryMatrix& obj) = delete;
```
atilag: Just asking: Why do we explicitly remove copy constructor? I guess because we don't want copy from other UnitaryMatrixes, but is there a reason?

src/simulators/unitary/unitarymatrix.hpp
```diff
@@ -0,0 +1,232 @@
...
+  //-----------------------------------------------------------------------
+
+  // Set the size of the vector in terms of qubit number
+  inline void set_num_qubits(size_t num_qubits);
+
+  // Return the number of rows in the matrix
+  size_t num_rows() const {return rows_;}
+
+  // Returns the number of qubits for the current vector
+  inline uint_t num_qubits() const { return num_qubits_;}
```
atilag: no `inline`

src/simulators/unitary/unitarymatrix.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+  // Returns a copy of the underlying data_t data as a complex vector
+  AER::cmatrix_t matrix() const;
```
atilag: hmmm... so `AER` is the namespace of the common utilities (framework), and because this class is not part of the common framework is not included in the `AER` namespace, am I right?

src/simulators/unitary/unitarymatrix.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+  // Return JSON serialization of UnitaryMatrix;
+  json_t json() const;
```
atilag: At some point in the future, we have to review all the responsibilities our classes have because they might be too many.
In general, enforcing the `Single Responsibility Principle` is quite beneficial for having a flexible, debuggable, hence maintainable code base.

atilag: In this case for example, the `UnitaryMatrix` should know nothing about `json` or serialization... this is other's component responsibility, that will probably know how to serialize any of our components, not only a `UnitaryMatrix`.

src/simulators/statevector/qubitvector.hpp
atilag: You are right. Nevermind :)

src/simulators/statevector/qubitvector.hpp
chriseclectic: I didn't std::copy work on raw c-arrays? These functions were coded this way to be agnostic to the data type (as long as it has an accessor []).

atilag:
Tests are failing because of this commit from Terra: https://github.com/Qiskit/qiskit-terra/commit/f6ec0134d88377895264a8f8867961d7ea0216d7
I reverted this commit locally, and everything is working, so I'm gonna merge this PR and re-run all the tests once Terra is fixed.

PR #72, opened 2019-02-28
Comment derivation depolarising probability from gate error

fanizzamarco:
### Summary
p_depol = (dim) * gate_error / (dim-1) should not be instead
p_depol = (dim-1) * gate_error / dim ? 
I changed it but the other formulas should be also corrected

### DIFF
--- a/qiskit/providers/aer/noise/device/models.py
+++ b/qiskit/providers/aer/noise/device/models.py
@@ -308,13 +308,19 @@ def _depol_error_value_one_qubit(gate_error, gate_time=0, t1=inf, t2=inf):
         raise NoiseError("Invalid T_2 relaxation time parameter: T_2 greater than 2 * T_1.")
 
     # If T1 or T2 we have only a depolarizing error model
-    # in this case p_depol = (dim-1) * gate_error /  dim
-    # with dim = 2 for 1-qubit
+    #The gate error is the average gate infidelity e = 1 - F.
+    #For the depolarizing channel we have:
+    #F(E_dep) = (1 - p) * F(I) + p * F(D) = (1 - p) * 1 + p * (1 / d)    # F(I) = 1, F(D) = 1/d = 1 - p * (d - 1) / d
+    # => p = (1 - F(E_dep)) / ((d - 1) / d) 
+    # = d * (1 - F(E_dep)) / (d - 1)
+    # = d * e / (d - 1)
+    # therefore p_depol = (dim) * gate_error / (dim-1)
+    # with dim = 4 for 2-qubit
     if gate_time is None:
         gate_time = 0
     if gate_time == 0 or (t1 == inf and t2 == inf):
         if gate_error is not None and gate_error > 0:
-            return gate_error / 2
+            return 2 * gate_error 
         else:
             return 0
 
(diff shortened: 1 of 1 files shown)


chriseclectic:
Hi @fanizzamarco, thanks for checking this. I just rechecked the calculation and I believe it's correct:

The depolarizing channel probability in Aer is defined as `E_dep = (1-p) * I +  p * D` where `I` is the identity channel and `D` is the completely depolarizing channel (`D(rho) = id / d`)

The gate error is the average gate infidelity `e = 1 - F`. 

For the depolarizing channel we have:

```
F(E_dep) = (1 - p) * F(I) + p * F(D) 
         = (1 - p) * 1 + p * (1 / d)    # F(I) = 1, F(D) = 1/d
         = 1 - p * (d - 1) / d
=> p = (1 - F(E_dep)) / ((d - 1) / d) 
     = d * (1 - F(E_dep)) / (d - 1)
     = d * e / (d - 1)
```
Hence for `d=2` if all gate error is due to the depolarizing channel we have `p_depol = 2 * gate_error`

Maybe you could update your PR with the more complete derivation in the code comments and we could merge that?

fanizzamarco:
> Hi @fanizzamarco, thanks for checking this. I just rechecked the calculation and I believe it's correct:
> 
> The depolarizing channel probability in Aer is defined as `E_dep = (1-p) * I + p * D` where `I` is the identity channel and `D` is the completely depolarizing channel (`D(rho) = id / d`)
> 
> The gate error is the average gate infidelity `e = 1 - F`.
> 
> For the depolarizing channel we have:
> 
> ```
> F(E_dep) = (1 - p) * F(I) + p * F(D) 
>          = (1 - p) * 1 + p * (1 / d)    # F(I) = 1, F(D) = 1/d
>          = 1 - p * (d - 1) / d
> => p = (1 - F(E_dep)) / ((d - 1) / d) 
>      = d * (1 - F(E_dep)) / (d - 1)
>      = d * e / (d - 1)
> ```
> Hence for `d=2` if all gate error is due to the depolarizing channel we have `p_depol = 2 * gate_error`
> 
> Maybe you could update your PR with the more complete derivation in the code comments and we could merge that?

Great, sorry for having used your time! I exchanged the definitions, and my confusion came from the fact that there is no warning on gate_error in the calculation of p_depol, and it could come out p_depol>1. But the code works perfectly then, since p_depol is checked afterwards. Thank you very much for your effort!

fanizzamarco:
Now that I understand how this works, I still have a doubt on the last formula... When the relaxation times go to infinity p_depol should go to the value that we have already checked. It seems that this does not hold for the last formula

```
denom = (q0_par1 + q1_par1 + q0_par1 * q1_par1 +
             4 * q0_par2 * q1_par2 +
             2 * (q0_par2 + q1_par2) +
             2 * (q1_par1 * q0_par2 + q0_par1 * q1_par2))
    p_depol = 1 + (5 / 3) * (4 * gate_error - 3) / denom
```
when q0_par1=1,q0_par2=1,... denom=15, so maybe it is
p_depol = 1 + (5 ) * (4 * gate_error - 3) / denom
so that when denom=15, p_depol=4/3 *gate_error. Moreover, when gate_error=zero and denom=15, p_depol=1/3, while it should be zero.
sorry if I am misunderstanding again

chriseclectic:
Yes, good catch! The correct formula should indeed be

```python
p_depol = 1 + 5 * (4 * gate_error - 3) / denom
```

fanizzamarco:
Corrected!

chriseclectic:
Thanks @fanizzamarco !

PR #74, opened 2019-03-04
Fix segfault on Mac OS when running the standalone qasm_simulator.

wh5a:


atilag:
Thanks for the fix @wh5a .
I'm gonna merge it, but this has brought a flaw in the way we load the config from the Qobj. I'll create a new issue with this and address it in another PR.

PR #81, opened 2019-03-11
[WIP] Add circuit equivalence checker

chriseclectic:
### Summary

Adds an identity checker mode to the `UnitarySimulator` that can be used as a basis for a circuit equivalence checker.

This mode is activated by using `backend_options={"identity_checker": True}` for the `UnitarySimulator` and in this case instead of returning the final unitary matrix the result will say if the final unitary is `exp(i * theta) * U_target` where `U_target` is the identity matrix, or a custom target unitary matrix. If result is `True` it also returns the global phase `theta`. 

### Details and comments

To manually activate the identity checking mode one can do the following

```python
# Get ideal circuit
target_circuit = ... # Target quantum circuit
target_unitary = qiskit.execute(circuit_target, UnitarySimulator()).result().get_unitary(target_circuit)

# Check new circuit
test_circuit = ... # new quantum circuit
opts = {
    "identity_checker": True, 
    "target_unitary": target_unitary
}
job = qiskit.execute(test_circuit, UnitarySimulator(),
    backend_options=opts)
results = job.results()
print(results.data(0))
# prints 
{"identity": [True, 0.0]}  # output is [pass, global_phase]
```

A more efficient method is to compares $U_{target}^\dagger . U = id$ (`inverse(target_circuit).circuit` to the identity):

```python
target_circuit = ...
test_circuit = ... 
check_circuit = target_circuit.inverse() + test_circuit
opts =  {"identity_checker": True }
job = qiskit.execute(check_circuit, UnitaryVerifier(),
    backend_options=opts)
results = job.results()
print(results.data(check_circuit))
```

### TODO

Decide how to expose this functionality to the user for a circuit equivalence checker such as `check_equivalent(circuit1, circuit2)`

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -17,6 +17,7 @@ Qiskit Aer qasm simulator backend.
 import json
 import logging
 import datetime
+import os
 import time
 import uuid
 from numpy import ndarray
@@ -32,6 +33,10 @@ from ..aererror import AerError
 
 # Logger
 logger = logging.getLogger(__name__)
+
+# Location where we put external libraries that will be loaded at runtime
+# by the simulator extension
+LIBRARY_DIR = os.path.dirname(__file__)
 
 
 class AerJSONEncoder(json.JSONEncoder):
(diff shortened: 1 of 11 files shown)


atilag:
I like the idea of these simulators being backends of the `UnitarySimulator`, and use them by passing a backend option.

chriseclectic:
Revisiting this PR after the 0.2 release and updated the description. I have removed the separate backends so that now there is an identity checker mode for the unitary controller / simulator. How to expose this functionality to the end user still needs to be discussed.

PR #83, opened 2019-03-12
add circuit optimization framework

hhorii:
### Summary
This PR provides circuit optimization for QasmController.

### Details and comments
Current design is here.

1. Add optimization in constructor.
```
QasmController::QasmController() {
  add_optimization(ReduceNop());
}
```
2. Apply optimization in `optimize_circuit()`.
```
template <class State_t>
void QasmController::optimize_circuit(const Circuit &input_circ,
                                      Circuit &output_circ) const {

  Circuit working_circ = input_circ;
  for (std::shared_ptr<CircuitOptimization> opt: optimizations)
    opt->optimize_circuit(working_circ);

  output_circ = working_circ;
}
```


src/framework/circuitopt.hpp
```diff
@@ -0,0 +1,42 @@
...
+  virtual void optimize_circuit(Circuit& circ) const = 0;
+  void set_config(const json_t &config);
+
+protected:
+  json_t config_;
+};
+
+void CircuitOptimization::set_config(const json_t& config) {
+  for ( auto it = config.begin(); it != config.end(); ++it )
+    config_[it.key()] = it.value();
```
chriseclectic: I think this loop can just be replaced with `config_ = config;`.

src/simulators/qasm/basic_optimization.hpp
```diff
@@ -0,0 +1,51 @@
...
+class Debug : public CircuitOptimization {
+public:
+  void optimize_circuit(Circuit& circ) const;
+};
+
+void Debug::optimize_circuit(Circuit& circ) const {
+
+  std::vector<Operations::Op>::iterator it = circ.ops.begin();
+  while (it != circ.ops.end()) {
+    std::cerr << it->name << std::endl;
```
chriseclectic: Can you change this to `std::clog << it->name << std::endl;` ?

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -73,6 +79,14 @@ class QasmController : public Base::Controller {
   // Clear the current config
   void virtual clear_config() override;
 
+  // Add circuit optimization
+  template <typename Type>
+  inline auto add_optimization(Type&& opt)
```
chriseclectic: This should be added to the `Base::Controller` class so it can be used by all the controller subclasses. I would also rename it to be `add_circuit_optimization` to be specific about what optimizations we are talking about

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -186,13 +200,22 @@ class QasmController : public Base::Controller {
   // Initial statevector for Statevector simulation method
   cvector_t initial_statevector_;
 
+  std::vector<std::shared_ptr<CircuitOptimization>> optimizations;
```
chriseclectic: `optimizations_;` with underscore for private/protected class member

Also could you add a short documentation comment above for what this is.

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -299,8 +322,12 @@ void QasmController::initialize_state(const Circuit &circ,
 template <class State_t>
 void QasmController::optimize_circuit(const Circuit &input_circ,
```
chriseclectic: This function should also be moved to the `Base::Controller` class

chriseclectic: Maybe we should also change the function definition to be:

```c++
Circuit optimize_circuit(const Circuit &circ) const
```

So that it returns the output circuit object and is used like

```c++
auto opt_circ = optimize_circuit<State_t>(circ);
```

rather than 

```c++
Circuit opt_circ;
optimize_circuit<State_t>(circ, opt_circ);
```

chriseclectic:
@hhorii If you make the changes above (and update the conflicting files) we can get this merged

PR #90, opened 2019-03-15
allow various qubit vector classes in qasm controller

hhorii:
### Summary
Current `AER::Simulator::QasmController` cannot change its `statevec_t` class in `AER::Statevector::State` because `AER::Statevector::State` explicitly specifies `QV::QubitVector` as `statevec_t` in its class definition.
This PR changes`AER::Statevector::State` not to specify `statevec_t` and `AER::Simulator::QasmController` specifies `QV::QubitVector` in `run_circuit()`.

### Details and comments
Originally, `AER::Statevector::State` explicitly specifies `QV::QubitVector` as `statevec_t` in its class definition.
```
template <class data_t = complex_t*>
class State : public Base::State<QV::QubitVector<data_t>> {
public:
  using BaseState = Base::State<QV::QubitVector<data_t>>;

```
This PR changes`AER::Statevector::State` not to specify `statevec_t`.
```
template <class statevec_t>
class State : public Base::State<statevec_t> {
public:
  using BaseState = Base::State<statevec_t>;

```

### DIFF
--- a/src/simulators/statevector/statevector_state.hpp
+++ b/src/simulators/statevector/statevector_state.hpp
@@ -40,7 +40,7 @@ enum class Snapshots {
 // QubitVector State subclass
 //=========================================================================
 
-template <class statevec_t>
+template <class statevec_t = QV::QubitVector<complex_t*>>
 class State : public Base::State<statevec_t> {
 public:
   using BaseState = Base::State<statevec_t>;
--- a/src/simulators/statevector/statevector_state.hpp
+++ b/src/simulators/statevector/statevector_state.hpp
@@ -40,10 +40,10 @@ enum class Snapshots {
 // QubitVector State subclass
 //=========================================================================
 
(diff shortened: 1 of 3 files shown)


src/simulators/statevector/statevector_state.hpp
```diff
@@ -40,10 +40,10 @@ enum class Snapshots {
 // QubitVector State subclass
 //=========================================================================
 
-template <class data_t = complex_t*>
-class State : public Base::State<QV::QubitVector<data_t>> {
+template <class statevec_t>
```
chriseclectic: Can you change it to still has the original default class:

```c++
template <class statevec_t = QV::QubitVector<complex_t*>>
```

hhorii: Thanks. I added a default class for the template and remove class specification in callers.

PR #104, opened 2019-03-26
Noise transform

gadial:
### Details and comments

**TODO**
- [ ] Add a line in the Changelog

### DIFF
--- a/qiskit/providers/aer/noise/__init__.py
+++ b/qiskit/providers/aer/noise/__init__.py
@@ -67,3 +67,4 @@ canonical types of quantum errors:
 from .noise_model import NoiseModel
 from . import errors
 from . import device
+from .noise_transformation import NoiseTransformer
--- a/qiskit/providers/aer/noise/noise_transformation.py
+++ b/qiskit/providers/aer/noise/noise_transformation.py
@@ -1,0 +1,405 @@
+# coding: utf-8
+
+import numpy
+import sympy
+import itertools
+import copy
+
+from functools import singledispatch
+
+
+class NoiseTransformer:
+    """Transforms one quantum noise channel to another based on a specified criteria.
+
+    A quantum 1-qubit noise channel is represented by Kraus operators: a sequence (E0, E1,...,En) of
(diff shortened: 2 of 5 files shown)


chriseclectic:
Hi @gadial here are some general ideas for restructuring the code.

From a high level individual noise objects are represented in Aer as `QuantumError` objects, and these are placed into `NoiseModel` objects for a simulation. I think this code should basically consist of two functions that can convert these objects (`QuantumError` or `NoiseModel` object) into an approximate version that can then be used for simulation.

So basically restructuring into two functions:

```python
def approximate_quantum_error(error, <additional_args>):
    """Return an approximate QuantumError.
    
    Args:
        error (QuantumError): the error to be approximated.
    
    Returns:
        QuantumError: the approximate quantum error.
    """
    # This would accept as input anything that the QuantumError class can take as input
    # Eg: list of circuits, list of kraus matrices, a QuantumChannel sublcass object, or a 
    # QuantumError object. This could be done as:
    if not isinstance(error, QuantumError):
        error = QuantumError(error)
   # rest of code for approximating error
   return approx_error

def approximate_noise_model(model, <additional_args>):
    """Return an approximate noise model.
    
    Args:
        model (NoiseModel): the noise model to be approximated.
    
    Returns:
        NoiseModel: the approximate noise model.
    
    Raises:
        NoiseError: if the QuantumError cannot be approximated.
    """
    # This function would iterate over all QuantumErrors in the noise model
    # and call the `approximate_quantum_errror` function on them to generate
    # an approximate noise model
   return approx_noise_model
```

Some helper functions you might want to use so that there isn’t duplication:
* The `quantum_info` `QuantumChannel` module has functions for converting between channel representations (eg. can convert other representations into a Kraus representation and includes functions for checking Kraus representations are CPTP). I recently refactored the QuantumError object in Aer to use these for some checks.
* There are also some helper functions already in Aer that you can use for single-qubit Clifford gates (and append as needed). These are in the https://github.com/Qiskit/qiskit-aer/blob/master/qiskit/providers/aer/noise/errors/errorutils.py file

PR #118, opened 2019-04-02
Rename "ch" method to "extended_stabilizer"

chriseclectic:
### Summary

* Renames the `"ch"` `QasmSimulator` method to more descriptive `"extended_stabilizer"` method and updates options and tests to use this new convention.
* Updates `CH::Runner` OpenMP loops to be compatible with windows OpenMP 2.0.

@padraic-padraic I can't add you as a reviewer here, but could you look at these changes?


src/simulators/extended_stabilizer/ch_runner.hpp
```diff
@@ -192,8 +192,9 @@ bool Runner::check_omp_threshold()
 
 void Runner::apply_pauli_projector(const std::vector<pauli_t> &generators)
 {
+  const int_t END = num_states_;
   #pragma omp parallel for if(num_states_ > omp_threshold_ && num_threads_ > 1) num_threads(num_threads_)
-  for(uint_t i=0; i<num_states_; i++)
+  for(int_t i=0; i<END; i++)
```
atilag: I'd rather having a `static_cast<int_t>(num_states_)` than using another variable with different semantic meaning`END != num_states_`

PR #119, opened 2019-04-02
update install from source instructions in contributing.md

derivation:
### Summary
Refactored install from source instructions in contributing.md

### Details and comments
Also fixed some formatting that was in ReST to be in markdown.

**TODO**
- [ ] Update Changelog


.github/CONTRIBUTING.md
```diff
@@ -34,151 +30,191 @@ ensure that the community is free to use your contributions.
...
+
+    qiskit-aer$ python ./setup.py bdist_wheel
+
+We are using **scikit-build** as a substitute for **setuptools**. This is
+basically the glue between **setuptools** and **CMake**, so there are various
+options to pass variables to **CMake**, and the underlying build system
+(depending on your platform). The way to pass variables is:
+
+    qiskit-aer$ python ./setup.py bdist_wheel -- -DCMAKE_VARIABLE=Values -- -
+    Makefile_Flag
```
atilag: This `Makefile_Flag` should be right after the `-` of the above line:
```
qiskit-aer$ python ./setup.py bdist_wheel -- -DCMAKE_VARIABLE=Values -- -Makefile_Flag
```
... but it can be a Visual Studio like flag, so the right thing to say here should be:
```
qiskit-aer$ python ./setup.py bdist_wheel -- -DCMAKE_VARIABLE=Values -- [-Makefile_Flag | /VisualStudio_Flag]
```
... but all in the same line.

.github/CONTRIBUTING.md
```diff
@@ -34,151 +30,191 @@ ensure that the community is free to use your contributions.
...
 
+### macOS
 
+There are various methods depending on the compiler we want to use. If we want
+to use the **Clang** compiler, we need to install an extra library for
+supporting **OpenMP**: **libomp**. The **CMake** build system will warn you
+otherwise. To install it manually, in a terminal window, run:
 
-### Building
+    brew install libomp
```
atilag: We usually put the `$` sign to emphasize the fact that this is a command line tool that needs to be run in a terminal.

.github/CONTRIBUTING.md
```diff
@@ -34,151 +30,191 @@ ensure that the community is free to use your contributions.
...
+You further need to have **Xcode Command Line Tools** installed on macOS:
+
+    xcode-select --install
+
+There are two ways of building Aer simulators, depending on your goal:
+
+1. Build a Terra compatible add-on;
+2. Build a standalone executable.
+
+**Terra Add-on**
```
atilag: Are we repeating the same instructions here?
One of the  cool things of using `scikit-build / CMake` is that we always build in the same way. One build system for all platforms.

.github/CONTRIBUTING.md
```diff
@@ -34,151 +30,191 @@ ensure that the community is free to use your contributions.
...
+passing `-j8` flag to the underlaying build system, which in this case is
+Makefile, telling it that we want to build in parallel, using 8 processes.
 
-```
-qiskit-aer$ python ./setup.py bdist_wheel --plat-name macosx-10.9-x86_64 -- -DSTATIC_LINKING=False -- -j8
-```
+> You may need to turn off static linking and specify your platform name, e.g.:
 
-After this command is executed successfully, we will have a wheel package into the ``dist/`` directory, so next step is installing it:
+>   qiskit-aer$ python ./setup.py bdist_wheel --plat-name macosx-10.9-x86_64 -- >   -DSTATIC_LINKING=False -- -j8
```
atilag: This command has a typo, there should be no `>`, anyway let's change this and simplify, because STATIC_LINKING is disabled by default, so there's no need to specify it:
```
qiskit-aer$ python ./setup.py bdist_wheel --plat-name macosx-10.9-x86_64 -- -- -j8
```

.github/CONTRIBUTING.md
```diff
@@ -34,151 +30,191 @@ ensure that the community is free to use your contributions.
...
-qiskit-aer/out/Release$ ls
-aer_simulator_cpp
-```
+### Windows
+
+On Windows, you must have **Anaconda3** installed. We recommend also installing
+**Visual Studio 2017** (Community Edition). **Anaconda3** is required when
+searching for an **OpenBLAS** implementation. If **CMake** can't find a suitable
+implementation installed, it will take the **BLAS** library from the
+**Anaconda3** environment.
```
atilag: Ok, this is my fault. The instructions are very poor, I need to improve Windows instructions.

.github/CONTRIBUTING.md
derivation: @atilag This section about scikit-build in our doc seems to contradict [scikit-build doc](https://scikit-build.readthedocs.io/en/latest/usage.html#command-line-options).

According to scikit-build doc, the way to pass variables is actually 

```
qiskit-aer$ python ./setup.py bdist_wheel [skbuild_opts] \
[-- cmake_opts] [-- build_tool_opts]
```

I don't understand where this part of our syntax example comes from: -DCMAKE_VARIABLE=Values. A Google search for DCMAKE_VARIABLE only returns a hit on this contributing file!

The example in our doc near line 103:

`qiskit-aer$ python ./setup.py bdist_wheel -j8`

seems to be explained incorrectly.

According to the scikit-build doc, the -j8 flag is actually being passed as skbuild_opts, although our doc says -j8 is being passed as a bulid_tool_opts. 

Our doc mentions STATIC_LINKING as a cmake_opt, but the example command doesn't set STATIC_LINKING explicitly.

So, overall, this section about scikit-build is very confusing to me. Could you help clarify what's trying to be said?

.github/CONTRIBUTING.md
derivation: I was totally confused by the mixture of linux, macOS, windows caveats in the original version of the doc, so I figured separating the platforms out would make reading easier. I'll try to combine them and maintain clarity.

.github/CONTRIBUTING.md
derivation: ok, I'll put these marks back in

.github/CONTRIBUTING.md
derivation: yeah, I messed up line breaks. Those > marks were meant to be markdown syntax to create a block quote.

.github/CONTRIBUTING.md
derivation: For the example using -j8, the text says this is being passed to the underlying build system, so shouldn't the flag be entered as '-- -- -j8'? I'm basing this on [scikit-build docs](https://scikit-build.readthedocs.io/en/latest/usage.html#usage-setuptools-options).

atilag: The format is the following:
```
python ./setup.py <action> -- parameters passed to CMake -- parameters passed to underlaying build system
```
The `-j8` is a parameter for Makefile (the underlaying build system in Mac and Linux), so for example, if I want to pass no parameters to CMake but pass some to Makefile, the command would be:
```
python ./setup.py bdist_wheel -- -- -j8
```
But if I do want to pass something to CMake:
```
python ./setup.py bdist_wheel -- -DPARAMTER_TO_CMAKE=Value -- -j8
```
So the `--` is like the delimiter

atilag:
Paul, could you add a line in the Changelog advertising this change? 
I have added a TODO task in the description, check it when it's done.
Thanks!

.github/CONTRIBUTING.md
chriseclectic: I agree with Paul, seperate sections and repetition is much clearer. That way you can just jump to one section and read everything not have to go back and forth fishing for cmake flags and dependencies

PR #123, opened 2019-04-04
imporve u1 performance in statevector

hhorii:
### Summary
This PR improves singel-qubit phase operation in statevector.

### Details and comments
`AER::Statevector::apply_gate_phase` calls `AER::Statevector::apply_matrix(reg_t, cvector_t)`, which calls `QV::QubitVector::apply_diagonal_matrix(reg_t, cvector_t)`.

However, `apply_diagonal_matrix(uint_t, cvector_t)` is much better than `apply_diagonal_matrix(reg_t, cvector_t)`.

This PR bridges the both.

### DIFF
--- a/src/simulators/statevector/qubitvector.hpp
+++ b/src/simulators/statevector/qubitvector.hpp
@@ -1210,6 +1210,11 @@ template <typename data_t>
 void QubitVector<data_t>::apply_diagonal_matrix(const reg_t &qubits,
                                                 const cvector_t &diag) {
 
+  if (qubits.size() == 1) {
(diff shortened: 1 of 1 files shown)


chriseclectic:
We should also modify `apply_phase_gate` in `statevector` and `unitary` states to directly call the correct `apply_diagonal_matrix` function in qubitvector. Eg:

```c++
template <class statevec_t>
void State<statevec_t>::apply_gate_phase(uint_t qubit, complex_t phase) {
  cvector_t diag = {{1., phase}};
  BaseState::qreg_.apply_diagonal_matrix(qubit, diag);
}
```

PR #126, opened 2019-04-05
Fixes in qubitvector

chriseclectic:
### Summary

Adds several fixes to `QubitVector` class:

* `apply_matrix` checks if matrix is single-qubit and if so uses optimized function
* single qubit `apply_matrix` checks if matrix is diagonal and if so uses optimized function
* fixes bug in `apply_matrix2` using row-major order instead of col-major order for matrix
* fixes indexes of lambda function to be const ref instead of copy (which is bad for the unique pointer objects returned by the indexes function and only worked when it could use move-semantics).

### DIFF
--- a/src/simulators/statevector/qubitvector.hpp
+++ b/src/simulators/statevector/qubitvector.hpp
@@ -1151,7 +1151,6 @@ complex_t QubitVector<data_t>::apply_matrix_reduction_lambda(Lambda&& func,
  * MATRIX MULTIPLICATION
  *
  ******************************************************************************/
-
 template <typename data_t>
 void QubitVector<data_t>::apply_matrix(const reg_t &qubits,
                                        const cvector_t &mat) {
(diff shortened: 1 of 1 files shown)


src/simulators/statevector/qubitvector.hpp
```diff
@@ -1358,9 +1358,17 @@ void QubitVector<data_t>::apply_toffoli(const uint_t qubit_ctrl0,
...
 void QubitVector<data_t>::apply_matrix(const uint_t qubit,
                                        const cvector_t& mat) {
-  const int_t BIT = BITS[qubit];
+  // Check if matrix is actually diagonal and if so use 
+  // apply_diagonal_matrix
+  // TODO: this should be changed to not check doubles with ==
+  if (mat[1] == 0.0 && mat[2] == 0.0) {
+    const cvector_t diag = {{mat[0], mat[3]}};
+    apply_diagonal_matrix(qubit, diag);
+  }
```
hhorii: need `return;` here?

chriseclectic: Indeed!

PR #124, opened 2019-04-04
Add support for optimized multi-controlled gates

chriseclectic:
### Summary

Adds support for multi-controller x,y,z,u1,u2,u3 gates to the statevector QasmSimulator, StatevectorSimulator and UnitarySimulator.

These gates have an optimized implementation making them faster to execute than the gate being unrolled into a basis of `cx` and single qubit gates.

Closes #127 

### Details and comments

These gates are passed in as (N+1)-qubit qobj instructions such as:

```python
{"name": "mcx", "qubits": [c0, c1,...,cn, target]}
```
where the first `N` qubits are the controls, and the last qubit is the target.

For the controlled `u1,u2,u3` gates the instruction includes the same parameters as for the single qubit gates. Eg:

```python
{"name": "mcu3", "qubits": [c0, c1,...,cn, target], "params": [theta, phi, lam]}
```
TODO:
- [ ] Add tests
- [x] Update Changelog

PR #133, opened 2019-04-06
Add two-qubit lambda in qubitvector

hhorii:
### Summary
This PR optimizes two-qubit gate operations in qubitvector

### Details and comments
Currently two-qubit operations use index. An index is generated for each iteration in a loop.

This PR introduces a new `apply_lambda` for two-qubits operations. `cnot` is implemented as follows: 
```
  const int_t GAP_ctrl = BITS[qubit_ctrl];
  const int_t GAP_trgt = BITS[qubit_trgt];
  auto lambda = [&](const int_t i00)->void {
    auto i10 = i00 | GAP_ctrl;
    std::swap(data_[i10], data_[i10 | GAP_trgt]);
  };
  apply_lambda(lambda, qubit_ctrl, qubit_trgt);
```
In the above, `i01 = i00 | GAP_trgt` and `i11 = i 10 | GAP_trgt`. This PR may increases arithmetic operations, but they are cheap comparing with index generation.  

My local environment (MacBook Pro (Retina, 13-inch, Early 2015)), cnot x 25 for a 25-qubits circuit took `3.19 sec` with this optimization and `9.55 sec` without this optimization.


chriseclectic:
@hhorii PR #124 removes these two qubit gates and replaces them with general multi-controlled versions. 

If performance is a lot better can you start from that PR and look at adding this back in as special case for N=2 qubit gates.

hhorii:
Static index with std::array<N> can show the similar performance. This latest version add static index on top of #124.

src/simulators/statevector/qubitvector.hpp
```diff
@@ -680,17 +692,34 @@ cvector_t QubitVector<data_t>::vector() const {
...
+    retval >>= qubits_sorted[j];
+    retval <<= qubits_sorted[j] + 1;
     retval |= lowbits;
   }
   return retval;
 }
 
+template <typename data_t>
+template <size_t N>
+areg_t<1ULL << N> QubitVector<data_t>::indexes(const areg_t<N> &qs,
```
chriseclectic: Is there a significant speed difference between this old static indexes function and the new function that uses unique pointers? It is much less convenient to use which is why it was removed.

chriseclectic:
@hhorii I ran some tests comparing the static indexing you added back in to the new indexing using unique_ptrs and didn't see any difference in performance:

Using a single-shot test circuit of N CNOTS for N qubits (N=10,15,20,25)
```
from qiskit import *
sim = Aer.get_backend('qasm_simulator')

def make_qobj(nq):
    q = QuantumRegister(nq)
    qc = QuantumCircuit(q)
    for j in range(nq):
        qc.cx(q[j], q[(j + 1) % nq])
    return compile(qc, Aer.get_backend('qasm_simulator'), shots=1)

nqs = [10, 15, 20, 25]
qobjs = [make_qobj(nq) for nq in nqs]
```

and simple timing using Jupyter 
```python
%%timeit
result = sim.run(qobj, validate=False).result()
```

I get results on my laptop of

**Using PR 124 (mcx gates without special case for N=2):**
* N=5: 27.5 ms ± 6.16 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
* N=10: 29.4 ms ± 5.52 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
* N=15: 34.9 ms ± 7.22 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
* N=20: 40.5 ms ± 9.16 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

**Using this PR:**
* N=5: 30 ms ± 7.19 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
* N=10: 32.4 ms ± 7.65 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
* N=15: 35.4 ms ± 5.62 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
* N=20: 41.8 ms ± 6.51 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

chriseclectic:
Some further testing and I think we can remove the single-qubit apply matrix functions and use the old static indexing. Changing the above example to apply and `x` gate instead of `cx` and comparing the current 1-qubit `apply_lambda` to an `N=1` static indexed case I get that they are about the same (or that static indexing is slightly faster).

PR #136, opened 2019-04-11
add fusion optimization for qasm simulator

hhorii:
### Summary
This PR introduces a basic optimization to aggregate gates in qasm simulator.

### Details and comments
Some consecutive gates are aggregated with this optimization.

- [x] supports fusion of 2x2 and 4x4 unitary gates
- [x] add test cases
- [x] provide configuration for performance tuning

Available backend_options: 
- `fusion_enable`: enable fusion optimization with `True` (default: `False`). 
- `fusion_max_qubit`: set maximum number of qubits for an aggregated operation.
- `fusion_verbose`: output optimized circuits
- `fusion_threshold`: minimum number of qubits of circuits to enable this optimization.

- Will remain: supports fusion of large unitary


src/simulators/statevector/qubitvector.hpp
```diff
@@ -238,6 +238,10 @@ class QubitVector {
   // The matrix is input as vector of the column-major vectorized N-qubit matrix.
   void apply_matrix(const reg_t &qubits, const cvector_t &mat);
 
+  // Apply N-qubit matrixes to the state vector.
```
chriseclectic: Could you change this description to state it's applying fusion and only works for 1 and 2 qubit matrices.
Something like:
```
// Apply a N-qubit matrix constructed from composition of 1 and 2 qubit matrices.
// The sets of qubits and matrices are passed as vectors, where each individual matrix
// is input as a column-major vectorized matrix.
```

src/simulators/statevector/qubitvector.hpp
```diff
@@ -238,6 +238,10 @@ class QubitVector {
...
+  // The matrixes are inputs as vectors of the column-major vectorized N-qubit matrixes.
+  void apply_matrixes(const std::vector<reg_t> &qubitss, const std::vector<cvector_t> &mats);
```
chriseclectic: Can you change this to `apply_matrix_sequence`: its a bit more descriptive of what the function does

src/simulators/statevector/qubitvector.hpp
```diff
@@ -539,6 +525,10 @@ class QubitVector {
   // Swap cols and rows of vectorized matrix
   void swap_cols_and_rows(const uint_t idx1, const uint_t idx2,
                           cvector_t &mat, uint_t dim) const;
+
+  cvector_t expand_matrix(const reg_t& src_qubits,
```
chriseclectic: Can you add a description of what this function does and what the arguments should be.
I think its something like `src_qubits` are the qubits for the matrix, and `dst_sorted_qubits` are the sorted list of qubits in the output space it will be expanded to

src/simulators/statevector/qubitvector.hpp
```diff
@@ -1188,6 +1152,47 @@ void QubitVector<data_t>::apply_matrix(const reg_t &qubits,
   } // end switch
 }
 
+template <typename data_t>
+void QubitVector<data_t>::apply_matrixes(const std::vector<reg_t> &qubitss,
+                                         const std::vector<cvector_t> &mats) {
```
chriseclectic:
```c++
void QubitVector<data_t>::apply_matrix_sequence(const std::vector<reg_t> &qubits_seq,
	                                         const std::vector<cvector_t> &mat_seq)
```

src/simulators/statevector/qubitvector.hpp
```diff
@@ -1188,6 +1152,47 @@ void QubitVector<data_t>::apply_matrix(const reg_t &qubits,
...
+  if (mats.size() == 0)
+    return;
+
+  reg_t sorted_qubits;
+  for (const reg_t& qubits: qubitss)
```
chriseclectic: I would add a check here that if any set of qubits is > 2 it raises an error, rather than waiting to for the exception to be raised by the `expand_matrix` function.

src/framework/operations.hpp
```diff
@@ -29,7 +29,7 @@ enum class RegComparison {Equal, NotEqual, Less, LessEqual, Greater, GreaterEqua
 // Enum class for operation types
 enum class OpType {
   gate, measure, reset, bfunc, barrier, snapshot,
-  matrix, kraus, roerror, noise_switch, initialize
+  matrix, matrixes, kraus, roerror, noise_switch, initialize
```
chriseclectic: can you change this to `matrix_sequence`

src/framework/operations.hpp
```diff
@@ -83,6 +86,7 @@ struct Op {
   OpType type;                    // operation type identifier
   std::string name;               // operation name
   reg_t qubits;                   //  qubits operation acts on
+  std::vector<reg_t> qubitss;     //  list of qubits for matrixes
```
chriseclectic: I would move this down to be above mats and call it:

```
// List of registers
std::vector<reg_t> regs;
```

Since it could be used by other operations (like the multiplexer @jagunnels is working on)

src/framework/operations.hpp
```diff
@@ -369,6 +373,17 @@ inline Op make_mat(const reg_t &qubits, const cmatrix_t &mat, std::string label
   return op;
 }
 
+inline Op make_matrixes(const std::vector<reg_t> &qubitss, const std::vector<cmatrix_t> &mats, std::string label = "") {
```
chriseclectic: `make_matrixes->make_matrix_sequence`

Here you could loop over the `qubit_seq` arguments and sort them and store the sorted qubits as `op.qubits` rather than do it later. 

This would also make it easier to combine two `matrix_sequence` instructions as you already have the sorted list for each and could just convert them to a set take the union, and convert back to a vector for example.

src/framework/operations.hpp
```diff
@@ -369,6 +373,17 @@ inline Op make_mat(const reg_t &qubits, const cmatrix_t &mat, std::string label
...
+  Op op;
+  op.type = OpType::matrixes;
```
chriseclectic: OpType::matrix_sequence

src/framework/operations.hpp
```diff
@@ -369,6 +373,17 @@ inline Op make_mat(const reg_t &qubits, const cmatrix_t &mat, std::string label
...
+  op.name = "mats";
+  op.qubitss = qubitss;
```
chriseclectic: `op.regs = qubit_seq`

src/simulators/statevector/statevector_state.hpp
```diff
@@ -67,6 +67,7 @@ class State : public Base::State<statevec_t> {
       Operations::OpType::bfunc,
       Operations::OpType::roerror,
       Operations::OpType::matrix,
+      Operations::OpType::matrixes,
```
chriseclectic: `matrix_sequence`

src/simulators/statevector/statevector_state.hpp
```diff
@@ -165,6 +166,9 @@ class State : public Base::State<statevec_t> {
   // Apply a vectorized matrix to given qubits (identity on all other qubits)
   void apply_matrix(const reg_t &qubits, const cvector_t & vmat);
 
+  // Apply multiple gate operations
+  void apply_matrixes(const std::vector<reg_t> &qubitss, const std::vector<cmatrix_t>& mats);
```
chriseclectic:
```c++
// Apply a sequence of gate operations by first composing the gates and then applying
// the resulting matrix to the state vector
void apply_matrix_sequence(const std::vector<reg_t> &qubits_seq, const std::vector<cmatrix_t>& mat_seq);
```

src/simulators/statevector/statevector_state.hpp
```diff
@@ -436,6 +434,9 @@ void State<statevec_t>::apply_ops(const std::vector<Operations::Op> &ops,
       case Operations::OpType::matrix:
         apply_matrix(op.qubits, op.mats[0]);
         break;
+      case Operations::OpType::matrixes:
```
chriseclectic: `matrix_sequence`

src/framework/circuitopt.hpp
```diff
@@ -24,8 +24,12 @@ namespace AER {
 
 class CircuitOptimization {
 public:
-  virtual void optimize_circuit(Circuit& circ) const = 0;
-  void set_config(const json_t &config);
+  virtual void optimize_circuit(Circuit& circ,
```
chriseclectic: Here it would be cleaner to just pass in either
```c++
virtual void optimize_circuit(Circuit& circ, const Operations::Opset &opset, OutputData &data) cont
```
since allowed gates, snapshots, ops can all be gotten from the opset (which can be gotten from the state)

src/framework/operations.hpp
hhorii: In plan, there are some optimization based on a matrix in matrix_sequence.  Refactor around make_matrix_sequence when these optimization is implemented.

chriseclectic:
@hhorii looks like something changed with the tests. As soon as you can get them passing we can merge

PR #117, opened 2019-04-01
Add initialise simulator instruction to statevector_state

ShellyGarion:
### Summary
#82 

### Details and comments
Added some printouts for debugging - should be removed later

**TODO**
- [ ] Update Changelog

PR #137, opened 2019-04-15
Initialize gate

yaelbh:
### Summary

This is a continuation of #117

### DIFF
--- a/src/noise/noise_model.hpp
+++ b/src/noise/noise_model.hpp
@@ -433,12 +433,12 @@ void NoiseModel::sample_local_quantum_noise(const Operations::Op &op,
   // Convert qubits to string for table lookup
   std::string op_qubits = reg2string(op.qubits);
 
-  // Get the qubit error map for  gate name
+  // Get the qubit error map for gate name
   auto iter = local_quantum_error_table_.find(name);
   if (iter != local_quantum_error_table_.end()) {
     // Check if the qubits are listed in the inner model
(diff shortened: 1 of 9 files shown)


yaelbh:
I see two tasks:
- [ ] Write tests for `qasm_simulator`, similar to tests of `reset` there.
- [ ] Handle correctly in `check_measure_sampling_opt`, namely do not optimize in the presence of `initialize` operations. Similarly update `is_measure_or_reset` to be `true` also for `initialize`.

I don't understand why Terra needs on its side to reset before the initialization, as this is done by the simulator as part of the initialization process.

yaelbh:
Note that Travis will not pass because Terra and Aer are out of sync. Need to checkout Terra f6cfde67f1c.

chriseclectic:
@yaelbh if you rebase from master it should pass now. I've merged shellys PR so you can squash/drop those commits and just add your additions/changes.

yaelbh:
Ready.
In `noise_model.hpp` there is a special handling of the case where the gate is `measure` or `reset` (`is_measure_or_reset`). It looks like the handling should be for any multiple qubit gate, including `initialize` but not only. I did not change the code there.

chriseclectic:
@yaelbh great! I would say "initialize" doesn't really need to be handled by noise model because if you were trying to mimic noise in a device you should unroll the initialise to a reset and preparation gates which would have the noise model defined on them.

PR #141, opened 2019-04-15
add omp to probabilities in qubitvector

hhorii:
### Summary
`QubitVector<data_t>::probabilities(const reg_t &qubits)` does not have an OMP directive.

### Details and comments
This PR adds `omp parallel` to a loop in `QubitVector<data_t>::probabilities(const reg_t &qubits)`.

### DIFF
--- a/src/simulators/statevector/qubitvector.hpp
+++ b/src/simulators/statevector/qubitvector.hpp
@@ -2134,10 +2134,14 @@ rvector_t QubitVector<data_t>::probabilities(const reg_t &qubits) const {
     return probabilities();
 
   rvector_t probs(DIM, 0.);
-  for (size_t k = 0; k < END; k++) {
-    auto idx = indexes(qubits, qubits_sorted, k);
-    for (size_t m = 0; m < DIM; ++m) {
-      probs[m] += probability(idx[m]);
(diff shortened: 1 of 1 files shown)


src/simulators/statevector/qubitvector.hpp
```diff
@@ -2134,10 +2134,14 @@ rvector_t QubitVector<data_t>::probabilities(const reg_t &qubits) const {
...
-    auto idx = indexes(qubits, qubits_sorted, k);
-    for (size_t m = 0; m < DIM; ++m) {
-      probs[m] += probability(idx[m]);
+#pragma omp parallel if (num_qubits_ > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
+  {
+#pragma omp for
+    for (size_t k = 0; k < END; k++) {
+      auto idx = indexes(qubits, qubits_sorted, k);
+      for (size_t m = 0; m < DIM; ++m) {
+        probs[m] += probability(idx[m]);
```
chriseclectic: Is this thread-safe since it's not a reduction? different threads may be trying to modify the same variable at the same time

hhorii: Sorry. I missed this reduction... I added OpenMP for reduction.

PR #146, opened 2019-04-16
Add OSX and python 37 and 35 jobs

mtreinish:
### Summary

We support running aer on windows, osx, and linux but we only
currently test running things on linux and only under python36. This
commit adds the travis config for running tests under osx too so we can
verify that it works as expected. It also adds the missing python versions
so we verify aer works in all support environments on linux and osx.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,17 +41,17 @@ install_osx: &stage_osx
     - brew install libomp
     - brew install openblas
     - |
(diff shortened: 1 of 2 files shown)


atilag:
Ouch, there was some desynchronization this time :(
I was already working on this!
Sorry @mtreinish 
Superseed by https://github.com/Qiskit/qiskit-aer/pull/148

PR #148, opened 2019-04-16
* Adding MacOS to the CIs

atilag:
* Moving to a stage based Travis config file
* We only test for Python 3.7 (for now)


mtreinish:
I get that CI is slow, but we need to verify things work on all python versions for all python versions. My patch adding this (#146 ) caught that the tests for example don't work on py3.5 because of the use of syntax added in 3.6. These sorts of gaps are how we end up with the current situation with broken builds on some environments.

.travis.yml
```diff
@@ -3,32 +3,113 @@
...
+    - sudo apt-get -y install libopenblas-dev
+  script:
+    - python setup.py bdist_wheel -- -DCMAKE_CXX_COMPILER=g++-7 -- -j4
+    - pip install dist/qiskit_aer*whl
+    - python -m unittest discover -s test -v
+
+stage_osx: &stage_osx
+  <<: *stage_generic
+  os: osx
+  osx_image: xcode9.2
```
chriseclectic: why not use xcode10.2 image since it's the latest version?

atilag: This is what Terra has, we should sync about this. But moving to 10.2 seems legit

PR #159, opened 2019-04-19
Fix cmake build failing for travis/linux

chriseclectic:
### Summary

Attempting to fix failing build while Juan is sleeping. Something is wrong with setting link flags on Travis from the MacOS fix recently merged.


atilag:
hahahaha this was my Easter egg...

PR #162, opened 2019-04-22
Noise transform

gadial:
### Summary
This module transforms one error channel into another built from specified building blocks, such that the two channels are as close as possible. e.g. converting a general noise to a Clifford noise that a Clifford simulator can handle.

### Details and comments

This module adds two functions:
- **approximate_quantum_error**: given an input channel and a description of the desired building blocks, and output a "close" channel built from the building blocks.
- **approximate_noise_model**: applies **approximate_quantum_error** to all noises in the noise model, with the same set of building blocks used for all noises in the model.

Examples:

Inputs as quantum error and operator string:
```
gamma = 0.23
error = amplitude_damping_error(gamma)
results = approximate_quantum_error(error, operator_string="reset")
```

Inputs as Kraus matrices and operator list:
```
X = standard_gate_unitary('x')
Y = standard_gate_unitary('y')
Z = standard_gate_unitary('z')
p = 0.34
theta = numpy.pi / 7
E0 = numpy.sqrt(1 - p) * numpy.array(numpy.eye(2))
E1 = numpy.sqrt(p) * (numpy.cos(theta) * X + numpy.sin(theta) * Y)
results = approximate_quantum_error((E0, E1), operator_list=[X, Y, Z])

```


qiskit/providers/aer/noise/errors/quantum_error.py
```diff
@@ -361,43 +361,43 @@ def power(self, n):
         return ret
 
     def tensor(self, other):
-        """Return the tensor product quantum error channel self ⊗ other.
+        """Return the tensor product quantum error channel self ? other.
```
chriseclectic: Looks like something odd is going on with the math operator font characters here

gadial:
Should be ok now.

PR #164, opened 2019-04-23
add qubits-truncate optimization

### LINKED ISSUE #147: idle qubits are not optimized away
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

The simulation of noise becomes extremely slow for a 20-qubit circuit, even though only 3 qubits are active.

Edit: the 3 qubit one is equally slow. That's probably a separate bug.

Aer should be more efficient in simulating sparse circuits. With some recent changes, Terra now embeds a small circuit on the full device and returns a circuit the size of that device. So this is a real example I got trying to simulate the behavior of a small circuit on Tokyo.

```python
qasm = """OPENQASM 2.0;
include "qelib1.inc";
qreg q[20];
creg c[3];
u3(1.10207363987740,3.38905574496354,-2.33347511214731) q[0];
u3(0.208506566967889,0.259713258155296,-2.12932391686369) q[1];
u3(0.199241978886621,-2.74203199956075,2.25788452327037) q[5];
cx q[0],q[5];
cx q[5],q[0];
cx q[0],q[5];
cx q[1],q[0];
u1(3.27262754774519) q[0];
u3(-0.811363284435596,0.0,0.0) q[1];
cx q[0],q[1];
u3(1.48572479330530,0.0,0.0) q[1];
cx q[1],q[0];
u3(2.51699832056813,-2.73058127778721,1.58262917334812) q[0];
cx q[0],q[5];
u3(0.989024262401792,4.42691825885413,-4.57369861647542) q[1];
cx q[5],q[0];
cx q[0],q[5];
cx q[1],q[0];
u1(2.46289370014954) q[0];
u3(-1.66213330644361,0.0,0.0) q[1];
cx q[0],q[1];
u3(3.09782206744186,0.0,0.0) q[1];
cx q[1],q[0];
u3(2.23147548122233,-0.350255348380724,0.602491869202778) q[0];
cx q[0],q[5];
u3(0.270435327116765,0.933056308516186,0.128753386588826) q[1];
cx q[5],q[0];
cx q[0],q[5];
cx q[0],q[1];
u3(-1.61740988418189,0.0,0.0) q[0];
u1(0.762771715778533) q[1];
cx q[1],q[0];
u3(3.07583422211664,0.0,0.0) q[0];
cx q[0],q[1];
u3(1.70456258257962,-0.252456527824428,-2.36180183089211) q[0];
u3(0.538560668214003,0.930143172545291,1.93761476960866) q[1];
barrier q[1],q[0],q[5];
measure q[0] -> c[0];
measure q[5] -> c[1];
measure q[1] -> c[2];
"""

circ = QuantumCircuit.from_qasm_str(qasm)
circ.draw(output='mpl', style={'fold': 100})

d = {0: 0, 1: 1, 5: 2}
small_q = QuantumRegister(3)
small_circ = QuantumCircuit(small_q, circ.cregs[0])
for gate, qargs, cargs in circ:
    new_qargs = list(map(lambda x: (small_q, d[x[1]]), qargs))
    small_circ.append(gate, new_qargs, cargs)
small_circ.draw(output='mpl', style={'fold': 100})

from qiskit import IBMQ, execute
from qiskit.providers.aer.noise.device import basic_device_noise_model

IBMQ.load_accounts()
devicename = 'ibmq_poughkeepsie'
device = IBMQ.get_backend(name=devicename)

qasm_sim = Aer.get_backend('qasm_simulator')
```

```python
%%time

# 20 q noisy simulation, only 3 q active -- time: 65 s
noisy_counts = execute(small_circ, qasm_sim, shots=100,
                       noise_model=basic_device_noise_model(device.properties()),
                       coupling_map=device.configuration().coupling_map,
                       basis_gates=device.configuration().basis_gates).result().get_counts()
```

```python
%%time

# same circuit, remove idle wires -- time: 65 s
noisy_counts = execute(small_circ, qasm_sim, shots=100,
                       noise_model=basic_device_noise_model(device.properties()),
                       coupling_map=device.configuration().coupling_map,
                       basis_gates=device.configuration().basis_gates).result().get_counts()
```

```python
%%time 

# no noise -- time: 0.3 s
noisy_counts = execute(circ, qasm_sim, shots=100,
                       coupling_map=device.configuration().coupling_map,
                       basis_gates=device.configuration().basis_gates).result().get_counts()
```

hhorii:
### Summary
This PR truncates qubits in statevector to resolve #147 

### Details and comments
If some qubits are not used, qasm simulator does not use them in its qubitvector with this change.

- [x] Implement a new circuit optimization
- [x] Implement test cases


chriseclectic:
@hhorii why did you move the optimization passes out of the controller? It makes the most sense to me there as they are roughly equivalent to transpilation passes in terra and should be able to be applied to circuits regardless of which state they will be executed on (though the state they are intended to be being executed on can obviously effect the optimizations). I think how you had it in your last PR was fine.

With this PR I would say rather than check every op type for remapping you should only specify the op types that need a special case for remapping (like snapshots, multiplexer, multimatrix). For every other instruction all you need to do is remap their qubits parameters and as long as you don't change the relative order of the lists it will be consistent.

Multiplexer and multimatrix are special cases becuase you also need to remap the `op.regs` fields, and snapshot is a special base where you can only remap some snapshots (the ones that use qubit arguements like probabilities and expectation values), but not others (like statevector, stabilizer). To make it a bit simpler I would say to just not allow remapping for all snapshots for this first implementation.

hhorii:
OK. I used think that this optimization should be applied only to `Statevector::State`. Checking a `State` type in `AER::CircuitOptimization`  is not a good way, so I moved optimization list from `Controller` to `State`. Because this my assumption is wrong,  I will roll this back.

chriseclectic:
It will be useful to apply for any state type used in QasmSimulator (like extended stabilizer, stabilizer, and especially density matrix in the future). The point of it is that (except for snapshots) the QasmSimulator only uses the state type to simulate outcomes of measurements, and those measurement outcomes don't change by this remapping of qubits.

ishapoval:
@hhorii The truncation of idle qubits works as expected if I use the `execute` wrapper, but if I go for `transpile`/`assemble`/`run` chain with seemingly all other parameters being equal I get a bunch of empty results 

> {'data': {},
              'meas_level': <MeasLevel.CLASSIFIED: 2>,
              'seed_simulator': 0,
              'shots': 0,
              'status': 'EMPTY',
              'success': False,
              'time_taken': 0.0}

from the QasmSimulator with the overall status of 

>  'ERROR: QasmSimulator: Insufficient memory for 65-qubit circuit using "statevector" method. You could try using the "matrix_product_state" or "extended_stabilizer" method instead. '

Do I have to enable fusion explicitly (if truncation is part of it) in the `transpile`/`assemble`/`run` chain or it's just that construction of noisy backend with `QasmSimulator.from_backend(device)' turns a few-qubit circuit into a 65-qubit simulation and there is just nothing to truncate?

PR #168, opened 2019-04-23
Build wheels automatically in CI environment

mtreinish:
### Summary

Right now when it's time to push a relase of qiskit-aer we have to
manually build wheels for all our supported environments because we
included compiled code in the project. This involves setting up a proper
environment with portable librariers and compilers in Linux, Windows,
and Mac OSX with all 4 versions of python we support. Then manually
building the 12 wheels necessary to publish on pypi. There is no reason
this manual step needs to exist and we can just leverage CI to do our
wheel building when it's time to do a release.

This commit adds initial support for building our wheels in travis
(for linux and osx) and azure pipelines (for windows). It leverages the
cibuildwheel project to automate the wheel building on linux and osx. This
will ensure we setup the proper environment for all 4 python versions on
both environments (for example pulling in the manylinux docker container on
linux). For windows it leverages conda to setup a fresh environment for each
python version instead of using cibuildwheel because of specific configuration
needed there.

### Details and comments

TODO:

- [x] Add windows builds with azure-pipelines
- [x] Debug wheel building process until it succeeds
- [x] Uncomment tag conditions and twine upload to pypi.org


mtreinish:
The build process for linux (and I assume osx and windows once I start debugging that) is blocked by #172 because cibuildwheel uses venvs to created isolated python environments. If we can't build the sdist into a wheel inside a venv than it's going to fail every time.

mtreinish:
Well 1/3 of the way there. The osx wheels built successfully: https://travis-ci.com/Qiskit/qiskit-aer/jobs/229330750

chriseclectic:
@mtreinish are you still working on this or should I close this PR?

mtreinish:
@chriseclectic I am still working on it, higher priority work has come up since my last update here. I'm planning to circle back to it eventually.

mtreinish:
So the current state of this is that windows builds are failing on `find_blas`. Even after manually installing openblas from source and also installing a copy via anaconda. The linux builds are failing because it's trying to link against libpython which is explicitly prohibited by the manylinux 2010 (and manylinux1) specifications. Details on that can be found here: https://www.python.org/dev/peps/pep-0513/#libpythonx-y-so-1 and https://www.python.org/dev/peps/pep-0571/#the-manylinux2010-policy

It looks like to move forward on here we'll have to change the linux builds to not link against libpython and for windows figure out why `find_blas` can't find either openblas installation.

atilag:
Ok, now that we've merged pybin11 support and that we took advantage of it to change the way we were linking the extensions, we shall have no problems with libpython anymore. I'll resume this PR asap and fix the remaining issues.

mtreinish:
Finally, the linux builds are working now. OSX is working as well except for python 3.8:

https://travis-ci.com/Qiskit/qiskit-aer/jobs/273238792#L1023

which looks like it's related to what was fixed in: https://github.com/Qiskit/qiskit-terra/commit/af390a89833c67d4ffd08b2034d82a37540ab8ae

Windows also still needs work it's not finding openblas right now:

https://dev.azure.com/qiskit-ci/qiskit-aer/_build/results?buildId=6841&view=logs&j=88076fea-ae17-5b87-3f51-0f9c840a5695&t=4e9c5838-9918-5d21-d580-b0d6b55ac95f&l=242

mtreinish:
After merging the openpulse branch onto master things are failing on linux and windows now.

@atilag it looks like the linux failure is because things are attempting to link against libpython again: https://travis-ci.com/Qiskit/qiskit-aer/jobs/280091283#L615 we need to fix this.

atilag:
Yep, Cython code from Open Pulse sim branch was still being linked against libpython, but I have already fix it on `master` after merging #524. That shouldn't be a problem now.

mtreinish:
Ok, everything here is finally working now that #550 merged. All the wheel builds succeeded on https://github.com/Qiskit/qiskit-aer/pull/168/commits/ea96598ae39c1931a1770f2b2efa939069603ade so this is finally ready to go. I've removed the WIP from the PR title and made all the jobs only run on tags/releases. After this merges I will go into the travis and azure UI and set the password for twine as a secret so that for releases CI will automatically build and upload wheels.

If anyone is curious azure enables us to upload artifacts, the windows wheels that were built can be found here: https://dev.azure.com/qiskit-ci/qiskit-aer/_build/results?buildId=7712&view=artifacts&type=publishedArtifacts (it's the zip file labeled `drop`)

.travis.yml
```diff
@@ -106,6 +106,7 @@ stage_osx: &stage_osx
 
 # Define the order of the stages.
 stages:
+  - wheels
```
chriseclectic: Should this stage be called `deploy` and come last in the chain?

mtreinish: I did it first because it only gets triggered on tag events. We've already run the tests at least once (for the PR) and more realistically twice (for the merge event) for the state of master at that point. So having the travis build wait for the tests to run again was an unnecessary delay especially when debugging any failures during a release where we want the fastest turnaround. Ideally I would skip the test and compile stages on tag events (that's what I did on azure for windows) but I didn't know if there was a .travis.yml syntax for conditions on stages off the top of my head.

As for the name of the stage I'm fine with renaming it if you feel strongly about it.

chriseclectic: I see, I think you can skip the other stages then by having `if: tag IS blank AND (<any other conds>)` according to [travis docs](https://docs.travis-ci.com/user/conditions-v1).

chriseclectic: I would still call it deploy though if its deploying the wheels ( `deploy` seems to be the standard tag used in the Travis CI documentation)

mtreinish: ok, updated the log and renamed it to deploy in: https://github.com/Qiskit/qiskit-aer/pull/168/commits/f8b75c1d2acf63cee7e82eea8da95ba3b36507e7

PR #177, opened 2019-04-26
Improve benchmarking

atilag:
### Summary
* Add support for Linux and Mac
* Cleaner reports
* Add documentation

### Details and comments
Even though benchmarks are already working for MacOS platform, I'm going to keep this PR in WIP until I make sure every Linux/MacOS config works as expected. Also, the documentations is still on draft and need to be polished.


atilag:
Ok, I just tested that everything works on Linux and MacOS with all the supported Pyhton versions: 3.5, 3.6 and 3.7.

PR #181, opened 2019-04-29
reduced_basic_noise_model

nonhermitian:
### Summary
Converts the basic noise model for a full device and maps it to a reduced coupling_map that specifies a subgraph of the device.  This allows for doing noise simulations on sub-graphs of devices without needing to pad the size of a circuit to the full width of the device.

This is meant to be used in conjunction with https://github.com/Qiskit/qiskit-terra/pull/2225

### DIFF
--- a/qiskit/providers/aer/noise/device/__init__.py
+++ b/qiskit/providers/aer/noise/device/__init__.py
@@ -7,7 +7,7 @@
 
 """Device noise models module Qiskit Aer."""
 
-from .models import basic_device_noise_model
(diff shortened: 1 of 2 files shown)


nonhermitian:
This is failing for reasons that have nothing to do with my PR.

chriseclectic:
I think this would be better as an optional kwarg argument of `basis_device_noise_model` instead of a seperate function.

nonhermitian:
Ideally it would work on any NoiseModel, but they do not include enough info to work on their own.

nonhermitian:
This is superseded by #190

PR #182, opened 2019-04-29
Fix type on loop iterators in openmp for loops

mtreinish:
### Summary

Right now in the statevector simulator there were a few for loops that
were using size_t as the loop iterator type. However the type there
should int_t so it's signed. This commit just fixes this issue and
switches the type.

### Details and comments


atilag:
Tests are failiing because of a triage with Terra, I'm merging these changes.

PR #187, opened 2019-04-29
Forcing OSX deployment target to 10.9 via setup.py

atilag:
### Summary
With this change, we can build from sdist (pip install --no-binary) on MacOS

### Details and comments
Build prerequisites need to be installed in the system though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,8 +56,6 @@ stage_osx: &stage_osx
     pip: true
     directories:
       - ~/python-interpreters/
-  env:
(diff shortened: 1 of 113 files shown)


chriseclectic:
Should you update the osx travis stage to check that you no longer need `--plat-name=macosx-10.9-x86_64`?

chriseclectic:
It looks like tests are failing because `assemble` was recently removed from top-level qiskit import. 

If you edit `test_statevector_simulator.py` so that it has `from qiskit.compiler import assemble` the tests should pass

PR #208, opened 2019-05-10
Identity is not the same thing as equality in Python

cclauss:
### Summary

Identity is not the same thing as equality in Python.  We want the latter in these instances.

### Details and comments

Use ==/!= to compare str, bytes, and int literals.

$ __python__
```python
>>> id = "i"
>>> id += 'd'
>>> id
'id'
>>> id == 'id'
True
>>> id is 'id'
False
>>> 0 == 0.0
True
>>> 0 is 0.0
False
```


atilag:
You are totally right @cclauss 
Thanks for the contribution!

atilag:
Tests are failiing due to Terra upgrade, everything seems correct in you side though.

PR #214, opened 2019-05-16
Qiskit Aer Parallel: support for multi-process and multi-GPU

doichanj:
I added multi-process (MPI and multi-GPU (CUDA) support codes.
I added 2 subclasses in;
src/simulators/qasm/qasm_controller_par.hpp
src/simulators/statevector/qubitvector_par.hpp
and 1 main program in;
contrib/standalone/qasm_controller_par.cpp
for this implementation.

I also added specific codes for the parallelization in 
src/simulators/statevector/qsim_par

I modified CMakeLists.txt to select this implementation when using cmake, and how to build is written in README_parallel.md .


atilag:
These are 8.5K lines of code to review.
It's going to take A LOT time to review it.
As this is a major change, I'd have loved to have a previous discussion in an issue so we could saw things like general architecture, design, GPU targets, or even the necessity of introducing MPI (as well as the benefits it will bring).
Do you think is feasible to split this PR in at least two? One for the GPU implementation and another one for the MPI one.
Thanks!

doichanj:
MPI version has only about 100 lines of additional codes from non-MPI version, so I think it is very small change.

doichanj:
I temporary delete MPI implementations, now it only supports GPUs.
And I also deleted 2 derived classes and now there is only one derived class (QubitVectorPar) from the original version. 
I modified CMakeList.txt to detect if there is CUDA installed or not and if there is, it automatically build GPU version of state vector simulator.

chriseclectic:
Superseded by #363

PR #219, opened 2019-05-31
Enable python linter

atilag:
### Summary
I have enabled the Python linter and pep8 code style checker in Travis, so every commit/PR will verify that the code conforms with the rules.
There is some refactoring as a consequence of removing linter Warnings.
I'm leaving tests and benchmarking out of the linter for now, but I'll add them asap.

@chriseclectic I left some docstring comments with the tag: TODO, Could you take care of them and replace them with real descriptions?


qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -27,7 +28,7 @@
 
 
 class QasmSimulator(AerBackend):
-    """Aer quantum circuit simulator
+    """Aer quantumç circuit simulator
```
chriseclectic: Typo: quantumç

qiskit/providers/aer/noise/errors/readout_error.py
```diff
@@ -243,7 +244,7 @@ def _check_probabilities(probabilities, threshold):
             if abs(sum(arr) - 1) > threshold:
                 raise NoiseError("Invalid probabilities: sum({})= {} "
                                  "is not 1.".format(vec, sum(arr)))
-            if len(arr[arr < 0]) > 0:
+            if arr[arr < 0]:
```
chriseclectic: Numpy arrays can't be checked for non-empty like lists, this will give a ValueError if `arr[arr > 0]` is non-empty.

It should be `if arr[arr < 0].size:` or  `if arr[arr < 0].size > 0:`

A couple of lines above could also be changed to `len(arr)` -> `arr.size` if you don't want to use `len`

qiskit/providers/aer/noise/errors/standard_errors.py
```diff
@@ -231,6 +232,7 @@ def single_pauli(pauli):
             return {'name': 'y'}
         if pauli == 'Z':
             return {'name': 'z'}
+        return None
```
chriseclectic: This should match the `single_pauli` function in line 175 which has `raise NoiseError("Invalid Pauli string.")` if it reaches the end rather than return None.

qiskit/providers/aer/noise/errors/standard_errors.py
```diff
@@ -191,7 +191,7 @@ def single_pauli(pauli):
             # Pauli strings go from qubit-0 on left to qubit-N on right
             # but pauli ops are tensor product of qubit-N on left to qubit-0 on right
             # We also drop identity operators to reduce dimension of matrix multiplication
-            mat = 1
+            mat = np.identity(1)
```
chriseclectic: This would be better as `mat = None` and then later checking if `if mat is None` for the identity case.

chriseclectic: Actually this is wrong. I forgot that the init `mat` was used in `np.kron` below

PR #222, opened 2019-06-04
enable sampling for readout error

hhorii:
### Summary
This PR adds support of sampling for read-out errors in qasm simulator.

### Details and comments
Currently, qasm simulator enables sampling if
- `noise_model.ideal()` returns `true`, and
- all the measure operations are the bottom in a circuit

This PR will change as follows:
- `noise_model.ideal()` or `noise_model.only_readout_errors()` returns `true`, and 
- all the measure and readout-error operations are the bottom in a circuit.

- [x] Implement this feature
- [x] Add tests


chriseclectic:
I think it would be good to have the noise model objects to have methods that return True/False to check for readout errors, local quantum errrors, and non-local quantum errors (this will also be useful for optimisation passes later)

* `noise_model.ideal()`: returns True if there are no quantum errors nor readout errors, otherwise False.
* `noise_model.has_readout_errors()`: returns True if there are readout errors, or false if not (regardless of quantum errors present).
* `noise_model.has_quantum_errors()`: returns True if there are either local or non-local quantum errors (regardless of readout errors).
* `noise_model.has_local_quantum_errors()`: returns True if there are local quantum errors, or false if not (regardless of non-local quantum errors or readout errors present).
* `noise_model.has_nonlocal_quantum_errors()`: returns True if there are non-local quantum errors, or False if not (regardless of local quantum errors or readout errors present).

Then in this PR you could replace for example `noise_model_.ideal() || noise_model_.only_readout_errors()` with `noise_model_.has_quantum_errors() == false`

PR #223, opened 2019-06-04
Noise transformation 2qubit

gadial:
### Summary
Extending the noise transformation module to support approximations of 2-qubit noises.

### Details and comments
The module now throws exception only for 3-qubit or more errors; additionaly, support for premade 2-qubit Pauli and reset operators was added.

### DIFF
--- a/qiskit/providers/aer/noise/utils/noise_transformation.py
+++ b/qiskit/providers/aer/noise/utils/noise_transformation.py
@@ -241,17 +241,38 @@ def pauli_operators():
     }
     return [pauli_1_qubit, pauli_2_qubit]
 
+def reset_operators():
+    reset_0_to_0 = [{'name': 'reset', 'qubits': [0]}]
+    reset_0_to_1 = [{'name': 'reset', 'qubits': [0]}, {'name': 'x', 'qubits': [0]}]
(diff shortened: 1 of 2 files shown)


qiskit/providers/aer/noise/utils/noise_transformation.py
```diff
@@ -212,24 +216,65 @@ def approximate_noise_model(model, *,
...
     return approx_noise_model
 
+def pauli_operators():
+    pauli_1_qubit = {
+                    'X': [{'name': 'x', 'qubits': [0]}],
+                    'Y': [{'name': 'y', 'qubits': [0]}],
+                    'Z': [{'name': 'z', 'qubits': [0]}]
+                    }
+    pauli_2_qubit = {
+        'XI': [{'name': 'x', 'qubits': [0]}, {'name': 'id', 'qubits': [1]}],
```
chriseclectic: Qubit 0 and 1 need to be swapped here. The convention used in other functions is eg:

```
pauli_error([('XY', 1.0)]) --> QuantumError([([{'name': 'y', 'qubits': [0]}, {'name': 'x', 'qubits': [1]}], 1.0)])
```

PR #230, opened 2019-06-10
Debug thread management for shot-level and add explicit parallelization

hhorii:
### Summary
`max_parallel_shots` does not work correctly. Also, explicit parallelization is necessary.

### Details and comments
`parallel_shots` must not exceed `max_parallel_shots` in the following circuit.

```
from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit
from qiskit.compiler import assemble, transpile
from qiskit.providers.aer.noise.errors.standard_errors import pauli_error

qasm_simulator = QasmSimulator()

num_of_qubits = 10

qr = QuantumRegister(num_of_qubits)
cr = ClassicalRegister(num_of_qubits)

circuit = QuantumCircuit(qr, cr)
for i in range(num_of_qubits):
    circuit.u3(1.1, 2.2, 3.3, qr[i])
for i in range(num_of_qubits):
    circuit.barrier(qr[i])
for i in range(num_of_qubits):
    circuit.measure(qr[i], cr[i])

noise_model = NoiseModel()
error = pauli_error([('X', 0.25), ('I', 0.75)])
noise_model.add_all_qubit_quantum_error(error, 'reset')

backend_opts={}
backend_opts['max_parallel_shots'] = 4
backend_opts['noise_model'] = noise_model

qobj = assemble([circuit], qasm_simulator, shots=16)
result = qasm_simulator.run(qobj, backend_options=backend_opts).result()
print("parallel_shots: " + str(result.to_dict()['results'][0]['metadata']['parallel_shots']))
```
Currently, the above max_parallel_shots is ignored. This change will fix this issue.

In addition, this PR adds a feature to explicitly configure parallelization by specifying `parallel_experiments`, `parallel_shots`, or `parallel_state_update`. If one of them is configured, the others are set to 1 if they are not configured.


atilag:
Can we add a test for this?

chriseclectic:
Can you change it to only load from a user config the QasmSimulator config parameters:

* "max_parallel_threads"
* "max_parallel_experiments"
* "max_parallel_shots"

and make the others just used internally. 

The idea is that by calling it max if I set max threads to 100, but my system only has 4, it would use the system max (with this configs the max threads state doesnt have a user config setting because it is inferred based on the total max threads and how many are left over after exp and shot parallelization).

src/base/controller.hpp
```diff
@@ -106,6 +106,12 @@ namespace Base {
  * - "max_memory_mb" (int): Sets the maximum size of memory for a store.
  *      If a state needs more, an error is thrown. If set to 0, the maximum
  *      will be automatically set to the system memory size [Default: 0].
+ * - "parallel_experiments" (int): Explicitly set number of circuits that
+ *      are executed in parallel [Default: -1].
```
chriseclectic: The default value for automatic in the `max_parallel_*` is currently 0 instead of -1. Can you change them so that this is consistent for the different parallel variables (you'll need to update the doc string in the python backend files too).

hhorii: I changed it to 0.

Issue #211, opened 2019-05-13
Idle qubit schedule utility function


chriseclectic:
### What is the expected behavior?

I would like to see a function in the `noise.utils` folder that can be used to convert a circuit to a gate level "schedule" based on which gates could (in principle) be done in parallel, and insert identities on idle qubits during each time step. This is an important tool for doing more accurate noise simulations of T1 relaxation since an error can be added to the idle qubit identities to simulate relaxation.

This insertion should respect barriers (so gates with a barrier between them will always be in a seperate time step), and insert different identities for the different length gates. I think initially having just two identity types: one for CX gates and one for single qubit (u3) gates is a good start. This could be implemented by using the label option for the identity gates so we have "id_cx" and "id_u3" as their labels for reference in the noise model.

yirkajk:
I copied the new idle scheduler function from #237, and it's working well for my purposes. But, the different labels for the id gates were difficult to work with for two reasons. 
- First, there is almost no mention in the tutorials and documentation (beyond the source code) that `noise_model` depends on gate labels, not gate names. Only the [Custom Gate Noise Tutorial](https://github.com/Qiskit/qiskit-tutorials/blob/master/qiskit/aer/custom_gate_noise.ipynb) mentions it, so someone working with id gates can easily miss it. (Note that this idle utility could be one of the first experiences a user has with automatically generated labels.) Right now, it's non-obvious that the gates added by this utility will not respond to noise on generic "id" gates.
- Second, because of the many different labels, it's cumbersome to apply noise to all of the gates.

I would suggest making noise depend on names instead of labels, but the discussion in #207 (continuing in Qiskit/qiskit-terra#2393) suggests idle gates are an area where labelling is useful. So, I suggest the noise documentation and/or the idle scheduler documentation needs to be updated in order for the idle scheduler to be usable. 

Second, perhaps an argument could be added to the scheduler function which allows the user to provide a default label?

gadial:
Thanks for the helpful feedback! It seems to me the best course of action to add a parameter for choosing the label to use (can be a string or a dict from gates to strings) and improving the documentation.

gadial:
I've added the parameter in b238d15f33b377c21032978f38f0a95a80450a20.

atilag:
I think it's safe to close this issue, it's been a while without activity, also last comment suggest this was already tackled?

chriseclectic:
Equivalent functionality will be added soon once we finish scheduled circuits in Terra

PR #237, opened 2019-06-13
Idle scheduler

gadial:
### Summary
This modules addressed issue #211 and adds a utility to add idle gates to better
simulate passing time when gates for other qubits are being applied.

### DIFF
--- a/qiskit/providers/aer/noise/utils/idle_scheduler.py
+++ b/qiskit/providers/aer/noise/utils/idle_scheduler.py
@@ -70,6 +70,7 @@ class IdleScheduler():
         # convert to circuit and back as hack to prevent nondeterminism in DAGCircuit
         layers = [circuit_to_dag(dag_to_circuit(l['graph'])) for l in dag.layers()]
         for layer in layers:
+            # print("idle times before node {}: {}".format(node.name, self.idle_times.values()))
             new_layer_graph = self.add_identities_to_layer(layer)
             new_dag.extend_back(new_layer_graph)
         return dag_to_circuit(new_dag)
@@ -89,7 +90,7 @@ class IdleScheduler():
 
         for node in layer.op_nodes():
             for qubit in node.qargs:
-                if node.op.name == 'barrier':  # special case
+                if node.op.name in ['barrier', 'snapshot']:  # special cases
                     self.idle_times[qubit] = 0
                 else:
                     self.idle_times[qubit] -= self.op_times.get(node.name, self.default_op_time)
--- a/qiskit/providers/aer/noise/utils/idle_scheduler.py
+++ b/qiskit/providers/aer/noise/utils/idle_scheduler.py
@@ -22,7 +22,7 @@ from qiskit.converters import dag_to_circuit
 from qiskit.extensions.standard.iden import IdGate
 
 
-def schedule_idle_gates(circuit, op_times=None, default_op_time=1):
+def schedule_idle_gates(circuit, op_times=None, default_op_time=1, labels=None):
     """
     This function gets a circuit and returns a new one with the idle
     gates in place; each idle gate is labeled "id_<name>" where <name> is the name of
@@ -32,23 +32,27 @@ def schedule_idle_gates(circuit, op_times=None, default_op_time=1):
         circuit (QuantumCircuit): The circuit to add idle gates to
         op_times (dictionary): dictionary of times taken by specific gate types
         default_op_time (float or int): default time for gates not found in op_times
+        labels (string or dictionary): the label to give the id gate, either a uniform
+                label for all the id gates (string) or a dictionary mapping from the
+                gate type that caused the delay to the expected id gate label
 
     Returns:
         QuantumCircuit: The new circuit with the added idle gates
 
     """
-    scheduler = IdleScheduler(op_times, default_op_time)
+    scheduler = IdleScheduler(op_times, default_op_time, labels)
     return scheduler.schedule(circuit)
 
 
 class IdleScheduler():
     """Adds idle gates to given circuits"""
-    def __init__(self, op_times, default_op_time):
+    def __init__(self, op_times, default_op_time, id_gate_labels=None):
         if op_times is None:
             self.op_times = {}
         else:
             self.op_times = op_times
         self.default_op_time = default_op_time
+        self.id_gate_labels = id_gate_labels
         self.circuit = None
         self.idle_times = None
 
@@ -93,8 +97,17 @@ class IdleScheduler():
         id_time = self.op_times.get('id', self.default_op_time)
         for qubit in self.circuit.qubits:
             while self.idle_times[qubit] >= id_time:
-                id_gate = IdGate(label="id_{}".format(max_op_name))
+                id_gate = IdGate(label=self.id_gate_label(max_op_name))
                 layer.apply_operation_back(id_gate, [qubit], [])
                 self.idle_times[qubit] -= id_time
 
         return layer
+
+    def id_gate_label(self, op_name):
+        label = "id_{}".format(op_name)
+        if self.id_gate_labels is not None:
+            if isinstance(self.id_gate_labels, str):
+                label = self.id_gate_labels
+            if isinstance(self.id_gate_labels, dict) and op_name in self.id_gate_labels:
+                label = self.id_gate_labels[op_name]
+        return label
--- a/qiskit/providers/aer/noise/utils/idle_scheduler.py
+++ b/qiskit/providers/aer/noise/utils/idle_scheduler.py
@@ -1,0 +1,96 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2019.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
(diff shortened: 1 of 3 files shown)


qiskit/providers/aer/noise/utils/idle_scheduler.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        max_op_time, max_op_name = max(
+            [(self.op_times.get(node.name, self.default_op_time), node.name)
+             for node in layer.op_nodes()])
+        for qubit in self.idle_times.keys():
+            self.idle_times[qubit] += max_op_time
+
+        for node in layer.op_nodes():
+            for qubit in node.qargs:
+                if node.op.name == 'barrier':  # special case
```
ajavadia: i think `snapshot` is also a special case.

qiskit/providers/aer/noise/utils/idle_scheduler.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def add_identities_to_layer(self, layer):
+        """Adds the idle gates to a specific layer in the circuit
+            Args:
+               layer (dict): The layer to modify (contains a circuit graph)
+            Returns:
+               dict: The modified layer
+        """
+        max_op_time, max_op_name = max(
+            [(self.op_times.get(node.name, self.default_op_time), node.name)
```
ajavadia: The `op_times` is a dict that is keyed on gate names. Therefore it prohibits things like "u3 on qubit 0 is 200 ns and u3 on qubit 1 is 300 ns".
Is there a clean way to make this more general? How does Aer currently handle the difference in latency on gates of the same kind?

test/terra/noise/test_idle_scheduler.py
```diff
@@ -0,0 +1,168 @@
...
+"""
+
+from qiskit import QuantumRegister, QuantumCircuit
+from qiskit.providers.aer.noise.utils import schedule_idle_gates
+from qiskit.converters import circuit_to_dag
+import unittest
+
+class TestIdleScheduler(unittest.TestCase):
+    def assertCircuitsEqual(self, lhs, rhs):
+        self.assertEqual(circuit_to_dag(lhs), circuit_to_dag(rhs))
```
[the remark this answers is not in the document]
gadial: Fixed, thanks. I don't understand why it didn't work for me before.

qiskit/providers/aer/noise/utils/idle_scheduler.py
gadial: Should `snapshot` affect timing?

qiskit/providers/aer/noise/utils/idle_scheduler.py
gadial: This should not be too difficult; key the times using a tuple `(node_name, qubit_num)`. The question is - do we need this right now (if so, I'll implement).

test/terra/noise/test_idle_scheduler.py
```diff
@@ -0,0 +1,168 @@
...
+        circuit.x(qr[0])
+        circuit.y(qr[2])
+
+        target_circuit = QuantumCircuit(qr)
+        target_circuit.x(qr[0])
+        target_circuit.iden(qr[1])
+        target_circuit.y(qr[2])
+
+        circuit1 = schedule_idle_gates(circuit, op_times={'id': 2, 'x': 2})
+        circuit2 = schedule_idle_gates(circuit, op_times={'id': 2, 'x': 2})
```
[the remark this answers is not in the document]
gadial: Bad case of copy-paste work, fixed.

qiskit/providers/aer/noise/utils/idle_scheduler.py
ajavadia: yes snapshots are like barriers. you insert them at a specific layer of the circuit, and compiler optimizations should not alter that layer otherwise you will get an unexpected snapshot.

qiskit/providers/aer/noise/utils/idle_scheduler.py
ajavadia: Take a look at gate times for some devices to see if different physical U or CX gates have different latencies. If so, then I'd say this should match that.

chriseclectic: I would try and make this mimic the same format used for specifying custom gate times in the device noise model, which is a list with entries `(name, qubits, time)`. If qubits is `None` it applies to that gate on all qubits, otherwise it applies only to those qubits.

Usually devices will have the same gate time on all qubits for all single-qubit gates, but different `cx` gate times on different pairs of qubits. In the device noise model this could be entered like

```python
gate_times = [('u1', None, 0), ('u2', None, 100), ('u3', None, 200),
    ('cx', [0, 2], 678), ('cx', [1, 2], 547), ('cx', [2, 3], 721)]
```

qiskit/providers/aer/noise/utils/idle_scheduler.py
gadial: Addressed in 4a3327c74d1cdbbe18d4149fc1c21ace18481e7b

gadial:
I am not supporting partial barriers or snapshots (i.e. not acting on all qubits) since I'm not sure what is the expected behavior in this case. Do you have an example in mind?

qiskit/providers/aer/noise/utils/idle_scheduler.py
gadial: Chris, your example raises two separate questions.

1. What to do in case of conflicts, i.e. `gate_times=[('u1', None, 0), ('u1', [1], 17)]`
2. How to handle gates like `u1`: apply only to `u1` gates, or for any gate defined via such gates? How can this be checked in Aer/Terra?

gadial: I introduced the new interface at 89d1d8f30fe0c033d842b047134b0ebb588bb8c3.

Note that for cx gates, different **order** of qubits results in "different gate", i.e. `('cx', [0, 2], 678), ('cx', [2,0], 356)` is acceptable, and `('cx', [0, 2], 678)` will give the default (1) time to `('cx', [2,0])`.

chriseclectic: You could do one of two options: 

1. `("u1", None, 0)` would be a default for all qubits not otherwise specified, and `("u"`, [1], 17)` would override the default for qubit-0. 
2. Raise an error if there a situation like that.

I think the first case is more useful. I'm not sure what device noise model does in this situation, but we should make them match. And for cx gates you are right that different order of qubits should be considered distinct gates.

gadial: Glad to say option 1 is already what the code does, so I'm resolving this.

qiskit/providers/aer/noise/utils/idle_scheduler.py
```diff
@@ -0,0 +1,158 @@
...
+
+def schedule_idle_gates(circuit, op_times=None, default_op_time=1, labels=None):
+    """
+    This function gets a circuit and returns a new one with the idle
+    gates in place; each idle gate is labeled "id_<name>" where <name> is the name of
+    the slowest gate that runs in parallel
+
+    Args:
+        circuit (QuantumCircuit): The circuit to add idle gates to
+        op_times (dictionary): dictionary of times taken by specific gate types
```
chriseclectic: Can this be changed to `gate_times (list)` to match `basic_device_noise_model` argument. And if in a layer all gates have time set to 0 no id should be inserted

qiskit/providers/aer/noise/utils/idle_scheduler.py
```diff
@@ -0,0 +1,158 @@
...
+        default_op_time (float or int): default time for gates not found in op_times
```
chriseclectic: Can this be changed to `default_gate_time`. It looks like the function doesn't terminate if this is set to 0. I would expect setting a `default_gate_time=0` to not insert identities for any layers containing only gates *not* specified in the `gate_times` list.

gadial:
Hi @chriseclectic. I have to admit I did not understand the expected behavior for `op_times=[('cx', None, 400), ('u3', None, 100), ('u2', None, 50), ('u1', None, 0)]` and

```
qc = QuantumCircuit(3)
qc.u2(0, np.pi, 0)
qc.cx(0, 1)
qc.cx(1, 2)
```

Your suggestion is to add 1 `id` gate per "free" qubit in each layer. But the layers represent different runtimes; the `id` gate in the first layer represents a 50-nanoseconds wait, while in the other layers it is a 400-nanoseconds wait. So a different number of id gates are required.

Since we use different labels for `id` gates, use of multiple gates can be somewhat avoided. `id_u2` can represent 50-nanoseconds wait and `id_cx` can represent 400-nanoseconds wait. But this by itself is not sufficient. Let's consider the following case:


`op_times=[('cx', None, 400), ('u2', None, 150)]`

with the circuit

```
qc = QuantumCircuit(3)
qc.cx(0, 1)
qc.u2(2, np.pi, 0)
```

Here, qubit 2 is occupied by the `u2` gate for 150 nanoseconds, and than has to wait 250 nanoseconds for the `cx` gate on qubits 0,1 to finish. Which `id` gate to use here? Using `id_cx` represents a wait time that is too large (400) and using `id_u2` represents a time that is too small (150). Using `id_u2` **twice** results in a wait time that is too large (300). 

So we need some better way to specify id gates wait time. For now, if the `id` gate is not explicitly mentioned in `op_times`, its runtime is considered to be "1", resulting in hundreds of gates in your example.

PR #243, opened 2019-06-20
Fix depolarizing error and basic device noise model

### LINKED ISSUE #241: double check equations for obtaining basic_device_noise_model
Currently when getting a basic device noise model, any cx_error above 0.75 causes error, and when considering thermal relaxation, any cx_error above 0.5 causes error.

Is this correct?

Motivation:

There is currently an error with modeling noise from some devices, as they report non-calibrated gates as 100% error. We need to find out what should be reported, but first check to make sure we do the correct math for getting the depolarizing error probability etc.

To reproduce the current error:

```python
from qiskit import IBMQ
from qiskit.providers.aer.noise.device import basic_device_noise_model

IBMQ.load_accounts()
device = IBMQ.get_backend('ibmq_poughkeepsie')
properties = device.properties()
noise_model = basic_device_noise_model(properties, thermal_relaxation=False)
```

chriseclectic:
### Summary

1. Change `depolarizing_error` parameter to allow values greater than 1 if channel is still CPTP
2. Fix basic_device_noise_model to use correct depol param calculation for n-qubit gate with n>1.
3. Truncate depolarizing error param calculated in basic_device_noise_model to max allowed if gate error is too large (Fixes #241)

#### Depolarizing Error

The error parameter in depolarizing error was interpreted as a probability with maximum value of 1 corresponding to a completely depolarizing channel. This allows larger values than 1 (so no-longer a probability) which return error channels with a reduced identity component, up to a maximum where there is no identity component. Ie for single-qubit the Pauli error `[(X, 1/3), (Y, 1/3), (Z, 1/3)]`.

#### Device noise model
In `basic_device_noise_model` when `thermal_relaxation=False` the depolarizing error probability was being calculated using the single-qubit depolarizing channel fidelity in all cases rather than the n-qubit fidelity for an n-qubit gate error. This fixes it to use the correct n-qubit equation.

Also if the gate error is larger than the maximum for the depolarizing error it will simply truncate it to the maximum allowed error rate for the model rather than throw an exception that the resulting error objects are non-CPTP.

### DIFF
--- a/qiskit/providers/aer/noise/device/models.py
+++ b/qiskit/providers/aer/noise/device/models.py
@@ -266,25 +266,34 @@ def _device_depolarizing_error(qubits,
 
     error = None
     num_qubits = len(qubits)
+    dim = 2 ** num_qubits
+
+    # If the device reports an error_param greater than the maximum
+    # allowed for a depolarzing error model:
+    #   error_param <= dim / (dim + 1)
+    # we truncated the depolarizing error probability to the maximum value
+    if error_param is not None:
+        error_param = min(error_param, dim / (dim + 1))
+
     if not thermal_relaxation:
         # Model gate error entirely as depolarizing error
         if error_param is not None and error_param > 0:
             dim = 2 ** num_qubits
-            p_depol = dim * error_param / (dim - 1)
+            depol_param = dim * error_param / (dim - 1)
         else:
-            p_depol = 0
+            depol_param = 0
     else:
         # Model gate error as thermal relaxation and depolarizing
         # error.
         # Get depolarizing probability
         if num_qubits == 1:
             t1, t2, _ = relax_params[qubits[0]]
-            p_depol = _depol_error_value_one_qubit(
+            depol_param = _depol_error_value_one_qubit(
                 error_param, gate_time, t1=t1, t2=t2)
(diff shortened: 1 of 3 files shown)


qiskit/providers/aer/noise/device/models.py
```diff
@@ -265,21 +265,35 @@ def _device_depolarizing_error(qubits,
...
     error = None
+    num_qubits = len(qubits)
+    dim = 2 ** num_qubits
+
+    # If the device reports an error_param greater than the maximum
+    # allowed for a depolarzing error model:
+    #   error_param <= dim / (dim + 1)
+    # we truncated the depolarizing error probability to the maximum value
+    if error_param is not None:
+        error_param = min(error_param, dim / (dim + 1))
```
ajavadia: Should the truncating be done in `basic_device_gate_errors()` rather than `_device_depolarizing_error()`? Because the basic_device_gate_error explicitly assumes a depolarizing channel, however there could be more generic noise channels that result in higher gate errors (not due to depolarizing). It seems like the `_device_depolarizing_error()` is a more generic function, potentially useful in other non-basic noise models?

chriseclectic: Good idea

PR #258, opened 2019-07-02
Add controlled-u1 to basis_gates

chriseclectic:
### Summary

Adds Controlled-U1 (`"cu1"`) to simulator basis gates. This greatly improves performance of ideal circuits with a lot of cu1 gates (like qft circuits).

### Details and comments

* Fixes a bug in previous implementation of `mcu1` in terms of `mcu3` in QubitVector, since the phase convention of cu3 didn't give a proper cu1.
* Adds optimized `apply_mcphase` method to `QubitVector` which is used to implement `u1, cu1, mcu1, z, cz, ccz, mcz, s, std, t, tdg`.
* Adds check for optimized implementation of anti-diagonal single-qubit gates to `apply_matrix` in qubit vector. This should slightly speed up the post-measurement/reset state update, and some Kraus errors (such as amplitude damping).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ Changelog](http://keepachangelog.com/en/1.0.0/).
 
 Added
 -----
+-   Added multi-controlled phase gate to `QubitVector` and changed
+    multi-controlled Z and multi-controlled u1 gates to use this method (\# 258)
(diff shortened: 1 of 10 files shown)


ajavadia:
can you add QFT to the benchmark suite so that the difference pre- and post- this PR is captured?

PR #259, opened 2019-07-02
Set max_workers parameter to ThreadPoolExecutor in AerJob

### LINKED ISSUE #244: Qiskit Aer has a memory leak
<!-- ⚠️ If you do not respect this template, your issue will be closed -->
<!-- ⚠️ Make sure to browse the opened and closed issues -->

### Informations

- **Qiskit Aer version**:
qiskit                    0.10.4                   pypi_0    pypi
qiskit-aer                0.2.1                    pypi_0    pypi
qiskit-aqua               0.5.1                    pypi_0    pypi
qiskit-chemistry          0.5.0                    pypi_0    pypi
qiskit-ibmq-provider      0.2.2                    pypi_0    pypi
qiskit-ignis              0.1.1                    pypi_0    pypi
qiskit-terra              0.8.2                    pypi_0    pypi
- **Python version**:
Python 3.6.6 :: Anaconda, Inc.
- **Operating system**:
CentOS Linux 7 (Core)

Consider the following example code that runs QPE repeatedly (choice of circuit is of no consequence, the same is observable for other circuits, including QAOA):

```python
import numpy as np
from qiskit import Aer, execute, QuantumCircuit, ClassicalRegister, QuantumRegister
     
def qft(circ, q, n):
    """n-qubit QFT on q in circ."""
    for j in range(n):
        for k in range(j):
            circ.cu1(np.pi / float(2**(j - k)), q[j], q[k])
        circ.h(q[j])
     
#Takes in a circuit with an operator on qubit n and appends the qpe circuit
def x_qpe(circ, q, n):
    for i in range(n-1):
        circ.h(q[i])
    for j in range(0, n-1, 2): # Only place a CX^n on every other qubit, because CX^n = I for n even
        circ.cx(q[j], q[n-1])
    circ.barrier()
    qft(circ, q, n-1)
     
# build QPE circuit
n = 20
qr = QuantumRegister(n)
cr = ClassicalRegister(n)
x_qpe_qc = QuantumCircuit(qr, cr)
x_qpe_qc.rx(np.pi/2, qr[n-1])
x_qpe_qc.barrier()
x_qpe(x_qpe_qc, qr, n)
x_qpe_qc.barrier()
for i in range(n-1):
    x_qpe_qc.measure(qr[i], cr[i])
     
     
backend = Aer.get_backend("qasm_simulator")
p = 1
for i in range(1000):
    qobj = execute(x_qpe_qc, backend = backend, shots=1000)
    res = qobj.result()
    print("Finished {}th run".format(i))                                                                                                                                                                                                                                        
```

The code crashes with the following output:
```
Finished 0th run
Finished 1th run
<...>
Finished 198th run
Finished 199th run

libgomp: Thread creation failed: Resource temporarily unavailable
```

Occasionally, I get a different error message for the same crash:
```
OMP: Error #34: System unable to allocate necessary resources for OMP thread:
OMP: System error #11: Resource temporarily unavailable
OMP: Hint Try decreasing the value of OMP_NUM_THREADS.
```

You can confirm that the reason OMP fails to create threads at (roughly) 200th iteration is memory leak by running [memory_profiler](https://pypi.org/project/memory-profiler/):

```
mprof run --include-children python simple_test_aer.py
mprof plot --backend tkagg
```

<img width="1288" alt="Screen Shot 2019-06-21 at 2 03 24 PM" src="https://user-images.githubusercontent.com/11508369/59945563-9d990080-942d-11e9-87f9-465dc7d2a00d.png">



Run the above code.

No memory is leaked.

### Suggested solutions

N/A

hitomitak:
### Summary
To fix #244.

### Details and comments
Add max_workers parameter to ThreadPoolExecutor in AerJob. If this value is not set, AerJob creates threads endlessly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Fixed
     with no identity component. (\#243)
 -   Fixed 2-qubit depolarizing-only error parameter calculation in
     basic_device_noise_model (\#243)
+-   Set maximum workers to ThreadPoolExecutor in AerJob to limit thread creation (\#259)
 
 [0.2.1](https://github.com/Qiskit/qiskit-aer/compare/0.2.0...0.2.1) - 2019-05-20
 ================================================================================
--- a/qiskit/providers/aer/aerjob.py
+++ b/qiskit/providers/aer/aerjob.py
@@ -50,7 +50,7 @@ class AerJob(BaseJob):
         _executor (futures.Executor): executor to handle asynchronous jobs
     """
 
-    _executor = futures.ThreadPoolExecutor()
+    _executor = futures.ThreadPoolExecutor(max_workers=1)
 
     def __init__(self, backend, job_id, fn, qobj, *args):
         super().__init__(backend, job_id)


atilag:
So after talking with @hitomitak, we have decided to put `max_workers = 1`. That makes a lot of sense since the C++ part of Aer should be using all the CPU cores at their maximum capacity because of the nature of the simulation (CPU Bound). So we shall not allow more than 1 instance running locally, otherwise the performance will suffer, especially with larger simulations. For short simulations I guess that the only way to check that this hypothesis is true, is by testing... in the meantime, we are setting `max_workers` to 1.

PR #260, opened 2019-07-03
Avoid reuse in azure pipelines and split build out

mtreinish:
### Summary

In the azure pipelines runs for #258 the jobs are failing for what looks
like a mismatch with the installed aer python package and what should
have been built. Just to rule that out for runs this commit makes 2
changes. First it creates a conda environment/virtualenv with a unique
name for each run. Secondly it runs git clean to remove any old built
wheels (assuming there are any). This turned out not to be the cause
of the failure, but it still doesn't hurt to just rule this out in future builds
anyway. To fix the underlying issue this commit splits the test running
from the aer build into a separate stage in the job. It also changes the
pip command invocation to only install locally and not fallback to use
pypi if a wheel isn't built.

### Details and comments

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,8 +36,8 @@ steps:
       pip install --ignore-installed -r requirements-dev.txt
       python setup.py bdist_wheel -- -G "Visual Studio 15 2017 Win64"
     displayName: 'Install Dependencies and Build Aer'
-  - bash: |
+  - script: |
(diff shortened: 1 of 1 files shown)


mtreinish:
These steps aren't actually needed to fix #258 (but they don't hurt). It looks like the jobs are failing there because the wheel build fails. But it apparently doesn't exit non-zero and our pypi command is falling back to install aer from pypi. I'm going to update this to change pip to only install locally. We should also figure out why the build fails.

PR #262, opened 2019-07-05
Improve diagonal matrix multiplication

hhorii:
### Summary
This PR optimizes `apply_diagonal_matrix` by using sequential access

### Details and comments
`apply_diagonal_matrix`does not need indexes. This PR uses sequential access to apply a diagonal matrix to qubit vector.


atilag:
@hhorii this PR definitely makes the code cleaner and compact, I like this, but seems like the new code has branching into a loop, which is a typical "non-friendly cache" thing to avoid. The old code didn't have this branching on purpose. Did you benchmark this change? Otherwise we can just merge the PR, and see what the daily benchmarks say about it, and revert it back if we detect a performance regression.

hhorii:
@atilag We have tested it with qft and observed about 10% improvement.

src/simulators/statevector/qubitvector.hpp
```diff
@@ -1190,61 +1190,29 @@ void QubitVector<data_t>::apply_multiplexer(const reg_t &control_qubits,
...
-          data_[inds[i]] *= _mat[i];
-        }
-      };
-      apply_lambda(lambda, qubits, diag);
+  if (N == 1) {
+    apply_diagonal_matrix(qubits[0], diag);
+    return;
+  }
+
+  auto lambda = [&](const areg_t<2> &inds, const cvector_t &_diag)->void {
```
chriseclectic: Can you add comments explaining this function? Its hacking the apply_lambda method by using a 1-qubit lambda to apply an N-qubit operation using captures of the input qubit parameters so some comments on this would be very useful for anyone looking at the code in the future.

atilag:
As we want to release 0.2.2 asap, let's not stop it because of this. @hhorii can add the requested comments later in a PR :)

Issue #268, opened 2019-07-06
Classically controlled measurements not working


eliasrg:
### Informations

- **Qiskit Aer version**: 0.2.1
- **Python version**: 3.7.3
- **Operating system**: Manjaro GNU/Linux

### What is the current behavior?
When a measurement is controlled by a classical bit, e.g. `circuit.measure(q, c2).c_if(c1, 1)`, the measurement seems to be made regardless of the value of the control bit. This problem occurs with `Aer`, but `BasicAer` has the expected behaviour.

### Steps to reproduce the problem
This example shows the difference between `Aer` and `BasicAer`:
```python
from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit, \
    Aer, BasicAer, execute

q = QuantumRegister(1, 'q')
c = ClassicalRegister(1, 'c')
o = ClassicalRegister(1, 'o')
C = QuantumCircuit(q, c, o)

C.x(q[0])
C.measure(q, o).c_if(c, 1)

print("BasicAer: {}".format(
    execute(C, BasicAer.get_backend('qasm_simulator'),
                                    shots=1 << 15).result().get_counts()))
print("Aer: {}".format(
    execute(C, Aer.get_backend('qasm_simulator'),
                               shots=1 << 15).result().get_counts()))
```
Output:
```
BasicAer: {'0 0': 32768}
Aer: {'1 0': 32768}
```

### What is the expected behavior?
In the above example, `c` is initialized to 0 and `q` is initialized to |1⟩. Then `q` is measured into `o` if `c == 1`. Theoretically, the measurement should not be performed because `c == 0`, but it is.

### Suggested solutions
Classically controlled measurements are allowed by the OpenQASM standard (`if(c==1) measure q[0] -> o[0];`), but seem not to be supported by IBMQ yet (ibmqx2 did give results in agreement with Aer: `{'1 0': 947, '0 0': 77}`). If the goal is to be completely compatible with present-day IBMQ, this should not be fixed. But since controlled measurements are currently useless anyway, I think supporting them would be a good idea.

hhorii:
Sampling works unexpectedly in qasm simulator.

Current qasm simulator applies sampling optimization if all the measure operations are in the tail of a circuit. However, it should apply this optimization only if all the measure operations are _non-conditional and_ in the tail.

https://github.com/Qiskit/qiskit-aer/blob/0.2.1/src/simulators/qasm/qasm_controller.hpp#L576-L581 

[...]

PR #270, opened 2019-07-08
fix sampling bug in qasm simulator

hhorii:
### Summary
This PR fix issue #268.

### Details and comments
Sampling should not work if a measure is conditional.

### DIFF
--- a/test/terra/backends/qasm_simulator/qasm_measure.py
+++ b/test/terra/backends/qasm_simulator/qasm_measure.py
@@ -129,15 +129,3 @@ class QasmMeasureTests:
             qobj, backend_options=self.BACKEND_OPTS).result()
         self.is_completed(result)
         self.compare_counts(result, circuits, targets, delta=0.05 * shots)
-
-    def test_measure_conditional(self):
-        """Test QasmSimulator measure with conditions"""
-        shots = 2000
-        qobj = ref_measure.measure_conditional()
-        qobj.config.shots = shots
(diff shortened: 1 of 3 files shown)


test/terra/reference/ref_measure.py
```diff
@@ -309,3 +308,33 @@ def measure_counts_qobj_nondeterministic(shots, hex_counts=True):
...
+
+    circuits = []
+    qr = QuantumRegister(1)
+    cr = ClassicalRegister(1)
+
+    # Measure |++> state (sampled)
+    circuit = QuantumCircuit(qr, cr)
+    circuit.x(qr[0])
+    circuit.barrier(qr)
+    circuit.measure(qr, cr).c_if(c, 1)
```
atilag: Seems like `c` variable in: `.c_if(c, 1)` is not defined, I guess that you meant: `cr` ?

test/terra/backends/qasm_simulator/qasm_measure.py
```diff
@@ -129,3 +129,15 @@ def test_measure_nondeterministic_multi_qubit_without_sampling(self):
             qobj, backend_options=self.BACKEND_OPTS).result()
         self.is_completed(result)
         self.compare_counts(result, circuits, targets, delta=0.05 * shots)
+
+    def test_measure_conditional(self):
```
chriseclectic: Can you remove this test case since (for now) conditional measure shouldn't be allowed by the qobj spec

PR #286, opened 2019-07-17
support single precision in state vector simulator

hhorii:
### Summary
This PR adds a new parameter `precision` for `qasm simulator`.

### Details and comments
`QubitVector<std::complex<double>*>` is changed to `QubiveVector<double | float>`.
By using single-precision, simulation becomes faster with reducing precision.

### DIFF
--- a/src/simulators/qasm/qasm_controller.hpp
+++ b/src/simulators/qasm/qasm_controller.hpp
@@ -22,7 +22,7 @@
 #include "simulators/extended_stabilizer/extended_stabilizer_state.hpp"
 #include "simulators/statevector/statevector_state.hpp"
 #include "simulators/stabilizer/stabilizer_state.hpp"
-
+#include "simulators/tensor_network/tensor_network_state.hpp"
 
 namespace AER {
 namespace Simulator {
@@ -40,7 +40,7 @@ namespace Simulator {
  *   optimizations passes for a noisy circuit [Default: 12].
  * 
  * From Statevector::State class
- * 
+ *
  * - "initial_statevector" (json complex vector): Use a custom initial
  *      statevector for the simulation [Default: null].
  * - "zero_threshold" (double): Threshold for truncating small values to
@@ -129,7 +129,13 @@ protected:
   //-----------------------------------------------------------------------
 
   // Simulation methods for the Qasm Controller
-  enum class Method {automatic, statevector, stabilizer, extended_stabilizer};
+  enum class Method {
+    automatic,
+    statevector,
+    stabilizer,
+    extended_stabilizer,
+    tensor_network
+  };
 
   //-----------------------------------------------------------------------
   // Base class abstract method override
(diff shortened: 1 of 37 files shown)


src/simulators/qasm/qasm_controller.hpp
```diff
@@ -293,6 +302,16 @@ void QasmController::set_config(const json_t &config) {
     }
   }
 
+  std::string precision;
+  if (JSON::get_value(method, "precision", config)) {
```
chriseclectic: I think this should be `JSON::get_value(precision, "precision", config)`

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -137,6 +137,12 @@ class QasmController : public Base::Controller {
     tensor_network
   };
 
+  // Simulation precision
+  enum class Precision {
+    double_precision,
```
chriseclectic: This could just be `enum class Precision {single, double}` since it will be used a `Precision::double` / `Precision::single`

src/simulators/statevector/qubitvector.hpp
```diff
@@ -37,10 +37,13 @@ using uint_t = uint64_t;
...
 using reg_t = std::vector<uint_t>;
 using indexes_t = std::unique_ptr<uint_t[]>;
-using complex_t = std::complex<double>;
-using cvector_t = std::vector<complex_t>;
-using rvector_t = std::vector<double>;
 template <size_t N> using areg_t = std::array<uint_t, N>;
+template <typename T> using complex_t = std::complex<T>;
+template <typename T> using cvector_t = std::vector<complex_t<T>>;
+template <typename T> using rvector_t = std::vector<T>;
+using complexd_t = std::complex<double>;
```
chriseclectic: Rather than having so many aliases can we use default template arguments like: 

```c++
template <typename T = double> using complex_t = std::complex<T>;
template <typename T = double> using cvector_t = std::vector<complex_t<T>>;
template <typename T = double> using rvector_t = std::vector<T>;
```
Then remove the `complexd_t/cvectord_t/rvectord_t` etc and replace with `cvector_t<>`

src/simulators/qasm/qasm_controller.hpp
hhorii: `double` is reserved in C++.

chriseclectic: Ah true... could use `Double`/`Single`, but it's not that important

chriseclectic:
@hhorii @atilag when using single precision do we have to worry about narrowing of data types when applying matrices given in double precision? The lambda loops will be things like `data_[i] = mat(i, j) * data_[j]` where `mat(i,j)` is `complex<double>` and `data_[j]` is `complex<float>`. Do we need to static cast the matrices to the reduced precision first?

atilag:
This is valid narrowing conversion between two floating point types, it's stated in the standard as implicit so there's no need for `static_cast`. Having said that, I really hate implicit behaviors, it's great for programming faster but in the long term are less readable so maintainable. If there's not many conversions I'd explicitly use `static_cast<float>` and make clear that this is the behavior we want.
There's always a round in the last decimal digit though, so If this is mathematically correct for our use-case, then it's fine.

chriseclectic:
@hhorii I'm ready to merge this PR, but it needs to be rebased from master to get the build to work. The merge update failed due to some changes in recently merged PRs (I think in the UnitaryMatrix class).

PR #287, opened 2019-07-17
Remove persistent config and noise model from controller

chriseclectic:
### Summary

This PR reworks the controller class so that it doesn't store a single config or noise model.

* The config and noise model is now passed through execute functions so that the noise model or config can be modified in a thread safe way for each different circuit execution. 
* Adds nosie model to circuit optimization framework (though it is currently not used)
* Fixes an issue with the MacOS openmp hack that would cause a segfault if it wasn't loaded before any virtual function of the controller was called. This hack is now loaded by the `execute_controller` cython wrapped function, rather than in the `set_config` method.

### DIFF
--- a/src/base/controller.hpp
+++ b/src/base/controller.hpp
@@ -37,8 +37,8 @@
 #include "framework/data.hpp"
 #include "framework/rng.hpp"
 #include "framework/creg.hpp"
-#include "transpile/circuitopt.hpp"
 #include "noise/noise_model.hpp"
+#include "transpile/circuitopt.hpp"
 
 #ifdef _OPENMP
 #include <omp.h>
@@ -157,12 +157,14 @@ protected:
   // Parallel execution of a circuit
   // This function manages parallel shot configuration and internally calls
   // the `run_circuit` method for each shot thread
-  virtual json_t execute_circuit(Circuit &circ);
+  virtual json_t execute_circuit(Circuit &circ,
+                                 Noise::NoiseModel &noise);
 
   // Abstract method for executing a circuit.
   // This method must initialize a state and return output data for
   // the required number of shots.
   virtual OutputData run_circuit(const Circuit &circ,
+                                 const Noise::NoiseModel &noise,
                                  uint_t shots,
                                  uint_t rng_seed) const = 0;
 
@@ -195,8 +197,8 @@ protected:
   // Generate an equivalent circuit with input_circ as output_circ.
   template <class state_t>
   void optimize_circuit(Circuit &input_circ,
-                           state_t& state,
-                           OutputData &data) const;
+                        state_t& state,
+                        OutputData &data) const;
 
   //-----------------------------------------------------------------------
   // Config
@@ -207,9 +209,6 @@ protected:
 
   // Controller config settings
   json_t config_;
-
-  // Noise model
-  Noise::NoiseModel noise_model_;
 
   // Circuit optimization
   std::vector<std::shared_ptr<Transpile::CircuitOptimization>> optimizations_;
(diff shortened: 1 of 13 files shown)


chriseclectic:
@atilag while this seems to work for cython wrapped simulator, it looks like it is causing a segfault for the standalone qasm_simulator, probably something to do with the OpenMP loading hack.

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -379,31 +402,53 @@ OutputData QasmController::run_circuit(const Circuit &circ,
...
-  auto method = simulation_method_;
-  if (method == Method::automatic) {
-    // Check if Clifford circuit and noise model
-    if (validate_state(Stabilizer::State(), circ, noise_model_, false)) {
-      method = Method::stabilizer;
-    } else {
-    // Default method is statevector, unless the memory requirements are too large
+QasmController::Method
+QasmController::simulation_method(const Circuit &circ,
+                                  const Noise::NoiseModel &noise_model,
```
atilag: One of the problems that I see is that the `noise_model` is now traveling around across functions that should know nothing about noise models, for example this member function, needs the noise_model just because there's a validation that requires it. But the only purpose of this member function is to return the simulation method, there's no need for the noise model at all.
I'd move out validation from this member function (so the noise_mode argument). Haven't seen where to move it actually but I saw that we are already calling validation just before running a circuit, so this might be enough?

src/base/controller.hpp
```diff
@@ -334,16 +324,18 @@ void Controller::clear_parallelization() {
...
 }
 
-void Controller::set_parallelization_experiments(const std::vector<Circuit>& circuits) {
+void Controller::set_parallelization_experiments(const std::vector<Circuit>& circuits,
+                                                 const Noise::NoiseModel& noise) {
 
   if (max_parallel_experiments_ <= 0)
     return;
   // if memory allows, execute experiments in parallel
   std::vector<size_t> required_memory_mb_list;
```
atilag: just find out that we can set the initial size of the vector here, as we know it before hand: `circuits.size()`. This will avoid many allocations due to the `push_back`

src/simulators/statevector/statevector_controller.hpp
```diff
@@ -122,7 +125,8 @@ void StatevectorController::clear_config() {
   initial_state_ = cvector_t();
 }
 
-size_t StatevectorController::required_memory_mb(const Circuit& circ) const {
+size_t StatevectorController::required_memory_mb(const Circuit& circ,
+                                                 const Noise::NoiseModel& noise) const {
```
atilag: Why do we need `noise`  here? Seems like we are not using it at all.

src/simulators/qasm/qasm_controller.hpp
chriseclectic: Actually the simulation method *does* critically depend on both the noise model and the circuit. You can only use the stabilizer method if both the circuit and noise model is Clifford for example. This was being checked before, but was using the noise model stored as a controller member variable

src/base/controller.hpp
chriseclectic: I didn't change any implementation of code not concerning the stored config/noise model in this PR, but if you notice some more simple changes (like this) I can add them

atilag: Yeah, I know it's not in your changeset, just saw it while reviewing the code. If you can change this that would be great.

src/simulators/qasm/qasm_controller.hpp
atilag: Ok then.

atilag: Ok then!

src/simulators/statevector/statevector_controller.hpp
chriseclectic: It's used in the qasm controller because the required memory depends on simulation method (which depends on noise model). There should be a more elegant way of doing this but it would require some more serious refactoring of controller classes that I would like to postpone till later

PR #288, opened 2019-07-17
Add trace and check identity to UnitaryMatrix

chriseclectic:
### Summary

Adds methods to `UnitaryMatrix` class for computing trace and checking if the matrix is an identity. These are taken from the old identity-checker PR so they can be merged before bit-rotting away, though they are not used as an identity checker (that interface needs to be decided later).

### DIFF
--- a/src/simulators/unitary/unitarymatrix.hpp
+++ b/src/simulators/unitary/unitarymatrix.hpp
@@ -63,6 +63,9 @@ public:
   // Returns a copy of the underlying data_t data as a complex vector
   AER::cmatrix_t matrix() const;
 
+  // Return the trace of the unitary
+  complex_t trace() const;
+
   // Return JSON serialization of UnitaryMatrix;
   json_t json() const;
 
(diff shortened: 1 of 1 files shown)


src/simulators/unitary/unitarymatrix.hpp
```diff
@@ -224,6 +252,63 @@ void UnitaryMatrix<data_t>::set_num_qubits(size_t num_qubits) {
...
+// Check Identity
+//------------------------------------------------------------------------------
+
+template <class data_t>
+std::pair<bool, double> UnitaryMatrix<data_t>::check_identity() const {
+  // To check if identity we first check we check that:
+  // 1. U(0, 0) = exp(i * theta)
+  // 2. U(i, i) = U(0, 0)
+  // 3. U(i, j) = 0 for j != i 
+  auto failed = std::make_pair(false, 0.0);
```
ajavadia: i feel like this variable should be called `passed` instead of `failed`, because when passing, you set it to `true`.

chriseclectic: This variable is called `failed` because it's storing the return value for when it fails to be unitary ie: `return failed;` when it fails. A separate pair containing the phase is constructed and returned when it doesn't fail.

PR #292, opened 2019-07-22
Fixed bug in parallelization in MPS simulator. 

merav-aharoni:
### Summary
There was an "omp parallel" on a for loop that had a vector that was assigned values using push_back. That caused undeterministic behavior. Changed the push_back to assignment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@ Changelog](http://keepachangelog.com/en/1.0.0/).
 Added
 -----
 -   New simulation method for qasm simulator: tensor_network (\#56)
-
+-   Added superop qobj instruction and superoperator matrix utils (\# 289)
+-   Added support for conditional unitary, kraus, superop qobj instructions (\# 291)
 
 Changed
 -------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ Removed
 
 Fixed
 -----
-- Bug in handling parallelization in matrix_product_state.cpp
+- Bug in handling parallelization in matrix_product_state.cpp (PR \#292)
 
 
 [0.2.3](https://github.com/Qiskit/qiskit-aer/compare/0.2.2...0.2.3) - 2019-07-11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Removed
 
 Fixed
 -----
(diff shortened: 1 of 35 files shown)


src/simulators/tensor_network/matrix_product_state.cpp
```diff
@@ -92,6 +92,12 @@ cmatrix_t mul_matrix_by_lambda(const cmatrix_t &mat,
   if (lambda == rvector_t {1.0}) return mat;
   cmatrix_t res_mat(mat);
   uint_t num_rows = mat.GetRows(), num_cols = mat.GetColumns();
+
+  #ifdef _WIN32
+     #pragma omp parallel for
+  #else
+     #pragma omp parallel for collapse(2)
+  #endif
   for(uint_t row = 0; row < num_rows; row++) {
```
atilag: We need to change the type of `row` to `int_t` due to incompatibilities with Windows OpenMP support.

src/simulators/tensor_network/matrix_product_state.cpp
```diff
@@ -92,6 +92,12 @@ cmatrix_t mul_matrix_by_lambda(const cmatrix_t &mat,
...
       for(uint_t col = 0; col < num_cols; col++) {
```
atilag: We need to chage this type too: `uint_t col` => `int_t col`

src/simulators/tensor_network/matrix_product_state.cpp
merav-aharoni: Corrected both places as you pointed out.
It is a problem to catch failures that only happen non-deterministically. I can only suggest running the unittest a few times.
We also use another kind of testing here - we create random circuits and then compare results between the mps and the statevector. If you are interested, I can add these somewhere as well.

src/simulators/tensor_network/matrix_product_state.cpp
merav-aharoni: see above.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ Removed
 
 Fixed
 -----
+- Bug in handling parallelization in matrix_product_state.cpp
```
atilag: We need to add the number of the PR at the end of this line (see the other entries) :)

merav-aharoni: right. done.

Issue #218, opened 2019-05-31
Aer cannot simulate circuit using only few qubits if width is large


nonhermitian:
### Informations

- **Qiskit Aer version**: master (0.3)
- **Python version**:
- **Operating system**: osx

### What is the current behavior?
```python

circ = QuantumCircuit(50, 1)
circ.h(45)
for _ in range(100):
     circ.t(45)
circ.measure(45, 0)
job = execute(circ, Aer.get_backend('qasm_simulator'))
```

gives:

```python

AerError: 'ERROR: a circuit requires more memory than max_memory_mb.'
```

Also:

```python

import qiskit
from qiskit.quantum_info.random import random_unitary
from qiskit import Aer

U = random_unitary(4)
qr = QuantumRegister(50)
cr = ClassicalRegister(2)
qc = QuantumCircuit(qr, cr)
qc.append(U,[qr[48],qr[49]], [])
qc.measure(qr[48], cr[0])
qc.measure(qr[49], cr[1])
backend = Aer.get_backend('qasm_simulator')
job = qiskit.execute(qc, backend=backend)
job.result()
```

gives:

```python

AerError: 'ERROR: QasmController: Circuit cannot be run on any available backend. max_memory_mb=8192mb'
```

ajavadia:
This was working for me, but not for @nonhermitian. Our versions seemed identical, so not sure what the issue is. If others can try this and report the result with version number that'd be great.

chriseclectic:
I think the issue here is that the memory checks are happening before the qubit truncation pass. Currently all circuit optimiztion passes happen at a single step just before execution (but after all the memory checks for automatic configuration of paralellization). We need to change how these passes are implemented so that some, such as this truncation, can be applied before the memory checks.

chriseclectic:
Closed by #296

PR #296, opened 2019-07-26
Truncate/remap qubits before calculating memory requirements or noise sampling

### LINKED ISSUE #285: Improve qubit truncation and circuit optimization framework

Currently the circuit optimization pass for qubit truncation does not take the noise model into account. This means for noisy simulations truncation must be done *after* noise sampling which is very inefficient.

This pass needs to take into account the noise model and check both the circuit for qubits, and the non-local errors in the noise model for instructions in the circuit to find out which qubits are potentially active, and then truncate/remap both the circuit and the noise model.

After this the qubit truncation can be run immediately after loading a circuit and before memory requirements are checked for parallelisation configuration. We will also have to change the `CircuitOptimization` base class to allow for noise models.

chriseclectic:
### Summary

* Adds a remap qubits function to C++ `NoiseModel` class
* Updates the truncate qubit optimization pass to remap the noise model for the given  truncation
* Applies this optimization in the Controller class before calculating memory requirements for parallelization.

Closes #285, #218

### DIFF
--- a/src/noise/noise_model.hpp
+++ b/src/noise/noise_model.hpp
@@ -119,20 +119,6 @@ public:
   // If no nonlocal error exists an empty set is returned.
   std::set<uint_t> nonlocal_noise_qubits(const std::string label, const reg_t &qubits) const;
 
-  // Return vector of qubit registers for all readout errors in noise model.
-  // The empty reg is used to an all-qubit readout error.
-  std::vector<reg_t> readout_error_keys() const;
-
-  // Return vector of key pairs for all non-local quantum errors in the noise
-  // model. Vector elements are tuples (label, qubits, noise_qubits) where
-  // label is the string label for the error instruction, qubits is the reg of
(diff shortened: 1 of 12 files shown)


src/noise/noise_model.hpp
```diff
@@ -54,6 +54,39 @@ class NoiseModel {
   // can be done in a thread-safe manner
   Circuit sample_noise(const Circuit &circ, RngEngine &rng) const;
 
+  //-----------------------------------------------------------------------
+  // Checking if errors types are in noise model
+  //-----------------------------------------------------------------------
+
+  // Return True if noise model contains readout errors
+  inline bool has_readout_erros() const {
```
atilag:
```suggestion
  inline bool has_readout_errors() const {
```

chriseclectic:
@atilag added some new tests (and found and fixed some bugs in doing so)

PR #297, opened 2019-07-28
[WIP] Add many worlds simulator to Aer

eliarbel:


eliarbel:
This is the 1st phase of transferring the code from Terra/BasicAer into Aer. 
Currently the capability is fully implemented in python in Providers/Aer/backends/many_worlds_simulator.py, which by itself is an adaptation of qasm_simulator.py of BasicAer in Terra. 
The next step would involve moving the functionality out of many_worlds_simulator.py into the relevant C++ parts

eliarbel:
With the shift in focus towards Ignis, this seems even less valuable than before

PR #299, opened 2019-07-29
Partial measure in tensor_network simulation method

merav-aharoni:
### Summary
Added support for apply_measure in tensor_network.
Also changed sample_measure to use apply_measure since the new implementation is more efficient.

### Details and comments
I would like to use measure in the examples in the tutorial, therefore I hope you can perform this merge soon.


src/simulators/tensor_network/matrix_product_state.hpp
```diff
@@ -22,6 +22,8 @@ enum Gates {
   cx, cz, cu1, swap, su4 // two qubit
 };
 
+enum Direction {RIGHT, LEFT};
```
atilag:
```suggestion
enum class Direction {RIGHT, LEFT};
```

src/simulators/tensor_network/matrix_product_state.hpp
```diff
@@ -191,8 +193,23 @@ class MPS{
...
+  // the effects of measurement from the measured qubit to all other
+  // qubits
+  //----------------------------------------------------------------
+  void update_qubit_and_neighbor(uint_t index, 
+				 Direction direction,
+				 const cmatrix_t &mul_matrix);
   reg_t sample_measure(std::vector<double> &rnds);
 
+  reg_t apply_measure(const reg_t &qubits, 
+		      RngEngine &rng);
```
atilag:
```suggestion
	               RngEngine &rng) const;
```

src/simulators/tensor_network/matrix_product_state.hpp
```diff
@@ -191,8 +193,23 @@ class MPS{
...
+
+  uint_t apply_measure(uint_t qubit, 
+		      RngEngine &rng);
```
atilag:
```suggestion
	                RngEngine &rng);
```

src/simulators/tensor_network/matrix_product_state.hpp
```diff
@@ -191,8 +193,23 @@ class MPS{
     return 0;
   }
   
+  //----------------------------------------------------------------
+  // function name: update_qubit_and_neighbor
+  // Description: This function is used by apply_measure to propagate
+  // the effects of measurement from the measured qubit to all other
+  // qubits
+  //----------------------------------------------------------------
+  void update_qubit_and_neighbor(uint_t index,
```
atilag: I can't see where this function is being used

src/simulators/tensor_network/matrix_product_state.hpp
atilag: This function can be `const` as seems it's not changing the state of the object

merav-aharoni: apply_measure does change the state, so I can not turn it into 'const'.
sample_measure could in fact be const, but I removed it completely because it is implemented by saving the internal state and then using apply_measure on it.

src/simulators/tensor_network/matrix_product_state.hpp
merav-aharoni: you are right. I removed this function.

PR #307, opened 2019-07-30
Do not limit the MPS simulator by the number of qubits

yaelbh:
### Summary

When I try to run the MPS simulator with 40 qubits, I receive an error message stating that it exceeds the memory capability of my machine. The check that triggers the error assumes that I run the state vector simulator. This pull request fixes this.

### Details and comments

Strangely, @merav-aharoni told me that she is able to run the MPS simulator with hundreds of qubits without encountering this error.

### DIFF
--- a/qiskit/providers/aer/backends/qasm_simulator.py
+++ b/qiskit/providers/aer/backends/qasm_simulator.py
@@ -39,12 +39,13 @@ class QasmSimulator(AerBackend):
         ------------------------
         * "method" (str): Set the simulation method. Allowed values are:
             * "statevector": Uses a dense statevector simulation.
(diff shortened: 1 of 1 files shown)


chriseclectic:
I think we should remove all Python side memory checking and do it in C++. Especially with the  truncation/remapping in PR #296 that allows truncating unused qubits before memory checks are run.

@merav-aharoni may have been bypassing the memory check by using `validate=False`

atilag:
Agree, these kind of checks should belong to each of the simulators, not the python wrapper.

chriseclectic:
I think we can close this since these Python side checks were removed in 0.3

Issue #300, opened 2019-07-29
Add snapshot circuit extensions


chriseclectic:
### What is the expected behavior?

Add circuit extensions for specific snapshot types to make them easier to use. Initially we can start with the currently implemented snapshots and add additional ones in the future. The initial instructions should be:

- [x] `SnapshotProbabilities` #304 : measurement probability snapshot for specified qubits (can apply to all methods of qasm simulator).
- [x] `SnapshotExpectationValue` #305 : exp val snapshots for methods of qasm simulator that support it. Initially this should convert the input to a Pauli expectation value snapshot.
- [x] `SnapshotStatevector` #301 : statevector snapshot for qasm simulator and statevector simulator
- [x] `SnapshotStabilizer` #302 : snapshot for stabilizer method of qasm simulator 
- [x] `SnapshotDensityMatrix` #303 : snapshot for density matrix method of qasm simulator (could be added to other methods in future)
- [x] Add snapshot tests
- [ ] Make snapshot tutorial notebook for qiskit-tutorials

I will create separate issues for each snapshot type with more details and collect them here. 

There already exists a general `Snapshot` instruction in Terra, this can be used as the super class for these extensions. This base Snapshot instruction should probably be copied/moved to Aer for now.

 *Note: There is an issue with parameters and validation that they must be wrapped in numpy or sympy to validate*.

chriseclectic:
@kanejess will get started on this

atilag:
@chriseclectic I think we can close this issue, right?

PR #309, opened 2019-07-30
Add some missing snapshots

chriseclectic:
### Summary

Adds three new snapshots:

1. "stabilizer" snapshot for QasmSimulator stabilizer simulation method. This returns the current stabilizer state for each shot
2. "expectation_value_pauli_single_shot": per shot expectation values of Pauli matrix (instead of averaged over all shots)
3. "expectation_value_matrix_single_shot": per shot expectation values of matrix (instead of averaged over all shots)

### Details and comments

The single shot expectation value take the same params as the other expectation value snapshots, but instead of returning the average value over shots they return the list of expectation values for each shot. These kind of snapshots are needed for quantum error correction simulations. 

They are returned under the field `"expectation_values"` rather than `"expectation_value"`.

Example
```python
from qiskit import *
from qiskit.extensions.simulator import *
from qiskit.providers.aer import *

sim = Aer.get_backend("qasm_simulator")

exp_x = [np.array([[1, 0], 'X'])]
snap_ave = Snapshot('<X>_ave', 'expectation_value_pauli', 1, 0, exp_x)
snap_shot = Snapshot('<X>_shot', 'expectation_value_pauli_single_shot', 1, 0, exp_x)

nm = noise.NoiseModel()
nm.add_all_qubit_quantum_error(noise.errors.depolarizing_error(0.5, 1), 'h')

qc = QuantumCircuit(1, 1)
qc.h(0)
qc.append(snap_ave, [0])
qc.append(snap_shot, [0])
qc.measure(0, 0)
job = execute(qc, sim, shots=10, noise_model=nm)
result = job.result()
result.data(0)
```

Returns 

```python
{'snapshots': {'expectation_values': {'<X>_shot': [[1.0, 0.0],
    [-1.0, 0.0],
    [1.0, 0.0],
    [-1.0, 0.0],
    [1.0, 0.0],
    [1.0, 0.0],
    [1.0, 0.0],
    [1.0, 0.0],
    [1.0, 0.0],
    [1.0, 0.0]]},
  'expectation_value': {'<X>_ave': [{'memory': '0x0', 'value': [0.6, 0.0]}]}},
 'counts': {'0x1': 4, '0x0': 6}}
```
​
```

### CODE REVIEW

File: src/simulators/statevector/statevector_state.hpp [hunk 1]
```diff
@@ -578,16 +595,27 @@ void State<statevec_t>::snapshot_pauli_expval(const Operations::Op &op,
   }
   // add to snapshot
   Utils::chop_inplace(expval, json_chop_threshold_);
-  data.add_average_snapshot("expectation_value", op.string_params[0],
-                            BaseState::creg_.memory_hex(), expval, variance);
+  switch (type) {
+    case SnapshotDataType::average:
+      data.add_average_snapshot("expectation_value", op.string_params[0],
+                            BaseState::creg_.memory_hex(), expval, false);
```
atilag: What is the last `true` and `false` argument? I'd definitely put an enum here so I don't have to go the function definition to know what's this boolean means.

chriseclectic:
For tests we really need the instructions @kanejess and @merav-aharoni are working on in #300

Issue #305, opened 2019-07-29
Expectation Value snapshot extension


chriseclectic:
### What is the expected behavior?

This will be used by various methods of the QasmSimulator to return snapshots of expectation value outcomes `<M>` for some operator `M`.

Eventually we will allow different input classes for M but for now we will allow N-qubit `Operator` objects (which can be constructed from other Operators, lists, np.arrays, Paulis, etc). Behind the scenes these should be converted into either pauli expectation value snapshots, or dense matrix snapshots.

The qobj format for Pauli expectation value snapshot params is a list `[[coeff, pauli], ...]` where `coeff` is a complex number (JSON: `[re, im]`), and `pauli` is a string containing only `IXYZ` for the Pauli term (qubit 0 to N-1 are characters right to left).

```python
class SnapshotExpectationValue(Snapshot):
    __init__(self, label, op, pauli=True, variance=False):
        # num_qubits can be inferred form size of operator.
        # If Pauli is true convert to Pauli expval snapshot
       # type: "expectation_value_pauli" or ""expectation_value_pauli_with_variance"
       # otherwise we use general matrix ones
      # type: "expectation_value_matrix" or ""expectation_value_matrix_with_variance"
      # See `operations.hpp` for input format

# Monkey patch function
def snapshot_expectation_value(self, label, op, qubits=None, variance=False):
     # If qubits is None snapshot all qubits.

circuit.snapshot_expectation_value = snapshot_expectation_value
```

Issue #304, opened 2019-07-29
Probabilities snapshot extension


chriseclectic:
### What is the expected behavior?

This will be used by various methods of the QasmSimulator to return snapshots of the measurement outcome probabilities of the specified qubits.

```python
class SnapshotProbabilities(Snapshot):
    __init__(self, label, num_qubits, variance=False):
        if variance:
            super(label, "probabilities_with_variance", num_qubits, 0, None)
        else:
            super(label, "probabilities", num_qubits, 0, None)


# Monkey patch function
def snapshot_probabilities(self, label, qubits=None, variance=False):
     # If qubits is None snapshot all qubits.

circuit.snapshot_probabilities = snapshot_probabilities
```

kanejess:
Which QasmSimulator methods does this need to work for?

Issue #303, opened 2019-07-29
DensityMatrix snapshot extension


chriseclectic:
### What is the expected behavior?

This will be used by the density matrix method of the QasmSimulator.

```python
class SnapshotDensityMatrix(Snapshot):
    __init__(self, label, num_qubits):
        super(label, "density_matrix", num_qubits, 0, None)
 

# Monkey patch function
def snapshot_density_matrix(self, label, qubits=None):
     # see base snapshot monkey patch for converting qubits to
     # be all qubits in circuit

circuit.snapshot_density_matrix = snapshot_density_matrix
```

The `num_qubits` parameter is actually ignored by Aer (it is always an all-qubit instruction), however it is used by Terra to make the snapshot act as a barrier.

Issue #302, opened 2019-07-29
Stabilizer snapshot extension


chriseclectic:
### What is the expected behavior?

This is used by the stabilizer method of the QasmSimulator to return snapshots of the stabilizer state of the simulator. This should have signature

```python
class SnapshotStabilizer(Snapshot):
    __init__(self, label, num_qubits):
        super(label, "stabilizer", num_qubits, 0, None)
 

# Monkey patch function
def snapshot_stabilizer(self, label, qubits=None):
     # see base snapshot monkey patch for converting qubits to
     # be all qubits in circuit

circuit.snapshot_stabilizer = snapshot_stabilizer
```

The `num_qubits` parameter is actually ignored by Aer (it is always an all-qubit instruction), however it is used by Terra to make the snapshot act as a barrier.

chriseclectic:
Closed by #355

PR #317, opened 2019-08-02
Snapshot circuit extensions 

kanejess:
### Summary
Getting started on #300 - i.e. improving usability of different snapshot types. 

### Details and comments
Subclasses and methods written for `SnapshotStatevector` #301, `SnapshotStabilizer` #302, `SnapshotDensityMatrix` #303, `SnapshotProbabilities` #304, and `SnapshotExpectationValue` #305. 

So far, tests are only checking that the snapshots have been created, and do not check that snapshot data is correct.

#### Still needs fixing:
* `'density_matrix'` test gives error `'ERROR: Circuit contains invalid instructions ( invalid snapshot instructions: {density_matrix})'`


chriseclectic:
@kanejess To match terra I would suggest making a separate file for each type of snapshot and placing them inside `aer/extensions/snapshot.py`, `aer/extensions/snapshot_probabilities.py` etc, and then having an `__init__.py` inside the extensions folder that imports each one

qiskit/providers/aer/aer_snapshots.py
```diff
@@ -0,0 +1,273 @@
+# -*- coding: utf-8 -*-
+
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2017.
```
chriseclectic: The copyright should be `(C) Copyright IBM 2018, 2019.`

qiskit/providers/aer/aer_snapshots.py
```diff
@@ -0,0 +1,332 @@
...
+        super().__init__(label, snapshot_type, num_qubits, num_clbits, params)
+
+def snapshot_stabilizer(self,
+                        label,
+                        snapshot_type='stabilizer',
+                        qubits=None,
+                        params=None):
+
+    # Convert label to string for backwards compatibility
+    if not isinstance(label, str):
```
maddy-tod: Could all of the first part of this method (up to the return) be extracted into a separate method? That way we reduce the duplication(and so improve the maintainability of the code) as all the different types of snapshot can call on that one method instead of having their own copy of the code. It could be made a private method of `Snapshot`.

qiskit/providers/aer/aer_snapshots.py
```diff
@@ -0,0 +1,332 @@
...
+                 label,
+                 snapshot_type='density_matrix',
+                 num_qubits=0,
+                 num_clbits=0,
+                 params=None):
+        super().__init__(label, num_qubits, num_clbits, params)
+
+def snapshot_density_matrix(self,
+                            label,
+                            snapshot_type='density_matrix',
```
maddy-tod: Is there a reason the `snapshot_type` is taken as a parameter? It seems like this could allow you to do
`circuit.snapshot_density_matrix('label', snapshot_type='stabilizer') ` which means there are conflicting snapshot types.

qiskit/providers/aer/extensions/snapshot.py
```diff
@@ -0,0 +1,137 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Simulator command to snapshot internal simulator representation.
+"""
+
+import warnings
+
```
maddy-tod: Are these files a duplication of what is in `qiskit/providers/aer/aer_snapshots.py`? I think this is a better way to lay them out so I would suggest removing that file and keeping these ones (above comments still apply to these files :) )

qiskit/providers/aer/aer_snapshots.py
kanejess: Done! Thanks for the help

qiskit/providers/aer/aer_snapshots.py
kanejess: Good catch. I've fixed this

qiskit/providers/aer/extensions/snapshot.py
kanejess: Fixed 😄

qiskit/providers/aer/aer_snapshots.py
kanejess: Fixed!

atilag:
Hi @kanejess , what is left to remove the [WIP]?. We targeted this to be included in 0.3 release which will happen very soon.

qiskit/providers/aer/extensions/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+        instruction.snapshot_type = self._snapshot_type
+        return instruction
+
+    def inverse(self):
+        """Special case. Return self."""
+        return Snapshot(self.num_qubits, self.num_clbits, self.params[0],
+                        self.params[1])
+
+    @staticmethod
+    def define_snapshot_register(self,
```
maddy-tod: Can you rename `self` to `circuit` or similar? As this is a static method it doesn't take an instance of the class it belongs to, which is what `self` usually means.

qiskit/providers/aer/extensions/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+        else:
+            raise TypeError('label expects a string')
+
+
+def snapshot(self,
+             label,
+             snapshot_type='statevector',
+             qubits=None,
+             params=None):
+    """Take a statevector snapshot of the internal simulator representation.
```
maddy-tod:
```suggestion
    """Take a snapshot of the internal simulator representation.
```

qiskit/providers/aer/extensions/snapshot_density_matrix.py
```diff
@@ -0,0 +1,39 @@
...
+# that they have been altered from the originals.
+
+from qiskit import QuantumCircuit
+from qiskit.circuit.quantumregister import QuantumRegister
+from qiskit.providers.aer.extensions import Snapshot
+
+class SnapshotDensityMatrix(Snapshot):
+    def __init__(self,
+                 label,
+                 snapshot_type='density_matrix',
```
maddy-tod: I think it would be preferable not to take the `snapshot_type` as a parameter, as you know that it should be `'density_matrix'`. Could you remove this, and just make the call below `super().__init__(label, 'density_matrix', num_qubits, num_clbits, params)` instead? This applies to all the snapshot classes.

qiskit/providers/aer/extensions/snapshot_probabilities.py
```diff
@@ -0,0 +1,46 @@
...
+                 label,
+                 num_qubits,
+                 snapshot_type='probabilities',
+                 num_clbits=0,
+                 params=None,
+                 variance=False):
+
+        super().__init__(label, snapshot_type, num_qubits, num_clbits, params)
+
+        if variance:
```
maddy-tod: This needs to come above the call to `super()` so that it is used in that method.

atilag:
There's a lot of code repetition in this PR, and other things that could be simplified, but that would probably require a deep redesign on how snapshots are implemented so I'm not going to postpone it as far as it works.
We can have the redesign discussion after 0.3 release.

atilag:
I'd need confirmation from @kanejess for tomorrow to include this in the release.
Otherwise will be postponed to 0.3.1.

PR #318, opened 2019-08-02
Delay measure circuit optimization

chriseclectic:
### Summary

Follow on from #316 

This adds a circuit optimization pass that pushes measurements to the end of a circuit if no conditional operations, or other operations on the already measured qubits follow.

This is needed for measure sampling to work with the density matrix simulator when there are quantum errors on measure instructions.

### DIFF
--- a/src/transpile/delay_measure.hpp
+++ b/src/transpile/delay_measure.hpp
@@ -1,0 +1,133 @@
+/**
+ * This code is part of Qiskit.
+ *
+ * (C) Copyright IBM 2018, 2019.
+ *
+ * This code is licensed under the Apache License, Version 2.0. You may
+ * obtain a copy of this license in the LICENSE.txt file in the root directory
+ * of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Any modifications or derivative works of this code must retain this
+ * copyright notice, and modified files need to carry a notice indicating
+ * that they have been altered from the originals.
+ */
+
+#ifndef _aer_transpile_delay_measure_hpp_
+#define _aer_transpile_delay_measure_hpp_
+
+#include <unordered_map>
+
+#include "transpile/circuitopt.hpp"
+
+namespace AER {
+namespace Transpile {
+
+class DelayMeasure : public CircuitOptimization {
+public:
+
+  // DelayMeasure uses following configuration options
+  // - delay_measure_verbose (bool): if true, output generated gates in metadata (default: false)
+  // - delay_measure_enable (bool): if true, activate optimization (default: true)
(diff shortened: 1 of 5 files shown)


chriseclectic:
Actually I think this doesn't need to be a separate optimization pass and I can just put this into the part of qasm_controller that calls the sampler.

src/transpile/delay_measure.hpp
```diff
@@ -0,0 +1,133 @@
...
+                                    const Operations::OpSet &allowed_opset,
+                                    OutputData &data) const {
+  // Pass if we have quantum errors in the noise model
+  if (active_ == false || circ.shots <= 1 || noise.has_quantum_errors())
+    return; 
+
+  // Get position of first measure / readout error in circ
+  auto it = circ.ops.begin();
+  while (it != circ.ops.end()) {
+    const auto type = it->type;
```
hhorii: I guess this optimization will not be effective if a circuit includes a kraus operation. If so, some check codes are necessary here.

chriseclectic: Kraus operations should be fine. This pass will look for any non-conditional operation on a different qubit after the first measurement and see if it can be moved forward before the first measurement.

If there are Kraus operations the measure-sampler itself won't be able to sample (unless its a density matrix simulation) but that's checked in a different piece of code.

hhorii: In my understanding, this optimization is to expand a scope of circuits for sampler. So, if a kraus is included in a circuit, this optimization is not necessary. So I just thought that adding `if (type == Operations::OpType::kraus) return;` in this loop is an option to stop this optimization and reduce time of transpilation.
Anyway, its reduction will be tiny. So applying this optimization to circuits that includes kraus operators will not introduce serious overheads.

chriseclectic: Actually the density matrix method supports measure sampling when there are kraus errors.  The main reason is to expand the scope of measure sampler for the density matrix simulator when there are errors (including kraus) on measure instructions. In this case noise insertion would result in a circuit like
`[measure(0, 0), measure(1,1)]` -> `[kraus(0), measure(0, 0), kraus(1), measure(1,1)]` and then this pass converts it to  `[kraus(0), kraus(1), measure(0, 0), measure(1,1)]`

It also will expand the scope of sampling for the statevector and other methods, but only for the ideal case in those ones since, as you say they can't be done when there are kraus errors. In that case there shouldn't be any extra overhead because this optimization is only applied in the number of shots is > 1, which only occurs in the qasm controller for the noisy density matrix (which has previously been checked to support sampling before noise insertion) or ideal statevector.

PR #333, opened 2019-08-29
Add SIMD function for Mac

hitomitak:
### Summary
Replace apply_matrix() to SIMD instructions for only Mac environment. This implementation improves the performance only when a single NUMA node.

I measured the execution time on My MBP with this code.
QV sec (SIMD/NON-SIMD)
- QV22  0.5/0.66 
- QV23 1.1/1.48
- QV24 2.38/2.9
- QV25 4.0/5.28
- QV26 9.38/11.38
- QV27 19.50/23.50
- QV28 40.49/49.86

### Details and comments
This SIMD functions are enabled when the execution env is Mac and `data_t` is `double`.
I added `#ifdef SIMD` to switch non-simd and simd function but this implementation is not so good. 
Are there any good solutions to switch them smartly?


atilag:
Does all "modern" Mac CPUs support AVX2/AVX instructions?

hitomitak:
Yes. This code uses only AVX instruction. AVX is supported by Sandy Bridge equipped with Mac since 2012.

chriseclectic:
@hitomitak @hhorii with this and the gpu work I think we should consider pulling these matrix routines, which may have multiple different kernels for implementing such as SIMD/non-SIMD etc, outside of the qubit vector class.

The `apply_lambda` and `apply_matrix` family of functions could be their own BLAS style functions that take the array/vector and max dimension as arguments, and then they could be applied to any appropriately sized array/vector type object that has an access operator `obj[j]`

src/simulators/statevector/kernel.hpp
```diff
@@ -0,0 +1,60 @@
...
+        data_[ii] = 0.;
+    }
+    // update state vector
+    for (size_t i = 0; i < 4; i++)
+        for (size_t j = 0; j < 4; j++)
+        data_[inds[i]] += _mat[i + 4 * j] * cache[j];
+}
+
+template <typename data_t>
+inline void kernel3(std::complex<data_t> *data_, const areg_t<8> &inds,const cvector_t<data_t> &_mat){
```
chriseclectic: Can you add `kernels4` for 4-qubit? This is needed for noisy 2-qubit gates in density matrix simulator

src/simulators/statevector/qubitvector.hpp
```diff
@@ -400,6 +404,10 @@ class QubitVector {
   size_t data_size_;
   std::complex<data_t>* data_;
   std::complex<data_t>* checkpoint_;
+  // If data_t is double, SIMD function is enable
+#ifdef SIMD
+  bool type_double = ((typeid(data_t) == typeid(double)) ? true : false);
```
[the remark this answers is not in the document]
hitomitak: No. this code only works with double precession. So if the type is single precession, it executes Non-SIMD apply_matrix function.

src/simulators/statevector/kernel.hpp
jagunnels: Is this (failure of kernel4) true for both kernel.hpp and kernel_simd.hpp?  Could you please compile the file (with kernel4 in it) and transmit the results?  My concern is that there is a problem lurking in the other routines that has simply not tripped.

src/simulators/statevector/qubitvector.hpp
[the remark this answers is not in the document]
jagunnels: I think you could perform similar operations in single-precision with anything as new as Haswell.  https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=4919,2549,2553&text=single&techs=AVX,FMA .  Is the restriction to Macs because we think that these chips will be "new enough" Intel chips.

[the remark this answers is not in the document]
hitomitak: Ok. I try to add simd functions for single-precision.

src/simulators/statevector/kernel.hpp
hitomitak: I added the simd function for kernel4. If I use kernel.hpp, the calculation works correctly. Only simd function for kernel4 outputs always 0x0 as a result on Mac. 
For example, qft20 
`SIMD : "counts" {"0x0": 1}, Correct : "counts" {"0x181f8": 1}`
When the same code is executed on Linux server (Skylake), I can get the correct answer.

chriseclectic:
@hitomitak I sent a PR to your branch adding some basic unit tests for 3 and 4-qubit unitary gates.

hitomitak:
@chriseclectic Thank you for adding the test. I merged.

src/simulators/statevector/kernel_simd.hpp
```diff
@@ -0,0 +1,379 @@
...
+      auto badc = _mm256_permute_pd(mat_vec[i], 5);
+      mat_perm[i] = _mm256_mul_pd(badc, neg);
+    }
+}
+
+template<typename data_t=double>
+inline void kernel2(std::complex<data_t> *data_, const areg_t<4> &inds, __m256d *mat_vec, __m256d *mat_perm ){
+    __m256d vec[4];
+
+    vec[0] = loadbc((double *)&data_[inds[0]]);
```
chriseclectic: Is there a specific reason you are only using half available space in `__m256d` (If I understand correctly you are storing 2 doubles when they can hold 4 doubles)

hitomitak: This call stores 4 doubles. `loadbc` calls `_mm_load_pd` and `_mm256_broadcast_pd` so vec[0] has inds[0] and inds[0]. SIMD uses this 2 x inds[0] varibales.

chriseclectic: Ah I see!

hitomitak:
I shortened the kernel_init() code with for loop but I do not use the loop code in the kernel() code for the optimization. I fixed the bug of kernel4 and it works well so this code is called in qubit_vector.hpp.

chriseclectic:
Closing since @hhorii is working on a new SIMD implementation

PR #347, opened 2019-09-15
[WIP] Function to estimate running time of mps simulator

eladgoldman:
## Summary
A function that estimates the running time of the matrix product state simulator in linear time with respect to number of qubits and gates in the quantum circuit. Can be helpful for choosing the right simulator to use.

Function name: mps_running_time_estimator
Input: Quantum circuit.
Output: Array length n-1 with the estimated number of schmidt coefficients for every bipartition of the qubits. If d is the maximal number of schmidt coefficients, the estimated running time is O(d^3).

@yehuda-naveh 
@merav-aharoni


chriseclectic:
@eladgoldman python scripts shouldn't go in the src directory, I think this would make more sense as a utility function in the `aer.utils` directory if you clean it up and add some documentation.

PR #348, opened 2019-09-16
Add tox.ini to start using tox

mtreinish:
### Summary

This commit adds support for using tox by adding a tox.ini. This makes
it very simple for users and contributors to run tests and style checks
locally by just using the `tox` command. It will handle building and
installing aer and it's dependencies in a virtualenv to isolate it from
system python for testing. This is also a common tool for python
projects, as many (if not most) other python projects use it. Most of
the other qiskit elements also support using it.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,27 @@
+[tox]
+minversion = 2.1
+envlist = py35, py36, py37, lint
+skipsdist = True
+
+[testenv]
+install_command = pip install -U {opts} {packages}
+setenv =
+  VIRTUAL_ENV={envdir}
+  LANGUAGE=en_US
+  LC_ALL=en_US.utf-8
+deps =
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -0,0 +1,27 @@
...
+install_command = pip install -U {opts} {packages}
+setenv =
+  VIRTUAL_ENV={envdir}
+  LANGUAGE=en_US
+  LC_ALL=en_US.utf-8
+deps =
+  -r requirements-dev.txt
+  git+https://github.com/Qiskit/qiskit-terra.git
+commands =
+    python setup.py bdist_wheel -- -- -j4
```
atilag: Not quite sure if `-j4` would fail on Windows, anyway leave it there and I'll change it if I need to.

PR #352, opened 2019-09-19
Update device noise model to use gate_length

### LINKED ISSUE #336: `gate_time_values` fails to get correct gate times
### Informations

- **Qiskit Aer version**: qiskit-aer 0.3.0
- **Python version**: 3.7.3
- **Operating system**: macOS Mojave 10.14.5

Constructing gate times from properties does not return correct results.

It's possible I could just be using this function incorrectly, though I believe I'm understanding the function `gate_time_values` correctly based on the docstring. https://github.com/Qiskit/qiskit-aer/blob/master/qiskit/providers/aer/noise/device/parameters.py#L81

```
from qiskit import IBMQ
from qiskit.providers.aer.noise.device.parameters import gate_time_values
IBMQ.load_accounts()
pk = IBMQ.get_backend('ibmq_poughkeepsie')
props = pk.properties()
gate_times_out = gate_time_values(props)

# Compare gate time for U3 gate on qubit 0
res_1 = props.gates[3].parameters[1]
res_2 = gate_times_out[3]
```

I find that...
```
res_1 = Nduv(date=datetime.datetime(2019, 9, 2, 8, 39, 14, 753630, tzinfo=tzutc()), name='gate_length', unit='ns', value=206.2222222222222)
res_2 = ('u3', [0], None)
```

`res_2` should be `('u3', [0], 206.2222222222222)`

### Suggested solutions

Not sure, I haven't tried debugging it yet.

Also if it's relevant, here are other versions...
```
adapt-vqe-qiskit-algorithm   0.1.0
qiskit                       0.10.5
qiskit-aer                   0.3.0
qiskit-aqua                  0.5.3
qiskit-chemistry             0.5.1
qiskit-ibmq-provider         0.3rc1
qiskit-ignis                 0.2.0
qiskit-terra                 0.9.0
```

chriseclectic:
### Summary

(Most) Devices now return gate timing information in their properties under the key `"gate_length"`. Previously the device noise model was looking for `"gate_time"` key. This updates the device noise model to use the correct name.

It also renames the kwarg for adding custom gate times to `gate_lengths` and adds a deprecation warning if someone uses the old `gate_times` kwarg

Closes #336

### DIFF
--- a/qiskit/providers/aer/noise/device/parameters.py
+++ b/qiskit/providers/aer/noise/device/parameters.py
@@ -39,7 +39,6 @@ def gate_param_values(properties):
         qubits = gate.qubits
         # Check for gate time information
         gate_length = None  # default value
-        time_param = _check_for_item(gate.parameters, 'gate_time')  # deprecated
         time_param = _check_for_item(gate.parameters, 'gate_length')
         if hasattr(time_param, 'value'):
             gate_length = time_param.value
@@ -97,7 +96,6 @@ def gate_length_values(properties):
         name = gate.gate
         qubits = gate.qubits
         value = None  # default value
-        params = _check_for_item(gate.parameters, 'gate_time')  # deprecated
         params = _check_for_item(gate.parameters, 'gate_length')
         if hasattr(params, 'value'):
             value = params.value
--- a/qiskit/providers/aer/noise/device/parameters.py
+++ b/qiskit/providers/aer/noise/device/parameters.py
@@ -30,7 +30,7 @@ def gate_param_values(properties):
 
     Returns:
         list: A list of tuples (name, qubits, time, error). If gate
-              error or gate_time information is not available None
+              error or gate_length information is not available None
               will be returned for value.
     """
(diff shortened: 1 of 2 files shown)


qiskit/providers/aer/noise/device/parameters.py
```diff
@@ -96,7 +97,8 @@ def gate_time_values(properties):
         name = gate.gate
         qubits = gate.qubits
         value = None  # default value
-        params = _check_for_item(gate.parameters, 'gate_time')
+        params = _check_for_item(gate.parameters, 'gate_time')  # deprecated
```
atilag: This is useless, right? The line behind will always overwrite `params`

chriseclectic: Yeah you're right, should be a different variable name

chriseclectic: Since this wasn't actually used in old version of backends I'll just remove it

PR #355, opened 2019-09-23
Snapshot statevector and stabilizer extensions

### LINKED ISSUE #301: Statevector snapshot extension

This should have signature

```python
class SnapshotStatevector(Snapshot):
    __init__(self, label, num_qubits):
        super(label, "statevector", num_qubits, 0, None)
 

# Monkey patch function
def snapshot_statevector(self, label, qubits=None):
     # see base snapshot monkey patch for converting qubits to
     # be all qubits in circuit

circuit.snapshot_statevector = snapshot_statevector
```

The `num_qubits` parameter is actually ignored by Aer (it is always an all-qubit instruction), however it is used by Terra to make the snapshot act as a barrier.

chriseclectic:
### Summary

**EDIT: ** *breaking PR up into multiple smaller ones. This now just covers statevector and stabilizer snapshots.*

Fixes signatures of state snapshots:

Close #301, #302, #369 

### Details and comments

* Changes statevector, stabilizer, density matrix snapshot circuit function to not need qubit number since they are always all qubit
* Changes stabilizer snapshot to only return the list of stabilizer generators rather than the full clifford table.
* Changes order of stabilizer string serialization so that it works correctly with terra Pauli class qubit order.

- [x] Extension tests
- [x] Backend implementation tests
- [x] Add a line to the Changelog

### DIFF
--- a/test/terra/backends/qasm_simulator/qasm_snapshot.py
+++ b/test/terra/backends/qasm_simulator/qasm_snapshot.py
@@ -201,10 +201,6 @@ class QasmSnapshotStabilizerTests:
                 pauli_mat = Pauli.from_label(stab).to_matrix()
             val = statevector.conj().dot(pauli_mat.dot(statevector))
             if not np.isclose(val, 1):
-                print("IP VAL:", val)
-                print("STAB:", stabilizer)
-                print("GEN:", stab)
-                print("VEC:", statevector)
                 return False
         return True
 
@@ -331,8 +327,5 @@ class QasmSnapshotStabilizerTests:
                 for j, mem in enumerate(data['memory']):
                     statevec = statevec_targets[i].get(mem)
                     stabilizer = snaps[j]
-                    if not self.stabilizes_statevector(stabilizer, statevec):
-                        print("\n", circuit.draw())
-                        print("MEMORY:", mem)
                     self.assertTrue(
                         self.stabilizes_statevector(stabilizer, statevec))
--- a/test/terra/backends/qasm_simulator/qasm_snapshot.py
+++ b/test/terra/backends/qasm_simulator/qasm_snapshot.py
@@ -65,7 +65,9 @@ class QasmSnapshotStatevectorTests:
                                                          post_measure=False)
 
         qobj = assemble(circuits, self.SIMULATOR, shots=shots)
-        job = self.SIMULATOR.run(qobj, backend_options=self.BACKEND_OPTS)
+        job = self.SIMULATOR.run(qobj,
+                                 backend_options=self.BACKEND_OPTS,
+                                 validate=False)
         method = self.BACKEND_OPTS.get('method', 'automatic')
         if method not in QasmSnapshotStatevectorTests.SUPPORTED_QASM_METHODS:
             self.assertRaises(AerError, job.result)
@@ -94,7 +96,9 @@ class QasmSnapshotStatevectorTests:
                                                             post_measure=False)
 
         qobj = assemble(circuits, self.SIMULATOR, shots=shots)
-        job = self.SIMULATOR.run(qobj, backend_options=self.BACKEND_OPTS)
+        job = self.SIMULATOR.run(qobj,
+                                 backend_options=self.BACKEND_OPTS,
+                                 validate=False)
         method = self.BACKEND_OPTS.get('method', 'automatic')
         if method not in QasmSnapshotStatevectorTests.SUPPORTED_QASM_METHODS:
             self.assertRaises(AerError, job.result)
@@ -126,7 +130,9 @@ class QasmSnapshotStatevectorTests:
                                                          post_measure=True)
 
         qobj = assemble(circuits, self.SIMULATOR, memory=True, shots=shots)
-        job = self.SIMULATOR.run(qobj, backend_options=self.BACKEND_OPTS)
+        job = self.SIMULATOR.run(qobj,
+                                 backend_options=self.BACKEND_OPTS,
+                                 validate=False)
         method = self.BACKEND_OPTS.get('method', 'automatic')
         if method not in QasmSnapshotStatevectorTests.SUPPORTED_QASM_METHODS:
             self.assertRaises(AerError, job.result)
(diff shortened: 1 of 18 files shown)


src/simulators/statevector/statevector_state.hpp
```diff
@@ -634,18 +634,20 @@ void State<statevec_t>::snapshot_matrix_expval(const Operations::Op &op,
       first = false;
     else
       BaseState::qreg_.revert(true);
-
     // Apply each matrix component
     for (const auto &pair: param.second) {
-      const reg_t &qubits = pair.first;
+      reg_t sub_qubits;
+      for (const auto pos : pair.first) {
+        sub_qubits.push_back(qubits[pos]);
```
atilag: `pos` is not linear, right? ... otherwise a `std::copy()` is preferable for copying ranges of vectors

chriseclectic: `pos` could be any integers. Eg `[0, 3, 7]`

chriseclectic:
@atilag will break up into a couple of PRs. This one is ready for review and merging

test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -0,0 +1,338 @@
...
+        )
+        circuits = snapshot_state_circuits_deterministic(label,
+                                                         'statevector',
+                                                         post_measure=False)
+
+        qobj = assemble(circuits, self.SIMULATOR, shots=shots)
+        job = self.SIMULATOR.run(qobj, backend_options=self.BACKEND_OPTS)
+        method = self.BACKEND_OPTS.get('method', 'automatic')
+        if method not in QasmSnapshotStatevectorTests.SUPPORTED_QASM_METHODS:
+            self.assertRaises(AerError, job.result)
```
atilag: Can this really happen? As we are setting the `BACKEND_OPTS` to `{}`, the `method` variable will always contain the default value `automatic`, isn't it?

test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -0,0 +1,338 @@
...
+        else:
+            result = job.result()
+            self.is_completed(result)
+            self.compare_counts(result,
+                                circuits,
+                                counts_targets,
+                                delta=0.2 * shots)
```
atilag: Does `delta` need to grow with the number of shots?

test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -0,0 +1,338 @@
...
+                data = result.data(circuit)
+                snaps = self.statevector_snapshots(data, label)
+                self.assertTrue(len(snaps), 1)
+                target = statevec_targets[j]
+                value = snaps[0]
+                self.assertTrue(np.allclose(value, target))
+
+    def test_snapshot_statevector_post_measure_det(self):
+        """Test snapshot statevector after deterministic final measurement"""
+        shots = 10
```
atilag: These are 10 times less shots than "pre-measure" tests. What's the reason? (just want a better understanding) :)

test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -0,0 +1,338 @@
...
+                                delta=0.1 * shots)
+            # Check snapshots
+            for i, circuit in enumerate(circuits):
+                data = result.data(circuit)
+                snaps = self.stabilizer_snapshots(data, label)
+                for j, mem in enumerate(data['memory']):
+                    statevec = statevec_targets[i].get(mem)
+                    stabilizer = snaps[j]
+                    if not self.stabilizes_statevector(stabilizer, statevec):
+                        print("\n", circuit.draw())
```
atilag: We should be using the `logger` instead of `print`

test/terra/backends/test_qasm_matrix_product_state.py
```diff
@@ -18,16 +18,20 @@
...
 from test.terra.backends.qasm_simulator.matrix_product_state_measure import QasmMatrixProductStateMeasureTests
-
+# Snapshot tests
+from test.terra.backends.qasm_simulator.qasm_snapshot import QasmSnapshotStatevectorTests
 
 class TestQasmMatrixProductStateSimulator(common.QiskitAerTestCase,
                                           QasmMatrixProductStateMethodTests,
-                                          QasmMatrixProductStateMeasureTests):
+                                          QasmMatrixProductStateMeasureTests,
+                                          #QasmSnapshotStatevectorTests,  # FAILING
```
atilag: What's the plan for these tests then?

test/terra/extensions/__init__.py
```diff
@@ -0,0 +1,32 @@
+# This code is part of Qiskit.
```
atilag: What's the concept of `extensions` in this context?

atilag:
One of the test failed because of a timeout in MacOS provisioning... I'm rerunning the Job again

test/terra/backends/qasm_simulator/qasm_snapshot.py
chriseclectic: This is set by the test classes themselves, and checks that the unsupported instructions are caught and raise exception correctly (eg. stabilizer snapshot on density matrix method)

test/terra/backends/qasm_simulator/qasm_snapshot.py
chriseclectic: Delta needs to grow as number of shots gets smaller

test/terra/backends/qasm_simulator/qasm_snapshot.py
chriseclectic: They take longer to run because there is no measure sampling

test/terra/backends/qasm_simulator/qasm_snapshot.py
chriseclectic: Whoops, this shouldnt be there...

test/terra/backends/test_qasm_matrix_product_state.py
chriseclectic: See issue #371

test/terra/extensions/__init__.py
chriseclectic: Just copying terra/extensions which is where custom gates and instructions live there

PR #357, opened 2019-09-24
Added tests for the Fredkin gate

### LINKED ISSUE #274: Add tests for controlled-swap (Fredkin) gate

Add tests for controlled-swap gate for qasm simulator (statevector), statevector simulator, unitary simulator.

Like other tests in the qasm simulator these should test for execution in the standard basis gates (which will soon include `cswap` directly), the waltz basis gates (u1, u2, u3, cx), and the minimal basis gate set (cx, u3).

brandhsn:
### Summary

Fixes #274 

### Details and comments

The cswap gate (Fredkin) gate didn't have tests before. This PR adds tests for the cswap gate on the basis, minimal and waltz gate set and for the qasm (statevector), statevector and unitary simulator. Deterministic and undeterministic tests were added.

### TODO
- [x] Add a line to the changelog


atilag:
The CI is failiing because of the linter.
To pass the linter locally, just run:
```
$ pycodestyle --ignore=E402,W504 --max-line-length=100 qiskit/providers/aer
$ pylint -j 2 -rn qiskit/providers/aer
```

brandhsn:
Strange - the nondeterministic T gate test fails on OSX but not on Linux. I haven't touched these tests and it looks like the counts are off by 0.525% instead of the allowed 0.5%. Maybe the test run was unlucky? Should I increase the error margins to 0.6%?

atilag:
Yeah, we have faced similar situations before, tweaking the delta is the way to go as you suggested, so go ahead. I'd even say to increase it to 1.0%, maybe @chriseclectic can confirm.

chriseclectic:
For the failing test we should do what terra did and add a fixed seed for non-deterministic tests

chriseclectic:
@brandhsn looks like this needs rebasing after merging the cu1 test PR, after that it is good to go!

brandhsn:
> Thanks @brandhsn, this is great!

Sure, doing these tests was really a good introduction to Qiskit and Github! :-)

PR #359, opened 2019-09-25
Removing temporary files for the MPS sim

atilag:
### Summary
Cleaning up


atilag:
A new version of Pylint seems to have borken `master`, will fix new linter errors here.

PR #362, opened 2019-09-25
Fix lint failures caused by new pylint and pin

mtreinish:
### Summary

This commit fixes all the issues reported via the newest pylint release.
Several of the rules are bogus failures either because pylint is not
able to inspect c extensions or because it doesn't understand context
and are disabled. In other places they pointed out a couple of
optimizations that can be made to the code. After fixing all the
reported errors this also adds a version pin using a constraints file
to ensure that we have a stable base for lint checking. We can
investigate bumping that pin at the future on our terms instead of
new pylint releases breaking CI periodically.


atilag:
Ouch, I just merged something to fix the linter, but your approach is more complete. Can you fix the conflicts?

PR #363, opened 2019-09-26
GPU implementation of statevector simulator by using Thrust

doichanj:
### Summary
I totally rewrite the GPU implementation by using Thrust library.
So now the statevector simulator for GPU is implemented in single header file.

### Details and comments

It supports single node, multiple GPUs.
There is no support of auto configure by cmake currently.
I prepared separate Makefile example to build this implementation.
It should be built by using nvcc and using GCC 8.3.0 for backend compiler.


atilag:
Can we close #214 ?

chriseclectic:
Thanks @doichanj, there seems to be a rebase issue where a lot of files are showing up as "new" or "changed" that shouldn't be.

doichanj:
I only modified and added these 3 files from Qiskit Aer master

contrib/standalone/qasm_simulator.cpp
src/simulators/qasm/qasm_controller.hpp
src/simulators/statevector/qubitvector_thrust.hpp


> 
> 
> Thanks @doichanj, there seems to be a rebase issue where a lot of files are showing up as "new" or "changed" that shouldn't be.

doichanj:
I would like to close this request and I will make another request to clean up errors.

PR #365, opened 2019-09-26
Add Result and ExperimentResult structs

chriseclectic:
### Summary

Start of refactor of simulator result data types.  This adds C++ structs for simulator results rather than using pure json internally. The goal is that these classes can be wrapped into python using pybind or cython to directly build a python result object without going via json/str conversion for local simulation.

### Details and comments

* Adds `Result` class
* Adds `ExperimentResult` class
* Renames `OutputData` to `ExperimentData`

### DIFF
--- a/src/framework/results/experiment_result.hpp
+++ b/src/framework/results/experiment_result.hpp
@@ -15,7 +15,7 @@
 #ifndef _aer_framework_results_experiment_result_hpp_
 #define _aer_framework_results_experiment_result_hpp_
 
-#include "framework/results/data.hpp"
+#include "framework/results/experiment_data.hpp"
 
 namespace AER {
 
@@ -30,7 +30,7 @@ public:
   enum class Status {empty, completed, error};
 
   // Experiment data
-  OutputData data;
+  ExperimentData data;
   uint_t shots;
   uint_t seed;
   double time_taken;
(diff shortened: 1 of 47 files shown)


src/framework/results/experiment_result.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "framework/results/experiment_data.hpp"
+
+namespace AER {
+
+//============================================================================
+// Result container for Qiskit-Aer
+//============================================================================
+
+struct ExperimentResult {
```
atilag: This interface is very similar to `Result` one, Didn't want to inherit from it?

src/framework/results/experiment_result.hpp
```diff
@@ -0,0 +1,111 @@
...
+  // Serialize engine data to JSON
+  json_t json() const;
+};
+
+
+//------------------------------------------------------------------------------
+// Add metadata
+//------------------------------------------------------------------------------
+template <typename T>
+void ExperimentResult::add_metadata(const std::string &key, const T &meta) {
```
atilag: This is exactly the same method found in `Result::add_metadata()`... inheriting from 
`Result` will remove the need to duplicate code.

src/framework/results/experiment_result.hpp
chriseclectic: `Result` contains `ExperimentResult` objects as members though

atilag: Inherit by composition, perfect.

PR #373, opened 2019-10-01
Import sympy lazily.

kdk:
In the spirit of https://github.com/Qiskit/qiskit-terra/pull/3156 , it would be great to get `import sympy` out of the list of things that are run by default when `import qiskit` is run (since it takes >500ms). Since Aer is imported by default from qiskit (when installed,) it would be great to remove the top-level `import sympy` from `qiskit/providers/aer/noise/utils/noise_transformation.py`. This PR moves the sympy to function level imports.


atilag:
There seems to be a problem with MacOS jobs, not related to our code, so I'm merging this.

PR #374, opened 2019-10-01
Fix snapshot density matrix

### LINKED ISSUE #303: DensityMatrix snapshot extension

chriseclectic:
### Summary

Fixes density matrix snapshot and adds tests.

Closes #303 
~Blocked by #355~

### Details and comments

* Changes signature density matrix snapshot circuit function to not need qubit number, if it is not specified it will act on all qubits.
* Fixes qasm simulator to default to density matrix method if there is a density matrix snapshot (previously it would raise an exception of unsupported instruction for statevector simulator).
* Fixes density matrix to be an average snapshot instead of single shot snapshot
* Addes tests for density matrix extension and simulator implementation.


src/simulators/qasm/qasm_controller.hpp
```diff
@@ -527,10 +539,16 @@ QasmController::simulation_method(const Circuit &circ,
...
     // If we didn't select extended stabilizer above proceed to the default switch clause
     default: {
-      // Default method is statevector
+      // For default we use statevector followed by density matrix (for the case
+      // when the circuit contains invalid instructions for statevector)
+      if (validate_state(Statevector::State<>(), circ, noise_model, false)) {
+        return Method::statevector;
+      }
+      // If circuit contains invalid instructions for statevector throw a hail
+      // mary and try for density matrix.
```
atilag: Ave María!!

test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -30,6 +30,10 @@
     snapshot_state_pre_measure_statevector_nondeterministic,
     snapshot_state_post_measure_statevector_nondeterministic)
 
+def get_snapshots(data, label, snapshot_type):
+        """Format snapshots as list of Numpy arrays"""
```
atilag: I think this has one more level of indentation, right?

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -468,6 +468,18 @@ QasmController::simulation_method(const Circuit &circ,
                                   bool validate) const {
   // Check simulation method and validate state
   switch(simulation_method_) {
+    case Method::statevector: {
+      if (validate) {
+        if (simulation_precision_ == Precision::single_precision) {
+          Statevector::State<QV::QubitVector<float>> state; 
+          validate_state(state, circ, noise_model, true);
```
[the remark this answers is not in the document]
chriseclectic: the state arg is actually different (template param <float> or <double> in each branch);

test/terra/backends/qasm_simulator/qasm_snapshot.py
chriseclectic: indeed!

chriseclectic: Will merge here and fix in #380

PR #376, opened 2019-10-01
Add travis wait to brew installs

mtreinish:
### Summary

The brew installation of openblas is sometimes taking over 10min to
build and install in the osx travis jobs. To avoid the job timing out
waiting for brew to emit output this commit adds a travis_wait call to
all the brew calls to make the timeout independent from output from the
command.


atilag:
Thanks Matthew!

PR #378, opened 2019-10-02
Make qobj schema validation optional

chriseclectic:
### Summary

Fixes issue that using `validate=False` wasn't correctly skipping qobj schema validation, nor result object validation.

This is required for performance improvements in Aqua application algorithms.


qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -153,19 +154,21 @@ def _format_results(self, job_id, output, time_taken):
...
         return Result.from_dict(output)
 
-    def _validate_controller_output(self, output):
+    def _validate_controller_output(self, output, warn=True):
         """Validate output from the controller wrapper."""
         if not isinstance(output, dict):
-            logger.error("%s: simulation failed.", self.name())
-            if output:
-                logger.error('Output: %s', output)
+            if warn:
```
atilag: Ok, if we want to disable logging, the correct way to do it is by using python's ' logging' library. 
This is an example of a client code (the one importing Aer) that wants to disable logging:
```python
from qiskit import Aer
...
import logging
logging.getLogger("Aer").setLevel(logging.ERROR)
```
... this will only show ERRORs in the logging.

chriseclectic: How do you set it to not show any?

PR #381, opened 2019-10-03
GPU enabled Qiskit Aer by using Thrust (renewed)

doichanj:
### Summary
I totally rewrite the GPU implementation by using Thrust library.
So now the statevector simulator for GPU is implemented in single header file.

(I refreshed my repository because there were so many errors I could not handle.)

### Details and comments
It supports single node, multiple GPUs.
There is no support of auto configure by cmake currently.
I prepared separate Makefile (Makefile.Thrust) example to build this implementation.
It should be built by using nvcc and using GCC 8.3.0 for backend compiler.


src/simulators/qasm/qasm_controller.hpp
```diff
@@ -340,6 +347,8 @@ void QasmController::set_config(const json_t &config) {
   JSON::get_value(extended_stabilizer_measure_sampling_,
                   "extended_stabilizer_measure_sampling", config);
 
+  JSON::get_value(use_GPUs_,"GPU", config);
```
chriseclectic: Instead of checking for `"GPI"` in `backend_options` can we have it as an explicit method `"method": "statevector_gpu`

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -383,6 +392,19 @@ OutputData QasmController::run_circuit(const Circuit &circ,
     case Method::statevector:
       if (simulation_precision_ == Precision::double_precision) {
         // Double-precision Statevector simulation
+#ifdef QASM_THRUST
```
chriseclectic: If we add GPU as a separate method then it could be a separate case here. Something like:

```cpp
case Method::statevector_gpu:
#ifdef QASM_THRUST
  // usual execution code
#else
  // throw an exception that GPU not enabled
#endif
```

The rest of the controller would then need to be updated so that checks for `Method::statevector` also check for `Method::statevector_gpu`

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+ * that they have been altered from the originals.
+ */
+
+
+
+#ifndef _qv_qubit_vector_thrust_hpp_
+#define _qv_qubit_vector_thrust_hpp_
+
+#include <cuda.h>
+#include <cuda_runtime.h>
```
chriseclectic: Should these be hidden in an #ifdef guard?

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+#include <iostream>
+#include <sstream>
+#include <stdexcept>
+
+#include "framework/json.hpp"
+
+
+#include "simulators/statevector/qubitvector.hpp"
+
+#ifdef QASM_TIMING
```
chriseclectic: What is the purpose of this block? Is it for profiling?

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+			}
+			else{
+				cudaMallocManaged(&pData,sizeof(thrust::complex<data_t>) * data_size_);
+			}
+			m_useATS = 0;
+		}
+		data_ = reinterpret_cast<std::complex<data_t>*>(pData);
+
+		if(m_nDevParallel > 1){
+			/*
```
chriseclectic: Is this commented out code block supposed to be here?

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+	TimeEnd(QS_GATE_MULT);
+#endif
+
+}
+
+template <typename data_t>
+void QubitVectorThrust<data_t>::apply_multiplexer(const reg_t &control_qubits,
+                                            const reg_t &target_qubits,
+                                            const cvector_t<double>  &mat) {
+	printf(" apply_multiplexer NOT SUPPORTED : %d, %d \n",control_qubits.size(),target_qubits.size());
```
chriseclectic: Can you throw an exception here instead of printing a warning. This will let the simulator catch and return the error message to qiskit users in result object.

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+  // indexes([1,3], [1,3], k) -> [int(0b0a), int(0b1a), int(1b0a), (1b1a)]
+  // If the qubits were passed in reverse order it would swap qubit position in the list:
+  // indexes([3,1], [1,3], k) -> [int(0b0a), int(1b0a), int(0b1a), (1b1a)]
+  // Example: k=77, qubits=qubits_sorted=[1,4] ==> output=[297,299,313,315]
+  // input: k = 77  = 1001101
+  // output[0]: 297 = 100101001 (with 0's put into places 1 and 4).
+  // output[1]: 299 = 100101011 (with 0 put into place 1, and 1 put into place 4).
+  // output[2]: 313 = 100111001 (with 1 put into place 1, and 0 put into place 4).
+  // output[3]: 313 = 100111011 (with 1's put into places 1 and 4).
+  indexes_t indexes(const reg_t &qubits, const reg_t &qubits_sorted, const uint_t k) const;
```
chriseclectic: IT doesn't look like `index0` or `indexes` functions are used anywhere in code, so could they be removed?

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+			}
+		}
+	}
+};
+
+template <typename data_t>
+void QubitVectorThrust<data_t>::apply_mcy(const reg_t &qubits) {
+  // Calculate the permutation positions for the last qubit.
+  const size_t N = qubits.size();
+//		printf("  Y\n ");
```
chriseclectic: Forgot to remove debug line?

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+  // Calculate the swap positions for the last two qubits.
+  // If N = 2 this is just a regular SWAP gate rather than a controlled-SWAP gate.
+	const size_t N = qubits.size();
+
+	uint_t size = data_size_ >> N;
+	uint_t pos[2];
+
+	allocate_buffers(N);
+
+	pos[0] = (1ull <<
```
chriseclectic: forgot to remove?

chriseclectic: forgot to remove?

chriseclectic: unused debug

chriseclectic: These kinds of print statements will break being able to return a valid result object if it prints to `std::out`, it would be better if they could be added to result metadata somehow.

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+}
+
+template <typename data_t>
+void QubitVectorThrust<data_t>::apply_matrix(const reg_t &qubits,
+                                       const cvector_t<double> &mat)
+{
+	const size_t N = qubits.size();
+	uint_t size;
+
+//	printf(" Mat Mult : %d\n",N);
```
chriseclectic: delete

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+	}
+};
+
+template <typename data_t>
+void QubitVectorThrust<data_t>::apply_diagonal_matrix(const reg_t &qubits,
+                                                const cvector_t<double> &diag)
+{
+	const int_t N = qubits.size();
+
+//	printf(" Diag Mult : %d",N);
```
chriseclectic: delete

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -0,0 +1,3146 @@
...
+  // The matrix is input as vector of the column-major vectorized 1-qubit matrix.
+  void apply_matrix(const uint_t qubit, const cvector_t<double> &mat);
+
+  // Apply a N-qubit matrix to the state vector.
+  // The matrix is input as vector of the column-major vectorized N-qubit matrix.
+  void apply_matrix(const reg_t &qubits, const cvector_t<double> &mat);
+
+  // Apply a stacked set of 2^control_count target_count--qubit matrix to the state vector.
+  // The matrix is input as vector of the column-major vectorized N-qubit matrix.
+  void apply_multiplexer(const reg_t &control_qubits, const reg_t &target_qubits, const cvector_t<double> &mat);
```
chriseclectic: If the gpu doesn't support all the same gates for now as the CPU version, maybe we should add a separate `State` subclass for it that defines a different set of allowed gates. Hopefully this could be done with minimal code copying by something like:

```cpp
namespace AER {
namespace StatevectorThrust {

template <class statevec_t = QV::QubitVectorThrust<double>>
class State : public Statevector::State<statevec_t> {
  // Here we can override gate set methods that are different to
  // CPU statevector class to remove other gates
  
  // Return the string name of the State class
  virtual std::string name() const override {return "statevector_gpu";}

  virtual Operations::OpSet::optypeset_t allowed_ops() const override {
    return // list of osps if different to STatevector state
  }
  virtual stringset_t allowed_gates() const override {
    return // list of gates if different to Statevector::State;
  }

  // other methods should be same and inherited from base class
}
}}
```

src/simulators/qasm/qasm_controller.hpp
doichanj: I added "method": "statevector_gpu"

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: I added ifdef QASM_THRUST to cover whole of qubitvector_thrust.hpp

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: Yes it is used for optimization. I would like to keep them.

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: deleted

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: I added message by throwing an exception

src/simulators/qasm/qasm_controller.hpp
doichanj: implemented

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: removed

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: removed

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: removed

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: removed

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: I added statevector_gpu_state.hpp

src/simulators/statevector/qubitvector_thrust.hpp
jagunnels: There may be a good reason why this cannot be done, but it's sort of unusual in code that uses templates to have if-then statements regarding float vs. double.  If there is a good reason for this (for example, some decision that is made, not just sizeof-dependent choices), leave it, of course, but it is a bit cluttered.

doichanj:
I think Thrust only supports NVIDIA GPUs.
We have to implement with other C++ library to enable AMD GPUs (Unfortunately I do not have access to systems with AMD GPUs ...)
I think BOOST++ or C++ AMP are candidates of  libraries work on both NVIDIA and AMD.

doichanj:
I have added cmake configuration for GPU simulator. 
Now GPU simulator is automatically built by cmake if CUDA toolkit is installed in the system.

chriseclectic:
I managed to build and run the standalone simulator using OpenMP thrust backend on my Macbook,  but the Python version seems to be causing a seg fault.

With regards to the building I think we should do something like

1. Try and build the CUDA version of the simulator, and if CUDA isn't found, it shouldn't build the GPU method. In this case if a use tries to use `"method": "statevector_gpu"` an error should be raised. that it is not available.

2. Building with OpenMP/TBB backend should achieved by manual CMAKE flags that the user has to put in themselves. We can use these flags for the travis/azure builds for testing.

If we do this we don't need to include the thrust library in the source code (Thrust is included with CUDA, and for the debug code we can pass in a manual build flag to find the header files). However we will need at least two ifdef macros in the Source code: the whole gpu code can be hidden in an THRUST macro like before, with the added second one for THRUST_CUDA to enable the multi-gpu cuda code.

doichanj:
I found the combination of Clang + OpenMP has some issue. 
I think Clang 8 or later has native OpenMP supports that will work with Thrust, but Thrust does not work on Apple and Cygwin + Clang. 
I'm sorry but I do not have Mac for testing, so currently I disabled qubitvector_thrust when cmake does not find OpenMP and the compiler is Clang.

chriseclectic:
@doichanj I sent a PR to your fork for adding unittests for the gpu method

doichanj:
I modified cmake setting for Cython in qiskit/providers/aer/backends/wrappers/CMakeLists.txt
With this modification, I can build Cython modules with Thrsut GPU backend on Power9 + Tesla V100 machine, but I have not succeeded in building environment for testing yet.

I refferred to https://github.com/Qiskit/qiskit-aer/blob/master/CONTRIBUTING.md to setting up environment but it seems there are some more prerequisites to run python code with Aer.

chriseclectic:
@atilag @mtreinish this looks to be passing tests on Linux and Windows but build is failing for MacOS on travis. Could you look into the travis config for mac?

atilag:
Looks like a simple build error because some preprocessor definition (`AER_THRUST_NOT_SUPPORTED`) is not placed correctly.
I can fix this.

atilag:
A new `gpu` branch have been created starting from @doichanj changes. All the GPU related work should be merged there from now on, meaning new changes have to PR against `gpu` branch.
So, I'm closing this PR.

PR #393, opened 2019-10-07
Added support for reset() in MPS simulation method

merav-aharoni:
### Summary
Added support for the reset function that was missing before.
Updated the test file (as suggested by Chris), commenting out those tests that fail.


test/terra/backends/test_qasm_mps_simulator.py
```diff
@@ -0,0 +1,86 @@
...
+        QasmAlgorithmTests,
+        QasmAlgorithmTestsWaltzBasis,
+        QasmAlgorithmTestsMinimalBasis,
+        QasmUnitaryGateTests,
+        #QasmInitializeTests,  # CRASHES: uncaught exception
+        QasmReadoutNoiseTests,
+        QasmPauliNoiseTests,
+        QasmResetNoiseTests,
+        QasmSnapshotStatevectorTests,
+        QasmSnapshotStabilizerTests,
```
chriseclectic: I forgot to add the `QasmSnapshotProbabilitiesTests` which was recently merged to master

PR #399, opened 2019-10-15
pybind wrappers for C++ simulators

ereastman:
### Summary
This pull request contains:
- Pybind11 build procedures
- Pybind11 instantiation wrappers for controller_execute function 
- Pybind11 wrapper for nlohmann JSON library/class 

### Details and comments
[Pybind](https://pybind11.readthedocs.io/en/master/) is a header-only C++ library built specifically for exposing Python interfaces to existing C++ libraries. It can be installed with 
`$ pip install pybind11` or `$ conda install -c conda-forge pybind11`

The previous implementation of the [controller_execute](https://github.com/Qiskit/qiskit-aer/blob/2bd78751e6c68a861d898f0cb9e76c751c5f6a32/src/simulators/controller_execute.hpp#L30) function - now renamed to controller_execute_json - required the input qobj to be 1) converted to a python dictionary ([to_dict](https://github.com/Qiskit/qiskit-aer/blob/4f4fea8307335f70400f71756fda3aa009988b86/qiskit/providers/aer/backends/aerbackend.py#L59)) and then 2) dumped (serialized) to a json formatted [string](https://github.com/Qiskit/qiskit-aer/blob/4f4fea8307335f70400f71756fda3aa009988b86/qiskit/providers/aer/backends/aerbackend.py#L161) prior to being passed to the wrappers. Then, prior to simulation, the c++ simulators had to [parse](https://github.com/Qiskit/qiskit-aer/blob/2bd78751e6c68a861d898f0cb9e76c751c5f6a32/src/simulators/controller_execute.hpp#L32) (deserialize) the string into its own internal representation (json). After simulation, the Results had to be dumped to a [string](https://github.com/Qiskit/qiskit-aer/blob/2bd78751e6c68a861d898f0cb9e76c751c5f6a32/src/simulators/controller_execute.hpp#L42) on the c++ side, passed back through the wrappers, deserialized by the Python json parser into a python [dict](https://github.com/Qiskit/qiskit-aer/blob/4f4fea8307335f70400f71756fda3aa009988b86/qiskit/providers/aer/backends/aerbackend.py#L134), and then converted to a [Results](https://github.com/Qiskit/qiskit-aer/blob/4f4fea8307335f70400f71756fda3aa009988b86/qiskit/providers/aer/backends/aerbackend.py#L175) object using Results.from_dict(). 

With this PR, the wrappers are now capable of [casting](https://github.com/Qiskit/qiskit-aer/blob/2bd78751e6c68a861d898f0cb9e76c751c5f6a32/qiskit/providers/aer/backends/wrappers/qasm_bindings.cc#L10) python objects directly to the [json](https://github.com/Qiskit/qiskit-aer/blob/2bd78751e6c68a861d898f0cb9e76c751c5f6a32/src/simulators/controller_execute.hpp#L46) representation used by the simulators. This gives the option of bypassing the serialization step on both sides of the simulation but still requires that the qobj (results) can be converted to a dict (can be built from a dict) - the option to accept json directly has been kept in place.

Performance Testing Results - these tests compare the time it takes to serialize a Python dictionary (qobj) into a C++ nlohmann::json object (input) using the old method (labeled json) vs. the method used in this PR (labeled pybind). Custom scripts were used and may be available upon request. The x-axis should be labeled as size (in memory) of the Qobj and Results object:
image


.travis.yml
```diff
@@ -25,6 +25,7 @@ sudo: false
 stage_generic: &stage_generic
   install:
     - pip install cython
+    - pip install pybind11
```
atilag: Instead of adding pybind11 here, add it to the `requirements-dev.txt` file

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -110,20 +110,17 @@ def _run_job(self, job_id, qobj, backend_options, noise_model, validate):
...
             self._validate(qobj, backend_options, noise_model)
-        qobj_str = self._format_qobj_str(qobj, backend_options, noise_model)
-        output = json.loads(self._controller(qobj_str).decode('UTF-8'))
+        output = self._controller(self._format_qobj(qobj, backend_options, noise_model))
         self._validate_controller_output(output)
         end = time.time()
         return self._format_results(job_id, output, end - start)
 
-    def _format_qobj_str(self, qobj, backend_options, noise_model):
+    def _format_qobj(self, qobj, backend_options, noise_model):
```
atilag: Why changing the name? What's the link with integrating `pybind11`?

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -19,7 +19,7 @@
 from qiskit.providers.models import QasmBackendConfiguration
 from .aerbackend import AerBackend
 # pylint: disable=import-error
-from .qasm_controller_wrapper import qasm_controller_execute
+from .qasm_controller_wrapper import qasm_controller_execute_new
```
atilag: no `_new` postfix, we will entirely replace `qasm_controller_execute` because the change should has no side effects, right?

qiskit/providers/aer/backends/wrappers/CMakeLists.txt
```diff
@@ -26,11 +26,28 @@ if(APPLE)
     unset(PYTHON_LIBRARIES)
 endif()
 
+set(_find_pybind_includes_command "
```
atilag: CMake can be tough in the beginning :(
The correct way to implement any dependency on CMake is using: `find_package()` function.
There are instructions on how to implement pybind11 with CMake [here](https://pybind11.readthedocs.io/en/stable/compiling.html#advanced-interface-library-target).
Basically, `find_package(pybind11 REQUIRED)` should be smart enough to find the installation of pybind and export all the functions and set all the variables you need to use later. Then you just need to tell CMake to link with pybind using this other function:
```
target_link_libraries(qasm_controller_wrapper
    ${AER_LIBRARIES}
    ${PYTHON_LIBRARIES}
    pybind11::embed)
```
Try implementing this way, as the doc says and ping me if you run into problems.

qiskit/providers/aer/backends/wrappers/CMakeLists.txt
```diff
@@ -26,11 +26,28 @@ if(APPLE)
...
+    message(FATAL_ERROR "COULD NOT FIND PYBIND!")
+endif()
 
 # QASM Controller
 
-add_cython_target(qasm_controller_wrapper qasm_controller_wrapper.pyx CXX)
-add_library(qasm_controller_wrapper MODULE ${qasm_controller_wrapper})
+add_library(qasm_controller_wrapper MODULE qasm_bindings.cc)
+
+#add_cython_target(qasm_controller_wrapper qasm_controller_wrapper.pyx CXX)
```
atilag: if you can make it work as the doc says, we don't need to comment any code, just remove it. (We always have git history if we want to take it back)

qiskit/providers/aer/backends/aerbackend.py
ereastman: _format_qobj_str previously created a dict (.to_dict) and then converted that dict to a json object and returned a string (json.dumps) representation - the new implementation returns a python object (dict) which is then sent directly through to the c++ wrappers - that python object is then converted directly to the binary c++ json object (thus skipping the serialization and 'stringification' on the python side)
TLDR and to respond directly to the question, the function name was changed because _format_qobj_str no longer returns a string

qiskit/providers/aer/backends/qasm_simulator.py
ereastman: yep, very true - I plan to adjust this in some way but I was wondering how the dependency of pybind should be added 
It could be done such that both cython and pybind wrappers are generated as separate modules - this could be useful for testing, comparing the two implementations, as well as introducing the new dependency in a 'soft' way - i.e. not breaking the builds of everybody who is currently building with cython and does not have pybind installed

qiskit/providers/aer/backends/wrappers/CMakeLists.txt
ereastman: Yes - so that is how we used to deal with the pybind package when we were building with conda - the unfortunate part of using pip is that the pypi package does not include the cmake modules - if you have pybind11 pip installed already you can run 'pip show -f pybind11' to see that the only files included are the c++ header files 
I really don't know why pybind/pypi decided to do it this way - the conda package is different in that its installation includes the cmake modules (although perhaps in a non-standard location)

qiskit/providers/aer/backends/qasm_simulator.py
atilag: I guess that if all our test suite pass with the new pybind backends, everything should be fine. As the interface with Python is the same no matter what cython or pybind version of the backend we are using, let's keep the Cython files for a while and add a flag to the build system so we can build the cython version whenever we want to, thoughts?
```
python ./setup.py bdist_wheel -- -DCYTHON_BACKENDS=True
```
I can implement this and PR your branch.

qiskit/providers/aer/backends/wrappers/CMakeLists.txt
atilag: Oh! this very unfortunate... seems like this problem was already brought here: https://github.com/pybind/pybind11/issues/1628. 
Ok, I'd move this code into it's own `cmake/pybind11Config.cmake` file so we can use `find_package()` in config mode:
```
find_package(pybind11 CONFIG REQUIRED)
```
This will look into our  `cmake/` directory for `pybind11Config.cmake` file and export PYBIND_INCLUDE_DIRS from there. I can help you with this too.

qiskit/providers/aer/backends/qasm_simulator.py
ereastman: Shoot actually the interfaces aren't the same - *_control_execute_new doesn't take in a string (like the cython ones) right now - although I can probably add a relatively simple 'if isinstance(string)' to the pybind wrappers to sub-set the cython functionality (at hopefully little cost to performance)

ereastman: The pybind dependency is extremely lightweight - pip/conda install pybind11 in probably 90% of cases and a simple fix in any others - so I'm going to just remove the cython bindings - I did add *_controller_execute_json method in case we ever wanted to use the json string interface (and renamed the new one to *_controller_execute)

qiskit/providers/aer/backends/wrappers/CMakeLists.txt
ereastman: I moved the finding of the pybind headers to pybind11Config.cmake - I think we may want to implement some kind of exception for the shadowed-declaration warnings that are produced by pybind since they're quite excessive and probably (?) not relevant - thoughts?

qiskit/providers/aer/backends/wrappers/CMakeLists.txt
ereastman: I removed the commented sections

ereastman:
@atilag I'm stuck on the Windows error - looks like it can't find pybind11Config.cmake despite it clearly being there - perhaps I should rename to Findpybind11.cmake to match the others (FindCython.cmake, FindPythonExtensions.cmake, etc.)? I don't have VisualStudio installed so it's difficult to test 

CMake Error at qiskit/providers/aer/backends/wrappers/CMakeLists.txt:5 (find_package):
  Could not find a package configuration file provided by "pybind11" with any
  of the following names:

    pybind11Config.cmake
    pybind11-config.cmake

ereastman:
@atilag @chriseclectic Windows build problems appear to have been fixed by changing the pybind11config.cmake to FindPybind11.cmake (kind of weird based on the error messages I was getting)

There's a slight problem with the CLA - I think I was committing from my local machine while not having set up my config user.email properly for this account - fixed now but all the prior commits still don't have my email attached - any recommendations on how to fix? 

Anyway I think the code itself is ready for a second look - perhaps further improvements to results processing/complex numbers/exp parameterization should be made on a separate PR - we're already getting some pretty good performance improvements with these changes.

src/framework/json.hpp
```diff
@@ -256,6 +313,173 @@ void std::to_json(json_t &js, const std::map<uint64_t, T1, T2> &map) {
...
+    if (buf.ndim != 1) {
+        throw std::runtime_error("Number of dims must be 1");
+    }
+
+    T *ptr = (T *) buf.ptr;
+    size_t D0 = buf.shape[0];
+
+    //std::cout << "1-d conversion: " << X << std::endl;
+
+    std::vector<T> tbr;
```
atilag: What does `tbr` means? Can we change the name to something more descriptive?

src/framework/json.hpp
```diff
@@ -256,6 +313,173 @@ void std::to_json(json_t &js, const std::map<uint64_t, T1, T2> &map) {
...
+    py::buffer_info buf = arr.request();
+    if (buf.ndim != 2) {
+        throw std::runtime_error("Number of dims must be 2");
+    }
+
+    T *ptr = (T *) buf.ptr;
+    size_t D0 = buf.shape[0];
+    size_t D1 = buf.shape[1];
+
+    //std::cout << "2-d conversion: " << D0 << "x" << D1 << std::endl;
```
atilag: Remove this line

src/framework/json.hpp
```diff
@@ -256,6 +313,173 @@ void std::to_json(json_t &js, const std::map<uint64_t, T1, T2> &map) {
...
+    }
+
+    T *ptr = (T *) buf.ptr;
+    size_t D0 = buf.shape[0];
+
+    //std::cout << "1-d conversion: " << X << std::endl;
+
+    std::vector<T> tbr;
+    for (size_t n0 = 0; n0 < D0; n0++)
+        tbr.push_back(ptr[n0]);
```
atilag: hmm wondering if `py::array_t<T, py::array::c_style>` can be used with `std::copy()` so we just:
```c++
std::copy(arr.begin(), arr.end(), std::back_inserter(tbr));
```
Did you try it?
If this work, this method is a one liner
... and we could get rid of all the other dimensional `numpy_to_json` methods

src/framework/json.hpp
ereastman: 'to be returned' (sorry, I think that's a personal thing) - equivalent to 'ret' and 'out' (same file, to/from_json for complexes) - I'll change to 'ret'/'out' to be consistent and add a comment

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -130,37 +130,46 @@ def _run_job(self, job_id, qobj, backend_options, noise_model, validate):
...
+                    if hasattr(val, 'to_dict'):
+                        config['backend_options'][key] = val.to_dict()
+            elif not isinstance(config['backend_options'], list):
+                raise ValueError("config[backend_options] must be a dict or list!")
+        # Double-check noise_model is a dict type
+        if 'noise_model' in config and not isinstance(config["noise_model"], dict):
+            if hasattr(config["noise_model"], 'to_dict'):
+                config["noise_model"] = config["noise_model"].to_dict()
+            else:
+                raise ValueError("noise_model must be a dict : " + str(type(config["noise_model"])))
```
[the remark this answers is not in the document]
ereastman: addressed in https://github.com/Qiskit/qiskit-aer/pull/399/commits/988a8e668e0ab77ce8c9b9d402d8fc889d3e0d52

src/framework/json.hpp
```diff
@@ -256,6 +313,173 @@ void std::to_json(json_t &js, const std::map<uint64_t, T1, T2> &map) {
...
+json_t numpy_to_json_1d(py::array_t<T, py::array::c_style> arr) {
+    py::buffer_info buf = arr.request();
+    if (buf.ndim != 1) {
+        throw std::runtime_error("Number of dims must be 1");
+    }
+
+    T *ptr = (T *) buf.ptr;
+    size_t D0 = buf.shape[0];
+
+    //std::cout << "1-d conversion: " << X << std::endl;
```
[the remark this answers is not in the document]
ereastman: addressed in https://github.com/Qiskit/qiskit-aer/pull/399/commits/988a8e668e0ab77ce8c9b9d402d8fc889d3e0d52

src/framework/json.hpp
ereastman: addressed in https://github.com/Qiskit/qiskit-aer/pull/399/commits/988a8e668e0ab77ce8c9b9d402d8fc889d3e0d52 (I just added an explanation where it is used because any name 'ret', 'out' would be equally non-descript)

src/framework/json.hpp
ereastman: addressed in https://github.com/Qiskit/qiskit-aer/pull/399/commits/988a8e668e0ab77ce8c9b9d402d8fc889d3e0d52

src/framework/json.hpp
ereastman: So this one is a bit nuanced - for the 1-d case, yes, it's quite simple and there are for sure more efficient ways - for example just doing `return std::vector<T>(arr, arr + D0);` would avoid the dynamic resizing caused by initializing an empty vector and then appending (which both the current and your suggested solution suffer from).

I'll take another look at these and see if I can't find a better solution because I think I know what you're getting at. 

As an aside, the reason I was semi-okay with the way they are now is that it clearly shows how they could be extended to 4,5,...-d tensors by someone who has no previous knowledge. The three functions follow a similar pattern with the most significant difference being the indexing function in the middle of the nested loops: 
1-d: `n0`
2-d: `n1+D1*n0`
3-d: `n2 + D2*(n1 + D1*n0)`
So it basically shows the [c-style array indexing equation](https://eli.thegreenplace.net/2015/memory-layout-of-multi-dimensional-arrays/), which I think would make it easier for whoever wanted to add 4-d, 5-d, etc. 

Of course, it would be better if it just handled any dimension, perfectly efficiently, right now! Which is what I will look into.

qiskit/providers/aer/backends/wrappers/qasm_bindings.cc
```diff
@@ -0,0 +1,13 @@
+#include <pybind11/pybind11.h>
+
+#include "simulators/qasm/qasm_controller.hpp"
+#include "simulators/controller_execute.hpp"
+#include <iostream>
+#include "framework/json.hpp"
+
+PYBIND11_MODULE(qasm_controller_wrapper, m) {
```
chriseclectic: When switching to PyBind should these files remain in the qiskit folder or be moved somewhere into `src` folder? ie `src/pybind`?

src/framework/json.hpp
```diff
@@ -22,9 +22,20 @@
 #include <map>
 #include <vector>
 
+#include <pybind11/pybind11.h>
```
chriseclectic: Can we compartmentalize this so that the standalone simulator doesn't require Pybind to build?

qiskit/providers/aer/backends/wrappers/qasm_bindings.cc
ereastman: We could go either way - would just need to adjust the CMakeLists.txt 
- benefit of the current structure is that the source tree reflects the eventual install tree structure 
- benefit of moving would be keeping all c++ source code together 
There are pluses and minuses both ways but whichever you feel is better I can make the changes

src/framework/json.hpp
ereastman: yes - didn't realize this would have the effect of introducing pointless dependencies to the standalone but now I do - I can move the pybind stuff to its own header file and only include it from the wrappers

qiskit/providers/aer/backends/wrappers/qasm_bindings.cc
ereastman: Going along with this - we could also combine the three qasm, unitary, statevector wrappers into a single file / module - more benefits in terms of maintenance here, same drawbacks of no longer reflecting the install tree structure

src/framework/json.hpp
ereastman: addressed in https://github.com/Qiskit/qiskit-aer/pull/399/commits/7c0784fb72e7cbff2cc9b1d51ab10a2f991bb1ff

src/framework/pybind_json.hpp
```diff
@@ -0,0 +1,268 @@
...
+json_t numpy_to_json_1d(py::array_t<T, py::array::c_style> arr) {
+    py::buffer_info buf = arr.request();
+    if (buf.ndim != 1) {
+        throw std::runtime_error("Number of dims must be 1");
+    }
+
+    T *ptr = (T *) buf.ptr;
+    size_t D0 = buf.shape[0];
+
+    std::vector<T> tbr; // to be returned
```
atilag: oh, ok. There's no need for a comment if you call the variable: `to_be_returned` instead of ``tbr` :) see: https://en.wikipedia.org/wiki/Self-documenting_code
I think naminig it `ret` is fine, it's like a standard-de-facto.

atilag:
@ereastman there seems to be some tests failing, can you take a look?

ereastman:
@atilag - I'm looking into the Windows build errors - It's a weird one but I think its known https://github.com/nlohmann/json/issues/1408 

Also, making sure you see that I consolidated the wrappers in https://github.com/Qiskit/qiskit-aer/pull/399/commits/54f858cc8cfe2a65e22e57d6e6333d1c80b24d01 - I'm pretty sure it decreased the compile time by a good amount in addition to making maintenance/debugging easier (less boilerplate in wrappers and CMakeLists) and helping me in the result wrapper implementation.

atilag:
>@atilag - I'm looking into the Windows build errors - It's a weird one but I think its known nlohmann/json#1408 - appears to be fixed now but for our Windows users I would keep an eye out for snprintf compiler errors

Oh! great, I see that you managed to fix it :)

> Also, making sure you see that I consolidated the wrappers in 54f858c

Yes! I liked the idea.  Good job,  I think this is ready to merge.

atilag:
This PR has been superseded by #505.

PR #400, opened 2019-10-16
[WIP] Add prob snapshots to stabilizer method

chriseclectic:
### Summary

Adds snapshot types `"probabilities"` and `"probabilities_with_variance"` to the stabilizer QasmSimulator method.

### Details and comments

These methods were already implemented, however the snapshots were missing from the allowed_snapshots set for the simulation method so could not be used.

Also auto formatted cpp file with clang-format


chriseclectic:
Closing since this needs more work

PR #401, opened 2019-10-16
Fix sdist to always attempt to build

mtreinish:
### Summary

This commit fixes an issue with the sdist if you ran in an environment
where scikit-build was not installed. scikit-build is a build time
dependency for installing aer from source, however the sdist treated
the lack of scikit-build as not fatal. Instead it would silently build a
dummy package which would just build an empty package with only
metadata. This leave users attempting to build from sdist very confused
because if they were missing scikit-build for any reason there is no
output regarding that or any errors. But instead when they go to use aer
it's not found because it was not actually installed.

### Details and comments

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,10 @@
 import os
 try:
     from skbuild import setup
-    dummy_install = False
-except:
-    print(""" WARNING
-              =======
-              scikit-build package is needed to build Aer sources.
-              Please, install scikit-build and reinstall Aer:
-              pip install -I qiskit-aer """)
-    from setuptools import setup
-    dummy_install = True
+except ImportError:
+    import subprocess
+    subprocess.call([sys.executable, '-m', 'pip', 'install', 'scikit-build'])
+    from skbuild import setup
 from setuptools import find_packages
 
 requirements = [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -57,6 +52,7 @@ def find_qiskit_aer_packages():
         "Topic :: Scientific/Engineering",
     ],
     install_requires=requirements,
+    setup_requires=['scikit-build', 'cmake', 'ninja', 'Cython'],
```
atilag: We don't need `ninja`, not even on Windows. scikit-build tries `ninja` as the first option always (I think this is going to change soon) , but if `ninja` is not found, it jumps to the next toolchain.

mtreinish: I just copied what was in the pyproject.toml file I can change this list to be whatever set of tools you think makes sense.

atilag: Ok, just removes `ninja` and I'll be merging
Thanks!

mtreinish: Fixed in: https://github.com/Qiskit/qiskit-aer/pull/401/commits/f344eedd46038414a38da62d1b33b1de080c88a3

PR #404, opened 2019-10-21
Improve additional data containers in ExperimentData

chriseclectic:
### Summary

* Adds complex vector and complex matrix additional data containers to the `ExperimentData` class.
* Updates JSON data containers to be `unordered_map<string, json>` so that JSON is only used for the inner data type.

### Details and comments

This is a first step for speeding up the statevector and unitary simulators by only converting to JSON at the very end of simulation. Follow up work is to bypass this conversion step entirely and copy the vector and matrix data directly into a numpy array in python for local simulation.

### DIFF
--- a/src/framework/results/experiment_data.hpp
+++ b/src/framework/results/experiment_data.hpp
@@ -80,18 +80,30 @@ class ExperimentData {
 
   // Add new data at the specified key.
   // If they key already exists this will override the stored data
+  // This will use the json conversion method `to_json` for data type T
+  // if it isn't a natively supported data type.
+  // Current native types are: cvector_t, cmatrix_t.
   template <typename T>
-  void add_additional_data(const std::string &key, const T &data);
+  void add_additional_data(const std::string &key, T &&data);
+
+  // Erase additional data stored at the provided key.
+  // This will erase from all additional data containers
+  void erase_additional_data(const std::string &key);
 
   //----------------------------------------------------------------
   // Metadata
   //----------------------------------------------------------------
 
+  // Access metadata map
   stringmap_t<json_t> &metadata() { return metadata_; }
   const stringmap_t<json_t> &metadata() const { return metadata_; }
 
+  // Add new data to metadata at the specified key.
+  // This will use the json conversion method `to_json` for data type T.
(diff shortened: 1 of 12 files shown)


src/framework/results/experiment_data.hpp
```diff
@@ -298,84 +379,87 @@ ExperimentData& ExperimentData::combine(ExperimentData &data) {
...
-    tmp["memory"] = memory_;
+  if (return_counts_ && counts_.empty() == false) tmp["counts"] = counts_;
+  if (return_memory_ && memory_.empty() == false) tmp["memory"] = memory_;
   if (return_register_ && register_.empty() == false)
     tmp["register"] = register_;
-  // Average snapshot data
+
+  // Add additional data
+  for (const auto &pair : additional_json_data_) {
+    tmp[pair.first] = pair.second;
```
atilag: If `json_t` type implements move constructor, I guess we can use move semantics here to avoid copying:
```suggestion
    tmp[pair.first] = std::move(pair.second);
```

atilag: We might want to double check that `json_t` (a.k.a: `nlohmann::json`) actually has a move constructor/assignament operator.

src/framework/results/experiment_data.hpp
```diff
@@ -177,117 +223,152 @@ void ExperimentData::add_memory_count(const std::string &memory) {
...
+void ExperimentData::add_additional_data(const std::string &key, json_t &&data) {
+  check_reserved_key(key);
   if (return_additional_data_) {
-    json_t js = data; // use implicit to_json conversion function for T
-    if (JSON::check_key(key, additional_data_))
-      additional_data_[key].update(js.begin(), js.end());
-    else
-      additional_data_[key] = js;
+    erase_additional_data(key);
+    additional_json_data_[key] = data;
```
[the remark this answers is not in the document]
atilag: Yes it is! :)

src/framework/results/experiment_data.hpp
chriseclectic: The serialization method has to be `const` on original object so can't use move anyway

PR #406, opened 2019-10-21
Linear algebra operations for STL types

chriseclectic:
### Summary

Adds utility functions for `add`, `iadd`, `sub`, `isub`, `mul`, `imul`, `div`, `idiv`, `square` for generic types that support these functions, and specialized functions for `std::array`, `std::vector` `std::map`, `std::unordered_map`, `matrix` and `json` types.

### Details and comments

This is needed as a first step for adding improved average snapshot containers with non-JSON data storage types in #407

### DIFF
--- a/src/framework/linalg/linops/linops_map.hpp
+++ b/src/framework/linalg/linops/linops_map.hpp
@@ -31,8 +31,8 @@ namespace Linalg {
 //----------------------------------------------------------------------------
 template <class T1, class T2, class T3, class T4,
           typename = enable_if_numeric<T2>>
-std::map<T1, T2, T3, T4> add(const std::map<T1, T2, T3, T4>& lhs,
-                             const std::map<T1, T2, T3, T4>& rhs) {
+decltype(auto) add(const std::map<T1, T2, T3, T4>& lhs,
+                   const std::map<T1, T2, T3, T4>& rhs) {
   std::map<T1, T2, T3, T4> result = lhs;
   for (const auto& pair : rhs) {
     result[pair.first] = std::plus<T2>()(result[pair.first], pair.second);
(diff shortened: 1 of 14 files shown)


src/framework/linalg/linops/linops_map.hpp
```diff
@@ -0,0 +1,190 @@
...
+// This defines functions add, sub, mul, div and iadd, imul, isub, idiv
+// that work for numeric map types
+
+//----------------------------------------------------------------------------
+// Linear operations
+//----------------------------------------------------------------------------
+template <class T1, class T2, class T3, class T4,
+          typename = enable_if_numeric<T2>>
+std::map<T1, T2, T3, T4> add(const std::map<T1, T2, T3, T4>& lhs,
+                             const std::map<T1, T2, T3, T4>& rhs) {
```
atilag: There's a neat C++14 feature that will make this code a little less verbose:
```c++
decltype(auto) add(const std::map<T1, T2, T3, T4>& lhs,
                   const std::map<T1, T2, T3, T4>& rhs) {
```

src/framework/linalg/linops/linops_unordered_map.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+// This defines functions add, sub, mul, div and iadd, imul, isub, idiv
+// that work for numeric unordered_map types
+
+//----------------------------------------------------------------------------
+// Linear operations
+//----------------------------------------------------------------------------
+template <class T1, class T2, class T3, class T4, class T5,
+          typename = enable_if_numeric<T2>>
+std::unordered_map<T1, T2, T3, T4, T5> add(
```
atilag: Reminder: you can use `decltype(auto)` as a return type here as well :)

chriseclectic: Good to know

PR #418, opened 2019-10-31
Add automatic label detection to NoiseModel

chriseclectic:
### Summary

Improves noise model so that if gate objects are used the label and basis_gates can be automatically inferred when adding quantum errors. This simplifies adding custom noisy gates which can then be used with the transpiler using `NoiseModel.basis_gates` for unrolling.

### Details and comments

For example, if we wish to add noise only to a labelled X-gate we can now do the following:

```python
from qiskit.extensions import XGate
from qiskit.providers.aer import noise

# Create noisy gate object and error
noisy_x = XGate(label="x_noise")
error_x = noise.errors.depolarizing_error(0.1, 1)

# Add to error model
nm = noise.NoiseModel()
nm.add_all_qubit_quantum_error(error_x, noisy_x)
print(nm)

# prints
"""
NoiseModel:
  Basis gates: ['cx', 'id', 'u3', 'x']
  Instructions with noise: ['x_noise']
  All-qubits errors: ['x_noise']
"""
```

### DIFF
--- a/qiskit/providers/aer/noise/noise_model.py
+++ b/qiskit/providers/aer/noise/noise_model.py
@@ -699,20 +699,19 @@ class NoiseModel:
         """Return two lists of instruction name strings and label strings."""
         if not isinstance(instructions, (list, tuple)):
             instructions = [instructions]
-        inst_labels = []
-        inst_names = []
+        names_labels = []
         for inst in instructions:
             # If instruction does not have a label we use the name
             # as the label
             if isinstance(inst, Instruction):
-                inst_names.append(inst.name)
-                inst_labels.append(getattr(inst, 'label', inst.name))
+                name = inst.name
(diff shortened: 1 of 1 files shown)


qiskit/providers/aer/noise/noise_model.py
```diff
@@ -698,19 +695,24 @@ def from_dict(noise_dict):
                 raise NoiseError("Invalid error type: {}".format(error_type))
         return noise_model
 
-    def _instruction_names(self, instructions):
-        """Return a list of instruction name strings for input instructions."""
+    def _instruction_names_labels(self, instructions):
+        """Return two lists of instruction name strings and label strings."""
```
yaelbh: Returns one list of pairs. By the way this could be directly implemented accordingly, instead of creating two lists and zipping them, but two lists + zip is also OK.

Issue #389, opened 2019-10-04
Add Pauli expectation value snapshot to stabilizer qasm method


chriseclectic:
### What is the expected behavior?

Add support for the following expectation value snapshots to the `Stabilizer::State` class in `stabilizer_state.hpp`

* [ ] `"expectation_value_pauli"
* [ ] `"expectation_value_pauli_with_variance"

yaelbh:
# Pauli expectation value

Suppose that a circuit has 5 qubits and the observable is `0.2IIXZI + 0.7ZXYII`.
Iterate on the Pauli operators. In the first iteration, take care of IIXZI, and in the second iteration take care of ZXYII.

For IIXZI:
- Create a copy of the state.
- Turn qubit 2 to the standard basis by applying a Hadamard gate.
- Call the probabilities snapshot for qubits 1 and 2.
- For each of the 4 basis states in the probabilities snapshot, the value is 0.2 multiplied by 1 or -1, depending on the parity of the 0's.
- Sum the values, weighted by the probabilities.

Apply the same steps for ZXYII and sum the values of the Pauli operators.

# Probabilities snapshot

The algorithm for the Pauli expectation value calls the probabilities snapshot. According to @chriseclectic, the code for the stabilizer probabilities snapshot does not work properly for subsets of qubits. The following should work:

Assume a circuit of 5 qubits and a probabilities snapshot on qubits 0, 2, and 3.
- Create a copy of the state.
- Check if qubit 0 would have a deterministic outcome if we measure it.
  - If the outcome is deterministic 0 then simulate a measurement of 0 for qubit 0, and call recursively for qubits 2 and 3. The result of the recursive call is a probabilities vector: 4 probabilities assigned to 4 basis states. Concatenate "0" to each basis state and return.
  - Similarly if the outcome is deterministic 1.
  - If the outcome is random then:
    - Simulate a measurement of 0 for qubit 0, and call recursively for qubits 2 and 3. In the resulting probability vector, concatenate "0" to each basis state and multiply each coefficient by 0.5.
    - Similarly for a measurement of 1 (on a fresh copy of the original state).
    - Return a unified probabilities vector.

Complexity is a polynomial multiplied by 2 to the number of qubits in the subset. Consequently, Pauli expectation value is polynomial for local Hamiltonians.

yaelbh:
In the deterministic case there is in fact no need to go in recursion, just to set all deterministic qubits and branch on the random ones.

PR #423, opened 2019-11-05
Rewrite stabilizer snapshot_probabilities

yaelbh:
### Summary

The code for the stabilizer probabilities snapshot does not work properly for subsets of qubits. This pull request fixes this.

### Details and comments

More details can be found in https://github.com/Qiskit/qiskit-aer/issues/389

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Changelog](http://keepachangelog.com/en/1.0.0/).
 
 Added
 -----
+- Added support for probabilities snapshot and Pauli expectation value snapshot in the stabilizer simulator (\#423)
 
 Changed
 -------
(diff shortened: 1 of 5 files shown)


yaelbh:
@atilag There are several locations in this code, including in this pull request, where the assignment operator is used for qubit registers (i.e. statevector data structure), assuming that it performs a deep copy. This is indeed a deep copy, since the data structure does not contain pointer, at least currently. But I don't think it is a good practice to use an assignment operator that was not implemented explicitly. What do you think?

yaelbh:
Ready for review  @chriseclectic.   I'll continue from here with the expectation value snapshot.

yaelbh:
Pauli expval snapshot for stabilizer is now included and ready for review. Thanks to @ShellyGarion for advise. @chriseclectic

chriseclectic:
@yaelbh can you add a line to the changelog that you added support for probability and pauli expval snapshots to stabilizer method

yaelbh:
@chriseclectic Done

PR #424, opened 2019-11-06
Fix snapshot_probabilities in mps

merav-aharoni:
### Summary
Fully support snapshot_probabilities for mps. Test for snapshot now passes and results are identical to those of statevector.

### Details and comments
Note that the implementation for a subset of the qubits does not require computing the full vector of probabilities for all qubits. We only compute the probabilities for the qubits in the subset.


merav-aharoni:
I fixed the bug that was blocking this issue. It is ready for review now.

src/simulators/matrix_product_state/matrix_product_state_internal.hpp
```diff
@@ -261,7 +274,7 @@ class MPS{
   uint_t omp_threads_ = 1;     // Disable multithreading by default
   uint_t omp_threshold_ = 14;  // Qubit threshold for multithreading when enabled
   int sample_measure_index_size_ = 10; // Sample measure indexing qubit size
-  double json_chop_threshold_ = 0;  // Threshold for choping small values
+  double json_chop_threshold_ = 1E-8;  // Threshold for choping small values
```
chriseclectic: This is larger than the value in `matrix_product_state.hpp` (1e-15). Can you set them both to the same value? I think 1e-8 is also a little large, the other state classes use 1e-10.

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -645,7 +649,9 @@ void State::apply_measure(const reg_t &qubits,
 }
 
 rvector_t State::measure_probs(const reg_t &qubits) const {
-  return qreg_.probabilities(qubits);
+  rvector_t probvector;
+  qreg_.probabilities_vector(probvector, qubits);
```
atilag: I thnik is clearer if the `MPS::probabilities_vector` memeber function is renamed to:
```c++
rvector_t probvector = qreg_.get_probabilties_vector(qubits);
```
it makes explicit the intention of the memeber function: returning a new data structure.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -41,29 +41,105 @@ static const cmatrix_t one_measure =
...
 // Function name: reverse_all_bits
 // Description: The ordering of the amplitudes in the statevector in this module is 
 //    000, 001, 010, 011, 100, 101, 110, 111.
 //    The ordering of the amplitudes in the statevector in Qasm in general is 
 //    000, 100, 010, 110, 001, 101, 011, 111.
 //    This function converts the statevector from one representation to the other.
+//    This is a special case of reorder_qubits
 // Input: the input statevector and the number of qubits
-// Returns: the statevector in reverse order
+// Output: the statevector in reverse order
```
atilag: nit: The standatd is to use the: `Args`/ `Returns` wording instead of `Input`/`Output` (that's a minor issue of course).

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -41,29 +41,105 @@ static const cmatrix_t one_measure =
...
+  reg_t squeezed_qubits(qubits.size());
+  std::vector<uint_t> sorted_qubits;
+  for (uint_t index : qubits) {
+    sorted_qubits.push_back(index);
+  }
+  sort(sorted_qubits.begin(), sorted_qubits.end());
+  for (uint_t i=0; i<qubits.size(); i++) {
+    for (uint_t j=0; j<sorted_qubits.size(); j++) {
+      if (qubits[i] == sorted_qubits[j]) {
+	squeezed_qubits[i] = j;
```
atilag: nit: indent this line

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -41,29 +41,105 @@ static const cmatrix_t one_measure =
...
+  int_t current_pos = 0, current_val = 0, new_pos = 0, shift =0;
+  uint_t num_qubits = qubits.size();
+  for (uint_t i=0; i<num_qubits; i++) {
+    current_pos = num_qubits-1-qubits[i];
+    current_val = 0x1 << current_pos;
+    new_pos = num_qubits-1-i;
+    shift = new_pos - current_pos;
+    if (index & current_val) {
+      if (shift > 0) {
+	new_index += current_val << shift;
```
atilag: nit: indent

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -41,29 +41,105 @@ static const cmatrix_t one_measure =
...
+      } else if (shift < 0) {
+	new_index += current_val >> -shift;
```
atilag: nit: indent

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -41,29 +41,105 @@ static const cmatrix_t one_measure =
...
+    current_val = 0x1 << current_pos;
+    new_pos = num_qubits-1-i;
+    shift = new_pos - current_pos;
+    if (index & current_val) {
+      if (shift > 0) {
+	new_index += current_val << shift;
+      } else if (shift < 0) {
+	new_index += current_val >> -shift; 
+      } else {
+	new_index += current_val;
```
atilag: nit: indent

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -349,6 +427,47 @@ void MPS::apply_diagonal_matrix(const AER::reg_t &qubits, const cvector_t &vmat)
   apply_matrix(qubits, diag_mat);
 }
 
+void MPS::centralize_qubits(MPS & new_MPS, const reg_t &qubits,
```
atilag: In the same spirit of the previous member function (`probabilities_vector`) I would change the signature of this member function to:
```
MPS MPS::centralize_qubits(const reg_t &qubits, uint_t &new_first, uint_t &new_last) const {
```
I have removed the `ordered` output argument because it's not being used outside of this function.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -349,6 +427,47 @@ void MPS::apply_diagonal_matrix(const AER::reg_t &qubits, const cvector_t &vmat)
   apply_matrix(qubits, diag_mat);
 }
 
+void MPS::centralize_qubits(MPS & new_MPS, const reg_t &qubits, 
+			    uint_t &new_first, uint_t &new_last, bool & ordered ) const {
+  new_MPS.initialize(*this);
+  ordered = true;
```
atilag: We are always overwriting `ordered` argument here, so there's no point on passing it as an argument at all.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -349,6 +427,47 @@ void MPS::apply_diagonal_matrix(const AER::reg_t &qubits, const cvector_t &vmat)
...
+      break;
+    }
+  }
+  if (!ordered)
+      sort(internalIndexes.begin(), internalIndexes.end());
+
+  std::vector<uint_t> new_indexes = calc_new_indexes(internalIndexes);
+  
+  uint_t avg = new_indexes[new_indexes.size()/2];
+  std::vector<uint_t>::iterator it = lower_bound(internalIndexes.begin(), internalIndexes.end(), avg);
```
atilag: nit: There's a very useful type: `auto`, that will infere the type of the variable so you don't have to type: `std::vector<uint_t>::iterator`:
```c++
auto it = lower_bound(internalIndexes.begin(), internalIndexes.end(), avg);
```

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -364,10 +483,11 @@ void MPS::change_position(uint_t src, uint_t dst)
 cmatrix_t MPS::density_matrix(const reg_t &qubits) const
 {
   MPS temp_MPS;
-  uint_t front = 0, back = 0;
-  MPS_with_new_indices(qubits, temp_MPS, front, back);
-  MPS_Tensor psi = temp_MPS.state_vec(front, back);
+  uint_t new_first=0, new_last=0;
+  bool ordered = true;
```
atilag: seems like you can remove `ordered` safetly

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -383,6 +503,24 @@ cmatrix_t MPS::density_matrix(const reg_t &qubits) const
   return rho;
 }
 
+
+rvector_t MPS::trace_of_density_matrix(const reg_t &qubits) const
+{
+  MPS temp_MPS;
+  uint_t new_first=0, new_last=0;
+  bool ordered = true;
```
atilag: No need for `ordered`

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -568,25 +705,40 @@ std::vector<reg_t> MPS::get_matrices_sizes() const
   return result;
 }
 
-MPS_Tensor MPS::state_vec(uint_t first_index, uint_t last_index) const
+MPS_Tensor MPS::state_vec_as_MPS(const reg_t &qubits) const {
+  MPS temp_MPS;
+  uint_t new_first=0, new_last=0;
+  bool ordered = true;
```
atilag: No need for `ordered`

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -598,15 +750,74 @@ void MPS::full_state_vector(cvector_t& statevector) const
...
+  for (uint_t index=0; index < num_qubits-1; index++) {
+    if (qubits[index] > qubits[index+1]){
+      ordered = false;
+      break;
+    }
+  }
+  bool reverse_ordered = true;
+  for (uint_t index=0; index < num_qubits-1; index++) {
+    if (qubits[index] < qubits[index+1]){
```
atilag: Don't we just need to test one case to know whether is reversed or not?
like:
```c++
    reversed = (qubits[0] < qubits[1] ? false: true);
```

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -598,15 +750,74 @@ void MPS::full_state_vector(cvector_t& statevector) const
...
+  rvector_t ordered_probvector = trace_of_density_matrix(qubits);
+
+  // 2.1 subset of qubits is ordered
+  // reverse_bits to be consistent with output order in qasm
+  if (ordered) {
+    probvector = reverse_all_bits(ordered_probvector, num_qubits);
+    return;
+  }
+  // 2.1 subset of qubits is reverse-ordered
+  else if (reverse_ordered) {
```
atilag: no need for the `else`

src/simulators/matrix_product_state/matrix_product_state_internal.hpp
merav-aharoni: @chriseclectic , you are right. I have some old code here that was originally copied from one of the other simulation methods. I was just looking at this in the context of the approximation issue. I intend to set the default for chopping the Schmidt coefficient to 1e-16, and this will be controllable by the user. However, I prefer to handle in the `approx` issue, and not here. Let me know if this sound like a reasonable default threshold.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I changed the `Output` to `Returns`, as in other functions. I generally tried to create a standard template for all our function declarations in matrix_product_state. However, I didn't see a standard in other cpp files; the format is rather loose. In particular, I didn't see anyone using `Args`, instead of `Input`, so I didn't change that.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Done.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Done - at least I see it indented. Let me know if you don't. Same for previous comment.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: same as previous comment.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: removed.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: okay, removed.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I agree. Changed it.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: done.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: No. I wanted to check it **all** qubits are ordered, i.e., `qubits[i]<qubits[i+1]` for `all i`. 
Or if all qubits are in reverse order, i.e., `qubits[i]>qubits[i+1]` for `all i`.
I wrote these special cases initially as an optimization, in order not to need to compute the density matrix. But after looking at it again because of your comment, I realized I could remove all those special cases, because I only compute the **trace** of the density matrix anyway, so the special cases don't improve performance. 
In short, I rewrote the function and now it is a lot shorter and simpler.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: see above.

atilag:
ouch!, there seems to be failing tests. Can you take a look before merging, @merav-aharoni ?

merav-aharoni:
@atilag , I see the failing test - it is: `test.terra.backends.test_qasm_simulator.TestQasmSimulator.test_truncate_ideal_sparse_circuit` 
and the failure message is 
`qiskit.transpiler.exceptions.TranspilerError: 'Number of qubits (50) in circuit1007 is greater than maximum (28) in the coupling_map'"`.
This doesn't seem related to my code. Does this fail elsewhere as well?

chriseclectic:
@merav-aharoni The failing test is due to a Terra bug, it doesn't have anything to do with this PR

PR #429, opened 2019-11-11
Remove cvxopt from requirements

chriseclectic:
### Summary

Removes cvxopt package from requirements. It must be installed separately by the user due to its software license.


chriseclectic:
Superseded by #430

PR #430, opened 2019-11-11
Switch cvxopt usage to cvxpy

mtreinish:
### Summary

This commit switches the usage of cvxopt to use cvxpy instead.

### Details and comments


requirements-dev.txt
```diff
@@ -3,7 +3,7 @@ cmake
 scikit-build
 cython
 asv
-cvxopt
+cvxpy
```
chriseclectic: This should be `cvxpy>=1.0.0` in case someone is using an old version.

qiskit/providers/aer/noise/utils/noise_transformation.py
```diff
@@ -691,22 +692,13 @@ def solve_quadratic_program(self, P, q):
...
-            ImportError: If cvxopt external module is not installed
-
         Additional information
         ======================
-        This method is the only place in the code where we rely on the cvxopt library
+        This method is the only place in the code where we rely on the cvpy library
         should we consider another library, only this method needs to change
         """
-        try:
-            import cvxopt
```
chriseclectic: I think we should leave `import cvxpy` here so that it isn't imported by the top level qiskit aer init.

requirements-dev.txt
mtreinish: Done in: https://github.com/Qiskit/qiskit-aer/pull/430/commits/426bcfccc69684c4ba50848b4ec8605ca59ad78b

qiskit/providers/aer/noise/utils/noise_transformation.py
mtreinish: Done in: https://github.com/Qiskit/qiskit-aer/pull/430/commits/426bcfccc69684c4ba50848b4ec8605ca59ad78b

mtreinish:
It looks like a numpy release earlier today is causing the python 3.5 failure on windows.

chriseclectic:
@mtreinish how can we make test pass?

mtreinish:
> @mtreinish how can we make test pass?

I'm hoping it was just a temporary packaging issue because of the new version that would sort itself today. If that doesn't work we can just set a cap on the python 3.5 job.

mtreinish:
I was actually wrong as to the root cause, it was an artifact of conda vs pip. I wrote down the details of what's going on in: https://github.com/Qiskit/qiskit-aer/pull/430/commits/bde0e77a816f8ac07395ee4f2f566381b4ea001e 

Hopefully that will work, I seem to remember that we needed to conda install numpy to get aer to compile in the windows jobs, but we'll see if that's an issue I guess.

PR #431, opened 2019-11-12
Cleaned up code for the function that handles 2-qubit gates in MPS

merav-aharoni:
### Summary
The function apply_2_qubit_gate was previously recursive. Changed it to use change_position, so now it is simpler and not recursive.

### Details and comments
While doing this, I found a bug in apply_matrix, for cases when the order of qubits is swapped. So I rewrote this function as well.


chriseclectic:
Looks good @merav-aharoni, is there a test we could add to the two-qubit gates for the qasm_simulators that would catch the fixed bug?

merav-aharoni:
@chriseclectic , I simply used a randomly generated matrix:
`from qiskit.quantum_info.random import random_unitary`
`qc.unitary(random_unitary(4), [0, 1])`
and also switch the order of the qubits:
`qc.unitary(random_unitary(4), [1, 0])`
I couldn't find any tests for apply_matrix. 
I could add this tests, but I am not sure how, because it does not fit the regular template for tests in Aer. What do you think?

chriseclectic:
You could add them to the unitary gate tests: 

https://github.com/Qiskit/qiskit-aer/blob/master/test/terra/backends/qasm_simulator/qasm_unitary_gate.py
https://github.com/Qiskit/qiskit-aer/blob/master/test/terra/reference/ref_unitary_gate.py

merav-aharoni:
@chriseclectic , I added a test with random unitary gates, using a fixed seed, as you suggested.

merav-aharoni:
@chriseclectic I think this issue can be merged. Please let me know if you see anything else that should be fixed.

PR #438, opened 2019-11-13
[WIP] [1 of 3] Improving build system for Pulse simulator

atilag:
### Summary
This is the first of a series of PRs for C++ migration of the ODE solver function, originally wrote on Cython.

### Details and comments
This adds all needed dependencies and scaffolding for coming PRs.
One of the dependencies is the MuparserX, a math expression evaluator. We need to add the .h files even though original instructions are not clear about it. I have added an static version of this library for MacOS, rest of the platforms will be added later.
Another external C++ library (header-only) added but not strictly required is `spdlog` for logging purposes.


chriseclectic:
@atilag can you make an issue about moving C++ logging to use this logger library, and maybe some documentation on style/format/best practice for how we should use for it?

PR #442, opened 2019-11-13
Migrate ODE function solver to C++

atilag:
### Summary
Now the Scipy ODE solver will use an static function built with C++, so there's no compilation on the flight nor the need for the user to have a specific set of toolchain nor files laying around.

### Details and comments
I haven't benchmarked anything yet, because I focused on making it work instead of having best performance (things should go faster though for small simulations, longer simulations should not at the moment).
There's a flag in the rutime config: `use_cpp_ode_func` that can be enabled to use the C++ version of the solver or the Cyhton one. I want to keep both for a while just in case the C++ version is buggy. Cyhton one will be removed eventually.

### About the code
The main files are: 
**- numeric_integrator.cpp**: Here is where all the logic resides. It's a direct translation of the algorithm that was being generated dynamically, plus some changes to generalize it (or be able to make it static)
**- python_to_cpp.hpp**: A set of function helpers to wrap Python C API with C++ containers (some STL and some defined by me).
**- jlog.hpp**: This is not being used in the code at the moment, but are a bunch of convinient functions for logging complex structures

### TODO
(We may want to address some of these items in following PRs)
- [x] Fix the linter
- [x] Fix a crash that exists in both versions running some Pulse tests
- [x] Restructure files in the project
- [x] Remove unccessary cython files
- [x] Linux support
- [x] Windows support


qiskit/providers/aer/backends/pulse_simulator.py
```diff
@@ -48,7 +48,8 @@ class PulseSimulator(AerBackend):
         'max_shots': 50000,
         'description': 'A pulse-based Hamiltonian simulator',
         'gates': [],
-        'basis_gates': []
+        'basis_gates': [],
+        'use_cpp_ode_func': True  # Use the C++ ODE func or the Cyhton one
```
chriseclectic: typo `Cython`

I feel like this should be in `backend_options`, not the config, if you expect a user to change it.

qiskit/providers/aer/openpulse/qobj/digest.py
```diff
@@ -199,6 +176,8 @@ def digest_pulse_obj(qobj):
 
     out.global_data['freqs'] = list(out.freqs.values())
 
+    # TODO: Here is the crash! pulses_idx is a list of indexes for
```
chriseclectic: If it crashes should we have a `try` or `if` to catch exception and return error message?

qiskit/providers/aer/backends/pulse_simulator.py
atilag: I wasn't thinking about the user, but the fact that I want both methdos (Cython/C++ Solvers) coexist for a while. Only the C++ solver will eventually remain. But you got a point, I'd rather make it easier for the user to change it, so I'm moving it to the backend_options.

PR #448, opened 2019-11-14
Switch to running tests in parallel with stestr

mtreinish:
### Summary

This commit switches the test runner for running the aer tests to
stestr. stestr is a unittest compatible test runner that executes tests
in parallel by default. This can provide significant speeds ups for
running a test suite. It also provides a rich set of scheduling [1] and
test selection [2] features which we can use to fine tune how we execute
tests. Based on local testing 2 classes of tests run in parallel already, to
optimize the run time thees need to be scheduled serially so that they don't
compete for resources. This is done using a group_regex [3] which is used
to provide a scheduler hint to keep these tests together.

### Details and comments

[1] https://stestr.readthedocs.io/en/stable/MANUAL.html#test-scheduling
[2] https://stestr.readthedocs.io/en/stable/MANUAL.html#test-selection
[3] https://stestr.readthedocs.io/en/stable/MANUAL.html#grouping-tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ contrib/standalone/version.hpp
 test/.asv
 
 .tox/
+.stestr/
 
 docs/_build/
 docs/stubs/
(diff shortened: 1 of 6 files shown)


mtreinish:
In my local testing (with a warm cache) this improved run time of a full run of the aer unit tests from ~105-115secs to ~30-40 seconds

PR #451, opened 2019-11-16
fix indexing error in FC

### LINKED ISSUE #450: Framechanges cause the simulator to hang
If a framechange shares the same start time or end time as a pulse, the simulator can hang (it doesn't seem to be universal behavior). I don't know why, but I suspect it has to do with the pulse interpolation in the simulator. 

An example would be a schedule created from the qasm for the fine amplitude cal.  

### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

nonhermitian:
### Summary
fixes #450


atilag:
oh, great!
Hope it fixes the seg. fault as well :)
Thanks @nonhermitian

PR #454, opened 2019-11-18
Added support for ccx in MPS simulation method

merav-aharoni:
### Summary
Added support for ccx in MPS simulator. Based on this implementation, we can add other 3-qubit gates as well.


merav-aharoni:
I fixed the bug that was blocking this from being merged. It was actually the same problem as in PR #424. So this PR is ready for review now.

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ Fixed
 Added
 -----
 - Added support for probabilities snapshot and Pauli expectation value snapshot in the stabilizer simulator (\#423)
+- MPS simulation method: added support for ``mcx`` (\#454)
```
chriseclectic: `mcx` is a multi-controlled-X gate with an arbitrary number of control qubits. I think in this case you have only added support for a the Toffoli gate, so it should just say `ccx` here.

```suggestion
- MPS simulation method: added support for ``ccx`` (\#454)
```

merav-aharoni: done

Issue #457, opened 2019-11-20
Automatic determination of `qubit_lo_freq` from Hamiltonian when simulating


DanPuzzuoli:
### What is the expected behavior?

When assembling the `qobj` to be submitted to the pulse simulator, have the option to automatically set `qubit_lo_freq` based on the energy levels in the Hamiltonian. The purpose is to streamline this aspect of simulation specification.

Currently, these need to be set by hand. For example, in `pulse_sim.ipynb`, the `qubit_lo_freq` terms are set by: 

- Calling `assemble` to construct the `qobj`
- Calling `backend_sim.get_dressed_energies` to get the eigenvalues of the Hamiltonian
- Hand-picking the eigenvalues, then calling `assemble` again to construct a new `qobj` with `qubit_lo_freq` specified in terms of the eigenvalues

### Proposed solution

The easiest way to incorporate this quickly is to pass something like `qubit_lo_freq = 'from_hamiltonian'` to `assemble`. When the the simulator calls `digest_pulse_obj`, if `qubit_lo_freq == 'from_hamiltonian'`, it can automatically set the `qubit_lo_freq` based on the eigenvalues of the Hamiltonian (which it already computes anyway).

*Update:* Trying to pass a string for `qubit_lo_freq` to `assemble` results in a validation error. I think now the easiest way to get this through is to pass it within the `sim_config` argument, as this way it can make it into `digest_pulse_obj` without any changes to `terra`.

DanPuzzuoli:
I'm working on this now, but comments are welcome.

PR #459, opened 2019-11-25
Calculate qubit_lo_freq from pulse hamiltonian

DanPuzzuoli:
### Summary

Resolving issue #457: When calling the pulse simulator, can add option `backend_options['qubit_lo_freq'] = 'from_hamiltonian'`, and the simulator will compute the `qubit_lo_freq` from the drift Hamiltonian.

Note: this currently only has an effect if the Hamiltonian only contains qubits; if it contains oscillators this will have no effect. (I don't know how the simulator determines what is a 'qubit' and what is an 'oscillator', but as far as I am aware all current demos/use-cases only work with what the simulator considers 'qubits'.)

### Details and comments

The way `qubit_lo_freq` is determined from the Hamiltonian:

- compute the spectral decomposition of the drift Hamiltonian
- for each qubit, construct the first excited state (i.e. the state where all other qubits are in 0, and the labelled qubit is in the first excited state)
- given the excited state, find the eigenvalue corresponding to the eigenspace of the drift Hamiltonian for which the excited state has maximal overlap
- the qubit_lo_freq for this qubit is set as the difference between this eigenvalue and the smallest eigenalue, divided by `2*pi` 

Code changes are:

- `qiskit/providers/aer/openpulse/qobj/digest.py`
  - added logic for handling the option
  - added 3 helper functions `_first_excited_state`, `_eval_for_max_espace_overlap`, and `_proj_norm` to perform the eigenvalue computations
- `qiskit/providers/aer/backends/pulse_simulator.py`
  - added a function `compute_lo_freqs_from_hamiltonian`, similar to the existing function `get_dressed_energies`, which a user can call to see the computed `qubit_lo_freq` values
- added test file `test/terra/openpulse/test_pulse_digest.py`
  - This is a place to test `digest.py`. Currently consists of a single test function for testing various behaviours of the option `backend_options['qubit_lo_freq'] = 'from_hamiltonian'


DanPuzzuoli:
The latest round of changes are:

- In `digest.py`, when checking whether to compute `qubit_lo_freq` from Hamiltonian, added the additional condition to do so if `config_dict['qubit_lo_freq'] = None`, i.e. if the user hasn't specified it, and the backend doesn't have a default `qubit_freq_est`, then compute from the Hamiltonian
- The function `_format_qobj_dict` has been moved from `pulse_simulator.PulseSimulator` into `digest.py`, and the function `digest_pulse_obj` now automatically calls this first. 
    - `_format_qobj_dict` had no dependencies on a PulseSimulator instance, and it was always called immediately before `digest_pulse_obj`, so there seemed to be no reason to have this inside `PulseSimulator` (as it was, it wasn't possible to examine the output of `digest` without instantiating a `PulseSimulator` object)
- The function `compute_lo_freq_from_hamiltonian` has been moved from `PulseSimulator` into a new file `qiskit/providers/aer/utils/pulse_sim_utils.py`

For the last, we may want to move more functions from `PulseSimulator` into `pulse_sim_utils`.

DanPuzzuoli:
To do in this branch:

- Move `get_dressed_energies` from `pulse_simulator` to the new `utils` folder. This is a more natural place for it, similar to the move of `lo_from_hamiltonian`. 
   - A consequence: the pulse sim example notebook, which uses `get_dressed_energies`, will need to be changed. This should be done anyway though, as it can now submit the option `backend_options[`qubit_lo_freq`] = 'from_hamiltonian'`

DanPuzzuoli:
Latest commit includes the changes described in the last comment.

PR #461, opened 2019-11-26
Remove qutip_lite math.pyx module

atilag:
### Summary
The module seems to be conflicting while cython builds, with the standard math module, in some specific configurations.

@lcapelluto can you check that this fixes the build for you? Thanks!

### Details and comments
It was not being used at the moment, but seems to contain a relevant function (`erf`) that could be useful for some pulse generation routines.
We can always take it back from the history if we want to.


atilag:
Ok, that's likely another unrelated issue. Going to merge this then.
Thanks Lauren!

PR #465, opened 2019-11-27
fixed pylint errors

DanPuzzuoli:
### Summary

Fixed the following errors:

- 3 `no-name-in-module` errors from importing cpython functions
  - fixed by disabling this error for the relevant import only
- 1 `bad-option-value` coming from an invalid `pylint` option
  - just removed this line from `noise_transformation.py`, and there were no further errors


DanPuzzuoli:
Somewhat confused by this, different errors are appearing locally vs on Travis CI

Issue #468, opened 2019-12-02
Replace usage of `==` for floating point compare by `Linalg::almost_equal` in svd.cpp


merav-aharoni:
### Informations

- **Operating system**: all

### What is the current behavior?
In the file svd.cpp there are several floating point comparisons using `==` and `!=`.
Replace them by `Linalg::almost_equal`.
This issue was raised by Chris in reviewing PR #344.

### What is the expected behavior?
This usage is more correct and will eliminate several warning from compilation.

atilag:
Flagging this issue as `bug` too, because although things are working fine at the moment, that could not be the case with some future change. Floating-point comparison using the canonical  operators it's dangerous (and erroneous depending on the needed precision).

PR #472, opened 2019-12-03
Changed usage of == for floating point comparison to Linalg::almost_equal in svd.cpp

merav-aharoni:
### Summary
Changed usage of `==` and `!=` for floating point comparison to `Linalg::almost_equal` in svd.cpp.

### Details and comments
1. Used casting to `long double` for all comparisons, to get sufficient precision 
2. Changed some constant values in the code to `#define`
3. Fixed indentation to more standard format

This PR resolves issue #468.

### DIFF
--- a/src/simulators/matrix_product_state/svd.cpp
+++ b/src/simulators/matrix_product_state/svd.cpp
@@ -24,6 +24,10 @@
 #include <cmath>
 #include <complex>
 #include <cassert>
+
+#include <limits>
+#include <type_traits>
+
 #include "svd.hpp"
 #include "framework/utils.hpp"
 #include "framework/linalg/almost_equal.hpp"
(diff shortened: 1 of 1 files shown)


atilag:
Thanks @merav-aharoni ! 
Test are failing because I think we have changes the `Result` object. Will try to fix this and merge!

merav-aharoni:
@atilag , thanks for letting me know, because I was just trying to figure out how I would run this on Windows, because my usual platform in Linux.

chriseclectic:
Is quadruple precision of `long double` required?

merav-aharoni:
@chriseclectic , unfortunately it is needed. At first I tried using double, but I had failures. Also in the code itself, there is a place where I resorted to cast to `long double` when I find that double precision is not sufficient to detect the difference between a very small number and 0. This works fine on Linux, but I think Windows does not support `long double`. Any suggestions?

atilag:
So the problem is that Microsoft Visual Studio compiler treats `double` and `long double` as the same type... it's not a bug, the standard allows this. I don't really know what other options we have for Windows at the moment... will investigate further.

Issue #449, opened 2019-11-15
dt isn't getting pulled into the openpulse solver correctly


dcmckayibm:
The `dt` parameters doesn't seem to be getting used by the solver correctly. In the example notebook it's changed it should affect the Rabi frequency in a deterministic way and it doesn't seem to be. This could be due to the pulse interpolation (i.e. the look up is not getting divided by dt). 

### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

dcmckayibm:
For example, we should be able to set dt=1e-9 and all the frequencies to GHz and this also doesn't seem to work

nonhermitian:
This is because the solver was designed to use natural units. This is a standard thing to do in numerics.

dcmckayibm:
Right, but since open pulse gives pulses as a list of samples there is a missing time unit (the time between samples) and that time is `dt`.

nonhermitian:
It is used here: https://github.com/Qiskit/qiskit-aer/blob/64e987d90e8847f31215bda05dfb6fc40849647e/qiskit/providers/aer/openpulse/qobj/digest.py#L412

to set the width of a pulse in time, and it the computation of the max_time needed in the simulations.

dcmckayibm:
Doesn't it need to be in chan_val (eg these lines https://github.com/Qiskit/qiskit-aer/blob/64e987d90e8847f31215bda05dfb6fc40849647e/qiskit/providers/aer/openpulse/cy/channel_value.pyx#L83) to determine which sample to pick?

nonhermitian:
I believe `start_time` and `stop_time` are used to determine the index that corresponds to the value of the channel in the discrete list of pulse values.

dcmckayibm:
Ok,I see what you're saying, but I was getting some non-intuitive answers by changing `dt`, so it might be good to double check.

DanPuzzuoli:
I think I've found the source of the error in `experiment_to_structs`: there are a few places where 'sample time' is being erroneously combined with 'real time' (i.e. `inst['t0']` vs  `inst['t0']*dt`). E.g.
https://github.com/Qiskit/qiskit-aer/blob/8b2cfa59e3006b75fbcd768ce5c3149a4e955c35/qiskit/providers/aer/openpulse/qobj/digest.py#L519-L520
`inst['t0']` is in 'sample time', but `dt*inst['duration']` is in 'real time'.

[...]

PR #473, opened 2019-12-03
Fix pulse schedule `dt` parameter not being applied in pulse simulator

DanPuzzuoli:
### Summary

This fixes issue #449 and adds a test case to detect errors with handling of `dt`.

### Details and comments

- Fixes the way `dt` is being handled in `digest.experiment_to_structs`. The changes can be summarized as: every appearance of `inst['t0']` was replaced with `inst['t0']*dt` to facilitate the conversion from 'time-step-number' of a pulse to 'time' occurring in this function.
- Adds a test case `test_dt_scaling_x_gate` to `TestPulseSimulator`, which checks that changing `dt` affects simulation outcomes in the expected way. It effectively runs the same steps as `test_x_gate`, but for different values of `dt`. For each value of `dt`, the test works by dividing all frequency parameters (including coefficients in the Hamiltonian) by `dt`, and then verifying that the result of the simulation is the same as the `dt=1` case (the scaling of all parameters should 'cancel out' the change in `dt`).


dcmckayibm:
I'm happy if @chriseclectic approves

dcmckayibm:
We need to fix that bug, we can refactor digest later

atilag:
Sure, was just a comment for later call to action

PR #475, opened 2019-12-04
Introducing approximation into the MPS simulation method

merav-aharoni:
### Summary
We allow using approximation when using the MPS simulation method.

### Details and comments
The idea of approximation is to limit the number of Schmidt coefficients in the lambda matrices of the MPS. This can be done either by a size threshold or by the number of coefficients.
The approximation is controlled by two parameters at the beginning of svd.cpp:
`APPROX_THRES` - we discard values in lambda whose norm is smaller than 
APPROX_THRES * norm(S[0]), where S[0] is the largest Schmidt coefficient.
`APPROX_LIMIT` - we discard values only if the size of lambda is greater than this value
These parameters are currently hard-coded, and for now can be changed only in the code.
We need to add a user interface for these parameters, and to experiment with this option.


merav-aharoni:
I removed the WIP - this is ready for review

merav-aharoni:
@chriseclectic , @atilag , this is ready for review.

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -82,6 +82,17 @@ namespace Simulator {
  *      parallel circuit or shot execution is enabled this will only
  *      use unallocated CPU cores up to max_parallel_threads. [Default: 100]
  *
+* From MatrixProductState::State class
+ * - "mps_approximation_threshold" (double): Set the cutoff value for the
```
chriseclectic: I think this should be called `"matrix_product_state_bond_truncation_threshold"`, and should always be used (with a default value) regardless of whether the bond dimension is also specified. If you don't want any approximation this can be set to 0 by default.

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -82,6 +82,17 @@ namespace Simulator {
...
+ *      Schmidt coefficients. Numbers smaller than the threshold will be discarded.
+ *      [Default: 1e-16]
+ *
+ * - "mps_approximation_num_sv" (uint): Set the cutoff value for the number
```
chriseclectic: Can you change the name of this option to `"matrix_product_state_bond_dimension"`, since that is (I believe) the standard naming convention in tensor network community. 


```
*-  "matrix_product_state_bond_dimension" (uint): Set the maximum size of the matrix product state bond dimensions. This sets a cutoff value for the number of Schmidt coeffs...
```

src/simulators/qasm/qasm_controller.hpp
```diff
@@ -82,6 +82,17 @@ namespace Simulator {
...
+ *      Schmidt coefficients. If there are more than this number of coefficient, 
+ *      the smallest values will be discarded.
+ *      [Default: none]
+ * Note that both approximation parameters must be set for approximation to take place.
```
chriseclectic: I would think approximation can happen for either option being used, or both.

* One option is to truncate almost-zero schmidt coefficients.
* The other is to limit the maximum number of non-zero schmidt coefficients

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -192,6 +192,31 @@ class QasmSimulator(AerBackend):
...
+    * ``"approximation_type"`` (int): Can be one of the following options:
+      NO_APPROX - means no approximation.
+      ABSOLUTE_APPROX - only the largest max_num_coefficients_for_approx
+      Schmidt coefficients are retained.
+      RELATIVE_APPROX - if the number of Schmidt coefficients is larger than
+      max_num_coefficients_for_approx, then the smallest values are
+      discarded (see approx_thresold).
+      (Default: NONE).
+
+    * ``"max_num_coefficients_for_approx"`` (int): Sets a limit on the number of Schmidt
```
chriseclectic: Can you call this `"matrix_product_state_max_bond_dimension"` (see comment below in controller.hpp)

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -192,6 +192,31 @@ class QasmSimulator(AerBackend):
...
+      coefficients retained at the end of the svd algorithm.
+      If the number of coefficients is greater than max_num_coefficients_for_approx,
+      small values may be discarded, depending on approximation_type and on approx_threshold.
+      (Default: UINT_64MAX).
```
chriseclectic: Isn't the default the full dimension of the n-qubit state space? Either way I think in documentation the default should be `Default: None`

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -192,6 +192,31 @@ class QasmSimulator(AerBackend):
       OpenMP parallelization. If parallel circuit or shot execution
       is enabled this will only use unallocated CPU cores up to
       max_parallel_threads (Default: 100).
+
+    These backend options only apply when using the ``"matrix_product_state"``
+    simulation method:
+
+    * ``"approximation_type"`` (int): Can be one of the following options:
+      NO_APPROX - means no approximation.
+      ABSOLUTE_APPROX - only the largest max_num_coefficients_for_approx
```
chriseclectic: I don't really understand what the different between these approximation types types are? It seems like we don't need this option, just two approximation options: a zero threshold, and a max bond dimension

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -192,6 +192,31 @@ class QasmSimulator(AerBackend):
...
+      discarded (see approx_thresold).
+      (Default: NONE).
+
+    * ``"max_num_coefficients_for_approx"`` (int): Sets a limit on the number of Schmidt
+      coefficients retained at the end of the svd algorithm.
+      If the number of coefficients is greater than max_num_coefficients_for_approx,
+      small values may be discarded, depending on approximation_type and on approx_threshold.
+      (Default: UINT_64MAX).
+
+    * ``"approx_threshold"`` (double): Sets a lower limit on the value of the Schmidt
```
chriseclectic: Same comment as for controller: I think this option should just be `"matrix_product_state_bond_truncation_threshold"` and should work independently of the max bond dimension truncation if set

merav-aharoni:
Hi @chriseclectic , thanks for the comments. I am fine with most of your comments (names of the parameters, updating the result metadata, and tests that were committed by mistake). 
Regarding the modes of approximation, I will try to explain my motivation. Often during the MPS algorithm, the arrays of Schmidt coefficients consists of identical entries. For example, we could get an array of length 64, where all entries are 0.0156. Using `truncation by max bond dimension` would remove an arbitrary number of entries, although these are no smaller than the others, so this might not be such a good approximation. Using a `truncation threshold of 0.01` would cause all entries to be removed, resulting in a crash. 
Therefore I introduced the third mode of approximation which is a more careful approximation: this is a `relative threshold` - i.e., percentage relative to the max entry in the array.
Let me know what you think, and if you still think I should remove the `relative threshold` or if I should modify it in some other way.

merav-aharoni:
@chriseclectic , I implemented the changes in MPS approximation as we discussed.
Regarding the `metadata`, I couldn't find any place where it is natural to add the specific metadata for a `State`. So I added a new method `State::add_metadata` in the base class. Every `State` can implement this method and add its own `metadata`.
This PR is ready for review.

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -375,6 +392,17 @@ void State::set_config(const json_t &config) {
   }
 }
 
+void State::add_metadata(ExperimentData &data, const json_t &config) const {
+  double threshold;
+  if (JSON::get_value(threshold, "matrix_product_state_truncation_threshold", config))
+    data.add_metadata("matrix_product_state_truncation_threshold", threshold);
```
chriseclectic, line 398: These values could be stored as class variables, so you don't need to get them again from the json

src/simulators/matrix_product_state/svd.cpp
```diff
@@ -95,11 +96,50 @@ uint_t num_of_SV(rvector_t S, double threshold)
...
+    // values in S, and discard all the rest
+    new_SV_num = max_bond_dimension;
+  } 
+  //  if (new_SV_num < SV_num)
+    //    std::cout << "Reduced1: after approx 1, new SV num = " << new_SV_num <<std::endl;
+  // Remove the lowest Schmidt coefficients such that the sum of 
+  // their squares is less than trunction_threshold
+  double sum_squares = 0;
+  for (int_t i=new_SV_num-1; i>0; i--) {
+    if (sum_squares + std::norm(S[i]) < truncation_threshold)
```
atilag, line 117: Can you put open and clone curly braces here?
```c++
{ 
}
```

src/simulators/matrix_product_state/svd.cpp
```diff
@@ -95,11 +96,50 @@ uint_t num_of_SV(rvector_t S, double threshold)
...
+  // After approximation, we may need to re-normalize the values of S
+  if (new_SV_num < SV_num) {
+    double sum=0;
+    for (uint_t i=0; i<S.size(); i++) {
+      sum += std::norm(S[0]);
+    }
+    if (1-sum > THRESHOLD) {
+      for (uint_t i=0; i<S.size(); i++) {
+	double square_i = std::norm(S[0])/sum;
+	S[i] = sqrt(square_i);
```
atilag, line 139: Can you indent this code?

src/simulators/matrix_product_state/svd.cpp
```diff
@@ -523,7 +563,7 @@ status csvd(cmatrix_t &A, cmatrix_t &U,rvector_t &S,cmatrix_t &V)
 	  std::stringstream ss;
 	  ss << "error: wrong SVD calc: A != USV*";
 	  throw std::runtime_error(ss.str());
-	}
+	  }
```
atilag, line 566: nit: remove the extra space.

src/simulators/state.hpp
```diff
@@ -121,6 +121,14 @@ class State {
   // Load any settings for the State class from a config JSON
   virtual void set_config(const json_t &config);
 
+    //-----------------------------------------------------------------------
+  // Optional: Add information to metadata 
+  //-----------------------------------------------------------------------
+
+  // Every state can add information to the metadata structure
+  virtual void add_metadata(ExperimentData &data, const json_t &config) const {
```
atilag, line 129: This an empty method, we can just remove it, right?

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -352,6 +360,15 @@ size_t State::required_memory_mb(uint_t num_qubits,
 }
 
 void State::set_config(const json_t &config) {
+  // Set threshold for truncating Schmidt coefficients
+  double threshold;
+  if (JSON::get_value(threshold, "matrix_product_state_truncation_threshold", config))
```
atilag, line 365: nit: just for consistency I'd add curly braces to this conditional block

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -375,6 +392,17 @@ void State::set_config(const json_t &config) {
   }
 }
 
+void State::add_metadata(ExperimentData &data, const json_t &config) const {
+  double threshold;
+  if (JSON::get_value(threshold, "matrix_product_state_truncation_threshold", config))
```
atilag, line 397: nit: Add curly braces here as well.

src/simulators/matrix_product_state/svd.cpp
```diff
@@ -28,9 +28,10 @@
 #include "framework/utils.hpp"
 #include "framework/linalg/almost_equal.hpp"
 
+// default values
 #define mul_factor 1e2
 #define tiny_factor 1e30
-#define THRESHOLD 1e-9
+#define THRESHOLD 1e-9 // threshold for re-normalization after approximation
```
chriseclectic, line 34: Can these be made const vars or configurable vars rather than compile time defines?

chriseclectic:
@merav-aharoni Could you add a release note for this new feature?

merav-aharoni:
Added documentation for release notes.

Issue #478, opened 2019-12-05
Eliminate spurious errors from Pulse Simulator tests


DanPuzzuoli:
### What is the current behavior?

Some of the pulse simulator tests fail randomly. The tests will typically pass when rerun, but they should be solved for overall stability, as well as to eliminate the nuisance of failed checks for PRs.

The issues that i'm aware of are:
1. `test_persistent_value` consistently exits with a ZVODE error. It occasionally occurs with other tests as well, but it seems to happen with `test_persistent_value` far more regularly than with the other tests.
2. Due to random sampling, some of the other tests fail when verifying counts.

### Suggested solutions

1. While the ZVODE error is a legitimate bug that needs to be addressed, due to the frequency with which it occurs with `test_persistent_value`, along with the fact that persistent value pulses will not be initially supported, I propose commenting out this test for now.
2. As suggested by @chriseclectic, this can be solved by passing a seed to the simulator for random number generation, as is done in qasm simulator tests. The functionality for specifying a seed already exists in `digest_pulse_obj`, so modifying `test_pulse_simulator` to do this should be straightforward.

DanPuzzuoli:
Though we need to still solve this zvode issue, I would say this issue can be closed, as it was primarily about stabilizing the issue with randomness in the tests. I'm going to close this, but create another one about the zvode issue, along with the likely solution of adding soft-reseting the solver at all points that the RHS of the DE is discontinuous.

PR #479, opened 2019-12-05
Issue 478 pulsesim test errors

DanPuzzuoli:
### Summary

Addresses issue #478 by implementing the suggested changes.

### Details and comments

1. The test case `test_persistent_value` was commented out.
2. `digest_pulse_obj` was modified to look for key `'seed'` in `backend_options` and pass this to the pulse simulator. (Code for this already existed, but it was looking in the standard 
pulse `qobj` configuration. I think `backend_options` is a more natural place for this as it is an option for how simulation is done.) The helper functions, `backend_options_1q` and `backend_options_2q`, were updated to include a seed.

### DIFF
--- a/test/terra/backends/test_pulse_simulator.py
+++ b/test/terra/backends/test_pulse_simulator.py
@@ -263,6 +263,7 @@ class TestPulseSimulator(common.QiskitAerTestCase):
         backend_options['qubit_list'] = [self.qubit_0]
         backend_options['dt'] = 1.0  # makes time = self.drive_samples
         backend_options['ode_options'] = {}  # optionally set ode settings
+        backend_options['seed'] = 90841
         return backend_options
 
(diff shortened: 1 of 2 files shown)


test/terra/backends/test_pulse_simulator.py
```diff
@@ -684,6 +686,8 @@ def test_frame_change(self):
         exp_prop_shift = self._analytic_prop_fc(np.pi / 4 - np.pi / 8)
         self.assertDictAlmostEqual(prop_shift, exp_prop_shift, delta=0.01)
 
+    '''
```
atilag: So the proper way to do this, is by using the `@unittest.skipTest("PerisitentValue pulses are currently not supported")`

DanPuzzuoli: Ah nice, thanks. I've made this change :).

PR #480, opened 2019-12-05
[WIP] AerExpectationValue algorithm using snapshot and Parameterized Qobj

dongreenberg:
v0, untested. Submitting here for comment and work tracking.

This is a prototype for an Aer-provided expectation value, encapsulating all the Aer-specific logic to wrap up the snapshot and pass parameters for Qobj-reparameterization. It is motivated by performance and code cleanliness.


atilag:
Not sure this should belongs to Aer, Aqua seems like a better fit? (not even Terra).

t-imamichi:
Not only expectation value but also standard deviation (or standard error) is important when `qasm_simulator` is used.

Issue #462, opened 2019-11-26
Unsupported features of pulse simulator for initial merge discussion


DanPuzzuoli:
### Description

As per the discussion this morning in the OpenPulse meeting, some of the less well-tested/developed pieces of the pulse simulator will be designated as 'unsupported' for the initial merge to the master branch.

The current list of features, as well as proposed approaches for handling unsupported features are below. Please comment, and we can update this list as necessary.

### Current list of features to be designated as 'unsupported':
- Noise (i.e. Lindblad operators, as per issues with T1 simulation #452)
- Persistent value pulses (as per issue #403)
- Oscillator system types

### Proposed solution

A few options were suggested:
- Add warnings in `digest.py`, notifying the user that the feature is unsupported (and therefore may be buggy)
- Raise errors in `digest.py`, preventing the user from attempting to run simulations with the specified options
- Remove the relevant segments of code pertaining to the options before merging.

The first option seemed to be preferred in the meeting.

chriseclectic:
I believe this has been addressed, with the exception of the oscillator system case.

PR #484, opened 2019-12-06
Added warning messages for Pulse Simulator untested features

DanPuzzuoli:
### Summary

Added warning messages to alert users that they have specified pulse simulations with untested features, as per issue #462. 

### Details and comments

Warnings are issued in `digest_pulse_obj` for:

- PersistentValue pulses
- Oscillator-type systems
- Noise models

These warnings do not halt execution; they just let the user know that the outcomes are not necessarily reliable.

### DIFF
--- a/qiskit/providers/aer/openpulse/qobj/digest.py
+++ b/qiskit/providers/aer/openpulse/qobj/digest.py
@@ -17,6 +17,7 @@
 into something we can actually use.
 """
 
+from warnings import warn
 from collections import OrderedDict
 import numpy as np
 import numpy.linalg as la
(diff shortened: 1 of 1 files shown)


qiskit/providers/aer/openpulse/qobj/digest.py
```diff
@@ -61,6 +62,15 @@ def digest_pulse_obj(qobj_input, backend_options, noise_model):
         raise ValueError('Qobj must have hamiltonian in config to simulate.')
     hamiltonian = config_dict_sim['hamiltonian']
 
+    # Warnings for untested features
```
atilag: Can you move all this code to a function? The `digest_pulse_obj` is already huge enough:
Maybe something like:
```
warnings(hamiltonian, noise_model, qobj['experiments'])
```
And do the check/`warn()` there.

DanPuzzuoli: Makes sense, thanks! I've made this change.

PR #485, opened 2019-12-06
Add support for parameterized qobj

chriseclectic:
### Summary

Allows specifying a parameterized qobj which will be expanded into multiple circuits when loading the qobj JSON.

### Details and comments

qobj paramterizations are specified in the `QobjConfig` as 

`"parameterizations": [exp_0_params, ... ]`

where `exp_k_params = [[[i, j], [val_0, val_1, ....]], ...]`
where `i` is the instruction position in the `k` qobj circuit, `j` is the param position for that instruction, and `val_n` is the value of the n-th parameterization.

For a given exp_param each parameterization instruction must have the same number of parameterization value, however the number of parameterizations can be different for each circuit in the qobj.

If non-empty the length of `"parameterizations"` must match the number of experiments in the qobj or an invalid qobj exception will be raised (if one of the circuits shouldn't be parameterized then it's `exp_k_params` should be set to an empty list `[]`).

### Example

```python
import numpy as np
from qiskit import *

sim = Aer.get_backend("qasm_simulator")
n_params0 = 5
n_params1 = 10
parameterizations = [
    [[[0, 0], np.linspace(0, 2*np.pi, n_params0)]],
    [[[1, 2], np.linspace(0, 2*np.pi, n_params1)]]
]

qc1 = QuantumCircuit(1, 1)
qc1.ry(1, 0)
qc1.measure(0, 0)

qc2 = QuantumCircuit(2, 2)
qc2.h(1)
qc2.u1(1, 1)
qc2.h(1)
qc2.measure([0, 1], [0, 1])

qobj = assemble(transpile([qc1, qc2], basis_gates=['h', 'u3'],optimization_level=0),
               parameterizations=parameterizations)
result = sim.run(qobj).result()

print('Status: {}'.format(result.status))
print("Results for parameterized circuit 0:")
for j in range(n_params0):
    print(result.get_counts(j))

print("Results for parameterized circuit 0:")
for j in range(n_params0, n_params0 + n_params1):
    print(result.get_counts(j))
```
returns
```
Status: COMPLETED
Results for parameterized circuit 0:
{'0': 1024}
{'0': 502, '1': 522}
{'1': 1024}
{'0': 502, '1': 522}
{'0': 1024}
Results for parameterized circuit 0:
{'00': 1024}
{'10': 111, '00': 913}
{'10': 412, '00': 612}
{'10': 755, '00': 269}
{'10': 990, '00': 34}
{'10': 990, '00': 34}
{'10': 755, '00': 269}
{'10': 412, '00': 612}
{'10': 111, '00': 913}
{'00': 1024}
```

### DIFF
--- a/src/framework/qobj.hpp
+++ b/src/framework/qobj.hpp
@@ -29,7 +29,7 @@ namespace AER {
 //============================================================================
 
 class Qobj {
-public:
+ public:
   //----------------------------------------------------------------
   // Constructors
   //----------------------------------------------------------------
@@ -51,13 +51,12 @@ public:
   json_t config;                  // (optional) qobj level config data
 };
 
-
 //============================================================================
 // JSON initialization and deserialization
 //============================================================================
 
 // JSON deserialization
-inline void from_json(const json_t &js, Qobj &qobj) {qobj = Qobj(js);}
+inline void from_json(const json_t &js, Qobj &qobj) { qobj = Qobj(js); }
 
 Qobj::Qobj(const json_t &js) {
   // Check required fields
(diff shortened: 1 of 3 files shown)


src/framework/qobj.hpp
```diff
@@ -83,23 +82,76 @@ Qobj::Qobj(const json_t &js) {
...
+      // Non parameterized circuit
+      circuits.push_back(circuit);
+    } else {
+      // Load different parameterizations of the initial circuit
+      const auto circ_params = param_table[i];
+      const size_t num_params = circ_params[0].second.size();
+      const size_t num_instr = circuit.ops.size();
+      for (size_t j=0; j<num_params; j++) {
+        // Make a copy of the initial circuit
+        Circuit param_circuit = circuit;
```
ereastman: Is it necessary to copy the circuit here? You create `circuit`, and then in the case that there are params, you copy the circuit, edit that copy, and then push the copy and throw out the original - why not just edit the original?

Also, I might try using reserve, emplace_back, and in-place editing of the circuits in circuits but perhaps that's too far

src/framework/qobj.hpp
```diff
@@ -83,23 +82,76 @@ Qobj::Qobj(const json_t &js) {
...
-  // Parse experiments
   const json_t &circs = js["experiments"];
-  for (const auto &circ : circs) {
-    Circuit circuit(circ, config);
-    // Override random seed with fixed seed if set
-    // We shift the seed for each successive experiment
-    // So that results aren't correlated between experiments
-    if (seed >= 0) {
+  const size_t num_circs = circs.size();
+
```
ereastman: circuits.reserve(num_circs)

src/framework/qobj.hpp
```diff
@@ -83,23 +82,76 @@ Qobj::Qobj(const json_t &js) {
...
+          }
+          auto &op = param_circuit.ops[instr_pos];
+          if (param_pos >= op.params.size()) {
+            throw std::invalid_argument(R"(Invalid parameterized qobj: instruction param position out of range)");
+          }
+          if (j >= params.second.size()) {
+            throw std::invalid_argument(R"(Invalid parameterized qobj: parameterization value out of range)");
+          }
+          // Update the param
+          op.params[param_pos] = params.second[j];
```
ereastman: how does the typing work here? Do we only allow a single parameter type? Or am I missing some clever templating?

chriseclectic: At the moment this only works for complex or double params (basically u1, u2, u3 gates).

chriseclectic:
@kdk It looks like an extra set of brackets because there is only 1 parameterized instruction in each of the example circuits. 

Also yes, it is sensitive to order in the assembled qobj and assumes whomever made the parameterization configuration knew the index of each instruction in the qobj experiments lists.

As you say, even with optimization level zero there is no guarantee this will be the same order as circuits (Maybe if you call assemble on the circuit without any transpiling you can guarantee this though?)

kdk:
> As you say, even with optimization level zero there is no guarantee this will be the same order as circuits (Maybe if you call assemble on the circuit without any transpiling you can guarantee this though?)

This will be true as long as there aren't any classical conditionals (which would add instructions for the qasm to qobj conditional bfunc conversion: https://github.com/Qiskit/qiskit-terra/blob/2ee7a3a/qiskit/assembler/assemble_circuits.py#L121 ).

PR #491, opened 2019-12-11
Fix build error in gpu

hhorii:
### Summary
Standalone build is failed in the current GPU branch.
CLang 11 should be enabled on Mac

### Details and comments
```
mkdir build/
cd build/
cmake -D AER_THRUST_BACKEND=OMP ..
make
```


CMakeLists.txt
```diff
@@ -41,8 +41,6 @@ set(AER_SIMULATOR_CPP_EXTERNAL_LIBS
 	"${AER_SIMULATOR_CPP_SRC_DIR}/third-party/win64/lib"
 	"${USER_LIB_PATH}")
 
-set(AER_COMPILER_DEFINITIONS "")
```
atilag: This was meant to be a placeholder for all Aer compiler definitions, not only Thrust related ones. Every new compiler definition should be added here.
We don't really want to pollute the space with specific variables for every use case, we can end up with dozens of variables which is harder to understand and maintain.
I'm suggesting to stick with AER_COMPILER_DEFINITIONS variable, and add all Thrust required definitions in here, instead of creating AER_THRUST_SUPPORTED_DEFINITION / AER_THRUST_DEVICE_SYSTEM_DEFINITION

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -1206,7 +1206,7 @@ class MatrixMult4x4 : public GateFuncBase
 };
 
 template <typename data_t>
-class MatrxMult8x8 : public GateFuncBase
+class MatrixMult8x8 : public GateFuncBase
```
atilag: Good catch!

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -1808,7 +1808,7 @@ void QubitVectorThrust<data_t>::apply_multiplexer(const reg_t &control_qubits,
 }
 
 template <typename data_t>
-class DiagonalMult2x2 : public GetFuncBase
+class DiagonalMult2x2 : public GateFuncBase
```
atilag: Ouch, good catch too!

CMakeLists.txt
hhorii: The current cmake generates wrong parameters.
```
$ cmake -version
cmake version 3.14.5
$ cat CMakeFiles/qasm_simulator.dir/flags.make | grep CXX_DEFINES
CXX_DEFINES = -D THRUST_DEVICE_SYSTEM="THRUST_DEVICE_SYSTEM_OMP AER_THRUST_SUPPORTED"
```
 I guess that use of `target_compile_definitions` is wrong.

atilag: Oh Ok, then maybe we have to use ";" or ":"  for splitting between variables... let me take a look.

atilag: Ok, we only need to remove the quotes, so `AER_COMPILER_DEFINITIONS` is now a list instead of a string:
Change this:
```
set(AER_COMPILER_DEFINITIONS "${AER_COMPILER_DEFINITIONS} THRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_OMP")
```
to this:
```
set(AER_COMPILER_DEFINITIONS ${AER_COMPILER_DEFINITIONS} THRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_OMP)
```
... in all the places.

atilag: Another option would be using CMake `list()` function like:
```
list(APPEND AER_COMPILER_DEFINITIONS THRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_OMP)
```
Whatever you like the most... I really don't mind

CMakeLists.txt
```diff
@@ -245,7 +249,8 @@ else() # Standalone build
 		RUNTIME_OUTPUT_DIRECTORY_DEBUG Debug
 		RUNTIME_OUTPUT_DIRECTORY_RELEASE Release)
 	target_include_directories(qasm_simulator PRIVATE ${AER_SIMULATOR_CPP_SRC_DIR})
```
atilag: @hhorii this is not your fault (is mine actually), but I find out that we need to fix this line to something like:
```
target_include_directories(qasm_simulator
		PRIVATE ${AER_SIMULATOR_CPP_SRC_DIR}
		PRIVATE ${AER_SIMULATOR_CPP_EXTERNAL_LIBS})
```
Could you change this as well? Thanks!

PR #493, opened 2019-12-11
Add HamiltonianModel class for configuring PulseSimulator

chriseclectic:
### Summary

Adds a `HamiltonianModel` class to store the hamiltonian configuration parts of the `OPSystem` class and hamiltonian utility functions from pulse qobj digest.

### Details and comments

Currently I just refactored digest to use this class without actually modifying `OPSystem` (it builds the HamiltonianModel, then copies entries back to OPSystem) so as to minimize code changes. In a future PR this could be refactored so OPSystem stores the HamiltonianModel.

The pulse utility functions should also be refactored (or removed) to use this class.


qiskit/providers/aer/openpulse/hamiltonian_model.py
```diff
@@ -0,0 +1,313 @@
...
+        system.parse(qubits)
+        self._system = system.compiled
+
+        # Step #2: Determine Hamiltonian channels
+        self._get_hamiltonian_channels()
+
+        # Step 3: Calculate diagonal hamiltonian
+        self._get_diag_hamiltonian()
+
+    def initial_state(self):
```
DanPuzzuoli: Along the lines of what you had suggested, I think a good name/argument for this method would be either `eigenstate(level=0)` or even `drift_eigenstate(level=0)`. It's a bit more verbose but `drift_eigenstate` is very descriptive.

qiskit/providers/aer/openpulse/hamiltonian_model.py
```diff
@@ -0,0 +1,313 @@
...
+        """Return the initial state of the time-independent hamiltonian"""
+        # Set initial state
+        ground_state = 0 * op.basis(len(self._evals), 1)
+        for idx, estate_coef in enumerate(self._estates[:, 0]):
+            ground_state += estate_coef * op.basis(len(self._evals), idx)
+        return ground_state
+
+    def calculate_frequencies(self, qubit_lo_freq=None, u_channel_lo=None):
```
DanPuzzuoli: I think the name of this is good for this iteration. Something that might make sense next, as we refactor so that OPSystem holds a HamiltonianModel, is to break this function up: one (or several) in HamiltonianSystem that compute eigenvectors/eigenvalues/dressed energies for the drift Hamiltonian, and then another in OPSystem called something like `channel_frequencies`, which, if passed the argument `from_hamiltonian` will draw the relevant info from `HamiltonianModel`.

chriseclectic: Do you think this one should be made a private function for now? `def _calculate_frequencies` ?

qiskit/providers/aer/openpulse/hamiltonian_model.py
```diff
@@ -0,0 +1,344 @@
...
+                for u_lo_idx in u_channel_lo[chidx]:
+                    if u_lo_idx['q'] < len(qubit_lo_freq):
+                        qfreq = qubit_lo_freq[u_lo_idx['q']]
+                        qscale = u_lo_idx['scale'][0]
+                        freqs[key] += qfreq * qscale
+            else:
+                raise ValueError("Channel is not D or U")
+        return freqs
+
+    def _get_hamiltonian_channels(self):
```
atilag: `get` semantics would expect the function to return something but it doesn't, what about:
```python
def _calculate_hamiltonian_channels(self):
```
?

qiskit/providers/aer/openpulse/hamiltonian_model.py
```diff
@@ -0,0 +1,344 @@
...
+                    channels.append(temp_chan)
+        channels.sort(key=lambda x: (int(x[1:]), x[0]))
+
+        channel_dict = OrderedDict()
+        for idx, val in enumerate(channels):
+            channel_dict[val] = idx
+
+        self._channels = channel_dict
+
+    def _get_diag_hamiltonian(self):
```
atilag: same here:
```python
def _calculate_diag_hamiltonian(self):
```

qiskit/providers/aer/openpulse/hamiltonian_model.py
```diff
@@ -0,0 +1,344 @@
...
+    # compute overlaps to unique evals
+    overlaps = np.zeros(len(unique_evals))
+    for idx, val in enumerate(unique_evals):
+        overlaps[idx] = _proj_norm(evecs[:, val == rounded_evals], u)
+
+    # return eval with largest overlap
+    return unique_evals[np.argmax(overlaps)]
+
+
+def _proj_norm(A, b):
```
atilag: This is a war I lose long time about.. but here I go: I guess that `A` and `b` are very common in math contexts, but extremely ugly for readability, would some other naming be acceptable here? :)

qiskit/providers/aer/openpulse/hamiltonian_model.py
```diff
@@ -0,0 +1,344 @@
...
+        dim_qub (dict): a dictionary with keys being qubit index as a string, and
+                        value being the dimension of the qubit
+
+    Returns:
+        vector: the state with qubit_idx in state 1, and the rest in state 0
+    """
+    vector = np.array([1.])
+
+    # iterate through qubits, tensoring on the state
+    for idx, dim in enumerate(dim_qub):
```
DanPuzzuoli: Btw `enumerate` causes an error here, as for some reason it just iterates over the keys, setting both `idx` and `dim` to the current key. Also, I had it iterating over `idx in range(len(dim_qub))` as this way it made sure that the qubit indexes are iterated over in order.

Also, I just realized my old way of doing it is also liable to raise errors, e.g. if a user specifies qubits `[0,2]` for a 3 qubit device. Maybe need to iterate over the list `dim_qub.keys()` after it has been sorted

qiskit/providers/aer/openpulse/hamiltonian_model.py
DanPuzzuoli: I do find it readable purely due to the commonality to math contexts, as well as it primarily being a call to `la.lstsq`, which uses `A` and `b` as arguments. I can't think of anything else, but I may just be too heavily indoctrinated :P .

qiskit/providers/aer/openpulse/hamiltonian_model.py
```diff
@@ -0,0 +1,344 @@
...
+        system.parse(qubits)
+        self._system = system.compiled
+
+        # Step #2: Determine Hamiltonian channels
+        self._get_hamiltonian_channels()
+
+        # Step 3: Calculate diagonal hamiltonian
+        self._get_diag_hamiltonian()
+
+    def drift_eigenstate(self, energy_level=0):
```
DanPuzzuoli: One comment here: the eigenvalues are not necessarily sorted from lowest to highest, so either this function needs to order the eigenvectors according to the eigenvalue first, or this needs to be done when `_evals` and `_estates` are set initially.

qiskit/providers/aer/openpulse/hamiltonian_model.py
chriseclectic: Fixed by writing as `dim_qub.items()` loop

qiskit/providers/aer/openpulse/hamiltonian_model.py
chriseclectic: Ah, thought they were already sorted. How did we assume the 0 element was ground state / initial state before?

qiskit/providers/aer/openpulse/hamiltonian_model.py
chriseclectic: Change to `mat` `vec` to make Juan happy

qiskit/providers/aer/openpulse/hamiltonian_model.py
DanPuzzuoli: It's been like that since I started working on it; I think with the matrix being close to diagonal already, and the lowest eval in the top left, it has been working by coincidence.

chriseclectic: I ended up just deleting this function

PR #494, opened 2019-12-13
Remove pip pin in azure jobs

mtreinish:
### Summary

In the past we needed to pin the pip version we used in the ci jobs to
avoid a pip regression introduced in pip 19.0. But that has since been
fixed so pinning is no longer necessary. This commit updates the azure
pipelines configuration to remove the pip pinning so we can use the
latest version.


atilag:
The failing test seems to be a temporay problem of Azure, so I'm merging this.

PR #503, opened 2019-12-23
Removed `noise_model` and `sim_config` as an argument in the pulse simulator

DanPuzzuoli:
### Summary
As it won't be supported, removed `noise_model` argument from the pulse simulator backend, as well as the `digest` file. It can still be passed and detected through `backend_options`, but an `AerError` will now be raised.

In addition: 

- all reference to `sim_config` has been removed; any options should be passed through `backend_options`
- The warning for PersistentValue pulses has also been changed to an `AerError`

### DIFF
--- a/qiskit/providers/aer/openpulse/qobj/digest.py
+++ b/qiskit/providers/aer/openpulse/qobj/digest.py
@@ -29,7 +29,7 @@ from ..cy.utils import oplist_to_array
 from . import op_qobj as op
 
 
-def digest_pulse_obj(qobj, system_model, backend_options=None, noise_model=None):
+def digest_pulse_obj(qobj, system_model, backend_options=None):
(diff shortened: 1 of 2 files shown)


qiskit/providers/aer/openpulse/qobj/digest.py
```diff
@@ -212,14 +208,15 @@ def _unsupported_warnings(qobj_dict, noise_model):
         noise_model (dict): backend_options for simulation
     Returns:
     Raises:
+        AerError: for unsupported features
     """
 
     # Warnings that don't stop execution
     warning_str = '{} are an untested feature, and therefore may not behave as expected.'
     if noise_model is not None:
-        warn(warning_str.format('Noise models'))
```
chriseclectic: I think we should keep the noise model as a warning for now

PR #509, opened 2020-01-09
Update MPS expectation value snapshots

yaelbh:
### Summary

Updated MPS expectation value snapshots to support snapshots with variance and single shot.

### Details and comments

@merav-aharoni can you please check a failing test.

### DIFF
--- a/test/terra/backends/test_qasm_simulator_matrix_product_state.py
+++ b/test/terra/backends/test_qasm_simulator_matrix_product_state.py
@@ -76,8 +76,8 @@ class TestQasmMatrixProductStateSimulator(
         QasmResetNoiseTests,
         QasmSnapshotStatevectorTests,
         QasmSnapshotProbabilitiesTests,
-        QasmSnapshotStabilizerTests,
-        QasmSnapshotExpValPauliTests,
+        QasmSnapshotStabilizerTests
+        #QasmSnapshotExpValPauliTests,
         #QasmSnapshotExpValMatrixTests,
         ):
     """QasmSimulator matrix product state method tests."""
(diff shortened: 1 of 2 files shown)


yaelbh:
This pull request works fine for the Pauli expectation value but not for the matrix expectation value. I've removed the test of the matrix expectation value, so all tests should pass now, and the pull request is ready for review. @chriseclectic could you please make it to the upcoming release. It is important because it allows faster simulation of VQE instances.

yaelbh:
I see that tests are failing. They pass for me locally. Checking.

yaelbh:
Apparently there is a core dump in ` MPS_with_new_indices` (called from `MPS::expectation_value_pauli`). @merav-aharoni will check.

test/terra/backends/test_qasm_simulator_matrix_product_state.py
```diff
@@ -75,6 +77,8 @@ class TestQasmMatrixProductStateSimulator(
         QasmSnapshotStatevectorTests,
         QasmSnapshotProbabilitiesTests,
         QasmSnapshotStabilizerTests
+        #QasmSnapshotExpValPauliTests,
```
chriseclectic: What is the issue with the failing tests? Is it due to a bug in the expectation value method or something else?

merav-aharoni: There are actually 3 issues in this PR:
1. bug in the interface of MPS expectation value, following the addition of SnapshotDataType. This was fixed by Yael.
2. bug in Pauli expectation value when `qubits` parameter is not given in ascending order.
3. bug in matrix expectation value when `qubits` parameter is not given in ascending order.
These were discovered because Yael added new tests after her change.
I am in the process of fixing 2 and 3, but would rather address them as separate issues (I will open these). The interface change (1) is more urgent, because that enables the users to use MPS in VQE. I stress that when the qubits are given in the default order, the expectation value works fine.

atilag:
Superseed by #619 
Closing.

Issue #514, opened 2020-01-09
Transmon Pulse SystemModel/HamiltonianModel generation functions


chriseclectic:
### What is the expected behavior?

Add functions to generate a pulse Hamiltonian for a network of transmon qubits.

This should include specifying the system by

- Frequencies of the qubits
- Anharmonicities of the qubits
- Coupling strengths between coupled qubits

PR #516, opened 2020-01-10
Duffing system model generators for pulse simulator

DanPuzzuoli:
### Summary

Functionality requested in issue #514 . A user can now construct and simulate a transmon system by only specifying a list of frequencies and couplings, without any reference to the string format.

- Main addition is the file `transmon_model_generators.py`, containing functionality for generating `PulseSystemModel` objects for transmon system models
- Tests for all functions in `transmon_model_generators.py` have been added in `test_transmon_model_generators.py`
- A new example notebook as been added: `example/transmon_model_example.ipynb`, which demonstrates the usage. (This is for internal viewing of how it works, and also to do sanity check that some of the simulations in `example/pulse_sim.ipynb` work as expected.)

### Details and comments

The only "public" function in `transmon_model_generators.py` is `transmon_system_model`, though there are a variety of functions that it builds on (that may be of later use). To create the system model, the user needs to specify:
- number of transmons
- dimension of transmons
- transmon frequencies, anharmonicities, drive strengths
- a "coupling graph", i.e. a specification of what qubits are coupled, along with the coupling strengths
- dt (the sampling rate for pulses)

The function returns a tuple with:
- A `PulseSystemModel` object representing the specified system, that can be plugged into the simulator
- a `dict`, which in the code is called `cr_idx_dict` (short for cross-resonance index `dict`). This `dict` stores u channel index information for doing cross-resonance pulses. E.g. to drive a CR pulse on qubit `q_driven` (an `int`) with target qubit `q_targ`, use the u channel with index `cr_idx_dict[(q_driven, q_targ)]`.


DanPuzzuoli:
Note: previous attempts at tests were due to an error being raised in `assemble` in one of the pulse simulator tests. This error wasn't occurring on my machine locally, even after updating to the latest version of the master branch of terra (not sure if I made a mistake here).

The test was failing due to the pulse simulator `DEFAULT_CONFIGURATION` not having `meas_levels`. I added this as `{ 'meas_levels': [0,1,2]}`  and it seems to work fine.

haggaila:
Hi, if the code implements a nonlinear oscillator with a quartic nonlinearity (proportional to n^2 with n = a^\dag a), it may be more precise to refer to it as a "Duffing" (or "quartic") oscillator, as it is an approximation of a transmon model with a cosine-type nonlinearity, which may be implemented in the future.

chriseclectic:
@haggaila, yes you are right, I think we should rename this to *duffing oscillator model*

DanPuzzuoli:
I've converted all relevant files to use nomenclature relating to "duffing oscillator" instead of "transmon"

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+        freq_symbol (str): Frequency symbol for internal HamiltonianModel
+        anharm_symbol (str): Anharmonicity symbol for internal HamiltonianModel
+        drive_symbol (str): Drive symbol for internal HamiltonianModel
+        coupling_symbol (str): Coupling symbol for internal HamiltonianModel
+
+    Returns:
+        tuple[PulseSystemModel, dict]: The generated oscillator system model, and a dict specifying
+                                       ControlChannel indices for CR driving.
+
+                                       For example:
```
chriseclectic: The example should be moved to its own section in doc string

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+    1 is modeled with the Hamiltonian term:
+        2*pi*r0*X0*U(t),
+    where r0 is the drive strength for oscillators 0, X0 is the drive operator for oscillators 0,
+    and U(t) is the signal for the CR drive. U(t) is set to have carrier frequency equal to that of
+    the target oscillators.
+
+    Indices for ControlChannels corresponding to CR drives are provided in the returned cr_idx_dict.
+
+    Args:
+        num_oscillators (int): Number of oscillators in the model
```
chriseclectic: Is this necessary? Can't the number of oscillators be inferred from the lenght of `oscillator_freqs`

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+        coupling_dict (dict): Specification of the coupling graph with keys being edges, and values
+                              the coupling strengths.
+
+                              For example:
+
+                              * ``{(0,1): 0.02, (1,3): 0.01}`` specifies a system in which
+                                oscillators (0,1) are coupled with strength 0.02, and (1,3) are
+                                coupled with strength 0.01
+        dt (float): Pixel size for pulse instructions
+        freq_symbol (str): Frequency symbol for internal HamiltonianModel
```
chriseclectic: What is the purpose of these symbol kwargs? Will a user ever need to override them or can we just set them to their default values inside the function?

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+        drive_strengths (list): Drive strength values, assumed to be of length num_oscillators
+        coupling_dict (dict): Specification of the coupling graph with keys being edges, and values
+                              the coupling strengths.
+
+                              For example:
+
+                              * ``{(0,1): 0.02, (1,3): 0.01}`` specifies a system in which
+                                oscillators (0,1) are coupled with strength 0.02, and (1,3) are
+                                coupled with strength 0.01
+        dt (float): Pixel size for pulse instructions
```
chriseclectic: Can `dt` have a default value?

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+        freq_symbol (str): Frequency symbol for internal HamiltonianModel
+        anharm_symbol (str): Anharmonicity symbol for internal HamiltonianModel
+        drive_symbol (str): Drive symbol for internal HamiltonianModel
+        coupling_symbol (str): Coupling symbol for internal HamiltonianModel
+
+    Returns:
+        tuple[PulseSystemModel, dict]: The generated oscillator system model, and a dict specifying
```
chriseclectic: returning the idx as well as the pulse model seems messy. It makes me think that the coupling channel should be a method of the `PulseSystemModel` itself

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+                                       ControlChannel indices for CR driving.
+
+                                       For example:
+
+                                       * Given the coupling_dict ``{(0,1): 0.02, (1,3): 0.01}``,
+                                       the returned cr_idx_dict will be:
```
chriseclectic: If order matters for assigning channels, maybe the couping input should be a list `[[(i,j), val), ...]`, where list-position becomes channel number, rather than a dict

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+                              drive_strengths,
+                              drive_symbols,
+                              ordered_coupling_edges,
+                              coupling_strengths,
+                              coupling_symbols,
+                              cr_idx_dict):
+    """Creates a hamiltonian string dict for a duffing oscillator model
+
+    Note, this function makes the following assumptions:
+        - oscillators, oscillator_dims, oscillator_freqs, freq_symbols, anharm_freqs,
```
chriseclectic: As with above function assumption means `oscillators` is unnecessary

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+    if isinstance(arg, Iterable):
+        return arg
+
+    return [arg]
+
+
+# Helper classes
+
+
+class _coupling_graph:
```
chriseclectic: Class should be capitalized `class CouplingGraph`

qiskit/providers/aer/openpulse/duffing_model_generators.py
```diff
@@ -0,0 +1,438 @@
...
+    the target oscillators.
+
+    Indices for ControlChannels corresponding to CR drives are provided in the returned cr_idx_dict.
+
+    Args:
+        num_oscillators (int): Number of oscillators in the model
+        dim_oscillators (int): Dimension of truncation for each oscillator
+        oscillator_freqs (list): Oscillator frequencies, assumed to be of length num_oscillators
+        anharm_freqs (list): Anharmonicity values, assumed to be of length num_oscillators
+        drive_strengths (list): Drive strength values, assumed to be of length num_oscillators
```
chriseclectic: I'm not sure if this is a frequency or a scale value? What does its value mean relative to `oscillator_freqs` for example (ie is it same units)

DanPuzzuoli:
- With regards to units: 
    - there is no assumed timescale. Nothing in the simulator cares if it's GHz/ns or Hz/s; so long as the units are consistent, it will output the same results regardless of the units used. We *could* write in the doc string that we are assuming some units, but it seems odd to do so given that it has no effect on the behaviour.
    - Also, frequencies are radial, i.e. everything gets multiplied by `2*pi`. The multiplication by `2*pi` is described in the doc string, but it is worth adding a point right at the beginning to say that everything is assumed radial.
- With regards to defaults: 
    - I would prefer to not have default values, as 'natural' values for one parameter will depend on what the other parameters are, and it's not clear to me what value will be provided for the added complication of figuring these things out.

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: Will do.

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: Yeah I can just change it to do that.

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: I put them there due to the fact that, unfortunately, the symbols survive in the background of `HamiltonianModel`, and so it could be useful to have some control over this. You're right though that at the moment there is no use for this, so I can remove them.

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: This is related to the units. I lean towards no but we can discuss this.

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: The issue is that U channels are completely arbitrary; in this case we are using them for a particular type of two-qubit interaction, but in general they can be anything. As a result I'm not sure this index information should be put into `PulseSystemModel`, as the meaning of the identifying information (the ordered qubit pair) is strongly tied to the particulars of this model.

The options I can think of are:

1. Put the information into `PulseSystemModel`: I think to maintain generality of this class we would need to add a general dictionary to it for control channel index info, where the type of the keys is arbitrary, so that they can contain identifying info for channel indices of any format, to not make assumptions about the model.
2. Define a new class called `DuffingSystemModel` that inherits from `PulseSystemModel`, but that stores `cr_idx_dict`, and has an added method for looking things up in it. I do kind of like this option as it seems like a natural thing to do when more functionality is needed beyond simply containing info necessary for running the simulation.
3. Leave it as is. I think for now I prefer this option as it's pretty simple, and we can think more about what the `PulseSystemModel` should be.

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: The issue here is that for the purpose of specifying couplings there should be no difference between `(0, 1)` and `(1, 0)`; they should both just mean that qubits 0 and 1 are coupled. As such if we want the user to specify u channel orders themselves it should be in a separate argument (as otherwise `val` will need to be duplicated for both `(0,1)` and `(1, 0)`). In terms of method signature/output it would be simpler to make the user specify u channel orderings themselves, but in terms of actual use it will be more cumbersome.

It's somewhat annoying to deal with as for u channels, consistency in order matters, but the ordering itself is completely arbitrary.

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: Here oscillators is a list containing arbitrary indices.

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: I want to keep this private, how is `_CouplingGraph`?

qiskit/providers/aer/openpulse/duffing_model_generators.py
DanPuzzuoli: `oscillator_freqs`, `anharm_freqs`, and `drive_strengths` are all the same units. As described in the doc string, the `drive_strengths` elements enter as coefficients in the Hamiltonian. I called it `drive_strengths` instead of `drive_freqs` as 'drive freq' is potentially confusable with the lo frequency.

DanPuzzuoli:
As discussed, the changes are:
- Added a dictionary `PulseSystemModel.control_channel_dict` and a function `PulseSystemModel.control_channel_index` that queries the dictionary (and provides warnings if a key is non-existent).
- updated `duffing_system_model` to only return a `PulseSystemModel` 
- changed `_coupling_graph` to `CouplingGraph`
- updated all relevant test cases to reflect these changes
- updated the model generator example

Issue #433, opened 2019-11-12
Add PulseSimulator to API docs


chriseclectic:
Add the `PulseSimulator` to the API documentation.

Move the content from the pulse simulator readme.md into the `PulseSimulator` class doc string so it is visible in the API documentation.

PR #521, opened 2020-01-14
Doc strings and default behavior for pulse simulator

### LINKED ISSUE #501: Cleanup/clarify pulse simulation defaults behaviour and digest
Some tasks:

- Remove the `sim_config` step from digest (this is easy, but is perhaps conveniently done in conjunction with some other things). This could break some old notebooks, but they are likely already broken by other recent changes.
- Put a step in `digest` allowing `backend_options` to override any other setting. (This was there before, but was removed in the original effort to separate types of parameters.)
- Make handling of defaults more clear. A `PulseDefaults` object has been added to the pulse simulator backend, which contains some dummy values that will allow `assemble` to complete without throwing errors. Some of these parameters are necessary for simulation, and some aren't. For the ones that are, like `qubit_lo_freq`, the pulse default is given as `[np.inf]`. Steps should be added to digest to determine the `qubit_lo_freq` if it is passed in the `qobj` as `[np.inf]`: e.g. first check if there is a `qubit_freq_est` in the system model, and if not, compute it from the Hamiltonian. In this case, warn the user.

DanPuzzuoli:
### Summary

This PR contains both clarification of default behaviors in `digest.py`, as well as added doc strings for `HamiltonianModel`, `PulseSystemModel`, and `PulseSimulator`. (Both were done at the same time as default behavior influences doc strings.) The doc string for `PulseSimulator` will likely need to be edited, but this is at least a start, and contains all of the information from the pulse simulator readme. 

Closes #501 and #433 

### Details and comments

Specific changes for `digest`:
- Updated default handling of `qubit_lo_freq` in `digest.py`
   - First it tries to populate it from the `qobj` passed by the user
   - If not present, takes it from `system_model._qubit_freq_est`
   - If this is not present, it automatically computes it from Hamiltonian and uses that
- Updated the old tests for `digest` to test for the above behavior.
- Added a piece of code to `digest` that allows `backend_options` to override any key/value in `qobj_config`.

Changes for doc strings:
- Updated doc strings for `HamiltonianModel` and `PulseSystemModel`
- Added doc string for `PulseSimulator`. May need more work, but it at least contains everything from the original readme.


qiskit/providers/aer/backends/pulse_simulator.py
```diff
@@ -33,19 +33,87 @@
 
 class PulseSimulator(AerBackend):
     """Aer OpenPulse simulator
+
+    The `PulseSimulator` simulates pulse `Schedules` on a model of a quantum system, where a model
+    is specified by a `PulseSystemModel` object, which stores Hamiltonian and control channel
```
chriseclectic: you should be able to link documentation by using sphinx directives. Eg: 
```
:class:`~qiskit.path-to.PulseSystemModel`
```

qiskit/providers/aer/backends/pulse_simulator.py
```diff
@@ -33,19 +33,87 @@
...
+
+    `PulseSystemModel` objects can be constructed from backends with a hamiltonian description, or
+    from the `transmon_system_model` function.
+
+    Results are returned in the same format as when jobs are submitted to actual devices.
+
+    **Example of usage**
+
+    To use the simulator, `assemble` a `PulseQobj` object from a list of pulse `Schedules`, using
+    `backend=PulseSimulator()`.
```
chriseclectic: In sphinx you need to use double-ticks for "code" formatting eg:

```
``backend=PulseSimulator()``.
```

chriseclectic:
@DanPuzzuoli can you rebase this PR against master now that the `openpulse-sim` branch has been merged

PR #522, opened 2020-01-16
Chunk based optimization of GPU simulator

doichanj:
### Summary
I changed statevector storage from using unified memory to storing on the device memory of each GPU. So I implemented chunk based algorithm to parallelize by multiple GPUs, this algorithm also will be necessary for multiple-node parallelization.

### Details and comments
When all the statevector can be stored on single GPU, it only used 1 GPU (only 1 chunk is used), and it automatically uses multiple GPUs or also host's memory when statevector can not be stored on single GPUs or all GPUs. 

This implementation is faster (and more stable) than previous implementation if number of qubits is large that can not be stored on single GPU. However it is slower than previous implementation for smaller qubits, because initialization cost is higher than to use unified memory.


src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -615,34 +1181,92 @@ std::complex<data_t> &QubitVectorThrust<data_t>::operator[](uint_t element) {
...
   #endif
-  return data_[element];
+
+	uint_t lcid,lid;
+	int iPlace = GlobalToLocal(lcid,lid,element);
+
+	UpdateReferencedValue();
+
+	if(iPlace >= 0){
+		data_[0] = (std::complex<data_t>)m_Chunks[iPlace].GetState(lcid,lid);
```
hhorii: why not `data[element]` ?

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -453,7 +1049,7 @@ json_t QubitVectorThrust<data_t>::json() const {
   const int_t END = data_size_;
   const json_t ZERO = std::complex<data_t>(0.0, 0.0);
   json_t js = json_t(data_size_, ZERO);
-
+
```
hhorii: I guess `data_` is inconsistent in `QubitVectorThrust<data_t>::json()`.

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -1005,54 +1695,80 @@ std::complex<double> QubitVectorThrust<data_t>::inner_product() const
...
+	zero();
+
+	thrust::complex<data_t> t;
+	t = 1.0;
+
+	if(m_globalChunkIndex == 0){
+		m_Chunks[0].SetState(0,0,t);
+	}
+
+	thrust::complex<data_t> r = m_Chunks[0].GetState(0,0);
```
hhorii: not necessary?

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -380,16 +982,26 @@ class QubitVectorThrust {
...
-	thrust::complex<data_t>** m_ppBuffer;	//pointer to buffers
-	int m_matBits;							//max number of fusion bits
-	uint_t m_matSize;
+	mutable std::vector<QubitVectorChunkContainer<data_t>> m_Chunks;
 
-#ifdef DEBUG
+	int m_chunkBits;						//bits per chunk
+	uint_t m_numGlobalChunks;				//number of total chunks
+	uint_t m_numChunks;						//number of chunks in this process
+	uint_t m_globalChunkIndex;				//starting chunk ID for this process
```
hhorii: I guess this should be added when multi-process implementation is supported.

chriseclectic:
This PR can be merged as the changes have been squashed and cherry picked into #544

PR #525, opened 2020-01-20
Rewrote stabilizer Pauli expectation value

yaelbh:
### Summary

Replaced the stabilizer Pauli expectation value code with code that works directly on the tableau, instead of using the probability snapshot. The new code has polynomial run time, without imposing further constraints.


yaelbh:
Thanks @chriseclectic for the review. I fixed everything according to your comments.

PR #529, opened 2020-01-20
Fix CMakeLists.txt and FindPybind11.cmake to pass compiler options to…

atilag:
… pybind compilation

### Summary
I have improved the way compiler and linker options are passed to the compilation process and fixed the problem with not passing OpenMP and other compiler flags.

### DIFF
--- a/qiskit/providers/aer/backends/wrappers/CMakeLists.txt
+++ b/qiskit/providers/aer/backends/wrappers/CMakeLists.txt
@@ -1,7 +1,4 @@
 include(Linter)
-
-find_package(PythonExtensions REQUIRED)
-find_package(PythonLibs REQUIRED)
 find_package(Pybind11 REQUIRED)
 
 # We need to remove the -static flag, because Python Extension system only supports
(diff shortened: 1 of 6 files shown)


qiskit/providers/aer/backends/wrappers/CMakeLists.txt
```diff
@@ -1,36 +1,16 @@
 include(Linter)
-
-find_package(PythonExtensions REQUIRED)
-find_package(PythonLibs REQUIRED)
```
ereastman: Move these two lines to the FindPybind11.cmake since pybind still needs the variables set by these modules (mainly PYTHON_INCLUDE_DIRS)

atilag: They are already there: https://github.com/Qiskit/qiskit-aer/blob/master/cmake/FindPybind11.cmake#L17

ereastman: True! sorry, I meant move them up so that they're before the check

atilag:
Ok, I merged yours, solve a little conflict and rebased.

PR #535, opened 2020-01-21
Add benchmark for qft

mtreinish:
### Summary

This commit adds an initial method for running a benchmark of a qft
circuit. This creates an input qft circuit of increasing size and
times the qasm simulation with all the different methods available.

### Details and comments


mtreinish:
On hold because I have not run this locally to test it works yet

mtreinish:
Oh, also this is partially duplicated with: https://github.com/Qiskit/qiskit-aer/blob/master/test/benchmark/quantum_fourier_transform_benchmarks.py I'll respin this to combine the 2

ajavadia:
why add benchmarking to Aer rather than in the general Qiskit benchmarking platform?

mtreinish:
> why add benchmarking to Aer rather than in the general Qiskit benchmarking platform?

As of right now the aer benchmarks live in the aer repo. There was a PR open at one point to migrate it to Qiskit/qiskit (https://github.com/Qiskit/qiskit/pull/366) but it was abandoned (I don't know why). Asv benchmark suites only support tracking 1 repo at a time. The benchmark suite at https://github.com/Qiskit/qiskit/tree/master/test/benchmarks tracks terra so we can't use it to also track aer changes. I'd like these to live in Qiskit/qiskit alongside the terra benchmarks (so they're all in one place) but figured that was independent of pushing improvements to the existing suite.

atilag:
Yeah, we already have a QFT. If you want to change something in the particular PR, let's merge it here and once it's done we can integrate with all the other benchmarks in Qiskit.

PR #540, opened 2020-01-22
Pin pip<20.0.0 and bump osx python 3.8 to 3.8.1

mtreinish:
### Summary

With the recent release of pip 20.0.0 and 20.0.1 some packages fail to
pip install correctly (20.0.0 didn't even work at all). Until these
issues are resolved we should just cap the pip we use in CI to avoid
issues. This commit adds a cap to use pip < 20.0.0 until the upstream
issues are resolved. At the same time this commit updates the osx
version we use in CI for the 3.8 job to use 3.8.1 which contains bug
fixes on top of the previous version 3.8.0 we were using prior.

### Details and comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ sudo: false
 #   * stage_osx
 stage_generic: &stage_generic
   install:
+    - pip install -U 'pip<20.0.0'
     - pip install cython
     # Installing qiskit-terra master branch...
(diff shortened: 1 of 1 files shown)


mtreinish:
Looking through the bug tracker for pip (which has been quite busy post release) I believe the issue we're hitting here on pip 20 is caused by: 

https://github.com/pypa/pip/issues/7626

So we should track that for removing the pin

Issue #526, opened 2020-01-20
Hamiltonian model parsing fails with empty string


haggaila:
### Informations

- **Qiskit Aer version**: PulseSimulator 
- **Python version**: 3.7
- **Operating system**: Windows

### What is the current behavior?
Calling `HamiltonianModel.from_dict(hamiltonian)` with the dictionary in `hamiltonian['h_str']` containing an empty string causes an exception in file oppparse.py, `HamiltonianParser._token2qobj()` since variable tokens is empty and the line `return stack[0]` runs with an empty stack list.

### Suggested solutions
Either in the indicated function or in its caller `parse()`, handle empty tokens (None is already handled).

PR #541, opened 2020-01-23
3-qubit testing and a minor bug fix

haggaila:
### Summary
Fixed issue #526 and added some tests of the open pulse simulator, comparing the evolution of a 3-qubit Hamiltonian with 60 random driven terms (corresponding to all 4 * 4 * 4 Pauli products, minus the trivial drifts).


chriseclectic:
@haggaila can you rebase this PR against master now that we have merged the openpulse-sim branch

haggaila:
Added three new test functions that currently fail - there is one exception, and two other fails are unexplained or against what I expect (or I may be misusing the simulator).

DanPuzzuoli:
@haggaila I've just copied your changes locally and am playing around with things. I've found:

- For `test_3Q_X_drift`, I'm actually getting values of `nan` in the returned vector, regardless of the value of `h_abc`. Can you track this down?
- In the truncation tests: no truncation is occurring as you need to pass `qubit_list` when calling `HamiltonianModel.from_dict`. This admittedly is somewhat odd behaviour, as you're passing `qubit_list` to the `PulseSystemModel` constructor, so ideally the truncation could happen here as well. However, due to the peculiarities of the parsing, we're currently treating `HamiltonianModel` as being immutable, so for now just pass `qubit_list` when constructing the Hamiltonian. However, this also produces an error: as the identities are included in the Hamiltonian string in every term in the Hamiltonian you are constructing, the parser thinks every qubit is involved in every Hamiltonian term, so it drops every term when parsing without a full qubit list, returning an empty `system`, which causes further errors. I think that you will need to modify the string to not have identity terms for this truncation to work. Once this is done though, the tests will still return an error, due to vector dimension issues. You'll need to set the initial state for the simulator to be of the appropriate dimension given the truncation. Also, I think it would be good for the test to compare the output state to the expected vector assuming the truncation/simulation has been done correctly (as opposed to checking that it is different from what would happen with no truncation).
- For the 3Q

atilag:
@haggaila I merged with current `master` branch and there should be no conflicts anymore. 
What's the status of this PR? Could you address/discuss @DanPuzzuoli comments?
Thanks!! :)

haggaila:
Thanks @atilag, this was discussed and solved offline, and as you say should already have been corrected in master.

Issue #546, opened 2020-01-24
Pip install is broken on master


mtreinish:
### Informations

- **Qiskit Aer version**: master
- **Python version**: any
- **Operating system**: anyy

### What is the current behavior?
Attempting to pip install a bare repo no longer works since the pulse simulator branch was merged into master with: https://github.com/Qiskit/qiskit-aer/commit/40d903797e497c281bbb6fb42d1e45ccfcc482d7

Running from a clean local clone of the aer repo yields:

```
Processing ./qiskit-aer
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... done
Collecting pybind11>=2.4
  Using cached pybind11-2.4.3-py2.py3-none-any.whl (150 kB)
Collecting numpy>=1.13
  Using cached numpy-1.18.1-cp37-cp37m-manylinux1_x86_64.whl (20.1 MB)
Building wheels for collected packages: qiskit-aer
  Building wheel for qiskit-aer (PEP 517) ... error
  ERROR: Command errored out with exit status 1:
   command: /tmp/aer-git-test-37/bin/python3.7 /tmp/aer-git-test-37/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py build_wheel /tmp/tmpw7pa94fd
       cwd: /tmp/pip-req-build-h5b2eahx
  Complete output (146 lines):
  Not searching for unused variables given on the command line.
  -- The C compiler identification is GNU 9.2.0
  -- Check for working C compiler: /usr/bin/gcc
  -- Check for working C compiler: /usr/bin/gcc -- works
  -- Detecting C compiler ABI info
  -- Detecting C compiler ABI info - done
  -- Detecting C compile features
  -- Detecting C compile features - done
  -- The CXX compiler identification is GNU 9.2.0
  -- Check for working CXX compiler: /usr/bin/c++
  -- Check for working CXX compiler: /usr/bin/c++ -- works
  -- Detecting CXX compiler ABI info
  -- Detecting CXX compiler ABI info - done
  -- Detecting CXX compile features
  -- Detecting CXX compile features - done
  -- Configuring done
  -- Generating done
  -- Build files have been written to: /tmp/pip-req-build-h5b2eahx/_cmake_test_compile/build
  -- The CXX compiler identification is GNU 9.2.0
  -- The C compiler identification is GNU 9.2.0
[...]

PR #547, opened 2020-01-25
Fix missing pulse simulator dependencies in setup.py

chriseclectic:
### Summary

The recently added pulse simulator adds numpy and scipy as build and install dependencies. These were missing from the setup.py causing an error if they weren't already installed on the system.

### Details and comments

Pybind is still a remaning issue. As noted in #546, if one tries to install directly in a clean environment (eg `pip install .` or `pip install git+https...`), setup will fail unless Pybind has already been pip installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,16 +13,22 @@ try:
 except ImportError:
     subprocess.call([sys.executable, '-m', 'pip', 'install', 'scikit-build'])
     from skbuild import setup
-try:
-    import pybind11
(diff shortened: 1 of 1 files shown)


mtreinish:
I've confirmed that this fixes the issue reported in #546 by running: `
pip install -U 'git+https://github.com/chriseclectic/qiskit-aer.git@fix-pulse-build'` in a clean venv.

PR #551, opened 2020-01-27
Fix bugs in parallelization thread management and improve testing

chriseclectic:
### Summary

Custom OpenMP parallelization settings were being set correctly with respect to maximum number of threads. This PR corrects the setting and adds several additional integration tests for testing that the default and custom thread setting behavior works as expected.

This PR also adds test decorators for tests that require OpenMP or a multicore CPU so that they will be skipped if run on a system without these.

### Details and comments

* By default all available threads will be used for parallel shots and state update, subject to available memory, noise, and measure sampling.
* Parallel experiments will only ever be run if this option is explicitly configured.
* Adds test decorator `requires_omp` which will skip tests if simulator is build without OpenMP
* Adds test decorator `requires_multiprocessing` which will skip tests if the number of cpus reported by pythons multiprocessing module is not > 1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,8 @@ Added
 - Added tests for probabilities snapshot (\#380)
 - Added support for reset() in MPS simulation method (\#393)
 - Added tests for matrix and Pauli expectation value snapshot (\#386)
+- Added test decorators for tests that require OpenMP and multi-threading(\#551)
+- Added tests for automatic and custom parallel thread configuration (\#511)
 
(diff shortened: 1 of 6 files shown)


nonhermitian:
Just note that multiprocessing returns num of cpus and hyperthreads on intel chips. If it is just physical cpus you want, you may want to use the terra checking routine.

chriseclectic:
I'm using at for number of openmp threads (which also counts hyperthreads), so I think this as correct.

PR #552, opened 2020-01-27
Doc strings, default behaviour, and minor bug fixes for pulse simulator

### LINKED ISSUE #501: Cleanup/clarify pulse simulation defaults behaviour and digest

DanPuzzuoli:
This consolidates a few PRs from before the merge to master. 

Note: the doc strings still need work, but it would be convenient to get the other changes merged asap.

### Summary

This PR contains: 
- Clarification of default behaviors in `digest.py`
- Added doc strings for `HamiltonianModel`, `PulseSystemModel`, and `PulseSimulator`
- Fixes for a handful of small bugs

Closes #501 and makes progress on #433 

### Details and comments

Specific changes for `digest`:
- Updated default handling of `qubit_lo_freq` in `digest.py`
   - First it tries to populate it from the `qobj` passed by the user
   - If not present, takes it from `system_model._qubit_freq_est`
   - If this is not present, it automatically computes it from Hamiltonian and uses that
- Updated the old tests for `digest` to test for the above behavior.
- Added a piece of code to `digest` that allows `backend_options` to override any key/value in `qobj_config`.

Changes for doc strings:
- Updated doc strings for `HamiltonianModel` and `PulseSystemModel`
- Added doc string for `PulseSimulator`

Bug fixes:
- Corrected checking of no oscillator systems in HamiltonianModel to accept an empty dictionary
- Fix errors in `duffing_model_generators` when the coupling dict is empty
- Added check in digest to make sure acquire command is set


DanPuzzuoli:
I've added some stuff to the changelog... not sure what level of detail is normal.

Issue #548, opened 2020-01-26
Excessive build warnings for pulse simulator cython files


chriseclectic:
### Informations

- **Qiskit Aer version**: Master
- **Python version**: All
- **Operating system**: Linux and MacOS on Travis

### What is the current behavior?

CI build scripts have become excessively long due to hundreds of warnings being emitted from building the pulse simulator cython files. In some cases this can cause the build to fail due to the log file becoming too large.

### What is the expected behavior?

These warnings should be suppressed from the build logs.

### Suggested solutions

Disable the offending warnings in the CMake build script for `openpulse/cy` and `openpulse/qutip_lite/cy` directories.

chriseclectic:
Close by #553

atilag:
PR #553 has been merged, so this should be fixed by now.

PR #554, opened 2020-01-28
Add compile stage to travis script

chriseclectic:
### Summary

Following on from #553, this PR helps clean up travis build logs and test by adding an extra `compile` stage to the travis tests.
* The compile stage tests building the Python wheels on Linux and MacOS
* The test stage is modified to only run the python unittest scripts so that the full CMake build log isn't returned in the output script.
* Move Python 3.8 MacOS tests to kron job since they are taking too long to run (3-4 x slower than 3.7 tests)

This helps fix issue #548 as we add more build files and more tests files leading to longer and longer logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,9 @@ stage_generic: &stage_generic
     - pip install cython
     # Installing qiskit-terra master branch...
     - pip install https://github.com/Qiskit/qiskit-terra/archive/master.zip
+    # Installing qiskit-ibmq-provider stable branch
+    # This is only needed to suppress warnings when importing Qiskit
+    - pip install https://github.com/Qiskit/qiskit-ibmq-provider/archive/stable.zip
     # Installing qiskit-aer...
     - pip install -U -c constraints.txt -r requirements-dev.txt
 
@@ -51,9 +54,10 @@ stage_linux: &stage_linux
       if [ ! "$(ls -A .stestr)" ]; then
         rm -rf .stestr
       fi
-  script:
+    # Build and install qiskit-aer package
     - python setup.py bdist_wheel -- -DCMAKE_CXX_COMPILER=g++-7 -- -j4
     - pip install dist/qiskit_aer*whl
+  script:
     - stestr run --slowest
 
 stage_osx: &stage_osx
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -112,10 +121,36 @@ jobs:
...
+    - stage: compile
+      name: Test Build MacOS
+      <<: *stage_osx
+      before_script: true
+      script:
+        - python setup.py bdist_wheel -- -- -j4
+    # Power Build
+    - stage: compile
+      name: Test Build Linux (Power)
+      arch: ppc64le
```
atilag: ouch! Does this exist on Travis???? Didn't know... great!

.travis.yml
```diff
@@ -86,10 +94,10 @@ stage_osx: &stage_osx
       if [ ! "$(ls -A .stestr)" ]; then
         rm -rf .stestr
       fi
-
-  script:
-    - python setup.py bdist_wheel -- -- -j4
+    # Build and install qiskit-aer package
+    - python setup.py bdist_wheel -- -DAER_THRUST_BACKEND=OMP -- -j4
```
atilag: did you leave `AER_THRUST_BACKEND` on purpose?

chriseclectic: Typo from gpu branch

.travis.yml
```diff
@@ -31,6 +31,9 @@ stage_generic: &stage_generic
     - pip install cython
     # Installing qiskit-terra master branch...
     - pip install https://github.com/Qiskit/qiskit-terra/archive/master.zip
+    # Installing qiskit-ibmq-provider stable branch
+    # This is only needed to suppress warnings when importing Qiskit
+    - pip install https://github.com/Qiskit/qiskit-ibmq-provider/archive/stable.zip
```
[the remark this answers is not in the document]
chriseclectic: This was just so `import qiskit` didnt print warnings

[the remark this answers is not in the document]
mtreinish: There was an open terra PR to add an env variable to disable this warning. Right now, you can use https://docs.python.org/3/library/warnings.html#warnings.filterwarnings to suppress it from python (as documented at the bottom of: https://qiskit.org/documentation/contributing_to_qiskit.html#installing-terra-from-source). But having an external option to do that  via and env variable would be good for cases like this (and via a user config file for local use too).

PR #561, opened 2020-01-30
Use different muparserx library in linux depending on the ABI

vvilpas:
### Summary

Solves problems related to muparserx when using CENTOS 7, as reported by John Smolin.

### Details and comments
The problem comes from the impossibility to use the new C++11_ABI for gcc compilers running under RHEL6 & RHEL7 ( and every linux distro based on those ones, like centos). Another version of the muparserx version compiled with the old ABI is provided in order to solve this issue.


CMakeLists.txt
```diff
@@ -200,16 +200,25 @@ else()
...
-	set(PLATFORM "win64")
-elseif(APPLE)
-	set(PLATFORM "macos")
-elseif(UNIX)
-	set(PLATFORM "linux")
-endif()
-execute_process(COMMAND ${CMAKE_COMMAND} -E tar "xvfj" "${AER_SIMULATOR_CPP_SRC_DIR}/third-party/${PLATFORM}/lib/muparserx.7z"
-	WORKING_DIRECTORY  "${AER_SIMULATOR_CPP_SRC_DIR}/third-party/${PLATFORM}/lib/")
-set(MUPARSERX_LIB_PATH "${AER_SIMULATOR_CPP_SRC_DIR}/third-party/${PLATFORM}/lib")
+#if(MSVC)
```
atilag: Ok, just remove all these commented lines. We can always go back in git history to see this code.

cmake/compiler_utils.cmake
```diff
@@ -56,4 +56,31 @@ function(get_muparserx_source_code)
                 Please download MuparserX library from https://github.com/beltoforion/muparserx.git \
                 and checkout latest stable release")
     endif()
-endfunction()
\ No newline at end of file
+endfunction()
+
+function(check_compiler_cpp11_abi)
+    # Consider also if -D_GLIBCXX_USE_CXX11_ABI has been passed as flag
```
atilag: Can we comment here why all this is needed? With the link to the original issue (bugzilla) and a short description... Our "old me" and newcomers to the project will thank some more clarification :)

cmake/compiler_utils.cmake
```diff
@@ -56,4 +56,31 @@ function(get_muparserx_source_code)
...
+    string(REGEX MATCH "-D_GLIBCXX_USE_CXX11_ABI=[(A-z)|(a-z)|(0-9)]+" CUSTOM_PREP_FLAGS ${CMAKE_CXX_FLAGS})
+    # Preprocessor run to check if CXX11_ABI is set
+    execute_process(COMMAND echo "#include <string>" COMMAND ${CMAKE_CXX_COMPILER} ${CUSTOM_PREP_FLAGS} -x c++ -E -dM -  COMMAND fgrep _GLIBCXX_USE_CXX11_ABI OUTPUT_VARIABLE C++11_ABI_OUT OUTPUT_STRIP_TRAILING_WHITESPACE)
+    string(REGEX REPLACE "#define _GLIBCXX_USE_CXX11_ABI " "" C++11_ABI "${C++11_ABI_OUT}")
+    set(C++11_ABI ${C++11_ABI} PARENT_SCOPE)
```
atilag: I'm surprised that cmake allows '+' as a variable name...

CMakeLists.txt
vvilpas: Yeah! I forgot!

cmake/compiler_utils.cmake
vvilpas: Totally agree!

cmake/compiler_utils.cmake
vvilpas: Actually, I've seen problems with '+' in names so I'll change them.

PR #574, opened 2020-02-03
Move Muparserx related things to Terra Add-on build

vvilpas:
### Summary

Move Muparserx library related things in CMake to Terra-Addon build.

### Details and comments
Muparserx libraries are not needed by standalone tool. Moved to Terra addon build as requested by Chris

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,19 +203,6 @@ else()
 	set(SPDLOG_LIB "")
 endif()
 
-message(STATUS "Uncompressing muparserx static library...")
-
-uncompress_muparsersx_lib()
-
-find_library(MUPARSERX_LIB NAMES libmuparserx.a muparserx HINTS ${MUPARSERX_LIB_PATH})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -268,11 +255,12 @@ set(AER_LIBRARIES
 	Threads::Threads
 	${SPDLOG_LIB}
 	${DL_LIB}
-	${MUPARSERX_LIB}
 	${THRUST_DEPENDANT_LIBS})
 
 # Cython build is only enabled if building through scikit-build.
 if(SKBUILD) # Terra Addon build
+	add_muparserx_lib()
```
atilag: Can we name it:  `find_muparserx()` so to keep consistency with the `find_` semantics CMake already have.

vvilpas: Not sure about this one, as we are not using the standard `FindLIB.cmake` mechanism. I didn't want anybody to misbelieve that. But I open to change it, though, I don't have a strong position about it.

atilag: Ok, let's leave it like this.

PR #575, opened 2020-02-03
Add compile stage to azure pipelines

mtreinish:
### Summary

In commit 55fa4adb5f7f5b983496819435d67cd8e8bf5116 we added a dedicated
stage and jobs to the travis ci for testing compilation on osx and
linux. This provided value by doing an initial stage of validation to
ensure that we can at least compile aer prior to spending the time to
run the full tests. It also separated the log output for building (which
can be quite verbose) from the test output making it easier to debug.
This commit adds the equivalent stage for windows via azure pipelines
(where we do windows testing) instead of travis. It adds a new ci stage
for compilation and adds equivalent jobs for building aer wheels and
a standalone binary.

### Details and comments


mtreinish:
This still is not working, the standalone compilation does not work yet.

mtreinish:
The sdist builds are failing on `No dl lib found`:

https://dev.azure.com/qiskit-ci/qiskit-aer/_build/results?buildId=7900&view=logs&j=211d4d41-7a4b-554c-46fb-b369c38d3653&t=41e2034a-81f5-5503-d240-be7312df5688&l=362

atilag:
Yep, is getting the MinGW compiler again.

chriseclectic:
@mtreinish Is this PR still active or can we close?

atilag:
I got this

atilag:
I think I finally got it.
Can't tell you exactly why but the combination of: Windows + virtualenv + the pip version of this vritualenv, was conflicting with the location where dependencies were being installed. So CMake was trying to locate Numpy (by running  a simple python script that imports it and gets the include direcotry), but `ModuleNotFound` error was being raised. 
So in order to fix this, I just needed to remove the virutalenv part for this stage.

mtreinish:
I'm not sure that's actually a viable solution. The point of the job is to install from sdist in a clean environment so that none of the build dependencies exist in the python environment. Basically we're trying to emulate what happens if a user were to run `pip install --no-binary qiskit-aer qiskit-aer` in a fresh system.

atilag:
Yes, I did install everything from scratch -- clean conda env, clean virtualenv, pip installing from `https://...master.zip` on my Windows, and everything worked. I couldn't reproduce the problem locally, not quite sure why numpy is not being recognized by the virtualenv python interpreter even though the log clearly says that it's installed.
I'd merge this so we have at least a compile stage working on Windows in the CIs.
The doc already says that building on Windows requires `Anaconda` though.

azure-pipelines.yml
```diff
@@ -65,8 +65,121 @@ stages:
...
+            displayName: Build sdist
+          - bash: |
+              set -x
+              set -e
+              source activate qiskit-aer
+              pip install dist/*tar.gz
+              pip install git+https://github.com/Qiskit/qiskit-terra
+              python tools/verify_wheels.py
+            env:
+              CMAKE_GENERATOR: "Visual Studio 15 2017 Win64"
```
mtreinish, line 152: Is this needed to install from sdist? If so we need to document this for installing from sdist via pip somewhere.

atilag: This is more like a corner-case. It's one of the options for someone on Windows with several toolchains installed on it, and not familiar with MSVC workflow (it usually requires running a batch file that setups some environment variables beforehand, so it becomes the default one).
We don't support other compilers on Windows, but we could certainly add something in the doc.

PR #576, opened 2020-02-03
 * Adding Thrust library as a submodule

atilag:
### Summary
Thrust library usually comes along with CUDA dirvers so in the common
case scenario, where we only want GPU support, there's no need to install
it in order to build. But because we are supporting other Thrust backends
(OMP and TBB) for testing purposes we want to add Thrust headers as a
submodule. We don't want fallbacks or implicit behaviors, so the build
system will gracefully fail if a Thrust backend is specified, but no
Thrust library is found, or the backend is not supported.


mtreinish:
Why are we using submodules? They typically cause way more trouble then they fix. If we want to bundle the headers in the repo as a hard requirement I would suggest that we just vendor a version we need and manually update it as needed. If it's just for testing as the pr summary says we should just make it an explicit build command in the test configuration for the CI system and add it to the local scripts we're using to run tests locally.

atilag:
We only have two libraries as `submodules`, which I agree is kind of cumbersome, but this is why added all the management burden into the build system. The main reason these two libraries are `submodules` is because none of them provide a `header` only version that I could just add to the source tree and maintain it manually (like we do with nlhomann json library).
Anyway, we have started working in integrating Conan package manager so my hope it's to delegate all dependencies to it and avoid having them in the source tree.

CMakeLists.txt
```diff
@@ -234,18 +234,22 @@ if(AER_THRUST_SUPPORTED)
...
 		set(THRUST_DEPENDANT_LIBS ${TBB_LIBRARIES})
-		list(APPEND AER_COMPILER_DEFINITIONS THRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_TTB AER_THRUST_CPU=TRUE)
+		set(AER_COMPILER_DEFINITIONS ${AER_COMPILER_DEFINITIONS} THRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_TTB AER_THRUST_CPU=TRUE)
 	elseif(AER_THRUST_BACKEND STREQUAL "OMP")
 		message(STATUS "Thrust library: Setting OMP backend")
-		list(APPEND AER_COMPILER_DEFINITIONS THRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_OMP AER_THRUST_CPU=TRUE)
+		if(NOT OPENMP_FOUND)
+			message(FATAL_ERROR "There's no OMP support. We cannot set Thrust backend to OMP!!")
+		endif()
+		get_thurst_source_code()
```
vvilpas: Small typo: `get_thurst_source_code` --> `get_thrust_source_code`

cmake/compiler_utils.cmake
```diff
@@ -67,22 +66,42 @@ function(get_muparserx_source_code)
...
+function(get_muparserx_source_code)
+    _get_library_source_code(
+        muparserx
+        https://github.com/beltoforion/muparserx.git
+        v4.0.8
+        muparserx/CMakeLists.txt
+    )
+endfunction()
+
+function(get_thurst_source_code)
```
vvilpas: Small typo: `get_thurst_source_code` --> `get_thrust_source_code`

mtreinish:
If we've isolated the management to the build system I would say then there is no need for submodules in the repo then. Why not just have the build system do a shallow clone of the necessary components at the version then? Adding submodules to the repo will only make this more complicated because you have to deal with whether they've been properly initialized locally or whether they've been checked out and at the right version, etc.

atilag:
It's not worth it changing everything now that it works with submodules, and that conan integration is underway.
Let's just fix what's remaining

PR #577, opened 2020-02-03
Small fix in cmake to try make travis work with sdist

vvilpas:
### Summary

Fix to make travis work with sdist

### Details and comments
Cmake was incorrectly considering the muparserx header dir empty although it was not.


atilag:
I hate CMake, did I say before? Yes, many times... and more to come.

PR #588, opened 2020-02-06
Added test in verify_wheels.py for pulse simulator

### LINKED ISSUE #584: Add pulse simulator test case to verify_wheels.py

The `tools/verify_wheels.py` script: https://github.com/Qiskit/qiskit-aer/blob/master/tools/verify_wheels.py is what we use to validate are built binaries prior to upload to pypi. It just quickly sanity checks each simulator and validates that everything compiled works and is accessible from python. However, it was originally written prior to the introduction of the pulse simulator. To make sure that we are building binaries that include a functional pulse simulator we should add a test case that exercises it to the script. This doesn't need to be exhaustive just something the puts together a small schedule and checks the results to validate it works.

The existing cases in there for the qasm, statevector, and unitary simulators just took the contents from a single unit test and converted them to a standalone test case.

DanPuzzuoli:
### Summary

Closes #584

Added a small test for the pulse simulator in `tools/verify_wheels.py`

### Details and comments

The test makes a single qubit model, constructs a schedule for doing a pi pulse, and verifies that the output state vector is correctly in the |1> state.


mtreinish:
Yep, looking at the output of the linux sdist job it ran successfully: https://travis-ci.com/Qiskit/qiskit-aer/jobs/284381534#L2802

DanPuzzuoli:
Hey sorry, just had to update it: I noticed when about to close the file that I hadn't deleted a method that ended up being unused. Just removed that.

mtreinish:
Also tagging this as a potential backport so we can include this check for any future releases on the 0.4.x branch

PR #591, opened 2020-02-06
Allow setting BLAS to NumPy BLAS (if available) using --with-numpy-blas

### LINKED ISSUE #18: Use same BLAS implimentation as NumPy
Currently, a BLAS implementation is needed Aer to build.  However NumPy, which is a requirement, is already linked to a BLAS library.  Under Anaconda, for example, this is the Intel MKL.  It is possible to use this BLAS for Aer as well, a la Cython and the old `qasm_simulator` that used it.

nonhermitian:
### Summary
Allows for simply doing `python setup.py install --with-numpy-blas`

Fixes #18


vvilpas:
Beware that `np.__config__.blas_opt_info` does not always report the right path to the blas library used.

For example, in a ubuntu distro (docker), `blas_opt_info` reports:

`{'libraries': ['openblas', 'openblas'], 'library_dirs': ['/usr/local/lib'], 'language': 'c', 'define_macros': [('HAVE_CBLAS', None)]}
`

But there are no libraries under `/user/local/lib` in this system.

If I get the path to `numpy` with `np.__path__` and I inspect, for example, the `_umath_linalg.cpython-37m-x86_64-linux-gnu.so` shared library with ldd, I get:

```
ldd lib/python3.7/site-packages/numpy/linalg/_umath_linalg.cpython-37m-x86_64-linux-gnu.so 
	libopenblasp-r0-34a18dc3.3.7.so => lib/python3.7/site-packages/numpy/linalg/../.libs/libopenblasp-r0-34a18dc3.3.7.so (0x00007fc197f92000)
```
which seems to be the real `openblas` library `numpy` is using, under a very different path than that reported by `blas_opt_info`.

atilag:
As the doc says, you can always use the BLAS library at your choice by just typing:
```
python ./setup.py -- -DBLAS_LIB_PATH=/path/to/my/blas
```

nonhermitian:
Yes, this looks to be set correctly only in `conda` or if installed from source.  I believe this can be modified to work correctly if the libs are not in the specified dir.  This is just a convenience function to automate the setting of the lib.  If there is resistance to including this, then I can just keep it on my branch.

atilag:
I just don't want to duplicate things and being consistent, this will open the door to having two different ways of doing basically the same thing, but in the other hand I totally understand the desire of simplifying common operations, with even more reason recently, as we are introducing many different combinations thus complexity. T
This would have broader implications so let's talk about this soon.

nonhermitian:
Closing as tricky

PR #593, opened 2020-02-06
Add terra to the aer requirements list

mtreinish:
### Summary

Installing the qiskit-aer python package always requires terra. This is
primarily because of the package namespacing the qiskit.providers package
doesn't exist outside of terra. But also because the majority of the
python functionality exposed via qiskit-aer depends on terra's api. We
should be enforcing that when you install aer terra is installed (and at
the correct version). Otherwise people could install aer and never get
it working.

### Details and comments

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,3 @@
-# Qiskit dep has been removed because we want to test against master Qiskit Terra branch
 cmake
 scikit-build
 cython
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ except ImportError:
 import setuptools
 
 requirements = [
+    'qiskit-terra>=0.12.0',
     'numpy>=1.13',
     'scipy>=1.0',
     'cython>=0.27.1',


chriseclectic:
I thought there was some issue why we didn't include Terra as a dependency @atilag ?

mtreinish:
Based on the comment in requirements-dev.txt it was for testing against terra master. But as long as you leave the requirement in without an upper cap you can always pull in a newer version from master. All the other elements are doing this. I've also had at least one other person ping me about aer 0.4.0 not working with terra 0.11.0 so getting a proper minimum set where there is that dependency would be good.

atilag:
Yep, the issue was that `requirements-dev.txt` was being used in our development workflow to install all dependencies, and at some point we wanted to use `Terra` master branch so we removed it as a dependency so people don't install `pypi` version.
By putting this dependency in the requirements, don't we fall in the same trap? Because our master branch couldn't work with  Terra `stable`  so users will still see weird errors due to incompatible things.
One thing we can do is show an error message if Terra is not installed?

mtreinish:
I don't understand the problem, all this is doing is telling pip that when you install aer at least terra 0.12.0 needs to be installed, newer versions still work fine. If you're manually installing something newer like we do in ci it still works. The point at which we break something on aer's master so that it needs terra master we can bump the minimum version to express that (although it breaks pip installs on a fresh clone of master).

PR #601, opened 2020-02-07
Switch linux wheel compile job to mirror deploy job

mtreinish:
### Summary

This commit updates the linux wheel compile job to mirror the
configuration we use for building wheels on linux at release time. This
adds 2 additional points of coverage. the first and most importantly is
that python packaging on linux has stringent requitements [1] that
require us to publish wheels in a specific docker image. To make sure
that we don't introduce a change that breaks compilation (since there
have been issues with that in the past) with that limitation we should
build wheels in that environment in CI. The second aspect of additional
coverage here is that it adds compilation of all supported python
versions.

### Details and comments

[1] https://www.python.org/dev/peps/pep-0571/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -139,10 +139,22 @@ jobs:
     # Linux Wheel
     - stage: compile
       name: Python Wheel Build Linux
-      <<: *stage_linux
-      before_script: true
+      language: python
+      python: 3.7
+      os: linux
+      services:
+        - docker
+      env:
+        - CIBW_BEFORE_BUILD="pip install -U Cython pip virtualenv pybind11 && yum install -y openblas-devel"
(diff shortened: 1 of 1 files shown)


mtreinish:
err, it looks like the pulse simulator is failing in this environment (we added coverage to the verify_wheels script in #584 ) https://travis-ci.com/Qiskit/qiskit-aer/jobs/285004587#L654 This isn't a good sign because we used this exact setup for publishing the wheels in 0.4.0

mtreinish:
After confirming this issue locally with the published wheels I've opened #602 once we have a fix for that in place, we definitely should merge this (and backport it to stable/0.4) so that we don't accidentally regress in the future. Otherwise we'll only catch this issue at release time which is when we don't want to have a failed build

PR #603, opened 2020-02-09
Fixed MPS expectation value when qubits are not given in ascending order

merav-aharoni:
### Summary
MPS expectation value is now working correctly for any ordering of the given qubits. 

### Details and comments
The fix includes the following:
1. Fix for Pauli expectation value
2. Fix for matrix expectation value
3. PR509, that includes 2 new tests in unittest
Since this PR includes PR509, the latter can be closed. @yaelbh , can you please review?


test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -632,6 +632,7 @@ def test_snapshot_expval_pauli_pre_measure(self):
         qobj = assemble(circuits, self.SIMULATOR, shots=shots)
         job = self.SIMULATOR.run(qobj, backend_options=self.BACKEND_OPTS)
         result = job.result()
+        print(result)
```
chriseclectic: This shouldn't be here :)

PR #605, opened 2020-02-11
Fix bug in QuantumError compose method

### LINKED ISSUE #580: QuantumError.compose sometimes returns incorrect results
### Informations

- **Qiskit version**: 0.14
- **Python version**: 3.7
- **Operating system**: macOS Catalina 10.15

When building a custom noise_model combining Qiskit Aer depolarization_error and thermal_relaxation_error in simulating final state density matrix after a Rx(pi*0.3) rotation, I obtain real part and imaginary part of density matrix that are opposite to density matrix calculated with Qiskit Ignis Tomography function. I.e.: element rho_real_00 is inverted with rho_real_11.

Calculate density matrix of a Rx(pi*3) with Qiskit Ignis Tomography function. Plot real and imaginary part of density matrix.
Reproduce the same density matrix calculation with Aer imposing depolarization_error and thermal_relaxation_error.

The expected behaviour is that the two calculated density matrices should be close each other.
Another expected behaviour is that by reducing the depolarization error near to zero, the two matrices should be close each other. The two density matrices are close each other only if one of the two depolarization or thermal_relaxation errors are set to zero (or deleted from noise model).

### Suggested solutions
Check error combination logic.

chriseclectic:
Composing certain kinds of quantum errors would fail when combining Kraus operators in the final error object. This rewrite the logic for simplifying Kraus terms

### Summary

Fixes bug in `QuantumError.compose` which would fail to combine certain Kraus error terms correctly.

Fixes #580 

### Details and comments

TODO: 
- [x] Add additional test coverage for bug


chriseclectic:
@atilag This one is ready now

PR #606, opened 2020-02-11
Fix gate fusion bug

### LINKED ISSUE #604: ERROR: Stabilizer::State::invalid instruction 'fusion'.
### Informations

- **Qiskit Aer version**: master
- **Python version**: 3.6
- **Operating system**: Linux

`ERROR: Stabilizer::State::invalid instruction 'fusion'.` for a simple innocent circuit.

```
import qiskit
from qiskit.providers.aer.extensions.snapshot_expectation_value import *
import numpy as np

n = 25
circ = qiskit.QuantumCircuit(n, n)
circ.h(0)
for i in range(n-1):
    circ.cx(0, i+1)
circ.snapshot_expectation_value('hi', [[1, 'Z']], [0])

sim = qiskit.Aer.get_backend('qasm_simulator')
res = qiskit.execute(circ, sim, shots=100).result()

print(res)
```

This works fine for `n<=19` (expectation value is 0). Starting from `n=20`, execution fails with the error that's in the issue's title.

Work always

chriseclectic:
### Summary

Fixes #604 

1. Fixes bugs in gate fusion transpilation pass where the pass was not checking that unitary gates (`OpType::matrix`) where in the allowed instructions of simulator. This was causing simulations on the stabilizer simulator to fail for example when fusion was enabled automatically.

2. Fixes name of fusion instruction to be "unitary" rather than "fusion", and instead uses "fusion" is as the unitary gate label.

3. Adds exception to MPS simulator if it tries to apply a larger than 2-qubit gate. Previously this was silently ignoring the instruction.

### DIFF
--- a/src/framework/operations.hpp
+++ b/src/framework/operations.hpp
@@ -197,7 +197,7 @@ public:
   void insert(const std::vector<Op> &ops);
 
   //-----------------------------------------------------------------------
-  // Check if operations are in all in the OpSet
+  // Check if operations are in the OpSet
   //-----------------------------------------------------------------------
 
   // Return true if an operation is contained in the current OpSet
--- a/src/framework/operations.hpp
+++ b/src/framework/operations.hpp
@@ -197,6 +197,13 @@ public:
   void insert(const std::vector<Op> &ops);
 
   //-----------------------------------------------------------------------
+  // Check if operations are in all in the OpSet
+  //-----------------------------------------------------------------------
+
+  // Return true if an operation is contained in the current OpSet
+  bool contains(const OpType &optype) const;
+
+  //-----------------------------------------------------------------------
   // Validate OpSet against sets of allowed operations
   //-----------------------------------------------------------------------
 
@@ -286,6 +293,12 @@ void OpSet::insert(const OpSet &opset) {
                 opset.gates.end());
   snapshots.insert(opset.snapshots.begin(),
                     opset.snapshots.end());
+}
+
+bool OpSet::contains(const OpType &optype) const {
+  if (optypes.find(optype) == optypes.end())
+    return false;
+  return true;
 }
 
 bool OpSet::validate(const optypeset_t &allowed_ops,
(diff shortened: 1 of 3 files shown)


src/framework/operations.hpp
```diff
@@ -196,6 +196,13 @@ class OpSet {
   // Add additional ops to the opset
   void insert(const std::vector<Op> &ops);
 
+  //-----------------------------------------------------------------------
+  // Check if operations are in all in the OpSet
+  //-----------------------------------------------------------------------
+
```
yaelbh: "in all" is a typo

chriseclectic:
I think we should just disable unitary matrices from MPS simulator. It doesn't actually support the instruction as the instruction is defined as a general N-qubit matrix, and the exception raising I added here for >2Q is basically a hack. Especially since you mentioned in the issue that it is much slower for applying 1 and 2Q unitaries than other gates anyway. We can add it back as a supported instruction if support is added for general N-qubit unitaries.

yaelbh:
@merav-aharoni what do you think about disabling unitaries (including 1Q and 2Q) for MPS?  @yehuda-naveh ?

merav-aharoni:
I think it would be a shame to disable the unitary gates for 1 and 2 qubits, since they do work, and it would mean many circuits would not be able to run on mps. I would rather add the functionality for N qubit unitaries. It shouldn't be too much trouble to implement, as it is a generalization of 3 qubit gates, which are already supported in principle (at least ccx is).

chriseclectic:
Any 1 and 2-qubit unitary can be synthesised into CX and u3 gates by terra compiler, so those circuits would still run if we removed it.

merav-aharoni:
So I guess it is okay to disable support for general unitaries. 
In addition, I agree with @yaelbh that we should disable fusion for MPS, because I believe that would harm performance.

src/transpile/fusion.hpp
```diff
@@ -367,7 +368,7 @@ op_t Fusion::generate_fusion_operation(const std::vector<op_t>& fusioned_ops) co
     U = u_tmp;
   }
 
-  return Operations::make_fusion(sorted_qubits, U, fusioned_ops);
+  return Operations::make_unitary(sorted_qubits, U, std::string("fusion"));
```
atilag: In C++14 we have a neat feature for literal strings:
```c++
 return Operations::make_unitary(sorted_qubits, U, "fusion"s);
```

atilag: Actually, one has to import from a namespace to use it:
```c++
using namespace std::string_literals;
```
... it's ugly.

PR #613, opened 2020-02-13
add comment of gpu options

hhorii:
### Summary
Add recent changes of GPU into qasmsimulator's docstring

### Details and comments
Added gpu's `method`s into the header.

### DIFF
--- a/qiskit/providers/aer/backends/qasm_simulator.py
+++ b/qiskit/providers/aer/backends/qasm_simulator.py
@@ -28,7 +28,7 @@ logger = logging.getLogger(__name__)
 
 class QasmSimulator(AerBackend):
     """
-    Noisy quantum circuit simulator backend
+    Noisy quantum circuit simulator backend.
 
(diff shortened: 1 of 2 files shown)


qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -192,6 +203,17 @@ class QasmSimulator(AerBackend):
       OpenMP parallelization. If parallel circuit or shot execution
       is enabled this will only use unallocated CPU cores up to
       max_parallel_threads (Default: 100).
+
```
atilag: There seems to be something wrong in here as linter is complaining.

atilag:
I don't know why this is not passing the Doc tests, because your changes are not affecting the line that has the error: qasm_simulator.py:32
Can you put a point in the end of line qasm_simulator.py:32?
```
Noisy quantum circuit simulator backend.<---
```
Thanks!!

PR #626, opened 2020-02-21
[WIP] Add travis compile jobs for building wheels on non-x86

mtreinish:
### Summary

This commit adds test jobs using cibuildwheel to test the ability of aer
to build and compile manylinux2014 wheels on non-x86 architectures. This
will mirror how we build the wheels at release time for publishing.

### Details and comments

Right now this is using the pending branch from a PR adding support to
cibuildwheel for building on non-x86 architectures, but when that is
merged and released this will need to be updated to use that instead.


mtreinish:
I don't think this will be feasible to do yet, pretty much every package has to be built from source since there aren't any wheels available on the non-x86 architecture. It spends too much time building dependencies, the fastest compiling one s390x failed after 33 minutes trying to compile scipy from source and never got to building aer.

chriseclectic:
I would just worry about testing on ppc. We should probably just use conda for installing dependencies like scipy/numpy though to avoid building every py package.

mtreinish:
Well this was mostly a test for building portable manylinux2014 wheels for pypi using the same build system we're using for the linux x86_64 wheels. I pushed a PR to add support for non-x86 to the tool we're using to do that in CI and wanted to see what would be required to use that for aer. 

The problem with the environment for building the manylinux wheels is that it doesn't have conda installed (https://github.com/pypa/manylinux#manylinux2014) It's a centos 7 docker image with python and system libraries setup in the proper way to build wheels compliant with the packaging spec. So using conda to build things for pypi might not be a viable option either.

chriseclectic:
Fair enough. But even if we can't get the build stage working with manylinux yet, it's still important that our CI test that Aer can be built from source and pass tests on PPC. If those tests require conda I think thats fine, and in that case we just don't add PPC to auto deployed wheels.

PR #634, opened 2020-02-24
MPS: Support multi-qubit apply matrix

merav-aharoni:
### Summary
Enabled applying unitary matrices on more than 2 qubits.

### Details and comments
This PR uses developments from PR #603, so must be merge after merging #603.


src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -520,14 +520,81 @@ void MPS::apply_matrix(const reg_t & qubits, const cmatrix_t &mat)
...
+  reg_t reversed_qubits = qubits;
+  std::reverse(reversed_qubits.begin(), reversed_qubits.end()); 
+
+  bool ordered = true;
+  for (uint_t index=0; index < qubits.size()-1; index++) {
+    if (reversed_qubits[index]+1 != reversed_qubits[index+1]){
+      ordered = false;
+      break;
+    }
+  }
```
atilag: just wondering, won't this be sufficient to know whether `reversed_qubits` are ordered or no?
```c++
bool ordered = num_qubits > 1 ? reversed_qubits[0]+1 == reversed_qubits[1] : true;
```

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -520,14 +520,81 @@ void MPS::apply_matrix(const reg_t & qubits, const cmatrix_t &mat)
...
+  if (ordered) {
+    target_qubits = reversed_qubits;
+  } else {
+    move_qubits_to_right_end(reversed_qubits, target_qubits, actual_indices);
+  }
+
+  uint_t first = target_qubits.front();
+  MPS_Tensor sub_tensor(state_vec_as_MPS(first, first+num_qubits-1));
+
+  sub_tensor.apply_matrix(mat, false /* swapped */);
```
atilag: The fact that `apply_matrix` needs to swap the contained matrix before computing the tensor, by passing a `true/false` value is kind of confusing. In order to preserve `Single Responsibility` principle I would definitely create a `swap` method in `MPS_Tensor` and express this like:
```c++
sub_tensor.swap();
sub_tensor.apply_matrix(mat);
```

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -520,14 +520,81 @@ void MPS::apply_matrix(const reg_t & qubits, const cmatrix_t &mat)
...
+
+
+  // state_mat is a matrix containing the flattened representation of the sub-tensor 
+  // into a single matrix. Note that sub_tensor will contain 8 matrices for 3-qubit
+  // gates. state_mat will be the concatenation of them all.
+  cmatrix_t state_mat = sub_tensor.get_data(0);
+  for (uint_t i=1; i<sub_tensor.get_data().size(); i++)
```
atilag: We are skipping the first element on purpose, right?

atilag: Is the sum of all `sub_tensor` matrices considered big? Then maybe parallelizing this loop will increase performance:
```c++
#pragma omp for
for(int_t i=1; i < sub_tensor.get_data().size(); ++i) {
...
}
```

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -520,14 +520,81 @@ void MPS::apply_matrix(const reg_t & qubits, const cmatrix_t &mat)
...
+    for (uint_t i=1; i<num_qubits-1; i++) {
+      lambda_reg_[first+i] = sub_MPS.lambda_reg_[i]; 
+    }
+    if (first+num_qubits-1 < num_qubits_-1)
+	q_reg_[first+num_qubits-1].div_Gamma_by_right_Lambda(lambda_reg_[first+num_qubits-1]);
+      
+  }
+  // need to move qubits back to original position, if they were moved
+  // at the beginning
+  if (!ordered) {
```
atilag: Just wondering if we can avoid all this moving qubits back and forth somehow.
Maybe having two functions like: `void MPS::apply_multi_qubit_gate()` and `void MPS::apply_ordered_multi_qubit_gate()` ?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -643,7 +708,28 @@ void MPS::move_qubits_to_right_end(const reg_t &qubits,
...
 
- void MPS::change_position(uint_t src, uint_t dst) {
+void MPS::move_qubits_back_from_right_end(const reg_t &qubits, reg_t &actual_indices) {
+  for (uint_t left_index=num_qubits_-qubits.size();  left_index< qubits.size(); left_index++) {
+    // find the qubit with the smallest index
+    //uint_t min_index = actual_indices[left_index];
+    uint_t min_index = left_index;
+    for (uint_t i = left_index+1; i < actual_indices.size(); i++) {
+      if (actual_indices[i] < actual_indices[min_index])
+	min_index = i;
```
atilag: nit: Indent this.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: 1. it is okay to skip the first element.
2. The size will depend on the size of the matrix applied. I don't expect it to be large, so not adding parallelization.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: No,  it is not sufficient, because I want to make sure **all qubits** are in order, not just the first 2.

Issue #619, opened 2020-02-18
MPS sudden performance degradation on measure


merav-aharoni:
### Informations

- **Qiskit Aer version**: 0.14
- **Python version**:
- **Operating system**: Linux

### What is the current behavior?
Since Qiskit version 0.14, measurement in MPS is very slow.

### Steps to reproduce the problem
```
import qiskit
n = 25 
circ = qiskit.QuantumCircuit(n, n)
circ.measure(range(n), range(n))
sim = qiskit.Aer.get_backend('qasm_simulator')
res = qiskit.execute(circ, sim, shots=100, backend_options={'method': "statevector"}).result()
print("qasm time " + str(res.time_taken))
res = qiskit.execute(circ, sim, shots=100, backend_options={"method": "matrix_product_state", "fusion_enable":False }).result()
print("mps time " + str(res.time_taken)) 
```
Results:
```
qasm time 0.3851964473724365
mps time 32.857303857803345
```
The run-time depends on the number of shots.  When I run 1000 shots, time goes up to 300 sec!

### What is the expected behavior?
Up to Qiskit 0.13.0, this circuit ran in less than 1 second.

### Suggested solutions
Did anything in Aer change that could cause this problem? 
Note that I disabled fusion, to make sure that is not causing the problem.

merav-aharoni:
This appears to be a bug in parallelization. When I remove the `#pragma omp parallel for collapse(2)` in the function `MPS_Tensor::contract_2_dimensions`, the performance on this tests goes down from 18 sec to 0.5 sec. @atilag , can you check what changed in the parallelization? adding also @yaelbh to this thread.

chriseclectic:
@merav-aharoni I would recommend adding OMP conditions like we use in the statevector and density matrix simulators.


Eg the QubitVector class has the following on all its pragmas:

```
#pragma omp parallel for if (num_qubits_ > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
```

The omp_threshold_ parameter can be set using config options, and a default value should be set based on your profiling of when it is beneficial to use multithreading (at the moment use 14 qubits by default for qubitvector/statevector_state).

chriseclectic:
Closed by #638

PR #638, opened 2020-02-26
MPS: Added thresholds to loops that use parallelization

merav-aharoni:
### Summary
Parallelization on for-loops without a threshold caused a severe deterioration of performance in MPS. I added reasonable thresholds that bring the performance back to the previous values. Determining the optimal values should be done in future work, but for now this is good enough.

### Details and comments
It would be great to include this in the current release because it dramatically affects performance of MPS. This closes issue #619.


src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -199,17 +203,14 @@ std::vector<uint_t> calc_new_indices(const reg_t &indices) {
...
   if (lambda == rvector_t {1.0}) return mat;
   cmatrix_t res_mat(mat);
   uint_t num_rows = mat.GetRows(), num_cols = mat.GetColumns();
 
-  #ifdef _WIN32
-     #pragma omp parallel for
-  #else
-     #pragma omp parallel for collapse(2)
-  #endif
+#pragma omp parallel for collapse(2) if (num_rows*num_cols > 64 && MPS::get_omp_threads() > 1) num_threads(MPS::get_omp_threads())
```
atilag: Ok,  you have removed the `#ifdef _WIN32` conditinal define, but this is needed unfortunately.
MSVC compiler doesn't support the  same version of OpenMP that the rest of supported compilers, so we still need it.
I'd suggest to rewrite this like:
```c++
#ifdef _WIN32
   #pragma omp parallel for if (num_rows*num_cols > 64 && MPS::get_omp_threads() > 1) num_threads(MPS::get_omp_threads()) 
#else
   #pragma omp parallel for collapse(2) if (num_rows*num_cols > 64 && MPS::get_omp_threads() > 1) num_threads(MPS::get_omp_threads()) 
#endif
```

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -666,11 +667,8 @@ cmatrix_t MPS::density_matrix(const reg_t &qubits) const
   MPS_Tensor psi = temp_MPS.state_vec_as_MPS(new_qubits.front(), new_qubits.back());
   uint_t size = psi.get_dim();
   cmatrix_t rho(size,size);
-  #ifdef _WIN32
-     #pragma omp parallel for
-  #else
-     #pragma omp parallel for collapse(2)
-  #endif
+
+#pragma omp parallel for collapse(2) if (size > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
```
atilag: Rewrite this as above

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -199,17 +203,14 @@ std::vector<uint_t> calc_new_indices(const reg_t &indices) {
 }
 
 cmatrix_t mul_matrix_by_lambda(const cmatrix_t &mat,
-			       const rvector_t &lambda)
+			       const rvector_t &lambda,
+			       uint_t omp_threads)
```
atilag: I think we are not using this variable anywhere, right?
Furthermore, shouldn't be a parameter of this function. Using concurrency in a multiplication operation should be an internal implementation detail and not exposed to the caller (through a parameter). The function itself, in it's implementation should have all the needed data to decide whether it can run concurrently.

src/simulators/matrix_product_state/matrix_product_state_internal.hpp
```diff
@@ -193,27 +193,39 @@ class MPS{
...
-  void set_sample_measure_index_size(int index_size){
+  static void set_sample_measure_index_size(uint_t index_size){
     sample_measure_index_size_ = index_size;
   }
+  static void set_enable_gate_opt(bool enable_gate_opt) {
+    enable_gate_opt_ = enable_gate_opt;
+  }
+
+  static uint_t get_omp_threads() {
+  return omp_threads_;
```
atilag: nit: indent

src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
```diff
@@ -442,6 +444,7 @@ MPS_Tensor MPS_Tensor::contract(const MPS_Tensor &left_gamma,
 //---------------------------------------------------------------
 void MPS_Tensor::contract_2_dimensions(const MPS_Tensor &left_gamma, 
 			               const MPS_Tensor &right_gamma,
+				       uint_t omp_threads,
```
atilag: Can we remove this parameter and use: `MPS::get_omp_threads()` instead?

src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
```diff
@@ -459,30 +462,23 @@ void MPS_Tensor::contract_2_dimensions(const MPS_Tensor &left_gamma,
...
-
-      for (int_t size=0; size<left_size; size++)
-	  for (int_t index=0; index<left_columns ; index++) {
- 	      result(l_row, r_col) += left_gamma.data_[size](l_row, index) *
-		                      right_gamma.data_[size](index, r_col);      
-
-	    }
-	  }
+  uint_t omp_limit = left_rows*right_columns;
+#pragma omp parallel for collapse(2) if ((omp_limit > 10) && (omp_threads > 1)) num_threads(omp_threads)
```
atilag: So instead of using `omp_threads` it would be great if we can use: `MPS::get_num_threads()` and avoid passing `omp_threads` as a parameter.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: done

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: done

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: you are correct. done.

src/simulators/matrix_product_state/matrix_product_state_internal.hpp
merav-aharoni: done

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -270,7 +270,7 @@ class State : public Base::State<matrixproductstate_t> {
   //-----------------------------------------------------------------------
   // Config Settings
   //-----------------------------------------------------------------------
-
+  /*
```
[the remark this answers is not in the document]
merav-aharoni: correct. done.

src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
merav-aharoni: I don't see how I can do this, because this file does not recognize the MPS class. I tried adding `class MPS` at the beginning, but it still doesn't work. Any suggestions?

atilag: Ok, leave it like this

PR #640, opened 2020-02-26
Fix QuantumError sometimes casting probabilities to complex

ereastman:
### Summary

[This PR](https://github.com/Qiskit/qiskit-terra/pull/3890) is currently blocked by bugs in Aer. This PR will hopefully fix some of those bugs. 

### Details and comments

Currently, Noise Models are being received on the C++ side with probability attributes having (what appears to be) complex types - ~~we have yet to identify why~~ but the existing changes will at least raise a reasonable exception when it does happen. 

Bug was caused by a missing cast from complex type to float type in a QuantumError factory. Thanks @chriseclectic

### DIFF
--- a/test/terra/noise/test_noise_model.py
+++ b/test/terra/noise/test_noise_model.py
@@ -24,7 +24,6 @@ from qiskit.providers.aer.noise.errors.standard_errors import pauli_error
 from qiskit.providers.aer.noise.errors.standard_errors import reset_error
 from qiskit.providers.aer.noise.errors.standard_errors import amplitude_damping_error
 from qiskit.test import mock
-#from qiskit.providers import aer
 
 
 class TestNoise(common.QiskitAerTestCase):
--- a/test/terra/noise/test_noise_model.py
+++ b/test/terra/noise/test_noise_model.py
@@ -16,13 +16,15 @@ NoiseModel class integration tests
 
 import unittest
 from test.terra import common
-from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit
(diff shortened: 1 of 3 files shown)


mtreinish:
Thanks for starting this, the other thing which would probably be good here is to add a test case that runs something like:

```
from qiskit import execute
from qiskit.test import mock
from qiskit.providers import aer

circ = QuantumCircuit(2)
circ.x(0)
circ.x(1)
circ.measure_all()

backend = mock.FakeSingapore()
noise_model = NoiseModel.from_backend(backend)
qobj = assemble(transpile(circ, backend), backend)
sim = aer.QasmSimulator()
sim.run(qobj, noise_model=noise_model).result()
```

to ensure we don't regress here moving forward.

test/terra/noise/test_noise_model.py
```diff
@@ -16,13 +16,15 @@
...
-from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit
+from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit, execute
 from qiskit.compiler import assemble, transpile
 from qiskit.providers.aer.backends import QasmSimulator
 from qiskit.providers.aer.noise import NoiseModel
 from qiskit.providers.aer.noise.errors.standard_errors import pauli_error
 from qiskit.providers.aer.noise.errors.standard_errors import reset_error
 from qiskit.providers.aer.noise.errors.standard_errors import amplitude_damping_error
+from qiskit.test import mock
+#from qiskit.providers import aer
```
mtreinish: Probably don't need this line anymore

ereastman: resolved in 24acc21

PR #641, opened 2020-02-27
MPS: Improvements to apply_matrix 

merav-aharoni:
### Summary
This PR is a followup to the comments that @atilag  suggested for PR #634.

### Details and comments
1. The main change is that I restructured apply_multi_qubit_gate. I split it into the `ordered` case and the `unordered` case, so it is more modular now.
2. I made a small optimization in move_qubits_to_right_end, so that the `qubits` (parameter) are moved only to the position of the largest qubit in `qubits` and not all the way to the right end.
3. @atilag suggested I remove the `swapped` parameter from apply_matrix. I prefer to leave it, because it is an optimization for the 2-qubit case. I did however remove it from the invocation in apply_matrix_to_target_qubits.


src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -677,53 +689,63 @@ void MPS::move_qubits_to_original_location(uint_t first, const reg_t &original_q
...
     // find "largest" element and move it to the right end
-    uint_t biggest = qubits[right_index];
-    num_moved++;
+    uint_t next_right = qubits[right_index];
     for (uint_t i=0; i<actual_indices.size(); i++) {
-      if (actual_indices[i] == biggest) {
-	for (uint_t j=i; j<actual_indices.size()-num_moved; j++) {
-	  //swap the qubits until biggest reaches right end
+      if (actual_indices[i] == next_right) {
+	for (uint_t j=i; j<right_end-num_moved; j++) {
```
atilag: nit: Can you indent this for?

PR #642, opened 2020-02-27
Start using reno for release documentation

mtreinish:
### Summary

This starts the process of using reno for handling release documentation
for qiskit-aer. Moving forward the expectation is that we'll require a
PR to include release documentation in the form of a reno release note
before we merge it (if it contains a user facing change). Release notes
differ from changelog in that they provide much more detail to end users
on what has changed, how they can adapt to it, and why it was changed (if
necessary). Accordingly the release notes are free form restructured text
that enables writing as much as needed and using any sphinx directive
since it will be compiled along with the documentation.

The changelog file is removed here because for future releases the
changelog will automatically be generated at release time and added
to the github release page using qiskit-bot. We will be required to
appropriately tag our PRs prior to a release to ensure the changelog
is generated correctly.

### Details and comments

You can look at qiskit-terra's release notes as an example of how this
should be used.


chriseclectic:
Lets wait till we put the 0.4.1 patch then merge this for the following release

chriseclectic:
@mtreinish can you update the contributing guide for how to use reno

mtreinish:
Ok, I've updated the contributing guide to include the new changelog and release note procedure in: https://github.com/Qiskit/qiskit-aer/pull/642/commits/c943b7e47780d2a2fdced07f33429b92e4de4d22
At the same time I cleaned up duplicate information that is already in the Qiskit-wide contributing guide.

releasenotes/notes/fix-quantumerror-compose-7c54090a0cad0797.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed bug in the :meth:`qiskit.providers.aer.noise.QuantumError.compose`
```
chriseclectic: I don't think this should be in this PR, it's already in stable.

chriseclectic:
@atilag can you review the update to contributing.

@mtreinish Do we need to port the old changelog to a reno so that it shows up? This seems to just delete changelog so all that information will be lost.

mtreinish:
The information for past release is on their release page for each release, for example: https://github.com/Qiskit/qiskit-aer/releases/tag/0.4.0 so it's not lost (it's also in https://qiskit.org/documentation/release_notes.html). Moving forward qiskit-bot will generate that changelog in the release page automatically at release time using tags. I deleted the changelog so there was no confusion about where to update things since there is nothing that file is used for anymore.

releasenotes/notes/fix-quantumerror-compose-7c54090a0cad0797.yaml
mtreinish: The fix will also apply to the 0.5.0 release too. Normally reno is smart enough to figure out the duplicate (because it looks for the filename in the git log) but in this case it will show up twice because 0.4.1 doesn't use reno. That being said we we need a file in the releasnotes/notes dir so git will track it. We can just remove this at release time if you don't want it in the 0.5.0 release notes

chriseclectic: But this makes it sound like there is an extra bug in compose that wasn't fixed last time.

If you need to have a note there can you just add a release note that says we have switched to using reno for release notes?

CONTRIBUTING.md
```diff
@@ -545,194 +740,3 @@ Now we can run the python interpreter and pass the arguments (the python file to
     Target 0: (python) stopped.
 
 After this, you can step through the code and continue with your debug session as always.
-
-
-## Style guide
```
atilag: Can you restore this info? 
- We will eventually (sooner than later) have a style guide.
- Also, I think it's important for contributors to take into account that we still need to be compatible with all supported platforms (as there's a lot of native code).
- Informing about the Good First Contributions program is interesting for newcomers.
- The guide on how to create PRs is useful as well.
- It's important to show how a good commit message should look like.

mtreinish: See the first section of the updated contributing guide. It says read the qiskit-wide contributing guide: https://qiskit.org/documentation/contributing_to_qiskit.html which contains all this information.

The only thing missing there from your list is the cross platform compatibility which I will add back.

releasenotes/notes/fix-quantumerror-compose-7c54090a0cad0797.yaml
mtreinish: ok changed to a release note about release notes in: https://github.com/Qiskit/qiskit-aer/pull/642/commits/db12c394b7546643319b6d40fd8aedb2948d873b

CONTRIBUTING.md
mtreinish: Done in: https://github.com/Qiskit/qiskit-aer/pull/642/commits/bc56e1ca0df2915196b37576686b4079e4912e93

atilag: Thanks Matt!

PR #648, opened 2020-03-04
Add Conan package manager to the buildsystem

### LINKED ISSUE #325: Use CONAN package manager under the hood, so we can forget about dependencies
If we integrate Conan within the build system, we can get rid of maintaining some of our dependencies (like OpenBLAS for Windows), and upgrade the compiler transparently so we can get rid of the pre-requisites (or make them way simpler).
This could have another great advantage, as we will not relay on the compiler preinstalled in the O.S., we could bump up C++ minimum standard to something like C++17 or even better, C++20.
These are the requirements we need to address:
* Conan wil be used to download our external dependencies for all supported OS (Windows, MacOS, Linux).
* Conan will be used to download the toolchain that will be used to build de sources.
* Conan needs to be called from within the `setup.py` script, so we don't have to change the current build workflow.

vvilpas:
### Summary
Introduce Conan as a package manager for C++ 

### Details and comments
Conan is a package manager for C++ (https://conan.io/). The idea is to use Conan to manage all external dependencies so that a user compiling doesn't have to worry about having the right library installed in their system. 
The commands to build will be the same so no modifications to the build process (besides having conan installed).

Right now, depending on options at build time, we can use these packages:

  **nlohmann_json
  spdlog
  openblas
  tbb
  muparserx
  thrust
  omp (Apple)**

The first four are already available as packages in conan repositories and can be used right now. The last three, on the other hand, has to be uploaded to conan-center and approved by conan to be used, so right now the build **WILL FAIL**. I'll initiate that process as soon as possible. 

There are some other dependencies that cannot be set up through Conan, like CUDA that has to be installed beforehand or other system libraries like threads and dl.

Things to do:

- [x] Upload packages to Conan-center.
- [x] Check ABI compatibility issues in RHEL 6&7

In the future we will check the possibility of including the toolchain as a dependency.

Fixes #325


vvilpas:
I have created the pull-requests to include muparserx and llvm-openmp in [conan-center](https://github.com/conan-io/conan-center-index):
- Muparserx: https://github.com/conan-io/conan-center-index/pull/1037
- llvm-openmp: https://github.com/conan-io/conan-center-index/pull/1075

I am still waiting to be included in the early access program. That would trigger the CI on those pull requests and allow to do the final merge. 
I'll do the pull-request for the Thrust recipe once the llvm-openmp one is approved and merged,  as it will be needed for testing the recipe. It is already finished and uploaded to github (https://github.com/vvilpas/conan-center-index/tree/thrust/recipes/thrust)

vvilpas:
After some weeks I've been accepted in the Early Access Program and both **Muparserx** and **llvm-openmp** have passed the CI builds. So, right now we are waiting them to do the merge.
As soon as that happens I'll do the pull request for **Thrust** and once it is approved and merged we can go on with this pull request.

vvilpas:
**Muparserx** is already a recipe in Conan!
**llvm-openmp** is already approved and has a green flag in CI, but still waiting for someone to merge.

vvilpas:
I have already done the pull-request to merge the recipe for **Thrust**. It is under review.

Regarding **llvm-openmp**,  although the PR was approved and ready to be merged, **Conan** people were concerned about introducing a package belonging to a framework (*llvm* in this case). They are deciding wether they merge this PR or they use another approach. They'll tell us something soon.

vvilpas:
This is now ready for review. Every recipe is in conan-center now. I have also added **catch2** dependency for C++ testing.

src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
```diff
@@ -18,6 +18,7 @@
 
 #include <cstdio>
 #include <iostream>
+#include <iomanip>
```
atilag, line 21: Why this? :)

vvilpas: The standalone doesn't compile now without this. Before, I guess, it ended up getting iomanip through some dependency

PR #650, opened 2020-03-06
Adding preprocessors to disable simulation methods for development

doichanj:
### Summary
I added some preprocessors to disable linking some of the simulation method to shorten compilation time for development and debugging of specific simulator.

### Details and comments
I added these preprocessors to disable simulation methods.

AER_DISABLE_DENSITY_MATRIX
AER_DISABLE_STABILIZER
AER_DISABLE_EXTENDED_STABILIZER
AER_DISABLE_MPS
AER_DISABLE_STATEVECTOR
AER_DISABLE_SUPEROPERATOR

I also added 
AER_DISABLE_SINGLE_PRECISION
to disable building single precision version of simulators 
(This will half the compilation time!)


vvilpas:
Do we really have that long compilation times? Aren't we adding too much noise to the code?

atilag:
Hi Doi, Victor is right. We have to think about another approach to tackle the building times for GPU. This is usually better achieved through configuration, so we'd want to refactor something so that we can build just a specific simulator. But I'd encourage not using preprocesors flags.

PR #651, opened 2020-03-06
Added postfix for linux on Raspberry Pi

mikaelhaglund:
armv7l is CMAKE_SYSTEM_PROCESSOR for at least the Raspberry Pi 3A and 4 that I have at hand to verify with

### Summary
Linux on Raspberry Pi should have its own entry i the CMake files for the pre-compiled library libmuparserx.a 

### Details and comments
Linux on Raspberry Pi should have its own entry. Today it's handled as Linux on x86, and the wrong libmuparserx.a is extracted. I tried installing the muparserx as a compiled package (that the CMAKE file is supposed to find - which it did), but there were linking errors using this approach too, so I suggest a scheme similar to that used with ppc64


mikaelhaglund:
There should of course also be a pre-compiled library (muparserx.armv7l.7z) added to qiskit-aer/src/third-party/linux/lib/

I have it ready, but am not sure where to put it to make this request complete. I can't upload it to qiskit-aer/src/third-party/linux/lib/

vvilpas:
Thanks for your contribution! For your info, right now we are migrating our dependencies management system to the Conan Package Manager (https://conan.io/). It will be available soon and I think it will make life easier in cases like this one. When that happens, we will no longer distribute the muparserx library.

atilag:
I'm merging this while we are transitioning to Conan.
Thanks @mikaelhaglund !

PR #653, opened 2020-03-09
Update Benchmark framework

atilag:
### Summary
There have been a lot of improvements in the build system, so I'm updating the benchmark frameworks for detecting performance regressions. 

### Details and comments
I'm removing static compilation, not forcing the platform name for MacOS, and using `-j` so the `Make` subsystem choose the better parallel strategy depending on the underlying architecture.

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -71,7 +71,7 @@ class AerBackend(BaseBackend):
         not available.
 
         Args:
-            controller (function): Aer cython controller to be executed
+            controller (function): Aer controller to be executed
             configuration (BackendConfiguration): backend configuration
             provider (BaseProvider): provider responsible for this backend
 
--- a/test/asv.linux.conf.json
+++ b/test/asv.linux.conf.json
@@ -43,7 +43,7 @@
     ],
     "build_command": [
         "python -mpip install -U scikit-build cython",
-        "python setup.py bdist_wheel --dist-dir={build_cache_dir} -- -DCMAKE_CXX_COMPILER=g++ -DSTATIC_LINKING=True -- -j8",
+        "python setup.py bdist_wheel --dist-dir={build_cache_dir} -- -DCMAKE_CXX_COMPILER=g++ -- -j",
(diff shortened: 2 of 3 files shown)


test/asv.linux.conf.json
```diff
@@ -43,7 +43,7 @@
     ],
     "build_command": [
         "python -mpip install -U scikit-build cython",
-        "python setup.py bdist_wheel --dist-dir={build_cache_dir} -- -DCMAKE_CXX_COMPILER=g++ -DSTATIC_LINKING=True -- -j8",
+        "python setup.py bdist_wheel --dist-dir={build_cache_dir} -- -DCMAKE_CXX_COMPILER=g++ -- -j",
         "python -mpip install https://github.com/QISKit/qiskit-terra/archive/master.zip"
```
mtreinish: For consistent benchmarking we should be using a fixed version of terra otherwise how do we know whether a performance improvement is caused by a recent change to terra or to aer?

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, controller, configuration, provider=None):
         not available.
 
         Args:
-            controller (function): Aer cython controller to be executed
+            controller (function): Aer controller to be executed
```
mtreinish: This doesn't seem related to the asv changes

test/asv.linux.conf.json
atilag: As we don't benchmark Terra compilation process, the only problem I could foresee here is Terra having an issue with the no-optimization parameter, or unrolling, and changing the final resulting circuit in an undesirable way.
Anyway, I don't really mind to stick to latest release, as long as Aer `master` branch is compatible with that release, something that we struggled with in the past.

qiskit/providers/aer/backends/aerbackend.py
atilag: Yep, that was an improvement I just did for free xD

test/asv.linux.conf.json
atilag: The new Qiskit release process seems to be working quite well, so it's true that we haven't had incompatibilities with Terra releases for a long time since then.

PR #656, opened 2020-03-11
Handle numpy array inputs from qobj

mtreinish:
### Summary

Starting in Terra 0.13.0 the qobj passed to aer will not undergo type
coercion to be something python's `json.dumps()` natively understands.
This means that input qobj objects will have numpy arrays of complex
numbers instead of `list(list(re, im))`. This should be more efficient for
aer since it doesn't have to undergo a conversion step. However, we were
not handling this input type correctly in the pulse simulator and it
would lead to a failure trying to convert the list format to a numpy
array. This commit fixes this issue by checking if we have a numpy array
in the qobj and just using that directly instead of trying to convert
it.

Additionally, for the snapshot expectation value instruction has to hack around
a terra bug that doesn't allow `list` as a valid type for parameters on an instruction,
by casting it as a numpy array. However, since now that terra isn't converting numpy
arrays to  lists automatically this breaks when we receive the qobj back from terra
with a snapshot expectation value instruction in it. To address this case, this adds
a custom `assemble()` method for the instruction which will build the qobj with a list
instead of a numpy array.

### Details and comments

This is really 2 separate commits/PRs combined together. This is necessary to unblock CI, since the recent changes to terra are breaking build for 2 distinct (but related reasons)

### DIFF
--- a/qiskit/providers/aer/extensions/snapshot_expectation_value.py
+++ b/qiskit/providers/aer/extensions/snapshot_expectation_value.py
@@ -18,7 +18,7 @@ import numpy
 from qiskit import QuantumCircuit
 from qiskit.circuit import Instruction
 from qiskit.extensions.exceptions import ExtensionError
-from qiskit.qobj.qasm_qobj import QasmQobjInstruction
+from qiskit.qobj import QasmQobjInstruction
 from qiskit.quantum_info.operators import Pauli, Operator
 from qiskit.providers.aer.extensions import Snapshot
 
--- a/qiskit/providers/aer/extensions/snapshot_expectation_value.py
+++ b/qiskit/providers/aer/extensions/snapshot_expectation_value.py
@@ -18,6 +18,7 @@ import numpy
 from qiskit import QuantumCircuit
 from qiskit.circuit import Instruction
 from qiskit.extensions.exceptions import ExtensionError
+from qiskit.qobj.qasm_qobj import QasmQobjInstruction
 from qiskit.quantum_info.operators import Pauli, Operator
 from qiskit.providers.aer.extensions import Snapshot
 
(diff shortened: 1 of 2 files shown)


qiskit/providers/aer/extensions/snapshot_expectation_value.py
```diff
@@ -18,6 +18,7 @@
 from qiskit import QuantumCircuit
 from qiskit.circuit import Instruction
 from qiskit.extensions.exceptions import ExtensionError
+from qiskit.qobj.qasm_qobj import QasmQobjInstruction
```
mtreinish: Want to wait for ci to finish, but this should be changed to:

```suggestion
from qiskit.qobj import QasmQobjInstruction
```
so that this will work for 0.12.0 and 0.13.0

mtreinish:
I have no idea why travis is listed as canceled, it ran and passed the tests here: https://travis-ci.com/github/Qiskit/qiskit-aer/builds/152851079

woodsp-ibm:
Was this anticipated to fix the failure of Aqua unit tests with Terra & Aer from master. We still have the failure even after this:
```
      File "/home/travis/build/Qiskit/qiskit-aqua/qiskit/aqua/utils/run_circuits.py", line 291, in run_qobj
        results.append(job.result(**qjob_config))
      File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/qiskit/providers/aer/aerjob.py", line 41, in _wrapper
        return func(self, *args, **kwargs)
      File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/qiskit/providers/aer/aerjob.py", line 94, in result
        return self._future.result(timeout=timeout)
      File "/opt/python/3.7.1/lib/python3.7/concurrent/futures/_base.py", line 432, in result
        return self.__get_result()
      File "/opt/python/3.7.1/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
        raise self._exception
      File "/opt/python/3.7.1/lib/python3.7/concurrent/futures/thread.py", line 57, in run
        result = self.fn(*self.args, **self.kwargs)
      File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/qiskit/providers/aer/backends/aerbackend.py", line 133, in _run_job
        output = self._controller(self._format_qobj(qobj, backend_options, noise_model))
    RuntimeError: Unable to cast Python instance to C++ type (compile in debug mode for details)
```

PR #659, opened 2020-03-12
Bump macos version in travis

mtreinish:
### Summary

Currently we run macOS CI on macOS 10.12 which was released ~4 yrs ago.
Terra added a new binary requirement retworkx which builds it's wheels
against macOS 10.13. This means that pip can't find the wheels for macOS
in CI because 10.13 is too new. This commit bumps the macOS/xcode
version we use in the travis jobs to one slightly newer using macOS 10.13
(which is currently the default travis env). This should unblock CI
because we don't need to install retworkx from source with a newer osx version.

### Details and comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,14 +90,11 @@ stage_osx: &stage_osx
           cd ../../..
           python-build $PYTHON_VERSION ~/python-interpreters/$PYTHON_VERSION
         fi
+        sudo pip2 install -U virtualenv pip setuptools
         virtualenv --python ~/python-interpreters/$PYTHON_VERSION/bin/python venv
         source venv/bin/activate
       fi
   before_script:
-    - |
-      if [ ! "$(ls -A .stestr)" ]; then
-        rm -rf .stestr
-      fi
     # Build with Thrust OpenMP CPU backend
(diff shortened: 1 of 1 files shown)


atilag:
I think it's fine upgrading to 10.13 for MacOS, even though our wheels target a minimum of 10.9 (we were already using 10.12 as you mentioned).

.travis.yml
```diff
@@ -90,14 +90,11 @@ stage_osx: &stage_osx
           cd ../../..
           python-build $PYTHON_VERSION ~/python-interpreters/$PYTHON_VERSION
         fi
+        sudo pip2 install -U virtualenv pip setuptools
```
atilag: sudo and pip2? Isn't pip3 already installed?

mtreinish: it's a weird conflict, system python is still python2 and we still don't really have an env setup yet, just a built version of python3. To use virtualenv on python2 system python to build a venv for the pyenv built python3 version. It's kinda hacky (but it's works) and what we were already using. I tried it without this line in https://github.com/Qiskit/qiskit-aer/pull/659/commits/5e5a6286766c792467f685337a2a4561c2c07a02 and the job failed because virtualenv was not installed for the system python.

Issue #477, opened 2019-12-05
MPS simulation method: reduce number of swaps needed


merav-aharoni:
### What is the expected behavior?
In the current implementation of the MPS simulation method, to apply a 2-qubit (or 3-qubit) gate, we must first bring the qubits to consecutive positions in the MPS structure. This is done by applying `swaps` on all qubits between `qubit_a` and `qubit_b` until they are next to each other. After applying the gate, the qubits are returned to their original positions by additional `swaps`.
These swaps have a penalty in performance both by performing the gate itself, and also by potentially increasing the tensor dimensions for each of the qubits.
Suggestion: rather than return the qubits to their original positions, maintain a structure that will keep track of the labels of the qubits, and swap them only on demand.

PR #665, opened 2020-03-19
MPS: Optimize swaps

merav-aharoni:
### Summary
The functions in MPS used `swap` operations to bring qubits together for any operation that involves more than 1 qubit. Every `swap` may double the size of the relevant tensors, causing a significant deterioration in performance. Therefore we try to minimize the number of `swap` operations. The idea in this enhancement is rather than swapping qubits back and forth, to use a `lazy swap`, i.e., only swap when needed. This is done by maintaining a data structure that stores the ordering of the qubits.

### Details and comments

#477


merav-aharoni:
Some explanation for the benefit of the reviewer: 

1. The main change is in the addition of two new private attributes to the class MPS:
```
reg_t qubit_order_;
reg_t qubit_location_;
```
These two structures keep track of the current location of the qubits since various operations can move them around. One of these vectors would have been enough, but the second one helps in terms of performance. 
 
2. To keep the changes at a minimum, and to hide the implementation, I separated all the API methods into 2 functions: the _outer_ function (wrapper function) computes the actual ordering of the qubits, and calls the _inner_ function with this ordering. So the _inner_ is pretty much identical to the implementation that was there before.

3. There is only one place where the `qubit_order_` is modified and that is in `internal_swap` . When 2 qubits are swapped, we need to swap them in `qubit_order_` as well. Unless this is actually a `swap` gate.

merav-aharoni:
I did some performance measurements using @hhorii 's benchmarks - comparing the master version of MPS with the version in this PR. `mps aer time` is for the current version of mps in master. `mps opt time` is for the new, optimized version.
bv-1000-qubits
qft-300-qubits

merav-aharoni:
From my side, this PR is ready for review. It is not passing the unittests, but it seems like some problem unrelated to the PR.

merav-aharoni:
Tests finally pass. I think this can be reviewed now.

chriseclectic:
@atilag can you review this PR?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -989,20 +1017,33 @@ complex_t MPS::expectation_value_pauli(const reg_t &qubits, const std::string &m
...
+	  out << lambda_reg_[i] << std::endl;
 	}
-	out << std::endl;
-	return out;
+    }
+  out << std::endl;
+  return out;
+}
+
+std::ostream& MPS::sort_and_print(std::ostream& out) {
```
atilag, line 1035: Is this function used? I'd remove it if it's not.
If you want to keep it, I'd strongly recommend removing the print part as we already have a print function, and changing the signature to: `void MPS::sort_qubits()`.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1043,28 +1082,43 @@ MPS_Tensor MPS::state_vec_as_MPS(uint_t first_index, uint_t last_index) const
...
-
 	return temp;
 }
 
-void MPS::full_state_vector(cvector_t& statevector) const
-{
-  MPS_Tensor mps_vec = state_vec_as_MPS(0, num_qubits_-1);
-  uint_t length = 1ULL << num_qubits_;   // length = pow(2, num_qubits_)
-  statevector.resize(length);
+void MPS::full_state_vector(cvector_t& statevector){
```
atilag, line 1088: This can be `const` , right?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1043,28 +1082,43 @@ MPS_Tensor MPS::state_vec_as_MPS(uint_t first_index, uint_t last_index) const
...
+  reg_t qubits(num_qubits_);
+  std::iota( std::begin(qubits), std::end(qubits), 0);
+  reg_t internal_qubits = get_internal_qubits(qubits);
+  full_state_vector_internal(statevector, internal_qubits);
+}
+
+void MPS::full_state_vector_internal(cvector_t& statevector,
```
atilag, line 1095: This can be `const` as well, can't be?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1043,28 +1082,43 @@ MPS_Tensor MPS::state_vec_as_MPS(uint_t first_index, uint_t last_index) const
...
+  //temp_statevector will contain the statevector in the ordering define in "qubits"
+  reorder_all_qubits(statevector, qubits, temp_statevector);
+  // reverse to be consistent with qasm ordering
+  statevector = reverse_all_bits(temp_statevector, num_qubits);
+}
 
-#ifdef DEBUG
-  std::cout << *this;
-#endif
+void MPS::get_probabilities_vector(rvector_t& probvector, const reg_t &qubits) {
```
atilag, line 1115: If we can make this one `const` would be great too.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1129,19 +1195,36 @@ uint_t MPS::apply_measure(uint_t qubit,
   }
 
   // and propagate the changes to all qubits to the left
-  for (int_t i=qubit; i>0; i--) {
+	 for (int_t i=qubit; i>0; i--) {
```
atilag, line 1198: nit: Can  you indent this code?

src/simulators/matrix_product_state/matrix_product_state_internal.hpp
```diff
@@ -238,26 +213,20 @@ class MPS{
     return enable_gate_opt_;
   }
 
-  //  void store_measure(const AER::reg_t outcome, const AER::reg_t &cmemory, const AER::reg_t &cregister) const{
-  //           cout << " store_measure not supported yet" <<endl;}
-
-  double norm(const uint_t qubit, cvector_t &vmat) const {
+  double norm(const uint_t qubit, cvector_t &vmat) {
```
atilag, line 216: Can the function and `vmat` arguments be `const` ?

src/simulators/matrix_product_state/matrix_product_state_internal.hpp
```diff
@@ -382,6 +386,15 @@ class MPS{
   uint_t num_qubits_;
   std::vector<MPS_Tensor> q_reg_;
   std::vector<rvector_t> lambda_reg_;
+
+  // qubit_order_ stores the current ordering of the qubits,
+  // qubit_location_ stores the location of each qubit in the vector. It is derived from qubit_order_ 
+  // at the end of every swap operation for performance reasons
+  // for example: starting position qubit_order_=qubit_location_=01234
+  // ccx(0,4) -> qubit_order_=04123, qubit_location_=02341
+  reg_t qubit_order_;
```
atilag, line 395: nit: A nice abstraction for `qubit_order_` and `qubit_location_` would be creating a simple class to encapsulate both, or a `struct` if it's very simple. This way we enforce the fact that they are tightly coupled.

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -189,7 +189,7 @@ class State : public Base::State<matrixproductstate_t> {
   // Return vector of measure probabilities for specified qubits
   // If a state subclass supports this function, then "measure"
   // must be contained in the set defined by 'allowed_ops'
-  rvector_t measure_probs(const reg_t &qubits) const;
+  rvector_t measure_probs(const reg_t &qubits);
```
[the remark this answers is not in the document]
merav-aharoni: You are right. I modified it previously because of a change that was cancelled in the end. Changed back to const.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -44,7 +44,18 @@ static const cmatrix_t one_measure =
...
 // local function declarations
 //------------------------------------------------------------------------
-
+//------------------------------------------------------------------------ 
+// Function name: print_array
+// Description: prints out the valus of a vector. For debug purposes.
+// Input: the vector name, and the vector.
+// Returns: void
+template <class T>
+  void print_array(std::string name, const std::vector<T> &array) {
```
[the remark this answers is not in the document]
merav-aharoni: Right. Removed this function.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: This can no longer by const because it involves moving the qubits before computation. This is true for all methods that do a computation directly on the MPS structure, without first copying it to a temporary structure.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Cannot be const - see comment above.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Returned the const.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: done

src/simulators/matrix_product_state/matrix_product_state_internal.hpp
merav-aharoni: I agree. done.

src/simulators/matrix_product_state/matrix_product_state_internal.hpp
merav-aharoni: Yes. Done.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Removed this function.

src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
```diff
@@ -140,6 +136,11 @@ static void contract_2_dimensions(const MPS_Tensor &left_gamma,
 				  uint_t omp_threads,
 				  cmatrix_t &result);
 
+  // public static class members
+  static const double SQR_HALF; 
+  static const uint_t NUMBER_OF_PRINTED_DIGITS;
+  static const uint_t MATRIX_OMP_THRESHOLD;
```
atilag, line 142:
```suggestion
static constexpr double SQR_HALF = sqrt(0.5);
static constexpr uint_t NUMBER_OF_PRINTED_DIGITS = 3;
static constexpr uint_t MATRIX_OMP_THRESHOLD = 8;
```

src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
```diff
@@ -151,6 +152,9 @@ static void contract_2_dimensions(const MPS_Tensor &left_gamma,
 //=========================================================================
 // Implementation
 //=========================================================================
+  const double MPS_Tensor::SQR_HALF = sqrt(0.5);
+  const uint_t MPS_Tensor::NUMBER_OF_PRINTED_DIGITS = 3;
+  const uint_t MPS_Tensor:: MATRIX_OMP_THRESHOLD = 8;
```
atilag, line 157:
```suggestion
```

merav-aharoni:
@atilag , there was a problem with `constexpr SQR_HALF = sqrt(0.5)` on Windows, so I changed that statement back to the original.

atilag:
🤦  std::sqrt() is not constexpr on Windows... :_(

PR #667, opened 2020-03-24
Improve gate fusion

### LINKED ISSUE #616: Improve fusion transpilation pass

- [x] Refactor the fusion transpilation pass to use the unitary_state simulator (or superoperator_state simulator) to allow fusion of arbitrary sized unitary blocks.

- [ ] Allow setting different default fusion parameters for different simulation methods if it provides an advantage (eg for cpu vs gpu, or statevector vs density matrix etc)

- [ ] Related to above, investigate if it is beneficial to run fusion on GPU unitary / superoperator simulator

- [ ] Allow fusion of Kraus and instructions for backends that support them. This will combine a bunch of gates with the more general Kraus instruction to be applied in a single case and should give performance improvements for noisy simulations.

- [x] Allow fusion of reset and SuperOp instruction on density matrix simulations. As above may provide significant performance improvements for noisy simulation.

chriseclectic:
### Summary

Refactor gate fusion transpilation to use the Unitary state simulator or superoperator state simulator for fusing instructions. This also allows fusion of superoperator instructions in noisy circuits for density matrix simulation.

### Details and comments

Partial closes #616

### DIFF
--- a/src/transpile/fusion.hpp
+++ b/src/transpile/fusion.hpp
@@ -98,6 +98,8 @@ private:
   void dump(const Circuit& circuit) const;
 #endif
 
+private:
+  bool allow_superop_ = true;
 };
 
 
@@ -117,8 +119,11 @@ void Fusion::set_config(const json_t &config) {
   if (JSON::check_key("fusion_threshold", config_))
     JSON::get_value(threshold, "fusion_threshold", config_);
 
-  if (JSON::check_key("fusion_cost_factor", config_))
-    JSON::get_value(cost_factor, "fusion_cost_factor", config_);
+  if (JSON::check_key("fusion_cost_factor", config))
+    JSON::get_value(cost_factor, "fusion_cost_factor", config);
+  
+  if (JSON::check_key("fusion_allow_superop", config))
+    JSON::get_value(allow_superop_, "fusion_allow_superop", config);
 }
 
 
@@ -132,16 +137,16 @@ void Fusion::optimize_circuit(Circuit& circ,
 
   // If we are doing superoperator fusion we have the threshold
   // for the density matrix simulator
-  bool density_matrix = allowed_opset.contains(optype_t::superop);
+  bool allow_superop = allow_superop_ && allowed_opset.contains(optype_t::superop);
 
   // Check if circuit size is above threshold
-  if (circ.num_qubits < (density_matrix) ? threshold / 2 : threshold) {
+  if (circ.num_qubits < (allow_superop) ? threshold / 2 : threshold) {
     return;
   }
 
   // Check if circuit contains kraus or superop instructions
   // snd backend allows superop instructions
-  bool use_superop_method = density_matrix && (
+  bool use_superop_method = allow_superop && (
     circ.opset().contains(optype_t::kraus) ||
     circ.opset().contains(optype_t::superop) ||
     circ.opset().contains(optype_t::reset));
@@ -151,7 +156,7 @@ void Fusion::optimize_circuit(Circuit& circ,
   metadata["cost_factor"] = cost_factor;
   metadata["max_fused_qubits"] = max_fused_qubits;
   metadata["method"] = (use_superop_method) ? "superop" : "unitary";
-  if (verbose)
+  if (verbose) {
     metadata["input_ops"] = circ.ops;
   }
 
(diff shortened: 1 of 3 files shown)


src/transpile/fusion.hpp
```diff
@@ -260,7 +277,36 @@ double Fusion::get_cost(const op_t& op) const {
...
+                                    superop_simulator.qreg().matrix());
+  } else {
+    QubitUnitary::State<> unitary_simulator;
+    unitary_simulator.initialize_qreg(qubits.size());
+    unitary_simulator.apply_ops(fusioned_ops, dummy_data, dummy_rng);
+    return Operations::make_unitary(qubits,
+                                    unitary_simulator.qreg().matrix(),
+                                    std::string("fusion"));
+  }
+}
```
vvilpas: I think this code can be made more generic to avoid code repetition. I've rewritten it using templates (see below). In this case it is a bit verbose because I've had to deal also with the `Operations::make...` which are themselves very specific.  Other problem is that, as we have most of the code in .hpp files, the free functions are exposed to everyone that includes the file. Although they can be put as private member functions in Fusion, I'd be careful not to start adding too many functions. I would normally put them in an anonymous namespace in the .cpp file.
Anyway, consider this a suggestion, as in this case both branches of the `if/else` are quite simple so feel free to keep it as it is now.

```
template <typename OPERATOR>
op_t make_fusion_op(const reg_t &qubits, OPERATOR& op_simulator);


template<> op_t make_fusion_op<QubitSuperoperator::State<>>(const reg_t &qubits,
                                                            QubitSuperoperator::State<>& op_simulator){
  return Operations::make_superop(qubits, op_simulator.qreg().matrix());
}


template<> op_t make_fusion_op<QubitUnitary::State<>>(const reg_t &qubits,
                                                      QubitUnitary::State<>& op_simulator){
  return Operations::make_unitary(qubits, op_simulator.qreg().matrix(),
                                  std::string("fusion"));
}

template <typename OPERATOR>
op_t get_fusion_operator(const std::vector<op_t>& fusioned_ops,
                         const reg_t &qubits) {
    RngEngine dummy_rng;
    ExperimentData dummy_data;
    OPERATOR op_simulator;
    op_simulator.initialize_qreg(qubits.size());
    op_simulator.apply_ops(fusioned_ops, dummy_data, dummy_rng);
    return make_fusion_op<OPERATOR>(qubits, op_simulator);
}

op_t Fusion::generate_fusion_operation(const std::vector<op_t>& fusioned_ops,
                                       const reg_t &qubits,
                                       bool use_superop_method) const {
  // Run simulation
  if (use_superop_method) {
      return get_fusion_operator<QubitSuperoperator::State<>>(fusioned_ops, qubits);
  }
  return get_fusion_operator<QubitUnitary::State<>>(fusioned_ops, qubits);
}
```

atilag: Nice, this is  the way to do it.
One single comment, we used before in the codebase the `type_t` nomenclature for our 
types, so I'd recommend  using: `operator_t`

chriseclectic: I did it that way originally but ended up changing it because found the template specialization made the code harder to follow. I think with one slightly verbose function it is easier to understand whats going on. I also had some issues with template specialization giving compiler errors if you didn't have the specialization before the function that called it etc.

In future I would also like to be able to add a third instruction output that uses superoperator simulator but converts its output to Kraus that I don't think will fit nicely with template parameter specialization based on the simulator class.

src/transpile/fusion.hpp
```diff
@@ -312,204 +370,41 @@ bool Fusion::aggregate_operations(oplist_t& ops, const int fusion_start, const i
 
     if (to != i) {
       std::vector<op_t> fusioned_ops;
+      std::set<uint_t> fusioned_qubits;
       for (int j = to; j <= i; ++j) {
         fusioned_ops.push_back(ops[j]);
+        std::copy(ops[j].qubits.cbegin(), ops[j].qubits.cend(),
+                  std::inserter(fusioned_qubits, fusioned_qubits.end()));
```
vvilpas, line 377:
```suggestion
        fusioned_qubits.insert(ops[j].qubits.cbegin(), ops[j].qubits.cend())
```
More compact and probably faster

PR #670, opened 2020-03-27
Add superop support to statevector method via conversion to Kraus

chriseclectic:
### Summary

Depends on #889.

* Start adding helper functions for converting between CPTP quantum error representations.
* Adds support for applying superop instructions in statevector method via conversion to Kraus (note that doing this every shot would be inefficient and all supeorops should be converted to Kraus before shot sampling).


chriseclectic:
Closing since this is superseded by #927

PR #673, opened 2020-03-30
[WIP] Implement a function to get the eigenvalues and eigenvector of a Hermitian PSD matrix

### LINKED ISSUE #669: C++ function for eigensystem decomposition of a PSD matrix.

A C++ function to compute the eigenvalues and eigenvectors of a postive-semidefinite (PSD) complex matrix. Something like

```c++
template <class float_t>
void eig_psd(matrix<std::complex<float_t>>& mat,
             std::vector<std::complex<float_t>> &evals,
             matrix<std::complex<float_t>> &evecs);
```
where `evels` stores the eigenvalues, and `evecs` stores the eigenvectors as rows or columns (depending on underlying array storage layout). 

From what I can gather the best way to do this would be to use the LAPACK routines (depending on template `float`/`double`):

1. `chetrd/zhetrd`: convert complex Hermitian matrix to a tri-diagonal matrix (which will be PSD if inital matrix is PSD).
2. `cpteqr/zpteqr`: compute eigenvalues and vectors for a PSD tri-diagonal matrix.

atilag:
### Summary
Fixes #669 
As stated in issue #669 , this PR will implement a new function in our `Linalg` library that will
return the eigenvalues/eigenvectors of a given PSD matrix. 

### Details and comments
As a result of setting up all the basic scaffolding for unit testing (C++), we have extended our `matrix` class so comparing two matrices of complex numbers is easier.
We have added a new `Linalg::complex_almost_equal` to compare complex numbers by approximation as well.

TODO
- [ ] The actual implementation of the `eigensystems_psd` function
- [ ] Check that C++ tests are correct
- [ ] Add Python tests


nonhermitian:
Not sure if this helps, but here is the cython version:

```python
cimport cython
cimport numpy as np
import numpy as np
from scipy.linalg.cython_lapack cimport zhetrd, zungtr, zpteqr
np.import_array()

cdef extern from "numpy/arrayobject.h" nogil:
    void PyDataMem_FREE(void * ptr)
    void PyDataMem_NEW(size_t size)


@cython.boundscheck(False)
@cython.wraparound(False)
cdef void EIG_PSD(complex * H, double * eigvals, int nrows):
    """
    Computes the eigenvalues and vectors of a dense Hermitian matrix.
    Eigenvectors are returned in H, values in eigvals

    Parameters:
        H (array_like): Input double complex Hermitian matrix.
        eigvals (array_like): Input dou3ble array.
    
    """
    # Use lower part of Hermitian matrix
    cdef char uplo = b'L'
    cdef char compz = b'V'
    cdef int lwork = 18 * nrows
    cdef int info=0
    
    #These need to be freed at end
    cdef complex * work = <complex *>PyDataMem_NEW(lwork * sizeof(complex))
    cdef double * e = <double *>PyDataMem_NEW((nrows-1) * sizeof(double))
    cdef complex * tau = <complex *>PyDataMem_NEW((nrows-1) * sizeof(complex))

    zhetrd(&uplo, &nrows, H, &nrows, eigvals, e, tau, work, &lwork, &info)

    if info != 0:
        if info < 0:
            raise Exception("Error in parameter : %s" & abs(info))
        else:
            raise Exception("Algorithm 1 failed to converge")
            
    # reset info
    info = 0 
    zungtr(&uplo, &nrows, H, &nrows, tau, work, &lwork, &info)
    
    if info != 0:
        if info < 0:
            raise Exception("Error in parameter : %s" & abs(info))
        else:
            raise Exception("Algorithm 2 failed to converge")
    
    
    PyDataMem_FREE(work)
    #These need to be freed at end
    cdef double * work2 = <double *>PyDataMem_NEW(lwork * sizeof(double))
    
    # reset info
    info = 0 
    zpteqr(&compz, &nrows, eigvals, e, H, &nrows, work2, &info)
    if info != 0:
        if info < 0:
            raise Exception("Error in parameter : %s" & abs(info))
        else:
            raise Exception("Algorithm 3 failed to converge")
    
    PyDataMem_FREE(work2)
    PyDataMem_FREE(e)
    PyDataMem_FREE(tau)
    

@cython.boundscheck(False)
@cython.wraparound(False)
def eig_psd(H):
    cdef int nrows = H.shape[0]
    cdef np.ndarray[complex, ndim=2, mode="c"] out = np.triu(H)
    cdef np.ndarray[double, ndim=1, mode="c"] eigvals = np.empty(nrows, dtype=float)
    
    EIG_PSD(&out[0,0], &eigvals[0], nrows)
    
    return eigvals[::-1], (out.conj().T)[:, ::-1]
```

src/framework/linalg/almost_equal.hpp
```diff
@@ -29,7 +29,7 @@ namespace Linalg {
 // If we have numbers closer to 0, then max_diff can be set to a value
 // way smaller than epsilon. For numbers larger than 1.0, epsilon will
 // scale (the bigger the number, the bigger the epsilon).
-template <typename T, typename = enable_if_numeric_t<T>>
```
chriseclectic: This function was specifically implemented to add complex to integral constant check (you don't really need to define an enable_if_numeric because it's basically a built in).  There are a lot of other areas of code you will need to change if you have modified this in all the linalg functions.

src/framework/linalg/almost_equal.hpp
```diff
@@ -39,6 +39,26 @@ bool almost_equal(T f1, T f2,
          max_relative_diff * std::max(std::abs(f1), std::abs(f2));
 }
 
+// Complex numbers are probably not considered scalar, so we need another
+// implementation to deal with them
+template <typename complex_t, typename = enable_if_complex_t<complex_t>>
+bool complex_almost_equal(complex_t f1, complex_t f2,
```
chriseclectic: Complex shouldn't need a special almost equal. The `abs` function handles that case.

src/framework/linalg/eigen_psd.hpp
```diff
@@ -0,0 +1,38 @@
...
+ * Returns the eigenvalues and eigenvectors of a Hermitian
+ * positive semi-definite (PSD) matrix.
+ * @param psd_matrix: The Hermitian PSD matrix.
+ * @param eigenvalues: The eignevalues of the matrix.
+ * @param eigenvectors: The eigenvectors of the matrix.
+ *
+ * @returns: true.
+ */
+template <class float_t>
+void eigensystems_psd(const matrix<std::complex<float_t>>& psd_matrix,
```
chriseclectic:
```suggestion
void eigensystem_psd(const matrix<std::complex<float_t>>& psd_matrix,
```

src/framework/matrix.hpp
```diff
@@ -513,6 +566,16 @@ inline void matrix<T>::SetOutputStyle(enum OutputStyle outputstyle) {
   // sets the outputstyle
   outputstyle_ = outputstyle;
 }
+
+template<class T>
+bool matrix<T>::AlmostEqual(const matrix<T>& rhs) const {
+  for(auto i = 0; i < rhs.size(); ++i){
+    if(complex_almost_equal<T>(mat_[i], rhs[i]) == false)
```
chriseclectic: Does this also work if matrices are real? 

Also could this be moved to a util function using templates: `almost_equal(matrix<T>&, matrix<T>&)` so we can have one `almost_equal` function that works with most types.

src/framework/matrix.hpp
```diff
@@ -35,6 +35,8 @@ Multiplication is done with the C wrapper of the fortran blas library.
 #include <iostream>
 #include <vector>
 #include <array>
+#include "framework/linalg/enable_if_numeric.hpp"
+#include "framework/linalg/almost_equal.hpp"
```
chriseclectic: Are dependencies are in danger of getting a bit circular here since matrix is one of the primitive classes imported included in a lot of places. I'm hesitant to add methods to the matrix class that require an import until we do a proper refactor.

src/framework/types.hpp
```diff
@@ -39,10 +39,16 @@ namespace AER {
   using int_t = int_fast64_t;
   using uint_t = uint_fast64_t;
   using complex_t = std::complex<double>;
+  using complexf_t = std::complex<float>;
```
chriseclectic: I'm hesitant to add more aliases here. Maybe instead look to change cvector and cmatrix to be templated as:
```
template <class FloatType = double>
using cmatrix_t = matrix<std::complex<FloatType>>
```
so we can use templates in all the other util functions to work for all float precisions (This would involve a pretty big refactor in a lot of places though)

src/framework/linalg/enable_if_numeric.hpp
```diff
@@ -19,20 +19,27 @@
 #include <type_traits>
 
 // Type check template to enable functions if type is a numeric scalar
-// (integer, float, or complex float)
```
chriseclectic: This was specifically to check all all numerics including complex.

If you wanted one that doesn't include complex I would make that the special case below with
`enable_if_numeric_t`  # real or complex
`enable_if_real_numeric_t` # real only
`enable_if_complex_numeric_t` # complex only

src/framework/linalg/almost_equal.hpp
```diff
@@ -39,6 +39,26 @@ bool almost_equal(T f1, T f2,
          max_relative_diff * std::max(std::abs(f1), std::abs(f2));
 }
 
+// Complex numbers are probably not considered scalar, so we need another
```
chriseclectic: Complex numbers are considered scalar, scalar just means not-matrix/vector valued

src/framework/linalg/almost_equal.hpp
atilag: Unfortunately, the `max_diff` (`std::numeric_limits<std::complex<double>>`) won't work. So as talked, let's create an overloaded function for complex types.

src/framework/linalg/almost_equal.hpp
atilag: Ok, let's keep it untouched because as you said is being used through all the linalg library, but I'm not sure that this will have consequences for functions that need specialization for complex numbers... maybe it's just semantics, or maybe these functions will just fail compiling at some point (which is acceptable).

src/framework/linalg/enable_if_numeric.hpp
atilag: oh! sure.. I just forgot to add the `complex floating-points` part. I just wanted to make sure that `double` types are included (and not just  `float`)

chriseclectic:
I forgot to add, we should include tests for this function that its decomposition recovers the original matrix. Eg something like:

```c++
eigensystem_psd(mat, evals, evecs);

cmatrix_t value(mat.size());

for (size_t j=0; j < evals.size(); j++) {
  value += evals[j] * Utils::projector(evecs[j]));
}
almost_equal(mat, value) == true
```

src/framework/matrix.hpp
atilag: Yeah, Agree., but that's why we put the include guards at the beginning of every header file, to avoid recursions.

src/framework/matrix.hpp
atilag: Yep! Better having overloaded functions.

src/framework/types.hpp
atilag: I found them helpful when I was writing the tests for double/float specialization... but yes, I do prefer your suggestion.

chriseclectic: You can always add them to the top of the test file

atilag: As this would require a large refactor, what about  adding just the 3 single-precision types: `complexf_t`, `cvectorf_t`, `cmatrixf_t` ? And we can tackle this refactor later in another PR (to not overcomplicate this one).

ereastman:
thanks for the pointer @nonhermitian - The function signatures were harder to get right than I thought - the output value that is supposed to tell you if something is wrong doesn't help if there's a seg fault inside the function. I got them down now I think. Just connecting them together and testing.

thanks for the testing framework @atilag - I was getting compiler errors on some of the enable_ifs at first but perhaps you've updated. I created a new PR #680 - once I work out a set of tests I'll merge with this branch and automate them.

If I try to run this on the Pauli matrices I'm getting an error code returned by cpteqr:
```
the Cholesky factorization of the matrix could not be performed because the i-th principal minor was not positive definite.
```
For the first Pauli matrix, I get i=1, for the second I get i=2. Is this expected? I can't really interpret the error message. 

Also, from the ?pteqr [docs](http://www.netlib.org/lapack/explore-html/d3/d7a/group__complex_p_tcomputational_ga0333b0f77b9e0af4ad1012b14d9d9d22.html): 

```
This routine computes the eigenvalues of the positive definite
 tridiagonal matrix to high relative accuracy.  This means that if the
 eigenvalues range over many orders of magnitude in size, then the
 small eigenvalues and corresponding eigenvectors will be computed
 more accurately than, for example, with the standard QR method.
```

Was wondering if/how this might impact results...? Not sure exactly how this will be used - perhaps we should check that we aren't accumulating errors. Would it be possible to demonstrate in tests?

atilag:
@ereastman yes, I fixed all the c++ tests in my last commit, everything should build by now.

chriseclectic:
@ereastman this will be used for PSD matrices where all eigenvalues are in the range `[0, 2^n]` (most often for N=1 or 2). The important part for accuracy is we want to tell when some eigenvalues are zero (or close to zero), so there will typically be several orders of magnitude between the values in that case.

Pauli matrices are not PSD: their eigenvalues are +1 and -1, so they can't be used for a test case. You can construct an  N x N PSD matrix with R-nonzero eigenvalues as `M = A.A^\dagger` where A is a N x R rectangular complex matrix.

chriseclectic:
@atilag What is the status of this PR?

ereastman:
From the [zheevx_ docs]()

```
Eigenvalues and eigenvectors can be selected by specifying either a range of values or a range of indices for the desired eigenvalues.
```
`eigenvalues can be in the range [0, 2^N]` - is n also the first (==second) dim of the matrix?
Can we make any assumptions on the number of eigenvals/vecs (that are of a certain magnitude) prior to calculating any?
My current code might be wrong since I currently only get the first N. 
Perhaps a performance boost could be had by not getting all values all at once. We could add the number of vectors to get as an argument.. but then the way we return by reference should probably be changed. Argument could be eigenvalue index and/or magnitude.
Also the allocation of working memory is probably a pretty big part of performance itself for these (would need benchmarking though) - if it were being called repeatedly we may want to think of ways to only allocate it once (there are A LOT of ways we could do this, function or library scale)

ereastman:
Also - see [here for further updates](https://github.com/atilag/qiskit-aer/pull/1/files)

chriseclectic:
Replaced by #889

PR #615, opened 2020-02-14
Avoid using AOL and RTOL for docstring in QuantumError and ReadoutError

hhorii:
### Summary
`sphinx-build` fails if file system does not support case-sensitive path.

### Details and comments
`QuantumError` and `ReadoutError` have `atol` methods, `ATOL` field, `rtol` methods, and `RTOL` filed. `sphinx-build` generates stub files for them. If the file system does not support case-sensitive path (as Mac does), `sphinx-build` fails (ref: https://github.com/sphinx-doc/sphinx/issues/4874).

PR #676, opened 2020-03-31
Fixes for terra 0.13 release

chriseclectic:
### Summary

* Changes `atol` and `rtol` methods in `QuantumError` and `ReadoutError` to batch terra `BaseOperator` class (supersedes #615 )
* Changes tests to use `circ.i` instead of deprecated `circ.iden` in tests.


qiskit/providers/aer/noise/errors/quantum_error.py
```diff
@@ -187,35 +187,37 @@ def copy(self):
...
-        """Set the relative tolerance parameter for float comparisons."""
-        max_tol = self.MAX_TOL
-        if rtol < 0:
-            raise NoiseError("Invalid rtol: must be non-negative.")
-        if rtol > max_tol:
+        return QuantumError._RTOL_DEFAULT
+
+    @classmethod
+    def set_atol(cls, value):
+        """Set the class default absolute tolerance parameter for float comparisons."""
```
vvilpas, line 200: I see the logic for tolerances follows the `base_operator` class in `terra`, but I find a bit weird to change class attributes through set methods in instantiated objects. Maybe a meta class would have been the way to go here.
This is just a comment, let's not halt the merge because of this.

chriseclectic: The idea was you call these on the class `QuantumError.set_atol(1e-7)` (of course since its python nothing stops you calling it on an object instance too)

vvilpas: Yes, you are right. I expressed myself quite bad.  My concern was why not use the `property` mechanism for the class members themselves through a metaclass, so that you do `QuantumError.atol = 1e-7` instead of `QuantumError.set_atol(1e-7)`. That way you can even avoid   the possibility of setting `atol` from an instance or mistakenly do `QuantumError.atol = 1e-7`, creating a new class variable `atol` and leaving `_ATOL_DEFAULT` unchanged.
But yes, it's Python and "_We are consenting adults_" here  :)
Jokes aside, I'm ok with this.

chriseclectic: Can you make an issue in terra about how to do this?

PR #685, opened 2020-04-09
[Stable] 0.4.2 backports

mtreinish:
### Summary

This PR is the backport of all the fixes on master that need to be included in 0.4.2. Most of the commits on master at this point are fixes that need to be included in the bug fix release. So instead of opening individual PRs for each of the commits that needs to be backported (per https://qiskit.org/documentation/contributing_to_qiskit.html#stable-branch-policy ) one PR is done here instead and a merge commit will be used to preserve the history.

### Details and comments

Backported from #651, #660, #485, and #683


mtreinish:
This should be merged as either rebase and merge or with a merge commit so we preserve the commits for the git history

PR #688, opened 2020-04-10
Disable jsonschema validation by default

mtreinish:
### Summary

Currently the aer backends are running jsonschema validation on the Qobj
objects passed in via terra by default. This is unnecessary overhead as
terra is programatically creating these objects and they're extremely
unlikely to be structurally invalid (which is all jsonschema will
catch). Even, if it was this will be caught later when something goes to
use it and the error message from that is more likely to be useful since
the way the json schemas are constructed precludes useful error message.
This commit changes the default value of this to be opt-in on the run()
method to avoid this unnecessary overhead.

### Details and comments

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -83,7 +83,7 @@ class AerBackend(BaseBackend):
         self._controller = controller
 
     # pylint: disable=arguments-differ
-    def run(self, qobj, backend_options=None, noise_model=None, validate=True):
+    def run(self, qobj, backend_options=None, noise_model=None, validate=False):
         """Run a qobj on the backend.
 
         Args:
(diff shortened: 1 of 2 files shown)


mtreinish:
The test failures here are unrelated. It looks like they're caused by Qiskit/qiskit-terra#4016 which merged this morning

atilag:
I agree to disable json schema validation by default temporary, while we come up with a better validation strategy (I'm currently working on it). Anyway, after analyzing several profiling reports, the time spent validating in large simulations can be considered negligible. I can't think of any real use-case where this is a real problem, apart from competitive benchmarkings of course :)

atilag:
Just a reminder: If we decide to disable schema validation by default, we have to coordinate with our cloud simulator team so they explicitly activate validation

mtreinish:
This is independent of the cloud simulator. As I said in the commit message this is just for the aer provider. The cloud simulator runs in standalone mode so this won't affect it. Additionally, the qobj is validated prior to the standalone aer simulator receiving it.

chriseclectic:
@atilag this disabling is fine as we basically validate twice anyway, once in python and again when we are parsing the JSON into C++. If we were disabling the checking of the C++ validation (which is not done via schema, but by explicitly checking instructions for things like right number of qubits and no duplicate qubits etc) we would need to coordinate with cloud team.

PR #690, opened 2020-04-10
Add templates to RngEngine class

chriseclectic:
### Summary

Code cleanup

Refactor of `rng.hpp` so that it uses int/float templates for its functions to work with different precision/width numbers.

### Details and comments

Note: I'm breaking up things in #671 into smaller separate PRs because I was getting some runtime errors with that PR.

### DIFF
--- a/src/framework/rng.hpp
+++ b/src/framework/rng.hpp
@@ -40,7 +40,7 @@ public:
   RngEngine() { set_random_seed(); }
 
   // Seeded constructor initialize RNG engine with a fixed seed
-  RngEngine(size_t seed) { set_seed(seed); }
+  explicit RngEngine(size_t seed) { set_seed(seed); }
 
   //-----------------------------------------------------------------------
   // Set fixed or random seed for RNG
--- a/src/framework/rng.hpp
+++ b/src/framework/rng.hpp
@@ -94,7 +94,7 @@ public:
   }
 
 private:
-  std::mt19937 rng; // Mersenne twister rng engine
+  std::mt19937_64 rng; // Mersenne twister rng engine
 };
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/framework/rng.hpp
```diff
@@ -18,117 +18,85 @@
...
-  explicit RngEngine(uint_t seed) { rng.seed(seed); };
+  //-----------------------------------------------------------------------
+  // Constructors
+  //-----------------------------------------------------------------------
+
+  // Default constructor initialize RNG engine with a random seed
+  RngEngine() { set_random_seed(); }
+
+  // Seeded constructor initialize RNG engine with a fixed seed
+  RngEngine(size_t seed) { set_seed(seed); }
```
vvilpas, line 43: Was `explicit` removed on purpose?

src/framework/rng.hpp
```diff
@@ -18,117 +18,83 @@
...
+  template <typename IntType>
+  IntType rand_int(IntType a, IntType b) {
+    return std::uniform_int_distribution<IntType>(a, b)(rng);
+  }
+
+  // Generate a pseudo random integer from a a discrete distribution
+  // constructed from an input vector of probabilities for [0,..,n-1]
+  // where n is the lenght of the vector. If this vector is not normalized
+  // it will be scaled when it is converted to a discrete_distribution
+  template <typename FloatType>
```
atilag, line 89: What about checking `FloatType` is actually a floating point type?
```c++
template<typename FloatType,
         typename = typename std::enable_if<std::is_floating_point<FloatType>::value >::type>
size_t rand_int(const std::vector<FloatType> &probs) {
        return std::discrete_distribution<size_t>(probs.begin(), probs.end())(rng);
}

```
(I really hope that Concepts and `if constexpr` will help us with these things)

PR #694, opened 2020-04-14
Handle complex numbers in pulse defaults

mtreinish:
### Summary

In Qiskit/qiskit-terra#4016 the pulse defaults (as well as backend
configuration and properties) were updated to not rely on marshmallow. A
side effect of this change is that the return from to_dict() on pulse
defaults no longer will convert complex -> [real, imag] since this is
typically an unecessary conversion that complicates using the output.
However, the pulse system model class was not handling this as expected
and will fail trying to access the the components from [real, imag].
This commit updates this to handle the case when a complex number is
passed in.

### Details and comments

This will be blocked until Qiskit/qiskit-terra#4124 merges.


chriseclectic:
I believe this was included in #695

PR #695, opened 2020-04-14
Avoid to_dict() when possible in pulse system model

mtreinish:
### Summary

As fall out from Qiskit/qiskit-terra#4016 there was a bug in the
to_dict() method of the backend configuration that was resulting in
UChannelLO objects not getting converted to their dict form in the
output of to_dict(). While this is being fixed in
Qiskit/qiskit-terra#4124 this commit side steps the issue by adapting
the access patterns in the pulse system model to rely on class attribute
access and getattr() instead of dict access and get().

### Details and comments

This is an alternative to the fix from a combination of #694 and Qiskit/qiskit-terra#4124

### DIFF
--- a/qiskit/providers/aer/pulse/pulse_system_model.py
+++ b/qiskit/providers/aer/pulse/pulse_system_model.py
@@ -151,7 +151,7 @@ class PulseSystemModel():
                 if drive_idx is not None:
                     # construct string for u channel
                     u_string = ''
-                    for u_term_dict in u_lo.to_dict():
+                    for u_term_dict in u_lo:
                         scale = getattr(u_term_dict, 'scale', [1.0, 0])
                         q_idx = getattr(u_term_dict, 'q')
                         if len(u_string) > 0:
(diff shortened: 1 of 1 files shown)


qiskit/providers/aer/pulse/pulse_system_model.py
```diff
@@ -151,12 +151,15 @@ def from_backend(cls, backend, subsystem_list=None):
                 if drive_idx is not None:
                     # construct string for u channel
                     u_string = ''
-                    for u_term_dict in u_lo:
+                    for u_term_dict in u_lo.to_dict():
```
chriseclectic, line 154:
```suggestion
                    for u_term_dict in u_lo:
```

qiskit/providers/aer/pulse/pulse_system_model.py
```diff
@@ -151,12 +151,15 @@ def from_backend(cls, backend, subsystem_list=None):
...
                         scale = u_term_dict.get('scale', [1.0, 0])
```
chriseclectic, line 155:
```suggestion
                        scale = getattr(u_term_dict, 'scale', [1.0, 0])
```

qiskit/providers/aer/pulse/pulse_system_model.py
```diff
@@ -151,12 +151,15 @@ def from_backend(cls, backend, subsystem_list=None):
...
                         q_idx = u_term_dict.get('q')
```
chriseclectic, line 156:
```suggestion
                        q_idx = getattr(u_term_dict, 'q')
```

Issue #692, opened 2020-04-13
Aer 0.5 not handling complex floats in state initialization


charleshadfield:
### Informations

- **Qiskit AER version**: 0.5.0
- **Operating system**: mac 10.14.6

### What is the current behavior?

If a QuantumCircuit is executed with an initialize command, a RuntimeError is thrown whenever the state chosen to initialize has complex amplitudes. (This problem has occurred upon updating to `qiskit==0.18` it does not occur for `qiskit==0.17`)

### Steps to reproduce the problem

    from qiskit import QuantumCircuit, execute
    from qiskit import Aer
    # from qiskit.quantum_info import random_statevector
    import numpy as np
    
    simulator = Aer.get_backend('qasm_simulator')
    
    circ = QuantumCircuit(1,1)
    # state = random_statevector(2).data
    state = (1/np.sqrt(2))*np.array([1, 1+0j]) # the presence of 0j throws the error
    circ.initialize(state, 0)
    circ.measure(0,0)
    
    job = execute(circ, simulator, shots=1)
    result = job.result()

The error:

    RuntimeError: Unable to cast Python instance to C++ type (compile in debug mode for details)

### Suggested solutions

chriseclectic:
The issue is that because of the way the terra `Instruction` checks param values the complex values of the array are being added to the params list as `numpy.complex128` instead of standard python `complex`, and these weren't caught by the pybind11 wrapper to C++. 

@ereastman is working on adding support for that to the wrappers, but probably the Instruction object should be modified to call `tolist` on Numpy arrays if it is storing the array as a param list, or better yet the param should store the full numpy vector since it is more efficient to pass Numpy arrays to C++ through Pybind11.

A work around in the meantime is to call `tolist` yourself like this:

```python
psi = np.array([1, -1j]) / np.sqrt(2)
circ.initiliaze(psi.tolist(), [0])
```

chriseclectic:
This should be fixed by #697

PR #697, opened 2020-04-14
Pybind numpy input

ereastman:
### Summary

Addresses #692  

This adds support for various numpy numeric types in the pybind interface.

### Details and comments

Within terra some numpy arrays are converted to python lists of NumPy types. Instead of doing ndarray.tolist() these components iterate and copy by value into a list. See [here](https://github.com/Qiskit/qiskit-terra/blob/99b83acc64238243684a0023abc9afd3fa53313f/qiskit/circuit/instruction.py#L59) and [here](https://github.com/Qiskit/qiskit-terra/blob/99b83acc64238243684a0023abc9afd3fa53313f/qiskit/qobj/qasm_qobj.py#L115) - the params is documented such that it indicates it must be a list but that is not enforced (i.e. raises if it is a numpy array) and also indicates the elements within must be one-of a set of types and that is also not enforced (until it hits pybind).


chriseclectic:
@ereastman Can you update the tests for `diagonal` and `initialize` to include cases where the params are real/complex numpy arrays or lists. Currently they are always lists I believe.

We could run the test with following cases. Eg for initialize if the initialize state is a real list `lst` you could test with 
* `np.array(lst)`
* `np.array(lst, dtype=float)` 
* `np.array(lst, dtype=np.float32)`
* `np.array(lst, dtype=complex)`
* `np.array(lst, dtype=np.complex64)`

Similarly for a complex list (but dropping the float cases).

mtreinish:
We should fix this in terra so that we're not converting numpy arrays to lists. Part of removing marshmallow from qobj was to facilitate local simulators getting the numpy array directly and avoid the conversion back and forth between arrays and lists.

chriseclectic:
@mtreinish this has already been fixed in Terra here https://github.com/Qiskit/qiskit-terra/pull/4152, but it won't help performance unless we change the signature of those instructions (`initialize` and `diagonal`) to store a a single Numpy array vector in their params, rather than extract the vector contents into params list.

Either way adding these edge cases in the pybind wrapper is still a good idea I think.

src/framework/pybind_json.hpp
```diff
@@ -257,38 +257,46 @@ json_t JSON::numpy_to_json(py::array_t<T, py::array::c_style> arr) {
...
+        for (auto item : py::cast<py::dict>(obj)) {
             js[item.first.cast<nl::json::string_t>()] = item.second;
         }
     } else if (py::isinstance<py::array_t<double> >(obj)) {
         js = JSON::numpy_to_json(obj.cast<py::array_t<double, py::array::c_style> >());
     } else if (py::isinstance<py::array_t<std::complex<double> > >(obj)) {
         js = JSON::numpy_to_json(obj.cast<py::array_t<std::complex<double>, py::array::c_style> >());
-    } else if (std::string(py::str(obj.get_type())) == "<class \'complex\'>") {
+    } else if (py::isinstance<py::float_>(obj)
+               || type_str == "<class \'numpy.float32\'>"
```
chriseclectic, line 280: You could separate the string check comparisons to all be done at the end (and sort them at the end in the order they are most likely to appear). This will prevent performance degradation for the most used types. Eg.

```cpp
else if (py::isinstance<py::float_>(obj)) {
  js = obj.cast<nl::json::number_float_t>();
} else if (py::isinstance<py::int_>(obj) {
  js = obj.cast<nl::json::number_integer_t>();
} else if (obj.is_none()) {
 return;
} else {
  // For performance reasons all string-match conversion cases are considered last
  auto type_str = std::string(py::str(obj.get_type()));
  if (std::string(py::str(obj.get_type())) == "<class \'complex\'>"
               || type_str == "<class \'numpy.complex128\'>"
               || type_str == "<class \'numpy.complex64\'>"
               || type_str == "<class \'numpy.complex_\'>" ) {
      auto tmp = obj.cast<std::complex<double>>();
      js.push_back(tmp.real());
      js.push_back(tmp.imag());
    } else if (type_str == "<class \'numpy.float64\'>" || type_str == "<class \'numpy.float32\'>") {
        js = obj.cast<nl::json::number_float_t>();
  } else if (type_str == "<class \'numpy.int64\'>"
               || type_str == "<class \'numpy.uint64\'>"
               || type_str == "<class \'numpy.int32\'>"
               || type_str == "<class \'numpy.uint32\'>" ) {
        js = obj.cast<nl::json::number_integer_t>();
    } else {
        throw std::runtime_error("to_json not implemented for this type of object: " + obj.cast<std::string>());
        throw std::runtime_error("to_json not implemented for this type of object: " + std::string(py::str(obj.get_type())));
    }
}
```

ereastman: h/t done

PR #698, opened 2020-04-14
Update transpilation pass interface in controller

### LINKED ISSUE #681: Wrong result when using QasmSimulator with matrix_product_state method
### Informations

- **Qiskit Aer version**: 0.4.0
- **Python version**: 3.7.5
- **Operating system**: Windows

When running the sample code in [Advanced Tutorials](https://qiskit.org/documentation/tutorials/advanced/aer/7_matrix_product_state_method.html)
The result has to be
{'00000000000000000000000000000000000000000000000000': 496,
 '11111111111111111111111111111111111111111111111111': 528}

However, I got
{'00000000000000000000000000000000000000000000000000': 513,
 '00000000000000000000000000000000000000000000000011': 511}

Just use the sample code as it is.

chriseclectic:
### Summary

* Removes the single list `optimize_circuits` from the base Controller class so that transpile passes must be called individually in controllers. This is important as certain passes (such as truncate qubits or fusion) should be called at different points of execution.
* Adds a `transpile_fusion` method to QasmController that returns a fusion pass that may be configured differently depending on the simulation method, circuit, and config. Currently it is set to use the same default values for all statevector and density matrix methods, and to disable fusion for stabilizer, mps and extended stabilizer methods.

### Details and comments

Fixes #681. Fusion was causing problems with MPS simulation method so this disables it entirely for this method. This could be re-enabled if benchmarks or config settings are found where it becomes favorable to use fusion.

### DIFF
--- a/src/controllers/qasm_controller.hpp
+++ b/src/controllers/qasm_controller.hpp
@@ -113,7 +113,7 @@ class QasmController : public Base::Controller {
   //-----------------------------------------------------------------------
   // Constructor
   //-----------------------------------------------------------------------
-  QasmController();
+  QasmController() = default;
 
   //-----------------------------------------------------------------------
   // Base class config override
@@ -183,6 +183,12 @@ class QasmController : public Base::Controller {
   virtual void set_parallelization_circuit(
       const Circuit &circ, const Noise::NoiseModel &noise) override;
 
+  // Return a fusion transpilation pass configured for the current
+  // method, circuit and config
+  Transpile::Fusion transpile_fusion(Method method,
+                                     const Circuit &circ,
+                                     const json_t &config) const;
+
   //----------------------------------------------------------------
   // Run circuit helpers
   //----------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/controllers/qasm_controller.hpp
```diff
@@ -839,35 +854,56 @@ void QasmController::run_single_shot(const Circuit &circ, State_t &state,
...
+
+  // Apply delay measure transpilation pass
+  Transpile::DelayMeasure measure_pass;
+  measure_pass.set_config(config);
+  measure_pass.optimize_circuit(opt_circ, dummy_noise, state_opset, data);
+
+  // Apply fusion transpilation pass
+  auto fusion_pass = transpile_fusion(method, circ, config);
+  fusion_pass.optimize_circuit(opt_circ, dummy_noise, state_opset, data);
+
```
vvilpas, line 906: This is just a comment. If the number of posible transpile optimizations grows I'd create an optimization composite where you add optimizations and just call optimize_circuit on that object.

src/controllers/qasm_controller.hpp
```diff
@@ -730,6 +726,28 @@ size_t QasmController::required_memory_mb(
   }
 }
 
+Transpile::Fusion QasmController::transpile_fusion(Method method,
+                                                   const Circuit &circ,
```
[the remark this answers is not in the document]
chriseclectic: Originally when I added this I was checking the number of qubits in the circuit, but ended up leaving that to the fusion pass itself. In principle you might want to decide the fusion pass parameters based on the circuit, but this can be removed for now an added back later if needed.

src/controllers/qasm_controller.hpp
```diff
@@ -839,35 +854,56 @@ void QasmController::run_single_shot(const Circuit &circ, State_t &state,
...
                                             const Initstate_t &initial_state,
+                                            const Method method,
                                             ExperimentData &data,
                                             RngEngine &rng) const {
-  // Sample a new noise circuit and optimize for each shot
+  // TODO: add opset to State class
+  Operations::OpSet state_opset;
+  state_opset.optypes = state.allowed_ops();
+  state_opset.gates = state.allowed_gates();
+  state_opset.snapshots = state.allowed_snapshots();
```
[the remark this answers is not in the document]
chriseclectic: This will be removed after the OpSet PR and replaced with `state.opset`

src/controllers/qasm_controller.hpp
chriseclectic: Thats what we originally had and this PR removes :)
The problem was you needed to call passes at different points in execution, and different passes if there was noise or no noise. 

Maybe we could add a separate class that can collect optimization passes rather than adding a single one to the controller like we had before, then you can build separate collections to run at different points in the controller.

vvilpas: Yeah, that's what I meant. Create an optimizer composite with the same interface as an optimizer, except that you can add optimizers to it and you just call `optimize_circuit` on this object. But, as you say, you can create them at any point and have different ones at different points of execution, is not a list for the controller. Something like this:

```
OptimizerComposite optimizations;
optimizations.add(optimizer1)
optimizations.add(optimizer2)
...
optimizations.optimize_circuit(opt_circ, dummy_noise, state_opset, data);
```
At this point, there is only one or two optimizers in this part of the code, so its ok. Just something to have in mind.

chriseclectic: Add it to the todo list!

PR #702, opened 2020-04-16
Refactor OpSet class and add OpSet to base State class

chriseclectic:
### Summary

* Adds an OpSet to the base State class that is set by the class constructor.
* Updates State subclasses to define opset via State constructor
* Removes `allowed_ops`, `allowed_gates`, `allowed_snapshots` methods from base State class as they are superseded by the opset.

### Details and comments

On-hold until #698 is merged


chriseclectic:
PR #698 should probably be merged before this PR so it can be backported to stable

PR #712, opened 2020-04-24
Density matrix snapshot not returned as Numpy array

### LINKED ISSUE #709: Density matrix snapshot not returned as Numpy array
### Informations

- **Qiskit Aer version**: 5.0
- **Python version**: All
- **Operating system**: All

When using the density matrix snapshot function the snapshot value should be returned as a complex numpy array matrix. However it instead returned in JSON nested list format.

Trying to fix this by explicitly setting the C++ type to matrix instead of using the built in JSON serialized for the class leads to the Pybind error: `TypeError: Unregistered type : matrix<std::__1::complex<double> >`

The snapshot code [here](https://github.com/Qiskit/qiskit-aer/blob/master/src/simulators/density_matrix/densitymatrix_state.hpp#L440L446) uses the default state serlizier which converts to JSON.

Changing it to the follow should store a matrix in the result instead of a JSON, however doing so leads to the above Pybind type error:

```C++
case Snapshots::densitymatrix: {
      data.add_average_snapshot("density_matrix",
                                op.string_params[0],
                                BaseState::creg_.memory_hex(),
                                BaseState::qreg_.matrix(),
                                false);
      break;
    }
```

Return density matrix snapshots as numpy arrays

### Suggested solutions

Fix Pybind handling of complex matrices in snapshots

ereastman:
### Summary

Fixes #709 - numpy arrays are returned correctly (although probably not perfectly efficiently). 
Since we can't register a custom matrix class and I couldn't get the custom type_caster to work correctly (see below) I had to overload the from_avg_data function so that when a matrix<T> is stored in an avg_data instance it gets converted through my own function (array_from_matrix) instead of through pybind's machinery. 
Basically, I ended up sticking with the existing approach although a custom type caster can probably be used - see [here](https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html) h/t @atilag .
In my attempts (both using pybind's numpy interface and the raw Numpy C-api) I wasn't able to get this to work perfectly - I end up with unexplainable (at least to me) BusErrors (on MacOS) - either upon casting or upon accessing the array (printing) in python (after being casted and returned). Although in principle it is likely a more general solution than the hacky manual result-by-result conversion I ended up using (and that we have been using since we started using JSON).

A detailed description (rant) on the problem is below.

### Details and comments

Pybind is capable of relatively efficiently casting (copying) many common STL containers (vectors, maps, lists, complex, etc.) into their corresponding python structures (lists, dicts, lists, complex, etc.). It also does this recursively - so a vector of maps will be automatically converted into a list of dicts. At the interface it allows us to return vectors from C++ and have them automatically converted to lists - we are not currently using it this way because all of the results are stored in custom classes (even though each class is really just a thin wrapper around some container, either a vector or map mostly). 
Instead we use it in a different way - essentially the manual way. Creating a python container on the c++ side is as simple as:
```
py::list l
```
then, appending arbitrary python-castable objects is as easy as:
```
std::map<string, string> m;
m["hello"] = "goodbye";
l.append(m);
```
We just used pybind's automatic implicit conversion from a std::map to a dict when we appended the map to the py::list l. This is used throughout the from_* functions in the pybind_json.hpp file. However, it was manual in that we built up the list as a python list - instead we could've built up the list as a vector and then converted everything at once...
```
std::vector<std::map<string,string>> l;
l.push_back(m);
py::list pylist(l);
```
accomplishes the exact same thing however the first part involves zero pybind. This is obviously weaker than the first because of c++'s typing. We can't define a container which can hold any type (which is what a py::list essentially is) RIGHT? Wrong - [we can](https://www.youtube.com/watch?v=QGcVXgEVMJg). The pattern in the video is actually stronger than pybind - Pybind's implicit casting does not work with custom types (such as matrix<T> or Result) because how would it know which python type to map to. For example:
```
py::list l;
l.append(matrix<double>());
```
will throw a pybind error - type not recognized - UNLESS one registers the matrix<double> type in pybind as I did previously in this PR. The problem with this approach for Results is that I don't actually want to RETURN a matrix instance anyway - I want to CONVERT the matrix<double> into a numpy array of np.complex64 (because terra and others probably use ```if isinstance(np.array)``` somewhere - and I'm not saying they shouldn't). 

Thus, pybind gives us the 'advanced' custom type_caster<> - which should allow us to register custom conversion functions for when pybind needs to cast a matrix<double/std::complex> - unfortunately I wasn't able to get this to work perfectly - I end up with unexplainable (at least to me) BusErrors (on MacOS) - either upon casting or upon accessing the array (printing) in python (after being casted and returned).
Thus, we (I) am currently between a rock and a hard-place - All of the simulators results are stored in Aer-specific classes which I can't register for automatic conversion because then Terra would complain about them not being just lists/dicts. And I unfortunately cannot register a custom type_caster because I just haven't been able to work out the return_value_policy semantics (I think). 
To get around this - I do all conversion manually (which is exactly how it was done with JSON - I essentially copied and pasted each to_json function). I make assumptions about how each result and nested result class is structured - and the types allowed in each (even though they are all general templates). This leads to many bad things happening - poor errors when things aren't right, very little help from the compiler, and a MESS of for loops and if statements at the interface. 

All of this could be avoided I think if instead of a custom Result class heirarchy, we instead do something like:
```
using Result = std::vector<model_t>;
using ExperimentResult = std::map<std::string, model_t>;
```
The model_t type is described in the video I link to above. This will allow ANY type of result to be added to the result / experiment result with the ONLY restriction being that it must support a certain set of operations - the most important for me at least is that it is convertible to a python object - perhaps for the standalone simulator we add a print function or a to_json function. WRT returning to Python, since pybind supports a lot out of the box we get basically any STL container for free and the only thing we have to worry about is when we want to go to a numpy array (from some custom container such as matrix<double>/qubitvector).


test/terra/extensions/test_snapshot_all.py
```diff
@@ -0,0 +1,54 @@
...
+from qiskit.providers.aer import QasmSimulator
+from qiskit.providers.aer import UnitarySimulator
+from qiskit.providers.aer.extensions import Snapshot
+
+from qiskit.providers.aer.extensions.snapshot_density_matrix import SnapshotDensityMatrix
+from qiskit.providers.aer.extensions.snapshot_expectation_value import SnapshotExpectationValue
+from qiskit.providers.aer.extensions.snapshot_probabilities import SnapshotProbabilities
+import qiskit.quantum_info as qi
+
+class TestSnapshotsExtension(unittest.TestCase):
```
chriseclectic, line 27: The `test.terra.extensions` directory is for tests of the snapshot instruction objects themselves (no execution).

There are already density matrix snapshot tests in `test/terra/backends/qasm_simulator/qasm_snapshot.QasmSnapshotDensityMatrixTests`. Any additional tests should be added there so they are run automatically on any qasm method that supports it (and those tests might need to be updated for these changes).

The unitary snapshot test should be added to the `test_unitary_simulator` tests.

ereastman:
All the fails right now are due to the random_unitary_gate tests (numpy RNG changes) - should be fixed with #727

test/terra/extensions/test_snapshot_all.py
ereastman: Addressed by 53a8b19

src/framework/pybind_json.hpp
```diff
@@ -374,6 +405,25 @@ py::object AerToPy::from_avg_data(const AER::AverageData<T> &avg_data) {
...
 }
 
+template<typename T> 
+py::object AerToPy::from_avg_data(const AER::AverageData<matrix<T>> &avg_data) {
+  py::dict d;
+  d["value"] = AerToPy::array_from_matrix(avg_data.mean());
+  if (avg_data.has_variance()) {
+    d["variance"] = AerToPy::array_from_matrix(avg_data.variance());
+  }
+  return d;
```
atilag, line 415: I think `std::move()` here will avoid a copy as the return value of the function and `py::dict` are different objects so he can't figure out how to do RVO (Return Value Optimization)

src/framework/pybind_json.hpp
```diff
@@ -374,6 +405,25 @@ py::object AerToPy::from_avg_data(const AER::AverageData<T> &avg_data) {
...
+}
+
+template<typename T> 
+py::object AerToPy::from_avg_data(const AER::AverageData<std::vector<T>> &avg_data) {
+  py::dict d;
+  d["value"] = AerToPy::array_from_vector(avg_data.mean());
+  if (avg_data.has_variance()) {
+    d["variance"] = AerToPy::array_from_vector(avg_data.variance());
+  }
+  return d;
```
atilag, line 425: `std::move()` ?

src/framework/pybind_json.hpp
```diff
@@ -384,146 +434,159 @@ py::object AerToPy::from_avg_snap(const AER::AverageSnapshot<T> &avg_snap) {
...
+
     if ( py::len(snapshots) != 0 )
-        pyresult["snapshots"] = snapshots;
+        pydata["snapshots"] = snapshots;
   }
-  //for (auto item : pyresult)
+  //for (auto item : pydatum)
   //  py::print("    {}:, {}"_s.format(item.first, item.second));
-  return std::move(pyresult);
+  return pydata;
```
atilag, line 556: `std::move()` ?

src/framework/pybind_json.hpp
```diff
@@ -384,146 +434,159 @@ py::object AerToPy::from_avg_snap(const AER::AverageSnapshot<T> &avg_snap) {
...
 
-  pyresult["data"] = AerToPy::from_exp_data(result.data);
+  pyexperiment["data"] = AerToPy::from_data(result.data);
 
-  pyresult["success"] = (result.status == AER::ExperimentResult::Status::completed);
+  pyexperiment["success"] = (result.status == AER::ExperimentResult::Status::completed);
   switch (result.status) {
     case AER::ExperimentResult::Status::completed:
-      pyresult["status"] = std::string("DONE");
+      pyexperiment["status"] = std::string("DONE");
```
atilag, line 570: in C++14 we have this beauty syntactic sugar construction: `"this is a std::string"s`

src/framework/pybind_json.hpp
```diff
@@ -384,146 +434,159 @@ py::object AerToPy::from_avg_snap(const AER::AverageSnapshot<T> &avg_snap) {
...
       break;
     case AER::ExperimentResult::Status::error:
-      pyresult["status"] = std::string("ERROR: ") + result.message;
+      pyexperiment["status"] = std::string("ERROR: ") + result.message;
```
atilag, line 573: `"literal std::string"s`

src/framework/pybind_json.hpp
```diff
@@ -384,146 +434,159 @@ py::object AerToPy::from_avg_snap(const AER::AverageSnapshot<T> &avg_snap) {
...
   }
   if (result.metadata.empty() == false) {
     py::object tmp;
     from_json(result.metadata, tmp);
-    pyresult["metadata"] = tmp;
+    pyexperiment["metadata"] = tmp;
   }
-  return std::move(pyresult);
-
+  return pyexperiment;
```
atilag, line 589: `std::move()` ?

src/framework/pybind_json.hpp
```diff
@@ -384,146 +434,159 @@ py::object AerToPy::from_avg_snap(const AER::AverageSnapshot<T> &avg_snap) {
...
-      pyresult["status"] = std::string("DONE");
+      pyexperiment["status"] = std::string("DONE");
       break;
     case AER::ExperimentResult::Status::error:
-      pyresult["status"] = std::string("ERROR: ") + result.message;
+      pyexperiment["status"] = std::string("ERROR: ") + result.message;
       break;
     case AER::ExperimentResult::Status::empty:
-      pyresult["status"] = std::string("EMPTY");
+      pyexperiment["status"] = std::string("EMPTY");
```
atilag, line 576: `"literal std::string"s`

test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -49,13 +49,14 @@ class QasmSnapshotStatevectorTests:
     ]
     BACKEND_OPTS = {}
 
-    @staticmethod
-    def statevector_snapshots(data, label):
+    #@staticmethod
```
atilag, line 52: Just remove the line, don't comment code.

atilag:
@ereastman good work here.
Do you happen to have the code of the pybind custom caster? I may give a try and see where this   BusErrors comes from.
In the meantime, I'm ok with merging using this approach if this fixes the problem (it does) and try to optimize it later.

src/framework/pybind_json.hpp
```diff
@@ -347,21 +372,27 @@ void std::from_json(const json_t &js, py::object &o) {
...
-    for (size_t c = 0; c < cols; c++)
-      mrow.emplace_back(mat(r, c));
-    tbr.emplace_back(mrow);
-  }
-  return py::cast(tbr);
+template<typename T>
+py::object AerToPy::array_from_vector(const std::vector<T> &src) {
+    std::array<py::ssize_t, 1> shape { static_cast<py::ssize_t>(src.size()) };
+    auto tbr = py::array_t<T, py::array::f_style>(shape);
+    auto buf = tbr.template mutable_unchecked<1>();
```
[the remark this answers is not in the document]
ereastman: yeah this line is weird - it's the one line that took hours to get right...
My first instinct was tbr.mutable_unchecked<T, 1>(); which didn't work... I probably also checked MANY other approaches but this seemed to work well...
The pybind array accessors are some very interesting template constructs... 10/10 would [recommend](https://github.com/pybind/pybind11/blob/3b1dbebabc801c9cf6f0953a4c20b904d444f879/include/pybind11/numpy.h#L411)

src/framework/pybind_json.hpp
```diff
@@ -347,21 +372,27 @@ void std::from_json(const json_t &js, py::object &o) {
...
+    for (size_t i = 0; i < src.size(); i++) {
```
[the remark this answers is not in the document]
ereastman: This definitely wouldn't work - as you mentioned once the vector is destroyed it will free the memory and we'll get a memory fault - btw I started messing around with the vector.data() function here - and I found out recently it does not in fact work (for different reasons) - my first instinct when finding .data() was that it was a bad idea to use it but I ignored that and now I'm rewriting a lot of code... don't repeat my mistake. 
vector is meant to abstract away the memory management when appending/removing elements - it does this by intelligently allocating more memory than necessary (in order to avoid reallocating on every append) but a consequence is that the capacity is often much larger than the number of elements... np.arrays wouldn't like that I don't think.

ereastman:
> @ereastman good work here.
> Do you happen to have the code of the pybind custom caster? I may give a try and see where this BusErrors comes from.
> In the meantime, I'm ok with merging using this approach if this fixes the problem (it does) and try to optimize it later.

This is basically what I was working with...
```
namespace pybind11 { namespace detail {
    template <> struct type_caster<matrix<double>> {
    public:
        PYBIND11_TYPE_CASTER(matrix<double>, _("matrix_d"));

        bool load(handle src, bool) {
             // we don't need Python->C++ conversion
             return false;
        }

        static handle cast(inty src, return_value_policy /* policy */, handle /* parent */) {
            // This part is the important part - the function signature itself is a bit complicated
            // I'm fairly sure the return_value_policy and parent handle don't matter
            // but getting to know what they are might help
            //  https://pybind11.readthedocs.io/en/master/advanced/functions.html#return-value-policies
             return PyLong_FromLong(src.long_value);
        }
    };
}} // namespace pybind11::detail
```

I had implemented the cast function a few different ways - using the [raw c NumPy API](https://numpy.org/doc/stable/reference/c-api/array.html#from-scratch) - and/or our array_from_vector function (which intuitively should work but apparently does not...)

src/framework/pybind_json.hpp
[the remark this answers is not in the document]
atilag: Ok yeah, makes sense.

src/framework/pybind_json.hpp
ereastman: you're right on all

src/framework/pybind_json.hpp
ereastman: ha - good catch - I think in the early days I was having trouble with the templating and at one point it was not able to recognize the old const "DONE" char literals as strings which is why I put the manual ctor call in - I'll take out the ctor but I'm hesitant to use the 14-specific feature, just for backwards compatibility (even though it is probably more efficient in some cases)

PR #718, opened 2020-04-27
Add subsystem density matrix snapshots to qasm simulator

chriseclectic:
### Summary

~Blocked by #712~

* Adds support for the density matrix snapshot to return the reduced density matrix of specified qubits
* Adds density matrix snapshot support to the statevector method, but returning density matrix sampled from the average of statevectors for each shots

### Details and comments

- [x] Include subsystem density matrix snapshot tests
- [x] Include statevector method density matrix snapshot tests
- [x] Add release note

### DIFF
--- a/src/simulators/density_matrix/densitymatrix_state.hpp
+++ b/src/simulators/density_matrix/densitymatrix_state.hpp
@@ -549,7 +549,7 @@ void State<densmat_t>::snapshot_density_matrix(const Operations::Op &op,
   data.add_average_snapshot("density_matrix",
                             op.string_params[0],
                             BaseState::creg_.memory_hex(),
-                            reduced_state,
+                            std::move(reduced_state),
                             false);
 }
 
--- a/src/simulators/density_matrix/densitymatrix_state.hpp
+++ b/src/simulators/density_matrix/densitymatrix_state.hpp
@@ -21,6 +21,7 @@
 
 #include "framework/utils.hpp"
 #include "framework/json.hpp"
+#include "framework/opset.hpp"
 #include "simulators/state.hpp"
 #include "densitymatrix.hpp"
 #ifdef AER_THRUST_SUPPORTED
@@ -198,6 +199,10 @@ protected:
   // should be left in the pre-snapshot state.
   //-----------------------------------------------------------------------
 
+  // Snapshot reduced density matrix
+  void snapshot_density_matrix(const Operations::Op &op,
+                               ExperimentData &data);
+                          
   // Snapshot current qubit probabilities for a measurement (average)
   void snapshot_probabilities(const Operations::Op &op,
                               ExperimentData &data,
@@ -437,11 +442,7 @@ void State<densmat_t>::apply_snapshot(const Operations::Op &op,
                                 op.name + "\'.");
   switch (it -> second) {
     case Snapshots::densitymatrix:
-      data.add_average_snapshot("density_matrix",
-                                op.string_params[0],
-                                BaseState::creg_.memory_hex(),
-                                BaseState::qreg_.matrix(),
-                                false);
+      snapshot_density_matrix(op, data);
       break;
     case Snapshots::cmemory:
       BaseState::snapshot_creg_memory(op, data);
@@ -516,6 +517,54 @@ void State<densmat_t>::snapshot_pauli_expval(const Operations::Op &op,
                             op.string_params[0],
                             BaseState::creg_.memory_hex(),
                             expval, variance);
+}
+
+
+template <class densmat_t>
+void State<densmat_t>::snapshot_density_matrix(const Operations::Op &op,
+                                               ExperimentData &data) {
+  cmatrix_t reduced_state;
+
+  // Check if tracing over all qubits
+  if (op.qubits.empty()) {
+    reduced_state = cmatrix_t(1, 1);
+    reduced_state[0] = BaseState::qreg_.trace();
+  } else {
+    const size_t N = op.qubits.size();
+    const size_t DIM = 1ULL << N;
+    reduced_state = cmatrix_t(DIM, DIM);
+    auto qubits_sorted = op.qubits;
+    std::sort(qubits_sorted.begin(), qubits_sorted.end());
+
+    // Return full density matrix
+    if ((N == BaseState::qreg_.num_qubits()) && (op.qubits == qubits_sorted)) {
+      reduced_state = BaseState::qreg_.matrix();
+    } else {
+      // Get superoperator qubits
+      const reg_t squbits = BaseState::qreg_.superop_qubits(op.qubits);
+      const reg_t squbits_sorted = BaseState::qreg_.superop_qubits(qubits_sorted);
+
+      // Return density matrix
+      const size_t VDIM = 1ULL << (2 * N);
+      const size_t END = 1ULL << (BaseState::qreg_.num_qubits() - N);  
+      const size_t SHIFT = END + 1;
+
+      // This is inefficient but we need to copy data to vector
+      // So we don't get errors with the Thrust / GPU backend
+      auto vmat = BaseState::qreg_.vector();
+      for (size_t k = 0; k < END; k++) {
+        const auto inds = QV::indexes(squbits, squbits_sorted, k * SHIFT);
+        for (int_t i = 0; i < VDIM; ++i) {
+          reduced_state[i] += complex_t(std::move(vmat[inds[i]]));
+        }
+      }
+    }
+  }
+  data.add_average_snapshot("density_matrix",
+                            op.string_params[0],
+                            BaseState::creg_.memory_hex(),
+                            std::move(reduced_state),
+                            false);
 }
 
 //=========================================================================
--- a/src/simulators/density_matrix/densitymatrix_state.hpp
+++ b/src/simulators/density_matrix/densitymatrix_state.hpp
@@ -549,7 +549,7 @@ void State<densmat_t>::snapshot_density_matrix(const Operations::Op &op,
   data.add_average_snapshot("density_matrix",
                             op.string_params[0],
                             BaseState::creg_.memory_hex(),
-                            std::move(reduced_state),
+                            reduced_state,
                             false);
 }
 
--- a/src/simulators/density_matrix/densitymatrix_state.hpp
+++ b/src/simulators/density_matrix/densitymatrix_state.hpp
@@ -218,6 +218,11 @@ protected:
                               ExperimentData &data,
                               bool variance);
 
+  // Return the reduced density matrix for the simulator
+  cmatrix_t reduced_density_matrix(const reg_t &qubits, const reg_t& qubits_sorted);
+  cmatrix_t reduced_density_matrix_cpu(const reg_t &qubits, const reg_t& qubits_sorted);
+  cmatrix_t reduced_density_matrix_thrust(const reg_t &qubits, const reg_t& qubits_sorted);
+
   //-----------------------------------------------------------------------
   // Single-qubit gate helpers
   //-----------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


src/simulators/statevector/statevector_state.hpp
```diff
@@ -634,6 +655,84 @@ void State<statevec_t>::snapshot_matrix_expval(const Operations::Op &op,
...
+  return vec2density(qubits, BaseState::qreg_.data());
+}
+template <>
+cmatrix_t State<QV::QubitVector<float>>::density_matrix(const reg_t &qubits) {
+  return vec2density(qubits, BaseState::qreg_.data());
+}
+
+template <class statevec_t>
+template <class T>
+cmatrix_t State<statevec_t>::vec2density(const reg_t &qubits, T &&vec) {
```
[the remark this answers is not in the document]
atilag: Ok, I see why you are using rvalue reference: Because of the `vector()` method that transforms a pointer to a vector into a vector... by copying everything. If we find out that using `data()` instead of `vector()` is an option, we have to change the signature to get a reference instead.

test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -348,154 +350,52 @@ class QasmSnapshotDensityMatrixTests:
...
+                    value = qi.DensityMatrix(snapshots[label][0]['value'])
+                    target = qi.DensityMatrix.from_label(3 * '0').evolve(circ, equbits)
+                    if num_qubits == 2:
+                        target = qi.partial_trace(target, [2])
+                    elif num_qubits == 1:
+                        target = qi.partial_trace(target, [1, 2])
+                    if value != target:
+                        print('FAILED:', squbits, equbits)
+                        print(value)
+                        print(target)
```
atilag, line 397: Use logging instead of print.

src/simulators/density_matrix/densitymatrix_state.hpp
```diff
@@ -518,6 +524,113 @@ void State<densmat_t>::snapshot_pauli_expval(const Operations::Op &op,
...
+  if (op.qubits.empty()) {
+    reduced_state = cmatrix_t(1, 1);
+    reduced_state[0] = BaseState::qreg_.trace();
+  } else {
+
+    auto qubits_sorted = op.qubits;
+    std::sort(qubits_sorted.begin(), qubits_sorted.end());
+
+    if ((op.qubits.size() == BaseState::qreg_.num_qubits()) && (op.qubits == qubits_sorted)) {
+      reduced_state = std::move(BaseState::qreg_.matrix());
```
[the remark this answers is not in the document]
chriseclectic: `qreg_.matrix` returns a copy, not a reference so there this isn't an issue with underfined state being left behind. Maybe move is unnecessary here like you mention below because of RVO

src/simulators/density_matrix/densitymatrix_state.hpp
```diff
@@ -518,6 +524,113 @@ void State<densmat_t>::snapshot_pauli_expval(const Operations::Op &op,
...
+    } else {
+      reduced_state = std::move(reduced_density_matrix(op.qubits, qubits_sorted));
+    }
+  }
+
+  data.add_average_snapshot("density_matrix",
+                            op.string_params[0],
+                            BaseState::creg_.memory_hex(),
+                            std::move(reduced_state),
```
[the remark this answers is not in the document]
chriseclectic: The data container has special functions for move semantics, so passing the std move here is needed to use that. Otherwise reduced_state will be copied (and it is potentially a very large matrix).

src/simulators/density_matrix/densitymatrix_state.hpp
```diff
@@ -518,6 +524,113 @@ void State<densmat_t>::snapshot_pauli_expval(const Operations::Op &op,
...
+template <>
+cmatrix_t State<QV::DensityMatrix<double>>::reduced_density_matrix(const reg_t &qubits,
+                                                                   const reg_t &qubits_sorted) {
+  return reduced_density_matrix_cpu(qubits, qubits_sorted);
+}
+template <>
+cmatrix_t State<QV::DensityMatrix<float>>::reduced_density_matrix(const reg_t &qubits,
+                                                                  const reg_t &qubits_sorted) {
+  return reduced_density_matrix_cpu(qubits, qubits_sorted);
+}
```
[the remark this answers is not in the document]
chriseclectic: I thought you couldn't do partial template specialization like this?

src/simulators/density_matrix/densitymatrix_state.hpp
```diff
@@ -518,6 +524,113 @@ void State<densmat_t>::snapshot_pauli_expval(const Operations::Op &op,
...
+
+  // Get dimensions
+  const size_t N = qubits.size();
+  const size_t DIM = 1ULL << N;
+  const size_t VDIM = 1ULL << (2 * N);
+  const size_t END = 1ULL << (BaseState::qreg_.num_qubits() - N);  
+  const size_t SHIFT = END + 1;
+
+  // Copy vector to host memory
+  auto vmat = BaseState::qreg_.vector();
```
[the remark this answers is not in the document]
chriseclectic: The Thrust backend returns an empty array from `data()` as the state is hosted on GPU device memory and needs to be copied back to host memory, which is what the `vector` function does.

This is why I had to go through the headache of template specialization and custom classes for the Thrust and non-Thrust case.

src/simulators/statevector/statevector_state.hpp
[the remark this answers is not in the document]
chriseclectic: For CPU case it uses `data`, for Thrust/GPU case it uses `vector`, which is why I passed by rvalue

test/terra/backends/qasm_simulator/qasm_snapshot.py
chriseclectic: These were for debugging, should have been removed

src/simulators/density_matrix/densitymatrix_state.hpp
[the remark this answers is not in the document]
atilag: My point is that we don't need move semantics here because there's no need for transfer ownership and using the overloaded `add_average_snapshot` method that passes `reduced_state` as a const reference would avoid the copy and will be even faster than casting to rvalue.

src/simulators/statevector/statevector_state.hpp
```diff
@@ -634,6 +655,82 @@ void State<statevec_t>::snapshot_matrix_expval(const Operations::Op &op,
...
+    reduced_state[0] = BaseState::qreg_.norm();
+  } else {
+    reduced_state = density_matrix(op.qubits);
+  }
+
+  // Add density matrix to result data
+  switch (type) {
+    case SnapshotDataType::average:
+      data.add_average_snapshot("density_matrix", op.string_params[0],
+                            BaseState::creg_.memory_hex(), std::move(reduced_state), false);
```
atilag, line 676: Same thing here, we don't really need move semantics for `reduced_state`, passing a reference would be fast and clear.

src/simulators/statevector/statevector_state.hpp
```diff
@@ -634,6 +655,82 @@ void State<statevec_t>::snapshot_matrix_expval(const Operations::Op &op,
...
+      break;
+    case SnapshotDataType::average_var:
+      data.add_average_snapshot("density_matrix", op.string_params[0],
+                            BaseState::creg_.memory_hex(), std::move(reduced_state), true);
```
atilag, line 680: `std::move()` -> `reduced_state` (by reference)

src/simulators/statevector/statevector_state.hpp
```diff
@@ -634,6 +655,82 @@ void State<statevec_t>::snapshot_matrix_expval(const Operations::Op &op,
...
+      break;
+    case SnapshotDataType::pershot:
+      data.add_pershot_snapshot("density_matrix", op.string_params[0], std::move(reduced_state));
```
atilag, line 683: `std::move()` -> `reduced_state` (by reference)

src/simulators/statevector/statevector_state.hpp
```diff
@@ -634,6 +655,82 @@ void State<statevec_t>::snapshot_matrix_expval(const Operations::Op &op,
...
+#ifdef AER_THRUST_SUPPORTED
+template <>
+cmatrix_t State<QV::QubitVectorThrust<float>>::density_matrix(const reg_t &qubits) {
+  return vec2density(qubits, BaseState::qreg_.vector());
+}
+
+template <>
+cmatrix_t State<QV::QubitVectorThrust<double>>::density_matrix(const reg_t &qubits) {
+  return vec2density(qubits, BaseState::qreg_.vector());
+}
```
atilag, line 702:
```c++
template <typename floating_t>
cmatrix_t State<QV::QubitVectorThrust<floating_t>>::density_matrix(const reg_t &qubits) {
  return vec2density(qubits, BaseState::qreg_.vector());
}
```
We can do this because both specializations have the same implementation, so the compiler will derive using `monomorphization` (specialize `double` and `float` if they are used in the code at compile time).

src/simulators/statevector/statevector_state.hpp
```diff
@@ -634,6 +655,82 @@ void State<statevec_t>::snapshot_matrix_expval(const Operations::Op &op,
...
+  const size_t N = qubits.size();
+  const size_t DIM = 1ULL << N;
+  auto qubits_sorted = qubits;
+  std::sort(qubits_sorted.begin(), qubits_sorted.end());
+
+  // Return full density matrix
+  cmatrix_t densmat(DIM, DIM);
+  if ((N == BaseState::qreg_.num_qubits()) && (qubits == qubits_sorted)) {
+    
+    for (size_t row = 0; row < DIM; ++row)
```
atilag, line 717: I think we could use OpenMP pragmas here to parallelize these loops.

src/simulators/statevector/statevector_state.hpp
```diff
@@ -634,6 +655,82 @@ void State<statevec_t>::snapshot_matrix_expval(const Operations::Op &op,
...
+      for (size_t col = 0; col < DIM; ++col) {
+        densmat(row, col) = complex_t(vec[row]) * complex_t(std::conj(vec[col]));
+      }
+  } else {
+    const size_t END = 1ULL << (BaseState::qreg_.num_qubits() - N);
+    for (size_t k = 0; k < END; k++) {
+      // store entries touched by U
+      const auto inds = QV::indexes(qubits, qubits_sorted, k);
+      for (size_t row = 0; row < DIM; ++row)
```
atilag, line 726: And these too

src/simulators/statevector/statevector_state.hpp
chriseclectic: This is a reduction so for one you would have to define a custom reduction type for OMP here, and I think that would result in memory for the matrix being duplicated across threads. For these types of snapshots which can be very large in memory usage I think minimizing memory is more important than speed.

src/simulators/statevector/statevector_state.hpp
chriseclectic: I tried this change but it resulted in compiler errors like: 

```
error: nested name specifier 'State<QV::DensityMatrixThrust<floating_t> >::' for declaration does not refer into a class, class template or class template partial specialization
cmatrix_t State<QV::DensityMatrixThrust<floating_t>>::reduced_density_matrix(...
```

src/simulators/statevector/statevector_state.hpp
chriseclectic: This one is a reduction so for one you would have to define a custom reduction type for OMP here, and I think that would result in memory for the matrix being duplicated across threads. For these types of snapshots which can be very large in memory usage I think minimizing memory is more important than speed.

PR #722, opened 2020-04-28
Fixing subsystem_list handling in pulse simulator

### LINKED ISSUE #719: Restriction to subsystem list in HamiltonianModel.from_dict incorrectly restricting to subsystem_list

When constructing a `HamiltonianModel` from a hamiltonian string, specifying `subsystem_list` is supposed to restrict the Hamiltonian in the string to only those subsystems. Currently, it is correctly restricting the terms in the list (i.e. a Hamiltonian term is only kept if the qubits it applies to is a subset of `subsystem_list`), but the excluded subsystems are not actually being removed from the system dimension description in `subsystem_dims`.

When the full Hamiltonian matrix is constructed, this results in every Hamiltonian term having a tensored identity on subsystems not specified in `subsystem_list`. This was discovered when a 20 qubit hamiltonian was restricted to 2 qubits, but enormous matrices were still being constructed due to the extra factor of `2**18` dimensions.

### Suggested solutions

Hopefully it will be as simple as updating the construction of `subsystem_dims` to only include terms from `subsystem_list` - though it is possible the indices will produce problematic behaviour.

DanPuzzuoli:
### Summary

Closes #719 

1. Corrected subsystem reduction to not include implicit identities on the subsystems not included in `subsystem_list` for `HamiltonianModel.from_dict` when representing `full()` operators. 
2. Also fixed `HamiltonianModel.get_qubit_lo_from_drift()` so that it works when `subsystem_list` is not `range(k)` for some `k`. 

Associated with the above fixes is a slightly awkward interface issue: when calling `assemble`, the frequencies in `qubit_lo_freq` are assumed to correspond to the frequencies of the drive channels with indices `range(len(qubit_lo_freq))`. I.e. the frequency of `DriveChannel(idx)` is always assumed to be `qubit_lo_freq[idx]` regardless of what `subsystem_list` is. This results in needing to pass in "dummy" values for drive channel frequencies of qubits that have been excluded from the model.
    - The fix to `HamiltonianModel.get_qubit_lo_from_drift()` is consistent with this - it now returns a frequency of `0` for any drive channel indices not in `subsystem_list`
    - An alternative is to have `qubit_lo_freq[idx]` correspond to the drive channel of `subsystem_list[idx]`, though this also has the potentially awkward aspect of requiring re-specifying `qubit_lo_freq` for different `subsystem_list`.
For now I'm fine with leaving it as is.

### Details and comments

1. Corrected reduction of `subsystem_dims` in `HamiltonianModel.from_dict` to only include the systems in `subsystem_list`.
2. Corrected subsystem referencing in construction of measurement operators.
3. Added a test for `HamiltonianModel.from_dict` to validate above change.
4. Added an integration test in `test_pulse_simulator` to validate the above change.

### DIFF
--- a/qiskit/providers/aer/pulse/controllers/pulse_controller.py
+++ b/qiskit/providers/aer/pulse/controllers/pulse_controller.py
@@ -116,7 +116,6 @@ def pulse_controller(qobj, system_model, backend_options):
     pulse_sim_desc.global_data['memory_slots'] = digested_qobj.memory_slots
     pulse_sim_desc.global_data['memory'] = digested_qobj.memory
     pulse_sim_desc.global_data['n_registers'] = digested_qobj.n_registers
-
     pulse_sim_desc.global_data['pulse_array'] = digested_qobj.pulse_array
     pulse_sim_desc.global_data['pulse_indices'] = digested_qobj.pulse_indices
     pulse_sim_desc.pulse_to_int = digested_qobj.pulse_to_int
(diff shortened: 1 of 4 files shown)


ntwalibas:
This pull request fixes the referenced issue which I am experiencing and it would be great if the merger happened soon. While it is possible to work from the branch that fixes the issue, company policy makes it unsafe.

PR #734, opened 2020-05-04
Fix stall of pulse simulator execution as described in issue #726

### LINKED ISSUE #726: Pulse simulator execution halting at parallel_map in Jupyter notebook
### Informations

- **Qiskit Aer version**: 0.6.0
- **Python version**: 3.7.6
- **Operating system**: Mac OS

When running current version of pulse simulator in master in a jupyter notebook nothing happens - execution seems to halt/stall at `qiskit.tools.parallel_map`. All unit tests work however.

Run the simulation portion of either pulse simulator tutorial notebook located [here](https://github.com/Qiskit/qiskit-tutorials/tree/master/legacy_tutorials/aer).

vvilpas:
### Summary

Fix pulse simulator execution

### Details and comments
Execution halts/stalls when executing unitary experiments in parallel. Problem is an inner function is sent to be executed in parallel. When we move the function out of the function where parallel is called everything seems to work.
Fixes #726

### DIFF
--- a/qiskit/providers/aer/pulse/controllers/pulse_controller.py
+++ b/qiskit/providers/aer/pulse/controllers/pulse_controller.py
@@ -116,7 +116,6 @@ def pulse_controller(qobj, system_model, backend_options):
     pulse_sim_desc.global_data['memory_slots'] = digested_qobj.memory_slots
     pulse_sim_desc.global_data['memory'] = digested_qobj.memory
     pulse_sim_desc.global_data['n_registers'] = digested_qobj.n_registers
-
     pulse_sim_desc.global_data['pulse_array'] = digested_qobj.pulse_array
     pulse_sim_desc.global_data['pulse_indices'] = digested_qobj.pulse_indices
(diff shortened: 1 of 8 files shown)


DanPuzzuoli:
I've pulled this locally and it is starting to work again, though I am running into another error now when running the tutorial:
```
BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```

This occurs in cell 18 (if all cells are executed in order) - which is the last run of simulations in that notebook.

nonhermitian:
https://github.com/matplotlib/matplotlib/issues/15410#issuecomment-620940403

DanPuzzuoli:
Ah right, thanks @nonhermitian , I forgot about the `matplotlib` issue after it turned out that the original issue was unrelated. Commenting out the import of `matplotlib` (and any lines that use it) results in the notebook running properly.

@vvilpas  I think with the test I've added in the PR to this branch this should be good then.

PR #738, opened 2020-05-07
Unblock ci for new jupyter-sphinx release

mtreinish:
### Summary

The latest jupyter-sphinx release deprecated the plugin path we were
using which will now emit a deprecation warning. However because
warnings are marked as fatal with -W this is breaking the docs builds.
This commit updates are jupyter-sphinx usage to use the new path and fix
the warning/failure.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -65,7 +65,7 @@ extensions = [
     'sphinx.ext.viewcode',
     'sphinx.ext.extlinks',
     'sphinx_tabs.tabs',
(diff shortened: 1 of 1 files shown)


mtreinish:
While this fixes the docs job, everything else is stuck on: https://github.com/Qiskit/qiskit-terra/pull/4412 once that merges this should be good to go and unblock things

mtreinish:
Qiskit/qiskit-terra#4412 has merged and this is now unblocked and ready to merge

PR #742, opened 2020-05-10
Update config access in calculate_channel_frequencies

### LINKED ISSUE #741: Pulse system model incosistently using dict based access
### Information

- **Qiskit Terra version**: master (Aer master and 0.5.1)
- **Python version**: 3.8
- **Operating system**: OSX

There is a pulse simulator mock backend test failure in `FakeBackendsTest.test_transpile_schedule_and_assemble`:
```python
        # Setup freqs for the channels
        freqs = OrderedDict()
        for key in self.hamiltonian._channels:
            chidx = int(key[1:])
            if key[0] == 'D':
                freqs[key] = qubit_lo_freq[chidx]
            elif key[0] == 'U':
                freqs[key] = 0
                for u_lo_idx in self.u_channel_lo[chidx]:
>                   if u_lo_idx['q'] < len(qubit_lo_freq):
E                   TypeError: 'UchannelLO' object is not subscriptable

/opt/miniconda3/lib/python3.8/site-packages/qiskit/providers/aer/pulse/pulse_system_model.py:220: TypeError
```

mtreinish:
### Summary

In #695 we updated the usage of the backend configuration to avoid
to_dict() and dict based access. All the access was supposed to be
updated to be class attribute based instead. However that PR neglected
to update the usage of the calculate_channel_frequencies method. This is
causing an error when using the method now. This commit corrects the
oversight and adds a unit test covering running that method when
building a pulse system model from a backend to avoid this in the
future.

### Details and comments

Fixes #741
Co-authored-by: Daniel Puzzuoli <dan.puzzuoli@gmail.com>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -65,7 +65,7 @@ extensions = [
     'sphinx.ext.viewcode',
     'sphinx.ext.extlinks',
     'sphinx_tabs.tabs',
-    'jupyter_sphinx.execute',
+    'jupyter_sphinx',
     'reno.sphinxext',
 ]
 html_static_path = ['_static']
--- a/qiskit/providers/aer/pulse/controllers/unitary_controller.py
+++ b/qiskit/providers/aer/pulse/controllers/unitary_controller.py
@@ -29,6 +29,40 @@ from ..de_solvers.pulse_utils import occ_probabilities, write_shots_memory
 dznrm2 = get_blas_funcs("znrm2", dtype=np.float64)
 
 
+def _full_simulation(exp, op_system):
+    """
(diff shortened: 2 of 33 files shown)


DanPuzzuoli:
@mtreinish thanks for these updates. 

There are actually a couple of other files that need updating - the `system_models/duffing_model_generators.py` and the corresponding tests are still in the `dict` format. No errors are being raised in the tests as no computations on the frequencies are being called.

I'll update these files and also add a test that would have caught this, then open a PR to your branch.

DanPuzzuoli:
I've just implemented the changes from the previous comments and have opened a PR to this branch @mtreinish

PR #743, opened 2020-05-11
Corrected Find NumPy in src/open_pulse

doichanj:
### Summary
I have corrected format of find NumPy in src/open_pulse/CMakeLists.txt to find correct NumPy installed on Anaconda instead of finding in /usr/include.

### Details and comments
find_package(Python REQUIRED COMPONENTS NumPy) can find NumPy installed by Anaconda and the path is set to Python_NumPy_INCLUDE_DIRS


atilag:
>  have a virtual environment (python3.8 -m venv ~/.virtualenv/test_38) in MacOS and instead of finding the numpy installed there, now it finds the default MacOS version. Any idea how can we preserve the previous behaviour as well?

@doichanj any comment on this?

chriseclectic:
Is there any activity on this? Otherwise I will close.

doichanj:
I still cannot compile without this change on Power environment without having root permission. But I think it is OK to close this.

vvilpas:
I think I might have found a solution. Let's see if CI pass.

vvilpas (approved):
Looks good

PR #744, opened 2020-05-11
Refactor ExperimentData class

chriseclectic:
### Summary

Refactor `ExperimentData` class to better handle custom data types

### Details and comments

Adds templated `DataContainer` class that contains all methods for storing and combining data of a fixed type. There is still some boiler plate code in the `ExperimentData` class to call the methods for each inherited DataContainer type, but a lot less than before.

### DIFF
--- a/src/framework/results/data/data_container.hpp
+++ b/src/framework/results/data/data_container.hpp
@@ -1,0 +1,299 @@
+/**
+ * This code is part of Qiskit.
+ *
+ * (C) Copyright IBM 2018, 2020.
+ *
+ * This code is licensed under the Apache License, Version 2.0. You may
+ * obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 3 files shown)


src/framework/results/data/data_container.hpp
```diff
@@ -0,0 +1,299 @@
...
+  }
+}
+
+template <typename T>
+void DataContainer<T>::add_additional_data(const std::string &key,
+                                           T &data) {
+  if (enabled_) {
+    additional_data_[key] = data;
+  }
+}
```
vvilpas, line 141: Shouldn't this work having only the one accepting the `const T&` (same for `add_pershot_snapshot` and `add_average_snapshot`?  Am I missing something?

chriseclectic: I did some testing and I found i needed to add both `const T&` and `T&`, if one was missing that would get converted via the default json conversion `T&&` method in the derived class

PR #745, opened 2020-05-12
Fixed compilation issues of GPU simulator on Windows with MSVC

doichanj:
### Summary
There were some issue on compiling Thrust simulators with CUDA backends on Windows.
This PR will fix these issues

### Details and comments
(1)
Following compilation error is caused by Windows specific directory naming
nvcc fatal : A single input file is required for a non-link phase when an outputfile is specified
I inserted double quart to specify compiler path of -ccbin option passed to nvcc.

(2)
MSVC does not accept unsigned value to the index of OpenMP iterators. So I replaced indices from uint_t to int_t in unitarymatrix_thrust.hpp

(3)
The combination of MSVC + pybind11 has issue in pybind11/cast.h
I modified cast.h and put in src/thirdparty/headers
I put all the headers of pybind11 in the directory, but I think there is more better way to do this, please someone make this better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,11 +220,15 @@ if(AER_THRUST_SUPPORTED)
 		find_package(CUDA REQUIRED)
 		message(STATUS "Thrust library: CUDA found!")
 		cuda_select_nvcc_arch_flags(AER_CUDA_ARCH Auto)
-		set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -ccbin ${CMAKE_CXX_COMPILER} ${AER_CUDA_ARCH} -DAER_THRUST_CUDA -std=c++14 -I${AER_SIMULATOR_CPP_SRC_DIR} -isystem ${AER_SIMULATOR_CPP_SRC_DIR}/third-party/headers -use_fast_math --expt-extended-lambda")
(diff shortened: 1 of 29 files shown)


atilag:
@doichanj We have recently removed all our in tree dependencies in favor of a package manager that deals with them (Conan).
I can't see what you actually changed in pybind, but this worries me, I would definitely not want to maintain patched versions of any dependency. Could you show me what you changed in pybind? We could try to contribute the upstream project if that is beneficial for everyone.

doichanj:
I think [https://github.com/pybind/pybind11/issues/2180](url) is the same issue but it is not solved in the upstream pybind. 

I replaced 
`    explicit operator type&() { return *(this->value); }`
to
`    explicit operator type&() { return *((type *)(this->value)); }`
in cast.h

vvilpas:
I think if we do the separated compilation of cuda and c++ code, we can avoid the need for a patched pybind11 version.

atilag:
Let's see if the upstream repo for pybind11 fixes the issue: https://github.com/pybind/pybind11/issues/2180

atilag:
Ok, there is progress: https://github.com/pybind/pybind11/pull/2240
If they merge, we can just wait for the new release.

vvilpas:
They merged! We just need the new version to be released.

atilag:
Ok, so I'm closing this and opening a new issue to upgrade pybind once they release.

PR #749, opened 2020-05-21
Avoid use of numpy prerelease candidate 1.19.0rc1

vvilpas:
### Summary

Avoid use of numpy prerelease candidate 1.19.0rc1

### Details and comments

Azure `Compile: Windows_Wheel_Builds Python35` and `Test: Windows Python35` fail because at some point it tries to install Numpy non-stable version 1.19.0rc1, and for some reason this breaks ([[1](https://dev.azure.com/qiskit-ci/qiskit-aer/_build/results?buildId=13867&view=logs&j=e75d9aed-3891-5f4d-4db6-188bb674529f&t=25841d21-5fa7-5cd2-f4f9-be64287d296a)] and [[2](https://dev.azure.com/qiskit-ci/qiskit-aer/_build/results?buildId=13867&view=logs&j=6cffd7fe-8559-5039-bd0f-14cdd5e20947&t=eb2406f7-5076-5344-3510-a196e012e349)]). In the first case it fails in our own `python ./setup.py` and in the second one when it tries to build one of the dependencies (`scs`).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -275,8 +275,8 @@ stages:
               else
                   pip install qiskit-terra
               fi
-              displayName: "Install dependencies"
-              condition: eq(variables['python.version'], '3.5')
+            displayName: "Install dependencies"
+            condition: eq(variables['python.version'], '3.5')
           - bash: |
               set -e
               source activate qiskit-aer-$(Build.BuildNumber)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -268,7 +268,7 @@ stages:
           - bash: |
               set -e
               source activate qiskit-aer-$(Build.BuildNumber)
-              pip install numpy<1.19.0
+              pip install 'numpy<1.19.0'
               pip install -r requirements-dev.txt
               if [[ $SYSTEM_PULLREQUEST_TARGETBRANCH == *"master"* || $BUILD_SOURCEBRANCH == *"master"* ]] ; then
                   pip install git+https://github.com/Qiskit/qiskit-terra.git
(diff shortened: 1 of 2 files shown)


vvilpas:
I have tried different options but SCS setup always tries to install numpy 1.19.0rc1. I have reproduced it on my windows machine, so I'll try to find a solution.

mtreinish:
@atilag I don't think this PR is ready yet, it doesn't pass testing, and we need to rework the requirements lists

vvilpas:
I agree, pip doesn't install pre release versions unless you explicitly say so.
However, `python ./setup.py` doesn't seem to follow this and tries to install the latest version, stable or pre release.  I solved that adding `'numpy>=1.16.3,!=1.19.0rc1',` to our `setup.py`, though I prefer the approach you suggest with the requirement based on the python version.
The other problem appears installing the reqs with `pip install -r requirement-dev.txt`. Here, the installation of `cvxpy` triggers the installation of `scs`. It seems `scs` installation is launching its `setup.py` script, which requires `numpy` (even if `numpy 1.18.4` is already installed), and tries to install the pre-release version.
I started to fix this in another pull-request, using the `constraints.txt file` approach, but it didn't work either ([check here](https://github.com/Qiskit/qiskit-aer/pull/648)).

atilag:
@mtreinish sure, I approved it because I knew it was not related to pip.
But I leave to you both to decide when it's ready :)

vvilpas:
@mtreinish @atilag I have found a workaround installing numpy before calling `pip install -r requirements.txt". I haven't been able to make it work any other way.

vvilpas:
It wasn't before and when I added there I had the same error, so I leave it as it was.

PR #753, opened 2020-05-26
Allow Simulator Seeds Smaller Than Zero

brandhsn:
### Summary

Currently, the simulator seed parameter is ignored if it is smaller than zero. This can lead to intermittent surprises if users choose to set the seed to a negative number e.g. by using python's random.getrandbits(64). The current RNG seems to work with negative seeds.

### DIFF
--- a/src/framework/qobj.hpp
+++ b/src/framework/qobj.hpp
@@ -76,12 +76,11 @@ Qobj::Qobj(const json_t &js) {
   JSON::get_value(header, "header", js);
 
   // Check for fixed simulator seed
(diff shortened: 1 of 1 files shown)


mtreinish:
FWIW, I would assume the RNG is just treating the signed integer as an unsigned int since IIRC all the `std::random` functions only deal with unsigned ints. But, it doesn't really matter since the seed just needs to be a fixed value and how it gets interpreted (negative vs large positive) doesn't really matter as long as it behaves the consistently for the same input.

Issue #754, opened 2020-05-26
Density matrix GPU does not  return correct probability snapshot


we-taper:
### Informations

- **Qiskit Aer version**:
```python
{'qiskit-terra': '0.14.1',
 'qiskit-aer': '0.5.1',
 'qiskit-ignis': '0.3.0',
 'qiskit-ibmq-provider': '0.7.1',
 'qiskit-aqua': '0.7.1',
 'qiskit': '0.19.2'}
```
- **Python version**: 3.6.9
- **Operating system**: Ubuntu 19.04

### What is the current behavior?

The `snapshot_probabilities` on 1 qubit returns un-normalized probability values for a 2-qubit system.

### Steps to reproduce the problem

Example:
```python
import qiskit as qk
import qiskit.providers.aer.extensions
qc = qk.QuantumCircuit(2, 1)
qc.snapshot_probabilities(label='before', qubits=[0])
qc.rz(0, 0)
qc.cz(0, 1)
qc.h(0)
qc.snapshot_probabilities(label='after', qubits=[0])
print(qc.draw(output='text'))
bkd = qk.Aer.get_backend('qasm_simulator')
res = qk.execute(qc, backend=bkd, shots=1, backend_options={"method": "density_matrix_gpu"}).result()
res = res.results[0]
res.data.snapshots.probabilities
```
Output:
```
     ┌────────┐┌───────┐   ┌───┐┌───────┐
q_0: ┤ before ├┤ RZ(0) ├─■─┤ H ├┤ after ├
     └────────┘└───────┘ │ └───┘└───────┘
q_1: ────────────────────■───────────────
                                         
c_0: ════════════════════════════════════
{'before': [{'value': {'0x0': 1.0}, 'memory': '0x0'}],
 'after': [{'value': {'0x0': 4.0, '0x1': 2.0}, 'memory': '0x0'}]}
```

### What is the expected behavior?

Return the correct probability values. Here it should be something like

```
'after': [{'value': {'0x0': 0.5, '0x1': 0.5}, 'memory': '0x0'}]
```

### Suggested solutions

Not sure. Maybe related to #388 ?

PR #760, opened 2020-05-29
Fixed cz gate of DensityMatrixThrust #754

doichanj:
### Summary
I modified cz gate implementation of DensityMatrixThrust class to fix issue #754
I noticed y gate also has issue so I fixed it.

### Details and comments
The matrix to be applied by apply_matrix in cz gate was wrong, diagonal elements should be set to 1 but set to 0.
Also the matrix set in y gate was wrong, I created separate kernel for y gate.


chriseclectic:
Hi @doichanj, can you separate the bug fix from the CMake changes (this also seems to include #743).

Also can you add a release note about the bug fix

chriseclectic:
I added this to PR #765 which fixes the tests which werent being run for density matrix simulator and it looks like a couple are still failing

doichanj:
Sorry I removed cmake files.

Also I fixed cz gate again. Using apply_diagonal_matrix instead of apply_matrix

PR #774, opened 2020-06-01
Reorganize azure pipelines jobs for better throughput

mtreinish:
### Summary

This commit reorganizes the azure pipelines flow to try and optimize it
for better throughput without losing any coverage. The big change here
is the dedicated wheel jobs are combined into a single job that is
exactly how we run the wheel builds at release time. That was the intent
of those jobs and it's good to ensure we don't break those by accident
at release. This is moved to the very end since we don't want to block
test results on that which is mostly testing the CI and automation
framework outside of the coverage the test jobs provide. Then the max
parallelism limits are removed from all jobs. This basically reorganizes
the windows builds into 3 stages:

### Details and comments

compile (sdist py37 and 38, and standalone) -> tests -> wheel builds


chriseclectic:
I think this is superseded by #881

PR #781, opened 2020-06-03
Add GPU build instructions to CONTRIBUTING guide

atilag:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -555,6 +555,31 @@ options we have on `Aer` to CMake we use it's native mechanism:
     (QiskitDevEnv) qiskit-aer\out> cmake -G "Visual Studio 15 2017" -DBLAS_LIB_PATH=c:\path\to\my\blas ..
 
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -555,6 +555,31 @@ options we have on `Aer` to CMake we use it's native mechanism:
...
+```
+
+For example,
+
+    qiskit-aer$ python ./setup.py bdist_wheel -- -DAER_THRUST_BACKEND=CUDA
+
+
+Few notes on GPU builds:
+1. Building takes considerable more time than non-GPU build, so be patient :)
+2. CUDA® >= 10.1 imposes the restriction of building with g++ version not newer than 8
```
chriseclectic, line 578: CUDA 10.1 requires g++ <= 8, but I think if you have CUDA 10.2 you can use g++ 9 right?

atilag: 8 unfortunately :(

PR #783, opened 2020-06-04
Select CUDA architecture(s) through Cmake option

### LINKED ISSUE #782: Specify cuda architecture to build system
Be able to select the CUDA architecture through a CMake option. This will allow to avoid compilation for all most common CUDA architectures when CMake fails to identify a specific arch.

vvilpas:
### Summary

Select CUDA architecture(s) through CMake option

### Details and comments

Specify the CUDA architecture through new option CUDA_ARCH. Specify a single architecture or a list of them: `-DCUDA_ARCH="5.2"` or `-DCUDA_ARCH="5.2; 5.3"`.

Closes #782


CONTRIBUTING.md
```diff
@@ -572,6 +572,13 @@ For example,
 
     qiskit-aer$ python ./setup.py bdist_wheel -- -DAER_THRUST_BACKEND=CUDA
 
+If we want to specify the CUDA® architecture instead of letting the build system 
+auto detect it, we can use the AER_CUDA_ARCH option. For example:
```
atilag, line 576: We should mention that this can be set via environment variable as well.

PR #784, opened 2020-06-04
Introduce abstraction of DE methods

### LINKED ISSUE #724: ODE framework in C++

Solve system of Ordinary Differential Equations using C++.
The idea is to implement a general framework where we can add different ode C/C++ libraries that best suit the different problems we want to solve.

At this time, our main goal is to introduce the library [**Sundials**](https://computing.llnl.gov/projects/sundials). This is a C library which provides several modules. The most interesting ones are **CVODE** and **CVODES**.  Both of them solve ODE systems, the latter including sensitivity analysis capabilities (forward and adjoint). The main drawback is that it doesn't support complex numbers natively. It can be adapted to work with them, although some functionality will be lost.

We are also testing [**OdeInt**](https://headmyshoulder.github.io/odeint-v2/). This is a native C++ Boost library which handles complex numbers natively. It lacks some of the functionality of **Sundials**, like the sensitivity analysis. However, in very early and limited testing has shown a leap in performance, being an order of magnitude faster.

DanPuzzuoli:
### Summary

This PR is the next step in refactoring the pulse simulator. The main point is to close #724 (or at least, do the initial work for this): abstract/modularize the solvers and the code that generates the RHS.

Goals: 
- Introduce an abstract class - currently called `ODE_Method` - providing a standardized interface for ODE solving routines that the rest of the code interacts with.
     - Purpose: if we want to interface a new solving routine/package, we just need to make a concrete `ODE_Method` object (which should have a clear procedure) and it should work with the rest of the code.
     - Will include automation of certain type handling
- Add a low level "model" class that holds all of the information necessary to generate the RHS function, and can generate the RHS
     - General Purpose: Abstract-the-interface/create-a-container surrounding specific numerical methods and data representations for generating the RHS.
 

### Details and comments

`ODE_Method`
- Will be instantiated with an RHS function, solver options, initial state, and initial time, and have class attributes specifying what type the state needs to be
      - will contain another object called `StateTypeConverter`, that, based on the initial state specified, and the internally required state of the method, will automatically handle conversion of state/rhs function so that a user need never do these translations themselves


DanPuzzuoli:
Current status notes:
- Initial introduction of `ODE_Method` and creating an `ODE_Method` subclass wrapper for `zvode` is almost successful
    - All tests pass except for the noisy simulation. Perhaps I'm introducing some overhead in the getting/setting of the state of the solver that makes the MC algorithm very slow (which gets/sets a lot). There may also be some weird parallelization issue, as the test just completely hangs if run with parallelization, but when parallelization is removed the simulations finish but fail.

DanPuzzuoli:
Previous issue luckily solved with changing one line:
- In the `integrate()` function for the `QiskitZVODE` method wrapper, I was setting the internal time variable `_t` through the "public" `t` property, which has the side effect of resetting the method's internal memory. This caused a huge slow down in the monte carlo routine tests, as the MC routine solves moving forward by taking single steps, and if the internal memory is reset every time it makes it much less efficient.

DanPuzzuoli:
Current status:
- DE_Methods are basically working (got the `QiskitZVODE` version working)
    - Still need some documentation work but will leave that to later
- Now I'm thinking about/generally attempting to break up the `PulseSimDescription` object in `pulse_controller`, with the goal compartmentalizing/clarifying flow of information. May separate into: 
    - DE model information (info specifically required to generate RHS)
    - Possibly further separate out experiment/signal information from the above
    - measurement information

qiskit/providers/aer/pulse/de/DE_Methods.py
```diff
@@ -0,0 +1,390 @@
...
+
+    def _reset_method(self, reset=True):
+        if reset:
+            self._ODE._integrator.call_args[3] = 1
+
+    def set_options(self, solver_options):
+        self.solver_options = solver_options
+
+
+class qiskit_zvode(zvode):
```
nonhermitian, line 319: This is QuTiP code so their license should be in the file.

DanPuzzuoli:
Currently attempting to add a DE-specific options class container to be used directly with `DE_Method` objects. Running into the issue that certain default values are not acceptable to all solvers even within `scipy`. E.g. even though the options have the same name, specifying `max_step==0` works as a default `None`-type argument for `zvode` but not for `solve_ivp`. Conversely, `max_step==None` works for `solve_ivp` but not for `zvode`.

I think I may make these types of `DE_Options` attributes default to `None` and the `DE_Method` itself will have to establish a valid default.

DanPuzzuoli:
Current status
- Moved a subset of data out of `PulseSimDescription` into a new class data-container class `PulseInternalDEModel` to separate the pure "model" parts (the parts involved in describing the DE) from the rest (e.g. measurement operators, # of shots, memory slots, seeds, etc)
- Creation of new `DE_Options` class that works directly with `DE_Methods`, and `PulseSimOptions` now contains an instance of `DE_Options` for DE specific options

DanPuzzuoli:
This PR is ready for review.

The main point of this PR is the introduction of an abstract interface for differential equation solver routines, as the next step in the modularization of the pulse simulator. 

### Summary of major changes

#### `pulse/de/DE_Methods.py`

This file is the main addition, containing the abstract class `ODE_Method`, which is the base class for ODE method wrappers. This class:
- Stores an internal time and state variable
- Stores an `rhs` function (and with further development should store `rhs_jac` functions as well)
- Has methods for integrating the system forward in time

This file contains three concrete subclasses:
- `ScipyODE` - wraps the function `scipy.integrate.solve_ivp`
- `QiskitZVODE` - a wrapper for the solver that is currently being used in the pulse simulator
- `RK4` - a simple hand-written runge-kutta 4 solver, that is mainly meant to demonstrate a minimal example of a concrete subclass of `ODE_Method`

The file also contains a function `method_from_string`, which returns one of the above classes based on a string specification, which is useful for interfaces.

#### Type handling and `pulse/de/type_utils.py`

This file contains a class `StateTypeConverter` that is a helper class for `DE_Methods.py`. The purpose of this class is to facilitate/automate type conversions for the `ODE_Method` class. 
- E.g. A user wants to solve a DE with a state given by a complex 2d array, and writes the rhs function to work with complex 2d arrays. The desired concrete `ODE_Method` however requires 1d arrays. The `StateTypeConverter` stores the "internal type" information - the numerical method requires 1d arrays - and the "external type" information  - the user is working with 2d arrays - and automatically converts back and forth when the state or rhs is get or set. The user never needs to be aware of this conversion.

Each `ODE_Method` contains a class attribute specifying the required internal state type, and a `StateTypeConverter` is instantiated when the DE state is set, with the "internal type" given by the `ODE_Method` specification, and the "outer type" given by whatever type of state the user supplied. The `StateTypeConverter` is used whenever the user gets/sets the state, and is also used to transform any supplied `rhs` functions from ones that receive/return arrays of the "outer type" to one that receives/returns arrays of the "inner type".

The current `StateTypeConverter` class and related functions only work for numpy arrays, and assume currently that everything is complex, but we can expand this to include better number type handling (e.g. real vs complex) as well as potentially broader input state types (e.g. if the user specifies a function using `terra` state types - if so desired).

Each concrete `ODE_Method` must contain a class attribute `method_spec`, which is a dictionary storing technical specification information. Currently, it only contains information regarding the required internal type for the method. E.g. for the `ScipyODE` method, this is:
- `method_spec = {'inner_state_spec': {'type': 'array', 'ndim': 1}}`

Which states that internally the method works with 1d arrays.

#### `DE_Options.py`

A new options container class that is solely about DE solving.

### Summary of other changes

- Everything that isn't specific to de solving has been moved out of the `pulse/de/` folder and into the `pulse/controllers/` folder. 
- `pulse_de_solver.py` originalyl contained a function setting up the `zvode` solver directly, and now it instantiates the `QiskitZVODE` concrete ODE method. The simulation methods have been modified slightly to work with this class as opposed to the `zvode` solver directly (this has simplified the MC solver in particular as the routine now doesn't need to include zvode-specific interface manipulations to reset the solver).
- The original `OPoptions` has been changed to `PulseSimOptions` and now contains an instance of `DE_Options`
- There have been several changes to how the information gets passed around in `pulse_controller.py`, though the main modification is:
     - "model specific" information (i.e. information for rhs generation) has been moved out of `PulseSimDesc` and into a new class called `PulseInternalDEModel`, which contains a method for constructing the rhs function for a given experiment.
- The method signatures of `unitary_controller` and `mc_controller` have changed as a consequence of the above point.

### Notes to reviewers

The "other changes" - i.e. the ones specific to the `pulse/controllers/` folder - are preliminary steps in changes to be implemented in a future PR to refactor the flow of information in files in the `controllers` folder. E.g. the `PulseInternalDEModel` class is likely just temporary. The main changes to review are the new files in the `pulse/de/` folder.

PR #792, opened 2020-06-11
Allow for suppressing warnings in NoiseModel.from_backend()

nonhermitian:
### Summary
The larger devices always have one or more bad gates that trigger warnings.  This allows for suppressing these.

### DIFF
--- a/qiskit/providers/aer/noise/device/models.py
+++ b/qiskit/providers/aer/noise/device/models.py
@@ -193,22 +193,24 @@ def _device_depolarizing_error(qubits,
         # Check if reported error param is un-physical
         # The minimum average gate fidelity is F_min = 1 / (dim + 1)
         # So the maximum gate error is 1 - F_min = dim / (dim + 1)
-        if error_param > error_max and warnings:
-            logger.warning(
-                'Device reported a gate error parameter greater'
-                ' than maximum allowed value (%f > %f). Truncating to'
-                ' maximum value.', error_param, error_max)
+        if error_param > error_max:
+            if warnings:
+                logger.warning(
+                    'Device reported a gate error parameter greater'
+                    ' than maximum allowed value (%f > %f). Truncating to'
+                    ' maximum value.', error_param, error_max)
             error_param = error_max
         # Model gate error entirely as depolarizing error
         num_qubits = len(qubits)
         dim = 2 ** num_qubits
         depol_param = dim * (error_param - relax_infid) / (dim * relax_fid - 1)
         max_param = 4**num_qubits / (4**num_qubits - 1)
-        if depol_param > max_param and warnings:
-            logger.warning(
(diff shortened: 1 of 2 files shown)


atilag:
One of the features we have with Python logging system, is that users can filter any Warning they want by doing:
```python
import logging
class WarningFilter(logging.Filter):
    def filter(self, record):
        return not record.getMessage().startswith('Device model')

logger = logging.getLogger("models")
logger.addFilter(WarningFilter())
```
This will filter all messages starting with "Device model" string, in the models.py file.
If you don't really mind about the warnings at all, you can just disable them:
```python
logging.disable(logging.WARNING)
```

nonhermitian:
Ok.  But it is unclear what that has to do with the current PR that just extends similar functionality that is already in the code base, e.g.:

https://github.com/Qiskit/qiskit-aer/blob/1cff17f248c1a2ba556e1962dc3b5bd3b6437e65/qiskit/providers/aer/noise/noise_model.py#L398

https://github.com/Qiskit/qiskit-aer/blob/1cff17f248c1a2ba556e1962dc3b5bd3b6437e65/qiskit/providers/aer/noise/noise_model.py#L455

https://github.com/Qiskit/qiskit-aer/blob/1cff17f248c1a2ba556e1962dc3b5bd3b6437e65/qiskit/providers/aer/noise/noise_model.py#L465

https://github.com/Qiskit/qiskit-aer/blob/1cff17f248c1a2ba556e1962dc3b5bd3b6437e65/qiskit/providers/aer/noise/noise_model.py#L528

https://github.com/Qiskit/qiskit-aer/blob/1cff17f248c1a2ba556e1962dc3b5bd3b6437e65/qiskit/providers/aer/noise/noise_model.py#L540

https://github.com/Qiskit/qiskit-aer/blob/1cff17f248c1a2ba556e1962dc3b5bd3b6437e65/qiskit/providers/aer/noise/noise_model.py#L609

https://github.com/Qiskit/qiskit-aer/blob/1cff17f248c1a2ba556e1962dc3b5bd3b6437e65/qiskit/providers/aer/noise/noise_model.py#L654

https://github.com/Qiskit/qiskit-aer/blob/1cff17f248c1a2ba556e1962dc3b5bd3b6437e65/qiskit/providers/aer/noise/noise_model.py#L664

nonhermitian:
Ok, now I see, the above does do what I need since I am an interested in using this in a module.    However, it still would be nice to extend the kwargs to allow suppressing the warnings without having to construct a whole class to do so.

atilag:
So you pointed out to something that I call technical debt. We shall not couple logging behavior with function parameters, it does only increases coupling between these components and makes things harder to test.
This is the reason why loggers are singletons (unique global instances).

Ok, let's merge this if you find it useful.

Issue #773, opened 2020-06-01
Some MPS method tests fail on Windows


chriseclectic:
### Informations

- **Qiskit Aer version**:  master
- **Python version**: 3.5-3.8
- **Operating system**: Windows

### What is the current behavior?

Some of the standard tests that have been fixed in #765 are failing on Windows for the matrix product state simulation method.

The failures all appear to be due to SVD errors `ERROR: error: wrong SVD calc: A != USV*` (likely due to lack of quad precision on windows).

The specific tests failure are related to CX, CZ, CCX, CSWap gates and grovers test:

- [ ] test_cx_gate_deterministic_minimal_basis_gates (circuit 7)
- [ ] test_cx_gate_deterministic_waltz_basis_gates (circuit 7)
- [ ] test_cz_gate_deterministic_minimal_basis_gates (circuit 4)
- [ ] test_cz_gate_nondeterministic_default_basis_gates (circuits 0, 1)
- [ ] test_cz_gate_nondeterministic_minimal_basis_gates (circuits 0)
- [ ] test_cz_gate_nondeterministic_waltz_basis_gates (circuit 0)
- [ ] test_ccx_gate_deterministic_minimal_basis_gates (circuits 4, 6, 7)
- [ ] test_ccx_gate_deterministic_waltz_basis_gates (circuits 4, 6, 7)
- [ ] test_ccx_gate_nondeterministic_minimal_basis_gates (circuits 2, 3)
- [ ] test_ccx_gate_nondeterministic_waltz_basis_gates (circuits 1, 2, 3)
- [ ] test_cswap_gate_deterministic_minimal_basis_gates (circuits 3, 5, 7, 8, 9)
- [ ] test_cswap_gate_deterministic_waltz_basis_gates (circuits 3, 7, 8, 9)
- [ ] test_cswap_gate_nondeterministic_minimal_basis_gates (circuits 1, 2, 6)
- [ ] test_cswap_gate_nondeterministic_waltz_basis_gates (circuits 6)
- [ ] test_grovers_default_basis_gates (circuit 0)
- [ ] test_grovers_minimal_basis_gates (circuit 0)
- [ ] test_grovers_waltz_basis_gates (circuit 0)

where test circuits are counted from 0.

### Steps to reproduce the problem

Run tests in Azure CI

### What is the expected behavior?

Tests pass

### Suggested solutions

Investigate the validation of SVD, maybe we are checking with too tight a tolerance?

[...]

PR #797, opened 2020-06-16
[WIP] MPS: Fixed bug related to floating-point precision on Windows

merav-aharoni:
### Summary
There were failures in the SVD algorithm on Windows, because of low precision of floating point numbers on Windows. This fix resolves issue #773.

### Details and comments
Some of the computations in the SVD algorithm produce very small fractions `(~10^-30)`. When comparing such a fraction to 0 on Windows, the precision of double floating point is not sufficient to detect that they are not 0.
I added a function that scales up the fractions by some large factor `(~10^30)` when doing the comparison. This function is not specific for SVD, so you may want to consider putting it in the Linalg::almost_equal library.
I updated the list of tests. Please check if these are all the tests that should pass at this stage. There is no longer a differentiation between platforms.


src/simulators/matrix_product_state/svd.cpp
```diff
@@ -114,33 +124,28 @@ status csvd(cmatrix_t &A, cmatrix_t &U,rvector_t &S,cmatrix_t &V)
...
-		}
-		b[k] = 0.0;
-		if ( tol < z )
-		{
+    b[k] = 0.0;
+    if ( tol < z ){
 			z = std::sqrt( z );
 			b[k] = z;
 			w = std::abs( A(k,k) );
-			if (Linalg::almost_equal(static_cast<long double>(w),
```
chriseclectic, line 138: For these sort of comparisons you can specify the type by the function template parameters instead of casting the argument `Linalg::almost_equal<long double>(a, b, max_diff)`

merav-aharoni: I started checking this, and while doing so, I noticed another problem. So I will need some more time to work on this PR. I am adding back the WIP.

merav-aharoni:
I don't know how to resolve this mistake with the missing CLA. So I will close this PR and open a new one instead.

Issue #736, opened 2020-05-06
Add simulator support for new Pulse instructions


DanPuzzuoli:
### What is the expected behavior?

Add support for new Pulse instructions: 

`Delay`, `SetPhase`, `ShiftPhase`, `SetFrequency`, `ShiftFrequency`

### Implementation

My initial thought is that `Delay`, `SetPhase`, and `ShiftPhase` will all naturally fit into the current interpretation scheme that the simulator uses for pulses in the `PulseQObj`. E.g. the deprecated `FrameChange` instruction was already supported in the PulseQobj interpretations, so updating that portion may be all that is necessary for the phase instructions.

`SetFrequency` and `ShiftFrequency` on the other hand will require a change in the way frequencies are dealt with; currently frequency is a static property of the "differential equation model", and there is currently no mechanism/process for updating this model mid-simulation. Some options:
1. I think the simplest option in terms of immediate implementation is the promotion of channel frequency from a constant to a time-dependent function, which can be constructed similarly to how the signals over channels are done now. The DE RHS function can then be updated to call the frequency function.
2. Add the ability to update/modify the "model" mid-simulation. This will amount to having types of instructions that induce an update of the `PulseSimDescription` object, in which case the ODE solver will also need to be updated with the new information.

Option 2 is more involved but more in-line with future plans - a solver's behaviour can make use of the carrier frequency (entering/exiting frames automatically, making RWA, etc) - but if the frequency is also treated as an arbitrary time-dependent function this won't be possible. Also, the ability to do such updates is necessary for future extensions involving more complicated simulation flows (e.g. handling control flow, performing measurements mid-experiment, etc).

DanPuzzuoli:
As a first step we can:
1. Add interpretation/support for `Delay`, `SetPhase`, and `ShiftPhase`.
    - Remove/replace internal `FrameChange` interpretation.
2. Add warnings for `SetFrequency` and `ShiftFrequency` commands.

eggerdj:
Perhaps we can easily implement `SetFrequency` and `ShiftFrequency` by multiplying the pulse envelope with a modulation.

DanPuzzuoli:
Not being actively developed.

amirebrahimi:
I'm getting an error when trying to make use of SetFrequency or ShiftFrequency on a DriveChannel of FakeArmonk:
```
[...]

PR #798, opened 2020-06-18
[WIP] Adding Delay, ShiftPhase, and SetPhase instruction handling to pulse simulator

DanPuzzuoli:
### Summary

This handles the first part of issue #736 , adding support/testing for the instructions in the title. 

### Details and comments

Current status:
- Luckily, `Delay` instructions don't need anything new - they don't even appear in the `qobj`.
    - A test has been added to verify the functioning of schedules with delays and it is currently passing
- `ShiftPhase` exactly replaces `FrameChange`, though in the qobj they have the same name, so `ShiftPhase` doesn't technically need anything new, however incorporating `SetPhase` requires tracking absolute phase.
     - I've changed the phase storage array constructed in `digest` to store absolute phase as opposed to relative phase, and it is now populated accordingly for `ShiftPhase` instructions. The C++ RHS-related function has also been changed accordingly (it previously added up all relative phase changes, now it just finds the relevant phase and uses that).
     - A new `ShiftPhase` test has been added and is working, and replaced the `FrameChange` test.

To do:
- Add more `ShiftPhase` tests
- Add `SetPhase` handling and tests

Encountered issue:
- While thinking of tests for phase handling, I've run into an issue where I think that the channel values may not be being evaluated consistently with the pulse spec - will make an issue for discussion/investigation of this.

### DIFF
--- a/qiskit/providers/aer/pulse/controllers/digest_pulse_qobj.py
+++ b/qiskit/providers/aer/pulse/controllers/digest_pulse_qobj.py
@@ -440,5 +440,5 @@ def experiment_to_structs(experiment, ham_chans, pulse_inds,
 
     if structs['tlist'][-1] > structs['acquire'][-1][0]:
         structs['can_sample'] = False
-
+    
     return structs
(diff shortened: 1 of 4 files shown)


DanPuzzuoli:
Closing this PR to replace it with another one.

PR #800, opened 2020-06-19
Moved the metadata into setup.cfg

KOLANICH:
### Summary
Moved the metadata into `setup.cfg`.
Vetsion is fetched automatically from git tags.

### Details and comments
IMHO this small change doesn;t worth any mention in changelogs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,10 @@
 # be linked statically per Apple's indications.
 
 cmake_minimum_required(VERSION 3.6)
-file(STRINGS "qiskit/providers/aer/VERSION.txt" VERSION_NUM)
+file(STRINGS "qiskit/providers/aer/VERSION.txt" VERSION_STRING)
+
+string(REGEX MATCH "[0-9]+(\\.[0-9]+)+" VERSION_NUM "${VERSION_STRING}")
+
 project(qasm_simulator VERSION ${VERSION_NUM} LANGUAGES CXX C)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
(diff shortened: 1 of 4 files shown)


mtreinish:
Thanks for opening this PR. I definitely like the idea of moving more of the metadata to a config file than having it encoded as a parameters in a python script (it reminds me of how it's done in pbr which I often use on personal python projects for this). I don't have any personal experience with setuptools scm, but it looks like it has a lot of useful features (which arguably should just be in setuptools). Unfortunately as CI is showing here the integration between versioning, aer's build system scikit-build with cmake and conan, and setuptools is very fragile. It took us a while before we got it as stable as it is now, and it's honestly still not great. I don't think we're realistically going to be looking at adding an extra dependency to the mix even if I personally like the direction it moves us. There are just too many moving pieces here. Maybe in the future if we simplify the aer build system and its integration with setuptools (ie remove the scikit-build piece) we can revisit this, but for right now I'm going to close this because it's not a great fit for right now.

Also, as an aside to accept any contributions you'll have to agree to and sign the CLA for the project. I see that you opened PRs against most of the qiskit projects, I'm definitely interested in looking at them, especially on the repos which are pure python. But before we can consider them you'll have to sign the CLA. Thanks.

KOLANICH:
>Also, as an aside to accept any contributions you'll have to agree to and sign the CLA for the project.

CLAs are legal documents, I prefer to stay away of any of legal stuff, because any legal stuff is a risk. I also prefer not to use doubtful services, like CLAbot. I also prefer not to use an own legal name anywhere in the Internet, and signing any documents is impossible without using it. I also consider CLAs as harmful and it is sad that it is getting fashionable to require CLAs, so I consider signing of CLAs as contributing to the trend that has been given the nickname of "New World Order" (not in the sense of conspiracy theories, but just a nickname for undesired changes, I mean 10 years ago (and even 5 years ago!) there were no such obession about CLAs, but in any big company there are lawers, who want to earn money, and one of the ways to create demand for their labour is to go to a boss and say "it's terrible, it's terrible, our company doesn't use CLA, our company strictly needs CLA, otherwise <very bad thing> can happen, we can create a CLA"). The result is that lot of companies now require CLAs even if they don't really need them.


But you can consider my contributions (the parts of them that Berne convention framework considers as mine) into Qiskit repos licensed under [Unlicense license](https://unlicense.org/), so anyone can do with them anything he wants (including putting own name on them and sending the PRs from own accounts). Or one can just close the PRs about the and redo the changes: the migration of metadata is done mostly automatically by `setuptools_py2cfg`, my contributions are mostly boilerplate `pyproject.toml` I use in own projects + some minor tweaks, all of this can be easily redone from scratch by anyone.

PR #805, opened 2020-06-23
[WIP] Wrapper for ODE Sundials solver in C++

vvilpas:
### Summary
Wrapper for ODE Sundials solver in C++

### Details and comments

This PR brings the capability to use ODE solvers from C/C++ domain. The first one introduced is [Sundials Cvodes](https://computing.llnl.gov/projects/sundials/cvodes). It adds a new feature that allows to compute sensitivities to different parameters.

Basically it has 2 flavors, regular solver and solver with sensitivities. For the first one, the initial conditions and the right hand side function are needed. For the second, in addition to those, we need a function to modify the parameters whose sensitivities we are interested in, and the initial values fo those parameters.

It also introduce the necessary bindings to use them from python, using python defined functions for right hand side, parameter perturbation, etc.


CMakeLists.txt
```diff
@@ -227,7 +227,9 @@ set(AER_LIBRARIES
 	Threads::Threads
 	CONAN_PKG::spdlog
 	${DL_LIB}
-	${THRUST_DEPENDANT_LIBS})
+	${THRUST_DEPENDANT_LIBS}
+	CONAN_PKG::sundials
+	CONAN_PKG::boost)
```
chriseclectic, line 232: :'(

Is it possible to only get the ODEINT part of boost?

vvilpas: We are using boost as a header only lib so we are only using what we include.

PR #806, opened 2020-06-23
optimize calculation of expectation value

hhorii:
### Summary

Reduce load of a state vector in `expval_pauli`

### Details and comments

Current calculation of expectation values loads one amplitude twice in a loop. `i` and `i ^ x_mask ^x_mask` are the same in `auto val = std::real(phase * data_[i ^ x_mask] * std::conj(data_[i]));`

This new implementation iterates a half times (`data_size_ >> 1` times).  

A new local variable `x_max` is the maximum qubit in `x_mask`. In the loop to calculate an expectation value, a value is 0 for the bit of `x_max` and the other is 1. All the larger or equal bits than `x_max` are shifted to left one and the lower bits remain.

For example, if `x_mask` is `00010001`, `x_max` is 4 (0-order from right), `mask_u` is `11100000` and `mask_l` is `00001111`. 
In the loop, `idxs[0]` (`= ((i << 1) & mask_u) | (i & mask_l)`) and `idxs[1]` (`= idx[0] ^ x_mask`) are:

| i | idx[0] | idx[1]
----|----|---- 
| 00000000 | 00000000 | 00010001
| 00000001 | 00000001 | 00010000
| 00000010 | 00000010 | 00010011
| 00000100 | 00000100 | 00010101
| 00001000 | 00001000 | 00011001
| 00010000 | 00100000 | 00110001
| 00100000 | 01000000 | 01010001
| 01000000 | 10000000 | 10010001


chriseclectic:
How much of an improvement in performance does this make @hhorii?

hhorii:
> How much of an improvement in performance does this make @hhorii?

UCCSD of LiH was improved from 56s to 21s.

PR #807, opened 2020-06-23
Improve performance of average snapshots

chriseclectic:
### Summary

Previously the mean and variance were returned as a copy of stored
accumulated data. This allows them to modify the stored data and return
a reference that can be moved into result containers which can help
avoid copies for large data objects like probability and density matrix
snapshots. To make this work it was also necessary to allow JSON conversion of
results to be non const.

### Details and comments

Memory profiler comparisons for following circuit:

```py
nq = 14
circ = QuantumCircuit(nq)
circ.snapshot_density_matrix('state')

backend = QasmSimulator()
backend_options = {
    "method": "density_matrix",
    "fusion_enabled": False,
    "truncate_enabled": False,
    "max_parallel_threads": 1
}
qobj = assemble(circ, shots=1)

start = time.time()
result = backend.run(qobj, backend_options=backend_options).result()
stop = time.time()
time_py = stop - start
time_cpp = result.metadata['time_taken']
```

For this PR: time_py = 29.2 s, time_cpp = 26.7 s
For master: time_py = 31.8 s, time_cpp = 58.3 s

And memory profiling PR:

mprof_dm_14q_pr

vs Master:

mprof_dm_14q_master

### DIFF
--- a/src/framework/results/data/average_data.hpp
+++ b/src/framework/results/data/average_data.hpp
@@ -26,11 +26,13 @@ class AverageData {
  public:
   // Return the mean of the accumulated data:
   // mean = accum / count
-  T mean() const;
+  // Calling this method normalizes the data and returns a reference
+  T& mean();
 
   // Return the unbiased sample variance of the accumulated data:
   // var = (1 / (n - 1)) * (sum_i (data[i]^2 / n) - mean^2) for n > 1
-  T variance() const;
+  // Calling this method normalizes the data and returns a reference
+  T& variance();
 
   // Add a new datum to the snapshot at the specified key
   // Uses copy semantics
@@ -70,6 +72,15 @@ class AverageData {
   // Return bool for in the container can compute variance
   bool has_variance() const { return variance_; }
 
+  // Noramlize data by dividing the accumulator by the number of
+  // shots to compute the mean, and computing the variance
+  void normalize();
+
+  // Convert container to JSON
+  // note that this method is not const Once it is converted it will 
+  // renormalized the data to compute the mean and variance
+  json_t to_json();
+
  protected:
   // Accumulated data
   T accum_;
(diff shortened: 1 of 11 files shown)


src/framework/results/data/average_snapshot.hpp
```diff
@@ -108,23 +111,21 @@ void AverageSnapshot<T>::combine(AverageSnapshot<T> &&other) noexcept {
...
+  json_t js = json_t::object();
+  for (auto &outer_pair : data_) {
+    for (auto &inner_pair : outer_pair.second) {
       // Store mean and variance for snapshot
-      json_t datum = inner_pair.second;
+      json_t datum = inner_pair.second.to_json();
       // Add memory key if there are classical registers
       auto memory = inner_pair.first;
-      if (memory.empty() == false) datum["memory"] = inner_pair.first;
+      if (memory.empty() == false) datum["memory"] = memory;
```
vvilpas, line 123: if (!memory.empty()) datum["memory"] = memory;

I think is more c++ style

src/framework/results/data/average_snapshot.hpp
```diff
@@ -108,23 +111,21 @@ void AverageSnapshot<T>::combine(AverageSnapshot<T> &&other) noexcept {
...
-    for (const auto &inner_pair : outer_pair.second) {
+json_t AverageSnapshot<T>::to_json() {
+  json_t js = json_t::object();
+  for (auto &outer_pair : data_) {
+    for (auto &inner_pair : outer_pair.second) {
       // Store mean and variance for snapshot
-      json_t datum = inner_pair.second;
+      json_t datum = inner_pair.second.to_json();
       // Add memory key if there are classical registers
       auto memory = inner_pair.first;
```
vvilpas, line 122: I know is not on this PR, but do we need to copy that string? Can't we just take a reference?

src/framework/results/data/average_data.hpp
```diff
@@ -82,39 +93,48 @@ class AverageData {
...
 //------------------------------------------------------------------------------
 // Implementation
 //------------------------------------------------------------------------------
-
 template <typename T>
-T AverageData<T>::mean() const {
-  return (count_ > 1) ? Linalg::div(T(accum_), double(count_)) : accum_;
+void AverageData<T>::normalize() {
+  if (!normalized_ && count_ >= 1) {
+    if (count_ > 1) {
```
vvilpas, line 109: If we move the `mean` and `variance` results, subsequent calls to those functions will yield wrong results. Probably we don't need to call them again but if that's the case we may need to implement something to recompute the values, as `normalize` won't be executed again as `normalized_` will be `true`. What do you think?

chriseclectic: It should be fine to call mean and variance multiple times, only the first call will result in normalizing the data (this is because `normalize` does nothing if `normalized_ = true`). But if you move them then yes, you've destroyed the container. I think this is like any other C++ object, after you used move you shouldn't use it again. If you want to use it again you should copy or reference the data rather than move, however i don't think this situation comes up in our current use case.

vvilpas: I think the movement is a bit more hidden because we are moving specific members instead of the whole object, but as we're not gonna reuse them it seems we are ok.

Issue #770, opened 2020-05-31
MPS: Improve performance of sample_measure


merav-aharoni:
### What is the expected behavior?
The current implementation of sample_measure in MPS creates a copy of the MPS, and invokes apply_measure. This is fine when the number of shots is small, but increases linearly with the number of shots. It is more efficient to implement something similar to what the statevector does. There the performance for a large number of shots is similar to the performance for a single shot.

PR #808, opened 2020-06-24
MPS: Improve performance of sample_measure

merav-aharoni:
### Summary
Performance of sample_measure in MPS was very slow - see issue #770 . This PR improves its performance.

### Details and comments
The previous version of sample_measure worked as follows: for every shot, create a temporary copy of the MPS structure, and invoked apply_measure on it. This works even for a large number of qubits, but is inefficient when the number of shots is large.
So for a large number of shots, I implemented an algorithm similar to what statevector does, but computing the probabilities is different as follows:
1. I store only the non-zero probabilities and the suitable base values
2. I store the accumulated probabilities
3. Since the accumulated vector of probabilities is increasing, I can search the current `rnd` using a binary search and not by searching linearly.
Following experimentation, with varying number of shots, number of qubits, I defined two algorithms. One of these will be invoked depending on the above parameters:
1) using apply_measure - for a small number of shots (<10) or large number of qubits (>26)
2) creating the vector of accumulated probabilities in all other cases.

I believe it is worth trying to use algorithm (2) for the qubitvector as well.


merav-aharoni:
comparing results of sample_measure

merav-aharoni:
sample_measure

merav-aharoni:
@hhorii , I made some improvements and I think this PR is now ready for review.

PR #809, opened 2020-06-24
MPS: Fixed bug related to floating-point precision on Windows

merav-aharoni:
### Summary
There were failures in the SVD algorithm on Windows, because of low precision of floating point numbers on Windows. This fix resolves issue #773.

### Details and comments
This PR replaces PR#797, which I couldn't merge, because of a CLA problem.
I found two problems when running the SVD algorithm.
1. Some of the computations in the SVD algorithm produce very small fractions (~10^-30). When comparing such a fraction to 0 on Windows, the precision of double floating point is not sufficient to detect that they are not 0. To resolve this, I added a parameter, `zero_threshold = 1e-50` for higher sensitivity of this comparison.
2. There was a validation check at the end of the SVD algorithm that was failing on high numbers of qubits, because the comparison was too sensitive (`1e-9`). I moved the check to `DEBUG` mode, since the algorithm is working correctly, it is only the precision that is problematic. 

Note that I also address comments provided by @chriseclectic in PR #797 that was closed.

### DIFF
--- a/src/simulators/matrix_product_state/svd.cpp
+++ b/src/simulators/matrix_product_state/svd.cpp
@@ -31,11 +31,11 @@
 namespace AER {
 
 // default values
-static const double mul_factor = 1e2;
-static const long double tiny_factor = 1e30;
-static const auto zero_threshold = 1e-50;  // threshold for comparing FP values
-static const double THRESHOLD = 1e-9; // threshold for cutting values in reduce_zeros
-static const int_t NUM_SVD_TRIES = 15;
+constexpr auto mul_factor = 1e2;
+constexpr long double tiny_factor = 1e30;
+constexpr auto zero_threshold = 1e-50;  // threshold for comparing FP values
+constexpr auto THRESHOLD = 1e-9; // threshold for cutting values in reduce_zeros
+constexpr auto NUM_SVD_TRIES = 15;
 
 cmatrix_t diag(rvector_t S, uint_t m, uint_t n);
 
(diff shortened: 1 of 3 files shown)


src/simulators/matrix_product_state/svd.cpp
```diff
@@ -136,6 +137,30 @@ void reduce_zeros(cmatrix_t &U, rvector_t &S, cmatrix_t &V,
...
+  std::cout << std::endl;
+  std::cout << "V* = " << std::endl << V ;
+  std::cout << "product= " << std::endl << product << std::endl;
+  for (uint_t ii=0; ii < nrows; ii++)
+    for (uint_t jj=0; jj < ncols; jj++)
+      if (!Linalg::almost_equal(std::abs(A(ii, jj)), std::abs(product(ii, jj)), THRESHOLD)) {
+	std::stringstream ss;
+	ss << "error: wrong SVD calc: A != USV*";
+	std::cout <<"error: wrong SVD calc: A != USV*"<< std::endl;
+	throw std::runtime_error(ss.str());
```
atilag, line 160:
```suggestion
        throw std::runtime_error("Error: Wrong SVD calculations: A != USV*");
```

src/simulators/matrix_product_state/svd.cpp
```diff
@@ -33,7 +33,8 @@ namespace AER {
 // default values
 static const double mul_factor = 1e2;
 static const long double tiny_factor = 1e30;
-static const double THRESHOLD = 1e-9; // threshold for re-normalization after approximation
+#define zero_threshold 1e-50  // threshold for comparing FP values
```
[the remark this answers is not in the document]
chriseclectic: or `constexpr` instead of `static const` for all these?

src/simulators/matrix_product_state/svd.cpp
```diff
@@ -136,6 +137,27 @@ void reduce_zeros(cmatrix_t &U, rvector_t &S, cmatrix_t &V,
   }
 }
 
+void validate_SVD_result(cmatrix_t &A, cmatrix_t &U,rvector_t &S,cmatrix_t &V) {
+  const uint_t nrows = A.GetRows(), ncols = A.GetColumns();
+  cmatrix_t diag_S = diag(S, nrows, ncols);
+  cmatrix_t product = U*diag_S;
+  product = product * AER::Utils::dagger(V);
+  std::cout << "matrices after SVD:" <<std::endl;
```
chriseclectic, line 145: All these debug print statements should be removed

src/simulators/matrix_product_state/svd.cpp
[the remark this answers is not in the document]
merav-aharoni: Done.

src/simulators/matrix_product_state/svd.cpp
merav-aharoni: Done.

src/simulators/matrix_product_state/svd.cpp
merav-aharoni: removed the debug prints.

PR #811, opened 2020-06-26
Fix channel value bug and revamp pulse sim testing

### LINKED ISSUE #803: Channel value evaluation in pulse simulator is inconsistent with the pulse spec
### Informations

- **Qiskit Aer version**: This is true in all previous versions, though the code used to demonstrate the issue is in PR #798 , as it has an added line of code enabling specifying initial state that is convenient for testing behaviour.

### Problem description

The [pulse spec](https://arxiv.org/abs/1809.03452) states that for a drive signal   on a channel with frequency , the value of the channel at a time  is:

and if a phase  is on the channel (e.g. from a `ShiftPhase` or `SetPhase` command), it multiplies the signal by , and hence the overall channel value at time  is

.

This does not appear to be what is actually evaluated in the RHS functions in the pulse simulator. First, when constructing the expression , the sign of the exponential appears to be flipped; the  contribution has a positive sign:
https://github.com/Qiskit/qiskit-aer/blob/c95bc6b390df9fd19ffa8a97472ae75f47563cea/src/open_pulse/numeric_integrator.cpp#L91
and the  term has a minus sign:
https://github.com/Qiskit/qiskit-aer/blob/c95bc6b390df9fd19ffa8a97472ae75f47563cea/src/open_pulse/numeric_integrator.cpp#L93

Also, at no point does it seem to take the real part of the above expression; this value is somehow used in the function `evaluate_hamiltonian_expression`, and the output of that function is used in constructing the RHS; with either the original value or complex conjugate being used.

A simple example demonstrating this issue is the following: Consider a 2 level system with Hamiltonian
,
where  is the standard Pauli X operator and  is drive channel 0. We run a pulse using this model with drive channel frequency 0 and a purely imaginary sample pulse . As  is purely imaginary and the channel frequency is 0, we will have  for all time. As a result, regardless of the sample pulse , the state of the system should undergo no change.

However, the following code demonstrates this is not the case in the pulse simulator:

```
hamiltonian = {}
hamiltonian['h_str'] = ['0.5*np.pi*X0||D0']
hamiltonian['qub'] = {'0': 2}
ham_model = HamiltonianModel.from_dict(hamiltonian)

u_channel_lo = []
subsystem_list = [0]
dt = 1.

system_model = PulseSystemModel(hamiltonian=ham_model,
                                u_channel_lo=u_channel_lo,
                                subsystem_list=subsystem_list,
                                dt=dt)

sched = pulse.Schedule()
sched += pulse.Play(pulse.SamplePulse([1j]), pulse.DriveChannel(0))
sched |= pulse.Acquire(1, pulse.AcquireChannel(0), pulse.MemorySlot(0)) << sched.duration

backend_sim = PulseSimulator()

qobj = assemble([sched],
                backend=backend_sim,
                meas_level=2,
                meas_return='single',
                qubit_lo_freq=[0.],
                shots=256)

backend_options = {'initial_state': np.array([1., 0.])}

result = backend_sim.run(qobj, system_model, backend_options).result()
```

As noted above, we expect the final state to be the same as the initial state, however the resulting state vector is: 
```
array([-4.64479624e-07+0.j,  1.00000000e+00+0.j])
```
i.e. a pi pulse (about some axis) is erroneously occurring. 

### Proposed solution

Some initial testing indicates that the following changes results in the correct expected behaviour:
- Switching the incorrect signs above
- Changing the function that evaluates the channel value to only return the real part

@taalexander Can you please verify my reading of the complex number, frequency, and phase conventions in the pulse spec (along with my expectation in the above example that the channel value should be `0`)? It is fairly plainly written there but I just want extra confirmation due to the importance of this.

DanPuzzuoli:
### Summary

Closes #803 , and restructures the integration tests of the pulse simulator to attempt to be more systematic. 

~This PR builds on PR #784 , so review should wait for that to be merged.~

### Details and comments

Changes
- The main functional change is correcting the error pointed out in issue #803: the `chan_value` function in `numeric_integrator.cpp` has been modified to have both complex exponentials have positive sign, and return only the real part of the computed value.
- Minor changes to `pulse_controller`: added `'method'` as a valid de option, and added the ability to set initial state.
- The `test_pulse_simulator.py` test file has been revamped; it was necessary to change all tests after  fixing `chan_value`, and this was a good opportunity to make things more rigorous. 
    - This file has been changed heavily, though in spirit all of the same tests are being done. The main addition is the file `pulse_sim_independent.py`, which contains simulation routines for some "standard" models, and besides using the `DE_Methods.py` file, is totally independent of the pulse simulator. Most of the tests in `test_pulse_simulator.py` now contain comparison of the results of the de solving with these independent simulations.  
    - The simulations in `pulse_sim_independent.py` take very little code and are meant to be fairly simple, and as such it is much easier to verify their correctness as compared to the full process of the pulse simulator itself.
    - Some of the tests also contain comparisons to approximate analytic solutions (obtained using the rotating wave approximation). In these comparisons, the tolerance for acceptance is lower than in the comparison to independent simulations as the quality of the approximation is not quantified (as compared to a DE solver). 

Notes
- Despite it being a seemingly very small change, the fix to `chan_value` has resulted in a notable slow down to the pulse simulator. I'm not totally sure why, but perhaps the corrected DEs are more challenging to solve.

Status
- ~Still need to clean things up (comments, linting, etc)~
- Done


DanPuzzuoli:
This PR is ready (though is waiting on PR #784).

DanPuzzuoli:
Yes good catch - I had noticed this but it slipped my mind. I was able to get the notebook to work by using the DE method `scipy-RK45` instead of zvode (this is one of the new methods available from the `DE_Methods` PR). This can be done through `backend_options`:
```
backend_options = {'solver_options' : {'method' : 'scipy-RK45'}}
sim_result = backend_sim.run(cr_rabi_qobj, two_qubit_model, backend_options).result()
```

I'm not totally sure what is going on with `zvode` and why it raises this error - it has been suggested that this may be due to discontinuities in the RHS function (as the pulses are piecewise constant), but I'm not sure I'm sold on this as it seems to work fine in most other instances (e.g. the simulations earlier in the notebook also contain such discontinuities). My _guess_ is zvode may just be hitting some lower limit on step size or an upper limit on number of steps (perhaps there is some option we can change?), as the cross-resonance simulations may have some fast oscillating terms in the solutions.

One simple option for now is to just change the default solver to `scipy-RK45`. This is rudimentary but comparing speeds of  running all of the pulse simulator tests using `scipy-RK45` on this PR seems to consistently take ~24-25 seconds, whereas with zvode they take ~21-22 seconds. So, zvode is faster but not by much (though admittedly this gap may appear larger after your RHS function improvements are included). Thoughts?

DanPuzzuoli:
Looking at the [zvode documentation](http://www.netlib.org/ode/zvode.f), the `-1 means excess work done on this call. (Perhaps wrong MF.)` is described as:

```
C          -1  means an excessive amount of work (more than MXSTEP
C              steps) was done on this call, before completing the
C              requested task, but the integration was otherwise
C              successful as far as T.  (MXSTEP is an optional input
C              and is normally 500.)  To continue, the user may
C              simply reset ISTATE to a value .gt. 1 and call again.
C              (The excess work step counter will be reset to 0.)
C              In addition, the user may increase MXSTEP to avoid
C              this error return.  (See optional input below.)
```

I'm not sure but `MXSTEP` may be the equivalent of `nsteps` in the `scipy` interface. I want to try increasing this value, though I'm currently unable to run the tutorial due to a bug I'm running into (before any simulator run) after updating python.

DanPuzzuoli:
Just increased value of `nsteps` to `10**6`, and the tutorial runs. I ended up ultimately deciding to stick with `zvode` as the simulations using either `zvode` or `scipy-RK45` seem to take about the same amount of time in the tutorial and `zvode` is a bit faster in testing. (This is odd given the observed discrepancies in the previous comment. These latest comparisons are the result of resetting the tutorial to its current state. Not sure what has caused the change, but resolving which solver to use should be part of more systematic speed tests anyway.)

PR #813, opened 2020-06-29
Adding Delay, ShiftPhase, and SetPhase instruction handling to pulse simulator

DanPuzzuoli:
### Summary

This handles the first part of issue #736 , adding support/testing for the `Delay`, `ShiftPhase`, and `SetPhase` instructions.

~~This PR depends on the changes in PR #811.~~

### Details and comments

- `Delay` didn't require any additional handling, as it doesn't enter the `qobj`. 
- A new test has been added to ensure delays are properly handled.
- Handling of phases in the function `experiment_to_structs` in `digest_pulse_qobj.py` and the function `chan_value` in `numeric_integrator.cpp` has been changed from relative to absolute. 
    - Previously the data structure storing phase changes stored relative changes, as the `FrameChange` instruction (which has been replaced by `ShiftPhase`) was relative to the current phase. However, `SetPhase` is an absolute instruction, and therefore requires absolute phase handling.
    - To handle the above point, `experiment_to_structs` has been changed to construct an array tracking absolute phases, and `chan_value` has been changed to look this value up (whereas previously it would add up all the relative phase changes). 
    - Handling of the `SetPhase` instruction has been added to `experiment_to_structs`.
- Several tests for `SetPhase` and `ShiftPhase` have been added.

### DIFF
--- a/src/open_pulse/numeric_integrator.cpp
+++ b/src/open_pulse/numeric_integrator.cpp
@@ -69,24 +69,25 @@ complex_t chan_value(
     // TODO floating point comparsion with complex<double> ?!
     // Seems like this is equivalent to: out != complex_t(0., 0.)
     if(out != 0.){
-        double phase = 0.;
         num_times = floor_div(fc_array.shape[0], 3);
-        for(auto i = 0; i < num_times; ++i){
-            // TODO floating point comparison
-            if(t >= fc_array[3 * i]){
-                bool do_fc = true;
-                if(fc_array[3 * i + 2] >= 0){
-                    if(!reg[static_cast<int>(fc_array[3 * i + 2])]){
-                       do_fc = false;
-                    }
-                }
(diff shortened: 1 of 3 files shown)


src/open_pulse/numeric_integrator.cpp
```diff
@@ -69,24 +69,25 @@ complex_t chan_value(
...
+        auto phase_idx = 0;
+        auto found = false;
+        while((found == false) && (phase_idx < num_times)){
+            if(t < fc_array[3 * phase_idx]){
+                found = true;
             }else{
-                break;
+                phase_idx++;
             }
         }
```
vvilpas, line 84: If you want to get rid of the `found` var you can do this:

```
while(phase_idx < num_times){
    if(t < fc_array[3 * phase_idx]) break;
    phase_idx++;
}
```

PR #814, opened 2020-07-02
Cache ODE RHS params

vvilpas:
### Summary

In normal ODE solving execution RHS function is called hundred of thousands of times. Conversion of function parameters from python to C++ are quite expensive and take most of the time. In this PR a functor is created to cache the converted parameters to C++ that do not change from call to call within the solver.

Testing with the tutorial [8_pulse_simulator_duffing_model](https://github.com/Qiskit/qiskit-tutorials/blob/master/legacy_tutorials/aer/8_pulse_simulator_duffing_model.ipynb):
- In the first simulation run (section 4.2), execution time goes down from **~10s** to **~2s**.
- In the second simulation (section 5.3, cell 15), execution times goes down from **~17s** to **~3.5s**.
- And finally, in the last simulation (section 5.3, cell 18), time goes down from **~40s** to **~8s**.

### DIFF
--- a/qiskit/providers/aer/pulse/controllers/pulse_controller.py
+++ b/qiskit/providers/aer/pulse/controllers/pulse_controller.py
@@ -25,8 +25,7 @@ from .digest_pulse_qobj import digest_pulse_qobj
 from .pulse_sim_options import PulseSimOptions
 from .unitary_controller import run_unitary_experiments
 from .mc_controller import run_monte_carlo_experiments
-from .pulse_utils import td_ode_rhs_static
-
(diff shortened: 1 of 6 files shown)


vvilpas:
Tested with PR #811 using `scipy-RK45`:

- In the first simulation run (section 4.2), execution time goes down from **~40s** to **~13s**.
- In the second simulation (section 5.3, cell 15), execution times goes down from **~250s** to **~78s**.
- And finally, in the last simulation (section 5.3, cell 18), time goes down from **~324s** to **~97s**.

PR #818, opened 2020-07-08
fix performance degradation due to nested omp

hhorii:
### Summary

nested omp was enabled in #796 

### Details and comments

OMP has significant overheads on nested omp loops.
#796 introduced nested omp loops of `parallel_experiments_` and `parallel_state_update_` even though `parallel_experiments_`  is `1`.


hhorii:
Simulation time for Mg2H was increased from 30s to 1630s.

vvilpas:
I don't have much experience with `openmp` so I don't fully understand. Why `#pragma omp parallel for if (parallel_experiments_ > 1) num_threads(parallel_experiments_)` is not equivalent to the `if/else` structure? I thought the `if` within the `pragma` did exactly that.

src/controllers/controller.hpp
```diff
@@ -522,12 +522,23 @@ Result Controller::execute(std::vector<Circuit> &circuits,
...
-    #pragma omp parallel for if (parallel_experiments_ > 1) num_threads(parallel_experiments_)
-    for (int j = 0; j < result.results.size(); ++j) {
-      // Make a copy of the noise model for each circuit execution
-      // so that it can be modified if required
-      auto circ_noise_model = noise_model;
-      execute_circuit(circuits[j], circ_noise_model, config, result.results[j]);
+    // then- and else-blocks have intentionally duplication.
+    // Nested omp has significant overheads even though a guard condition exists.
+    if (parallel_experiments_ > 1) {
+      #pragma omp parallel for if (parallel_experiments_ > 1) num_threads(parallel_experiments_)
```
vvilpas, line 528: Do we still need the `if (parallel_experiments_ > 1)` within the `pragma`? Wouldn't this suffice now:
```
#pragma omp parallel for num_threads(parallel_experiments_)
```

hhorii: We need this `if`. Even the guard `parallel_experiments_ > 1` exists in pragma, OMP recognizes it is nested inside the loop body.

vvilpas: So we need both `if`s?  In the first branch in the `pragma` we will never have `parallel_experiments_` <= 1,  right? Or am I missing something? Before #796 the code was:
```
if (parallel_experiments_ > 1) {
      // Parallel circuit execution
      #pragma omp parallel for num_threads(parallel_experiments_)
```
without `if` in the `pragma` statement. 
Sorry to insist, just want to be sure.

hhorii: sorry. You are right. `if (parallel_experiments_ > 1)` is not necessary in pragma. I will remove it.

chriseclectic:
Is this a bug in OpenMP? I was under the same impression as Victor, if parallel experiments was 1 the `#pragra omp for if` should have skipped using omp for that block.

hhorii:
Maybe this depends on implementation of OpenMP.

atilag:
Are we totally sure that this was the cause of the performance degradation?
By watching the code, it should definitely not, furthermore, there's a lot of projects out there using the same pragma expression.
If this is a bug in OpenMP, we can test on Linux (gcc) and MacOS (clang), as both have totally different implementations.

Issue #790, opened 2020-06-10
Get the state vector as a dictionary


yaelbh:
### What is the expected behavior?

We have a use case where we query the statevector, which in our case is a basis state. If we use `get_snapshot` or the statevector snapshot, we obtain a vector of length 2^n. We would like to have the option to represent the statevector as a dictionary that maps basis states to amplitudes, where zero amplitudes do not have to be present.

yaelbh:
Create a new snapshot, which will probably call `vec2ket` of `utils.hpp`.

chriseclectic:
This will be handled by the new `SaveStatevectorDict` instruction

PR #820, opened 2020-07-08
Support for saving statevector as ket dictionary

gadial:
### Summary
This PR adds the feature requested in #790 of adding a snapshot command to the Statevector simulator which returns the statevector in a dictionary for the nonzero entries, making it much more efficient than returning the full statevector in cases most of it is zero.

### Details and comments
The implementation is straightforward and relies on the `vec2ket` method. Other than that, it is identical to the standard statevector snapshot.


mtreinish:
The pip check failure here should be addressed by #829 it was caused by a new release of a test dependency.

src/simulators/statevector/statevector_state.hpp
```diff
@@ -503,6 +499,9 @@ void State<statevec_t>::apply_snapshot(const Operations::Op &op,
     case Snapshots::statevector:
       data.add_pershot_snapshot("statevector", op.string_params[0], BaseState::qreg_.vector());
       break;
+    case Snapshots::statevector_ket:
+      data.add_pershot_snapshot("statevector", op.string_params[0], Utils::vec2ket(BaseState::qreg_.vector(), json_chop_threshold_, 16));
```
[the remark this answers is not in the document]
gadial: I added this for `QubitVector` but I'm not familiar with `QubitVectorThrust`. How can I iterate over the statevector entries for this class?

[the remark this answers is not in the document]
gadial: For now I just added a vector to ket transformation. But for `QubitVector` we have the efficient solution.

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -622,6 +624,12 @@ cvector_t<data_t> QubitVectorThrust<data_t>::vector() const
   return ret;
 }
 
+template <typename data_t>
+cdict_t<data_t> QubitVectorThrust<data_t>::vector_ket(double epsilon) const{
+    // non-optimized version; relies on creating a copy of the statevector
+    return AER::Utils::vec2ket(vector(), epsilon, 16);
```
chriseclectic, line 630: I think this going to copy the full GPU vector to system memory before doing the conversion. It would be better if we could avoid this @doichanj

qiskit/providers/aer/extensions/snapshot_statevector.py
```diff
@@ -21,12 +21,14 @@
 class SnapshotStatevector(Snapshot):
     """ Snapshot instruction for statevector snapshot type """
 
-    def __init__(self, label, num_qubits=0):
+    def __init__(self, label, num_qubits=0, dict_form=False):
```
chriseclectic, line 24: I think this can be removed, not worth adding this to a function thats about to be deprecated

test/terra/backends/test_qasm_simulator.py
```diff
@@ -46,6 +46,7 @@
 from test.terra.backends.qasm_simulator.qasm_save import QasmSaveDataTests
 # Snapshot tests
 from test.terra.backends.qasm_simulator.qasm_snapshot import QasmSnapshotStatevectorTests
+from test.terra.backends.qasm_simulator.qasm_snapshot import QasmSnapshotStatevectorKetTests
```
chriseclectic, line 49: These can also be removed, and instead add tests to the `qasm_save_statevector.py` file (see the current statevector save examples). For tests you can compare the output to `qi.Statevector(circ).to_dict()`

chriseclectic (changes requested):
Needs some tests of the new instruction, and still some of the legacy snapshot code to be removed.

Can you also update the save instruction release note with an entry on this instruction that matches the other entries?

src/simulators/statevector/qubitvector_thrust.hpp
doichanj: Additional buffer will be required to make chopped vector on GPU (or on CPU) and I think it is not easy to parallelize making chopped vector. I think we can not avoid copying whole vector to CPU or accessing to whole vector from CPU to make chopped vector.

chriseclectic (approved):
LGTM

PR #822, opened 2020-07-10
Add BLAS_LIB_PATH as ENV var

vvilpas:
### Summary

Pass BLAS_LIB_PATH to the build system as ENV variable. This will help solving issues around OpenBlas compilation, as it seems to be sometimes problematic when binary lib is not available at Conan and compilation is triggered.


vvilpas:
As `BLAS_LIB_PATH` can be now an ENV var, I'm thinking that it would be better to prefix it with `AER`: `AER_BLAS_LIB_PATH`. It will affect people using the previous name though. What do you think?

atilag:
I like putting AER prefix to our variables, so we can make distinction between our variables and thoe potentially used in our third-party build system deps.

PR #823, opened 2020-07-10
[WIP] Add flag for distributable build

vvilpas:
### Summary

Add a flag to configure a build for distribution. At this moment it will enable the dynamic_arch for OpenBlas but more options can be added as needed.


vvilpas:
Closing as openblas will not be distributed with conan for now.

PR #684, opened 2020-04-09
Add SIMD (AVX) extension for QubitVector

hhorii:
### Summary
Add SIMD (AVX) extension for `QubitVector`

### Details and comments
Issue #646 

- [x] Pass tests in AVX env
- [x] Pass tests in non-AVX env
- [x] Add runtime check for AVX2 support

#### Next steps

- Support `apply_diagonal_matrix`
- Improve `Fusion::estimate_cost()` accuracy

PR #824, opened 2020-07-10
Add AVX with separation of binaries of AVX and non-AVX

hhorii:
### Summary

This is a revised version of #684 .

### Details and comments
Python:
`src/simulators/statevector/qvintrin.cpp` is compiled with `-mfma -favx2 -DQV_AVX_ISOLATION`.
`qiskit/providers/aer/backends/wrappers/bindings.cc` is compiled with `-DQV_AVX_ISOLATION`
Then, `bindings.cc` does not call any AVX and FMA if platform does not support them by checking CPU types.

Standalone:
`contrib/standalone/qasm_simulator.cpp` is compled with `-mfma -favx2` because standalone will be compiled on machines that run them.

If `__AVX2__` is not defined, AVX impl is not called in Intel platform and is not compiled in other platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,6 +218,12 @@ else()
 	message(STATUS "No Thrust support enabled")
 endif()
 
+if(SKBUILD) # Terra Addon build
+	enable_cxx_compiler_flag_if_supported("-DQV_AVX_ISOLATION")
+else() # Standalone build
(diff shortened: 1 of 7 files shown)


atilag:
Thanks for the PR @hhorii , I took some of the things you made here to fix non-SIMD environments.
As I merged https://github.com/Qiskit/qiskit-aer/pull/684 , we can close this.

Issue #779, opened 2020-06-03
Noise transformation fails with the latest cvxpy version


mtreinish:
### Informations

- **Qiskit Aer version**: Master
- **Python version**: Any
- **Operating system**: Any

### What is the current behavior?

Running tests with cvxpy 1.1.0 installed fails:

```
test.terra.noise.test_noise_transformation.TestNoiseTransformer.test_clifford

-----------------------------------------------------------------------------

Captured traceback:

~~~~~~~~~~~~~~~~~~~

    Traceback (most recent call last):

      File "/home/travis/build/Qiskit/qiskit-aer/test/terra/noise/test_noise_transformation.py", line 221, in test_clifford

    results = approximate_quantum_error(error, operator_string="clifford")

      File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/qiskit/providers/aer/utils/noise_transformation.py", line 102, in approximate_quantum_error

    probabilities = transformer.transform_by_operator_list(

      File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/qiskit/providers/aer/utils/noise_transformation.py", line 388, in transform_by_operator_list

    probabilities = self.transform_by_given_channel(

      File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/qiskit/providers/aer/utils/noise_transformation.py", line 669, in transform_by_given_channel

    return self.solve_quadratic_program(P, q)

      File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/qiskit/providers/aer/utils/noise_transformation.py", line 757, in solve_quadratic_program

    prob.solve()

      File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/cvxpy/problems/problem.py", line 395, in solve

    return solve_func(self, *args, **kwargs)

      File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/cvxpy/problems/problem.py", line 743, in _solve

    data, solving_chain, inverse_data = self.get_problem_data(

      File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/cvxpy/problems/problem.py", line 498, in get_problem_data

    solving_chain = self._construct_chain(

      File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/cvxpy/problems/problem.py", line 651, in _construct_chain

    return construct_solving_chain(self, candidate_solvers, gp=gp,

[...]

PR #832, opened 2020-07-14
Remove pin for broken CVXPY and update tests

gadial:
### Summary
This change addresses issue #779 which caused the Clifford-based transformations to fail, and so crashing a test. Currently there is no known solution to this problem (which might originate in changes in CVXPY) so we have to remove this test for now.

### DIFF
--- a/test/terra/noise/test_noise_transformation.py
+++ b/test/terra/noise/test_noise_transformation.py
@@ -267,7 +267,14 @@ class TestNoiseTransformer(unittest.TestCase):
         self.assertErrorsAlmostEqual(results_1, expected_results_1)
         self.assertErrorsAlmostEqual(results_2, expected_results_2)
 
-
+    def test_clifford(self):
+        x_p = 0.1
+        y_p = 0.2
+        z_p = 0.3
+        error = pauli_error([('X', x_p), ('Y', y_p), ('Z', z_p),
+                             ('I', 1 - (x_p + y_p + z_p))])
+        results = approximate_quantum_error(error, operator_string="clifford")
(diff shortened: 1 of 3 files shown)


test/terra/noise/test_noise_transformation.py
```diff
@@ -212,15 +212,6 @@ def test_approx_noise_model(self):
 
         self.assertNoiseModelsAlmostEqual(expected_result, result)
 
-    def test_clifford(self):
```
mtreinish, line 215: Instead of removing this test can we skip it with a skip message about the bug. Or is it not providing any value and we don't need the coverage?

gadial:
> So besides my inline comment about using a skip decorator instead of deleting. Does this mean we can remove the version cap https://github.com/Qiskit/qiskit-aer/blob/master/requirements-dev.txt#L7 or is there still a potential bug and we're not triggering it anymore?

There *is* still a potential bug; trying to use the Clifford elements will 100% trigger it, and there might be other, nonstandard element sets which will cause this as well. How should we proceed?

test/terra/noise/test_noise_transformation.py
gadial: The good news is that cvxpy update 1.1.3 eliminated the problem, as far as I can see. The bad news is that the performance of the optimization engine is worse than before; I had to change the test's parameters in order to get something which works relatively well. But this is not a correctness issue.

SteveDiamond:
@gadial if it's not too much trouble for you, we're happy to look into performance issues for your problems. We just need code to run/profile.

PR #833, opened 2020-07-14
Add proper support/testing for noise in pulse simulator and move noise interpretation and RHS generation into PulseSystemModel

### LINKED ISSUE #827: Remove support for PersistentValue instructions in pulse simulator

`PersistentValue` instructions are about to be removed from `pulse`, so code related to this in pulse simulator can be removed.

### LINKED ISSUE #830: Add testing/interface support for Lindblad noise models in pulse simulator

Add proper support for Lindblad noise models.

While being technically unsupported, Lindblad noise models can be included through the Monte Carlo algorithm in the pulse simulator via string format through the `backend_options` argument.

Steps for adding:
- Add proper tests
    - there is currently a very rudimentary test for this, but more stringent tests need to be added
- Add usable interface
    - E.g. add noise specification to duffing oscillator model constructor

DanPuzzuoli:
### Summary

Closes #830

Goals for this PR:

- Add tests for the monte carlo solver
- Move the noise parsing and RHS function generation into `PulseSystemModel` (fully compartmentalizing the "DE model" evaluation and making noise operators a proper part of the model)
- Add T1/T2 noise options to the duffing system model generators

### Details and comments

Currently working on adding tests for noise operators before rearranging any of the objects/computations (previous version of this PR tried opposite order but I broke something with the noise handling). 

- Added some tests for simple 2-level models that are analytically solvable
- Have begun working on a test for a 3 level duffing oscillator with T1 noise. In this case I haven't attempted to analytically solve it but have written an independent simulation of the Lindblad master equation, and have compared by eye the output counts from the MC solver to those expected by looking at the density matrix in the independent simulation, and things look good. 
     - Additional note: in doing this I have discovered that for systems with `dim > 2`, the measurement operators are `{P, 1-P}`, where `P` is the projection onto the first excited state, and as such states above the first excited state also return `0` when measured.


DanPuzzuoli:
Updates:
- Completed tests mentioned in original comment
- Successfully moved noise parsing and RHS generation into `PulseSystemModel`
     - Removed warning regarding noise models being untested
     - Kept the ability to specify noise model in `backend_options` - now it will override the one already in `PulseSystemModel`
- Added a deep copy method `copy()` to `PulseSystemModel` - the RHS generation stuff requires modifying state, so now the `pulse_controller` copies the input system model at the beginning

DanPuzzuoli:
Things are working though the MC simulations are very slow - I've been playing with a notebook to be a tutorial-like test of the simulator with a duffing model. I'm simulating a rabi experiment with:
- 3 level duffing oscillator
- T1 noise (that I've set to be very long)
- 32 rabi drive strengths
- 32 shots/drive strength
This takes about 4.67 minutes on my computer (with 8 cores). Part of this is that the MC method requires full simulation for each shot, and the above has a total of `1024` shots. However, one of the noise tests uses `1000` shots (for only a single experiment to simulate), and it only takes about 4 seconds.

The difference in the speeds may be due to the cost of setting up parallel execution, as parallelization is over shots in a given experiment (with parallelization set up each time for each experiment). I'm going to try parallelizing over experiments and shots simultaneously and see if that speeds things up. Another potential source for the difference in speeds is that the frequencies in the notebook I'm using are much higher, but changing these in the tests didn't result in any differences in test speed.

Another note: I've implemented an alternate file `mc_controller0` that runs the monte carlo algorithm using the `scipy-RK45` integrator, and using an event function to halt integration if the norm dips below a given value (the current monte carlo algorithm essentially handles the event "manually"). I had hoped this would eliminate some python loops and speed things up but as is typical, the `RK45` integrator is generally a bit slower than the `zvode` + manual event handling, though interestingly in the notebook I'm testing it is a bit faster (it takes about 4 mins - not drastic but not insignificant either).

DanPuzzuoli:
Modified parallelization scheme in `mc_controller` and `mc_controller0` to both parallelize over experiments and shots simultaneously and have seen a significant improvement in performance:
- `mc_controller` now takes about 2.5-2.67m to run the above mentioned simulations
- `mc_controller0` takes about 1.67m to run the above mentioned simulations

I've realized now that the correct test to compare these speeds to is `test_3d_osc_noise` as it contains noise, drift, and a driving term. The speeds here seem consistent compared to the time it takes to run that test (adjusting for shots).

DanPuzzuoli:
This PR is ready for review.

Summary of changes:

- Moved RHS generation into `PulseSystemModel`. This is involved:
     - Eliminating `PulseInternalDEModel` from `pulse_controller`, and moving the internal attributes of `PulseInternalDEModel` into `PulseSystemModel` (`_config_internal_data` and `_init_rhs`)
     - Added a `copy` method to `PulseSystemModel` - the moving of the RHS generation/configuration into `PulseSystemModel` results in modification of internal state, and the `copy` method allows the user-held object to remain unchanged.
     - Changed simulation routines to directly work with the `PulseSystemModel` object.
- Added `_noise_from_dict` method to `PulseSystemModel`. This allows the addition of a noise model (in `dict` format) directly to the model object (no longer required to pass through `backend_options` to the pulse simulator)
     - The untested warning for noise has been removed.
- Added T1 and T2 noise to the duffing system model generator, using [this paper supplement](https://arxiv.org/src/1712.06550v2/anc/threeq_supp.pdf) as a reference. 
     - Generators using T1 and T2 noise are set to lead to consistent density matrix element decay times as are given in that paper.
- Added tests:
     - `test_duffing_model_generators` - added tests to ensure proper construction of T1 and T2 noise operators.
     - `test_pulse_simulator` -
          - Added the tests `test_1Q_noise_depolarizing_no_drift`, `test_1Q_noise_dephasing_no_drift`, `test_1Q_noise_dephasing_with_drift`, `test_1Q_noise_x_gate`, `test_3d_osc_noise`, and `test_duffing_model_noise`
          - The first 3 are tested against analytic solutions, `test_1Q_noise_x_gate` is the same as `test_x_gate` but has such a high level of amplitude damping noise that it's expected that the final state should be driven back to 0, and  `test_3d_osc_noise` and `test_duffing_model_noise` test the results against an independent simulation added to `pulse_sim_independent`
- Other changes
    - Rewrote the parallelization in `mc_controller`. Previously, the experiments were looped over serially, and the parallelization occurred over the shots within a single experiment. Now, the parallelization is over all experiments/shots simultaneously. This seems to have reduced the cost of parallelization considerably (see previous comments).

Things I'm looking for comment on:
- I kept the ability to specify noise in `backend_options`. Should this be removed?
- I'm not totally happy with the `test_duffing_model_noise` test. I would have liked to directly measure the decay of each density matrix element, but this isn't possible with the MC simulator and measurement in the computational basis. The consistency with the independent simulation is perhaps enough though.

DanPuzzuoli:
Additional note: I have verified that the T1/T2 introduced in the duffing oscillator model is consistent (for the dim=2 case) with the Choi representation given in `standard_errors.thermal_relaxation_error`

vvilpas (commented):
LGTM!

Issue #789, opened 2020-06-10
Setting the initial state to a basis state


yaelbh:
### What is the expected behavior?

We have a use case where the initial state is a basis state. Right now we have two options:
1. Using the `initialize` instruction. This involves creating a 2^n state vector outside of the simulator and transmitting it to the simulator.
2. Inserting X gates at the beginning of the circuit. This has the cost of increasing the circuit depth, affecting run time.

I suggest to do at least one of the following:
1. Allow `initialize` to receive the vector in the form of a dictionary that maps basis states to amplitudes, where zero amplitudes do not have to be present.
2. Implement an optimization in the simulator that simulates a sequences of Pauli gates in one matrix loop.

yaelbh:
Option 2 above is related to #715.

yaelbh:
The rest of this issue is for Option 1. Issue #715 is for Option 2. We will probably need to write a function `ket2vec` in `utils.hpp`, similar to the existing `vec2ket`. It will be nice if Terra eventually supports `initialize` that accepts a ket; in the meantime we can create our own instruction and add it to the set of basis gates when transpiling. Alternatively, it may be possible to hack parameters and data types of `initialize`.

pradkrish:
Hello, is this issue still open? I am new to the project, happy to work on this with some guidance :)

PR #834, opened 2020-07-16
Added initialize_ket instruction to Aer

gadial:
### Summary
This PR addresses issue #789. It adds the instruction `initialize_ket` to `Circuit` which acts like the standard `initialize` instruction except that the target statevector is specified using dictionary and not vector; this results in a more compact representation that needs to be passed to the simulator.


qiskit/providers/aer/extensions/initialize_ket.py
```diff
@@ -0,0 +1,263 @@
...
+from qiskit.circuit import Instruction
+from qiskit.circuit.library.standard_gates.x import CXGate
+from qiskit.circuit.library.standard_gates.ry import RYGate
+from qiskit.circuit.library.standard_gates.rz import RZGate
+from qiskit.circuit.reset import Reset
+
+_EPS = 1e-10  # global variable used to chop very small numbers to zero
+
+
+class InitializeKet(Instruction):
```
chriseclectic, line 34: This function should be added to Terra. If it currently works with unrolling to reset + basis gates like regular initialization we could try and add it to the 0.15 release which is coming out any day now.

src/framework/operations.hpp
```diff
@@ -571,6 +574,44 @@ Op json_to_op_initialize(const json_t &js) {
...
+                                    comp.dump() + " invalid).");
+      // Get amplitude string
+      std::string key = comp[0];
+      // Get complex coefficient
+      complex_t coeff = comp[1];
+      ketmap[key] = coeff;
+    }
+  }
+
+  op.params = Utils::ket2vec(ketmap, nint, 16);
```
chriseclectic, line 603: Instead of converting to a dense vector here we could be more efficient by defining a new OpType `initialize_ket`, and storing just the non-zero coefficients and the index positions of the string labels in `params` and `regs` (or `string_params`) respectively: ie something like:

`{'00': 0.707.. , '11': 1j * 0.707}` -> `params = [0.707..., 1j * 0.707...], regs = [[0, 3]]`

Note `regs` here isn't a perfect field for storing this (it stores vectors of ints) so we just put all the coeffs as the first entry of `regs`. We could have used string params and stored the bit-strings but that would just delay conversion of strings to ints later which would be less efficient if being called over multiple shots.

To initialize a vector using this in the state class we can then initialize an empty vector/matrix, and then iterate over params/regs[0] and update the corresponding vector entries.

Implementing this might involve a bigger change to how general initialization is handled in the controller class so maybe we can leave this as parsing to a regular initialize instruciton now, and implement this ket OpType as a followup (and while doing so fix initialization for the non-statevector simulators to work properly)

qiskit/providers/aer/extensions/initialize_ket.py
```diff
@@ -0,0 +1,263 @@
...
+    flexible collection of qubit registers (assuming the qubits are in the
+    zero state).
+    Note that Initialize is an Instruction and not a Gate since it contains a reset instruction,
+    which is not unitary.
+    """
+
+    def __init__(self, params, num_qubits):
+        """Create new initialize composite.
+
+        params (list): vector of complex amplitudes to initialize to
```
chriseclectic, line 47: The types here look wrong, it should be a dict right?

src/framework/operations.hpp
gadial: We're getting the coefficients/index values already in this suggested format (via `js["params"]`) and can indeed use it in a new initialization op as you suggested; but I think it's best to delay this to the future, again as you suggested.

qiskit/providers/aer/extensions/initialize_ket.py
gadial: Yes, that's a silly copy-paste error.

qiskit/providers/aer/extensions/initialize_ket.py
gadial: It needs some sort of ket2vec to mimic what the regular initialize does. Do we have one in Python? If not, I'll write one myself and transfer to Terra.

yaelbh:
The following pull request may be related:  https://github.com/Qiskit/qiskit-terra/pull/5190
In Terra, probably the way to go is to allow initialization of `Statevector` from a dictionary. Then either to directly allow `initialize` to receive a dictionary; or users can indirectly construct their state vector from a dictionary, and send it to `initialize` (after https://github.com/Qiskit/qiskit-terra/pull/5190 is ready).

hhorii:
Please re-create a new PR with the latest branch if this requirement still exists.

PR #840, opened 2020-07-21
Move Aer to it's own package

mtreinish:
### Summary

Namespace packages are constant source of problems for users. The python
packaging ecosystem around splitting packages across namespaces is
fragile at the best of times and can often leave a you with an
environment that isn't recoverable (especially when mixing install
methods). There is also a performance hit whenever there is a piece of
the namespace we allow external packages to extend since it requires
doing a full python path search which can be slow depending on the
backing I/O and the number of paths in sys.path for an environment. This
commit addresses the piece from the aer perspective by moving
qiskit.providers.aer to it's own package and namespace 'qiskit_aer'.

This will be coupled with a change in terra that removes the arbitrary
namespace hook points and hard coding the element namespace maps via a
custom import loader at the root of the namespace. This has 2 advantages
it removes the use of namespace packages so the fragility and
performance impact are fixed since every element will be renamed to use
'qiskit_*' instead of 'qiskit.*', but it also makes it explicit where we
extend the namespace. The previous method allowed any package to extend
qiskit.* and qiskit.providers.* with whatever they wanted. This commit
will need to be coordinated with the terra change to ensure we don't
block development, because while it's not breaking for end users as a
coordinated code release piecewise it's a breaking change for each
element, including Aer.

### Details and comments

Depends-On: Qiskit/qiskit-terra#4767
Fixes Qiskit/qiskit#559


mtreinish:
This is failing unittests because of a weird circular import error. I originally never encountered this locally prior to merging Qiskit/qiskit-terra#4767 but by switching to `python setup.py bdist_wheel` instead of pip install I have been able to recreate this failure. I have no idea on the source of it as of yet. Since this will likely take me some time to debug I'm setting this as on hold as a revert is proposed at Qiskit/qiskit-terra#5006

qiskit_aer/extensions/snapshot_density_matrix.py
```diff
@@ -15,10 +15,10 @@
 """
 
 from qiskit import QuantumCircuit
-from qiskit.providers.aer.extensions import Snapshot
+from qiskit_aer.extensions import snapshot
 
 
-class SnapshotDensityMatrix(Snapshot):
+class SnapshotDensityMatrix(snapshot.Snapshot):
```
chriseclectic, line 21: Snapshot is used in the code below, which is why the CI is giving an import error.

PR #848, opened 2020-07-29
Rearrange CONTRIBUTING guide

### LINKED ISSUE #845: Modify CONTRIBUTING.md (for Linux) to indicate that g++ must be installed before installing from requirements-dev.txt
**requirements-dev.txt** includes the **cvxpy** package, which requires a C++ compiler. I was building Aer on WSL2 on my windows system and the `pip install -r requirements-dev.txt` failed to install cvxpy, because I was missing g++. 

I assume running `sudo apt install build-essential` will install g++ and mitigate this error, so maybe we could switch the order of commands and document that build-essential needs to be installed before installing the python packages?

atilag:
### Summary
Fixes #845 

Rearrange CONTRIBUTING guide to put in first place the need for installing the C++ toolchain before setting up the Python environment, as some Python dependencies would require having a C++ toolchain already installed, like CVXPY.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -232,36 +232,8 @@ fixes.
 
 ## Install from Source
 
->  Note: The following are prerequisites for all operating systems
-
-We recommend using Python virtual environments to cleanly separate Qiskit from
-other applications and improve your experience.
-
-The simplest way to use environments is by using *Anaconda* in a terminal
-window
-
-    $ conda create -y -n QiskitDevEnv python=3
-    $ conda activate QiskitDevEnv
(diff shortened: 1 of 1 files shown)


atilag:
@aditya-giri  ^

aditya-giri:
Ah, I was just about to raise a PR for this myself!

aditya-giri:
I also took the liberty of adding a few more links so that users could jump to their platform specific instructions instead of scrolling. Maybe you could take a look at this PR too: #849.

atilag:
Ok, I take yours :)

PR #849, opened 2020-07-29
fix #845: restructure contributing.md

### LINKED ISSUE #845: Modify CONTRIBUTING.md (for Linux) to indicate that g++ must be installed before installing from requirements-dev.txt

aditya-giri:
### Summary

fix #845 by asking developers to first install OS-specific dependencies before installing from requirements-dev.txt.
This is intended to solve the potential issue of g++ not being available to build `cvxpy` which is listed in requirements-dev.txt.

### Details and comments

- Separated the `dependencies` and `build` sections under each platform for easier linking.
- Reworded instructions so that users are pointed to platform specific instructions first.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -237,21 +237,26 @@ fixes.
 We recommend using Python virtual environments to cleanly separate Qiskit from
 other applications and improve your experience.
 
-The simplest way to use environments is by using *Anaconda* in a terminal
+- The simplest way to use environments is by using *Anaconda* in a terminal
 window
-
+```
     $ conda create -y -n QiskitDevEnv python=3
     $ conda activate QiskitDevEnv
+```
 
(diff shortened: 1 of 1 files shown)


aditya-giri:
The updated section will look like this: https://github.com/aditya-giri/qiskit-aer/blob/fix-845/CONTRIBUTING.md#install-from-source

aditya-giri:
Thanks @atilag 😃

PR #855, opened 2020-08-01
Versions of GCC before 9.1 lack the assignment operator in strstream.…

MarcusGDaniels:
…  Use a scoped constructor instead of an assignment.

### Summary

Versions of GCC before 9.1 did not declare an assignment operator for strstream.   Avoid the need for that by using a locally scoped constructor.


atilag:
Tests were failing because of one unrelated test, so I updated the branch with `master`, once all tests pass I'll merge.

Issue #843, opened 2020-07-24
Migrate current deprecated CUDA build system to native language support in CMake


atilag:
With the inclusion of SIMD in our codebase, we build parts of the code with SIMD support, and other parts without it.
There seems to be a bug in `FindCUDA.cmake` that doesn't allow us to build files with specific compiler flags, so we cannot set the SIMD flags for the parts of the code that supports it, we have to set these flags "globally" for the cmake target and this will conflict with systems running Aer with no SIMD support.

### What is the expected behavior?
The new way to build CUDA files in CMake allow us to set different flags to different source code files, so we should migrate our current CUDA build implementation to this new one.

vvilpas:
Implemented in #859

Issue #787, opened 2020-06-09
Fix Windows (MSVC) builds for GPU simulators


atilag:
### Informations

- **Qiskit Aer version**: All
- **Python version**: All
- **Operating system**: Windows

### What is the current behavior?
There are build errors in GPU code while trying to build on Windows (MSVC).

### Steps to reproduce the problem
```bash
> python setup.py bdist_wheel
```

### What is the expected behavior?
We should be able  to build without problems

### Suggested solutions
We need to wait for `pybind11` to release their new version and after testing that everything is still working and there are no regressions whatsoever, upgrade our dependencies.
There seems to be other errors related to the OpenMP version supported by MSVC and the presence of `unsgined` indices in parallelized loops.

MarcusGDaniels:
What fixes are you looking for in pybind11?   I have a Windows machine I could use to get it working.   Yeah, I see the travis tests are failing.  That's a drag.

MarcusGDaniels:
Here is one issue:

C:\Users\MarcusDaniels\.conda\envs\QiskitDevEnv\Include\pybind11\cast.h(1495): error: expression must be a pointer to a complete object type

That's an operator defined like this:

explicit operator type&() { return *(this->value); }

MarcusGDaniels:
Ah, I see it has been fixed on their master branch:  

https://github.com/pybind/pybind11/commit/c776e9ef937e3286c96f801ce2c4221f245e1bff

MarcusGDaniels:
With pull request #859 and this change, and copying pybind11 headers to a toplevel include/ directory, this build problem can be fixed.   I'll make another pull request if you are willing to temporarily copy the latest pybind11 headers as a workaround until Conda updates its pybind11 package.

diff --git a/qiskit/providers/aer/backends/wrappers/CMakeLists.txt b/qiskit/providers/aer/backends/wrappers/CMakeLists.txt
index bef44c9..fa150c4 100644
--- a/qiskit/providers/aer/backends/wrappers/CMakeLists.txt
+++ b/qiskit/providers/aer/backends/wrappers/CMakeLists.txt
@@ -16,6 +16,9 @@ basic_pybind11_add_module(controller_wrappers "${AER_SIMULATOR_SOURCES}")
 if(AER_THRUST_BACKEND STREQUAL "CUDA")
     set_source_files_properties(bindings.cc PROPERTIES LANGUAGE CUDA)
[...]

PR #859, opened 2020-08-05
Switch to using native support for CUDA instead of FindCUDA macros.

MarcusGDaniels:
### Summary

This switches to native support in CMake (versions >= 3.8) so that CUDA-specific CMake macros aren't used to set up the build.   It is to address issue #843 and make it possible to do per-CUDA file build options.


MarcusGDaniels:
I reopened this pull request because the original one had a bad e-mail in it.  (One that I no longer have.)
Also a small change to this will address issue #787.  It amounts to including the current pybind11 headers in the source tree and adding a conditional in the bindings CMakefile to reference them.  It appears that Qiskit's license and pybind11's are compatible, perhaps at least for the purposes of a temporary workaround until there is a newer Conda install for pybind11.

CMakeLists.txt
```diff
@@ -283,36 +309,48 @@ if(SKBUILD) # Terra Addon build
...
 	set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-	if(CUDA_FOUND)
-		cuda_add_executable(qasm_simulator ${AER_SIMULATOR_SOURCES})
+	if(CUDA_FOUND AND AER_THRUST_BACKEND STREQUAL "CUDA")
+		set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES LANGUAGE CUDA)
+		set_source_files_properties(${AER_SIMULATOR_SOURCES} PROPERTIES LANGUAGE CUDA)
+		set_source_files_properties(${AER_SIMULATOR_SOURCES} PROPERTIES COMPILE_FLAGS "${CUDA_NVCC_FLAGS}")
+                set(L ${SIMD_FLAGS_LIST})
+  		list(TRANSFORM L PREPEND " --compiler-options ")
+  		string(REPLACE ";" " " SIMD_FLAGS ${L})
```
vvilpas, line 326: It would be great for readability and maintainability to wrap in a function these kind of transformations that are used in several places and in different CMakeFiles

vvilpas (changes requested):
Thanks @MarcusGDaniels to take care of this! I have started reviewing, leave some comments.

CMakeLists.txt
```diff
@@ -129,7 +148,13 @@ setup_conan()
...
 # If we do not set them with a space CMake fails afterwards if nothing is set for this vars!
 set(AER_LINKER_FLAGS " ")
-set(AER_COMPILER_FLAGS " ")
+
+# This limit is hit building the Python plugin for CUDA
+if(APPLE OR UNIX)
+	set(AER_COMPILER_FLAGS " ")
+elseif(MSVC)
+	set(AER_COMPILER_FLAGS " /bigobj")
+endif()
```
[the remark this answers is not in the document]
MarcusGDaniels: > I think it is safe to remove this now and leave the original code, as some default windows flags are now set above (line 131)

I checked this on a Windows build and it is still necessary for CUDA builds.  Presumably because it is not C++ as far as CMake is concerned.  (The addition of /bigobj is for C++ files only.)   I have a follow-up branch that tweaks the Pybind11 header file so that WIndows builds work, and I reinstalled that change there for now.

CMakeLists.txt
```diff
@@ -223,18 +242,10 @@ if(AER_THRUST_SUPPORTED)
...
-		if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64" OR CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "AMD64")
-			if(APPLE OR UNIX)
-				set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS}; --compiler-options;-mfma,-mavx2")
-			elseif(MSVC)
-				set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} /arch:AVX2")
-			endif()
-		endif()
+
+		string(REPLACE ";" " "  AER_CUDA_ARCH_FLAGS_EXPAND "${AER_CUDA_ARCH_FLAGS}")
+		set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -ccbin \"${CMAKE_CXX_COMPILER}\" ${AER_CUDA_ARCH_FLAGS_EXPAND} -DAER_THRUST_CUDA -std=c++14 -I${AER_SIMULATOR_CPP_SRC_DIR} -isystem ${AER_SIMULATOR_CPP_SRC_DIR}/third-party/headers -use_fast_math --expt-extended-lambda")
```
vvilpas, line 247: Have you ever experience a build failure saying something like "redefinition of std"? I get that error on one linux machine (but it runs fine on others).  It seems the flag `-std=c++14` is added twice on the `CUDA` compile commands: `nvcc .... -std=c++14 ... -std=c++14 ...`. So I removed the c++14 flag here and it compiles everything fine but fails at linkage with undefined symbols errors.

MarcusGDaniels: Haven't seen that.   I'll try removing it to see if it breaks my builds.

MarcusGDaniels: Builds work for me without -std-c++=14 in that spot on Ubuntu 20.04 LTS with CUDA 10.1.  There, I have to use -DCMAKE_CXX_COMPILER=g++-8.  Also it builds ok (again that way) on Windows 10 with CUDA 11. A guess is that the unresolved symbols arose by building half with one option, and another half with another option?

vvilpas: I don't really know. Don't worry, I'll check what is different on this other linux machine and try to make it work.

vvilpas:
Hi @MarcusGDaniels. Can you merge this PR https://github.com/MarcusGDaniels/qiskit-aer/pull/2? With this commit we will temporarily enable our CI jobs to build GPU wheels. That way we can check that we will be able to build the pypi GPU package.

MarcusGDaniels:
> Hi @MarcusGDaniels. Can you merge this PR [MarcusGDaniels#2](https://github.com/MarcusGDaniels/qiskit-aer/pull/2)? With this commit we will temporarily enable our CI jobs to build GPU wheels. That way we can check that we will be able to build the pypi GPU package.

Sure, just let me know when.

vvilpas:
Good! Let's see if we are able to build the gpu packages.
Be careful when merging master again with this branch, because if we bring this commit with it, https://github.com/Qiskit/qiskit-aer/pull/998, we may accidentally publish the sdist wheel. 
To be safe, just revert my PR before merging master here.

MarcusGDaniels:
> Good! Let's see if we are able to build the gpu packages.
> Be careful when merging master again with this branch, because if we bring this commit with it, #998, we may accidentally publish the sdist wheel.
> To be safe, just revert my PR before merging master here.

Looks like it can't find the CUDA compiler in that environment.

MarcusGDaniels:
> Good! Let's see if we are able to build the gpu packages.
> Be careful when merging master again with this branch, because if we bring this commit with it, #998, we may accidentally publish the sdist wheel.
> To be safe, just revert my PR before merging master here.

Here's what I find:

1) it is necessary to get /usr/local/cuda/bin into the PATH somehow.  I hard-coded it for now.   I'm not sure where to insert this PATH change into the workflow manager (as opposed to CMakefiles).  

2) There this problem:

     + sh -c 'python3 /project/tools/verify_wheels.py'
terminate called after throwing an instance of 'thrust::system::system_error'
  what():  for_each: failed to synchronize: cudaErrorNoDevice: no CUDA-capable device is detected

MarcusGDaniels:
>     1. it is necessary to get /usr/local/cuda/bin into the PATH somehow.  I hard-coded it for now.   I'm not sure where to insert this PATH change into the workflow manager (as opposed to CMakefiles).

Ok, sorted out this problem.   Not sure what to do about runtime testing, though.  IIRC there was no solution GPU continuous integration?

MarcusGDaniels:
The test looks ok except for the #2 issue of verify_wheels.py failing due to CUDA-capable devices being unavaiable.

vvilpas:
hmm, not sure what is happening. @mtreinish can you help us with this? verify_wheel is failing with `CUDA-capable devices being unavailable` and @MarcusGDaniels had to add `nvcc` to the path.

mtreinish:
> hmm, not sure what is happening. @mtreinish can you help us with this? verify_wheel is failing with `CUDA-capable devices being unavailable` and @MarcusGDaniels had to add `nvcc` to the path.

The GPU wheel job works fine today as is. You can see the build logs for the recent 0.7.0 release gpu wheels here: https://github.com/Qiskit/qiskit-aer/runs/1261468922?check_suite_focus=true#step:5:23 (although since they're so long you'll have to download them since the gha log viewer won't show the full log)

The `CUDA-capable devices being unavailable` error is weird because that means something is now trying to use a GPU, which shouldn't happen unless you explicitly set a GPU enabled simulator method like `statevector_gpu`. verify wheels doesn't specify a backend method: https://github.com/Qiskit/qiskit-aer/blob/master/tools/verify_wheels.py#L446 but I don't think automatic method selection will ever pick a gpu simulator.

As for the nvcc path that is also weird because I thought the nvidia cuda RHEL/CentOS packages we install as the first part in the job add it to the default environment path for us. Like the current cmake files find nvcc find and build using it without issue, so I'm not sure what would change there.

MarcusGDaniels:
> As for the nvcc path that is also weird because I thought the nvidia cuda RHEL/CentOS packages we install as the first part in the job add it to the default environment path for us. Like the current cmake files find nvcc find and build using it without issue, so I'm not sure what would change there.

A change in this patch is getting away from the FundCUDA macros which guess at possible locations.

MarcusGDaniels:
I'm doing a test run now that just disables verify_wheels since it would be redundant with the non-GPU build, and it can't in principle run anyway, right?  Unless, as you say it exercises the CPU code path.

vvilpas:
I've been investigating a bit. At the beginning of execution the simulator checks which methods it can run. To do this it tries to simulate a small circuit for the different methods and checks if it was successful or not. If a method throws an exception, a catch block intercepts it and set the `success` field to `False`, setting the method to unavailable. This is what happened before when a 'gpu' method was selected and no device was available. Now, for some reason, we never reach the catch block and the program terminates when the exception is thrown.

MarcusGDaniels:
> Now, for some reason, we never reach the catch block and the program terminates when the exception is thrown.

Is there a way to run the code as generated where the test would run and get a stacktrace?  (e.g. under gdb)

vvilpas:
I've tried to get some info on the backtrace and in Release I haven't been able to clarify anything. 
So, I set up the docker image 'cibuildwheel' uses to do some debugging. Interestingly, in debug everything works as expected. I'll try to modify the code to get the same behavior as before.

vvilpas:
I have a potential solution. I asked our CUDA experts and I am waiting for their feedback.

vvilpas:
@MarcusGDaniels can you please merge https://github.com/MarcusGDaniels/qiskit-aer/pull/3? verify_tools should be working now with a small change I've made.

MarcusGDaniels:
> @MarcusGDaniels can you please merge [MarcusGDaniels#3](https://github.com/MarcusGDaniels/qiskit-aer/pull/3)? verify_tools should be working now with a small change I've made.

I did this merge, but now I'm having trouble wit urllib3 being the incorrect version.   Not sure what happened.

vvilpas:
> > @MarcusGDaniels can you please merge [MarcusGDaniels#3](https://github.com/MarcusGDaniels/qiskit-aer/pull/3)? verify_tools should be working now with a small change I've made.
> 
> I did this merge, but now I'm having trouble wit urllib3 being the incorrect version. Not sure what happened.

There is a compatibility problem between 2 conan dependencies: urllib3 and requests. They are fixing it and I've prepared a PR with a workaround in the meantime (#1031 ).

vvilpas:
I have limited CUDA arch to just 5.2 to check that CI works. It seems we are running out of time with all the archs enabled.

MarcusGDaniels:
> I have limited CUDA arch to just 5.2 to check that CI works. It seems we are running out of time with all the archs enabled.

Hmm, it's been running for hours.  I imagine it is stuck somehow but there is no log reported yet.

MarcusGDaniels:
> > I have limited CUDA arch to just 5.2 to check that CI works. It seems we are running out of time with all the archs enabled.
> 
> Hmm, it's been running for hours. I imagine it is stuck somehow but there is no log reported yet.

Yeah, timed out on veryify_wheels.py again.

vvilpas:
Yeah, I've already founded and hopefully fixed it (#1035). It was a change already introduced in master.

MarcusGDaniels:
> Yeah, I've already found it and hopefully fixed it (#1035). It was a change already introduced in master.

Looked like it worked except for a Mac build that ran out of time.   Is there an automatic heuristic that retriggers the build or is there a way to force one without doing a commit?

vvilpas:
> Looked like it worked except for a Mac build that ran out of time. Is there an automatic heuristic that retriggers the build or is there a way to force one without doing a commit?

Yeah, that was weird, but the full gpu build wheel action worked at last!!! 
I have disabled it so works again only for releases. If CI passes I think we can merge for upcoming release.

MarcusGDaniels:
> > Looked like it worked except for a Mac build that ran out of time. Is there an automatic heuristic that retriggers the build or is there a way to force one without doing a commit?
> 
> Yeah, that was weird, but the full gpu build wheel action worked at last!!!
> I have disabled it so works again only for releases. If CI passes I think we can merge for upcoming release.

Reminder:  There's still a small patch to get Windows working.  It involves copying a modified pybind header file into the distribution, though.

vvilpas:
> Reminder: There's still a small patch to get Windows working. It involves copying a modified pybind header file into the distribution, though.

No problem. I think we can address Win later as we don't officially support it yet. Pybind 2.6 is already released and fixes the error when compiling with nvcc.

chriseclectic (approved):

vvilpas (approved):
LGTM!

mtreinish (approved):

mtreinish:
We probably should add a release note about this because it bumps the minimum cmake version. But we can do that after it's added to #1036

PR #861, opened 2020-08-06
Add gate fusion to StatevectorSimulator and UnitarySimulator

chriseclectic:
### Summary

Adds the gate fusion circuit optimization used by the `QasmSimulator` to both the `StatevectorSimulator` and `UnitarySimulator` backends.

### DIFF
--- a/src/controllers/statevector_controller.hpp
+++ b/src/controllers/statevector_controller.hpp
@@ -303,13 +303,11 @@ void StatevectorController::run_circuit_helper(
   data.set_config(config);
 
   // Optimize circuit
-  const std::vector<Operations::Op>* op_ptr = nullptr;
-  Circuit opt_circ;
-  Transpile::Fusion fusion_pass(5, 20);
+  const std::vector<Operations::Op>* op_ptr = &circ.ops;
+  Transpile::Fusion fusion_pass(5, 20); // 20-qubit default threshold
   fusion_pass.set_config(config);
-  if (circ.num_qubits < fusion_pass.threshold || !fusion_pass.active) {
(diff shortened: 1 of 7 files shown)


src/controllers/statevector_controller.hpp
```diff
@@ -301,13 +302,26 @@ void StatevectorController::run_circuit_helper(
   // Output data container
   data.set_config(config);
 
+  // Optimize circuit
+  const std::vector<Operations::Op>* op_ptr = &circ.ops;
+  Transpile::Fusion fusion_pass(5, 20); // 20-qubit default threshold
+  fusion_pass.set_config(config);
+  Circuit opt_circ;
+  if (fusion_pass.active && circ.num_qubits >= fusion_pass.threshold) {
+    opt_circ = circ; // copy circuit
```
atilag, line 311: Do we really need a copy? Why not just apply the optimizations to `circ.ops`?

chriseclectic: Argument is const

atilag: We can remove "constness" here:
```c++
fusion_pass.optimize_circuit(const_cast<Circuit&>(circ), dummy_noise, state.opset(), data);
```
If that builds (I haven't tried but it should), we are getting rid of pointers and copies and temporary values.

chriseclectic: That seems sketchy if you are modifying what is supposed to be a const variable according to the function signature.

I'd rather rethink what should be const/non-const arguments in the base class API as a follow up.

PR #862, opened 2020-08-06
Revert management of openblas dependency through Conan

vvilpas:
### Summary

In the future we will need OpenBLAS+Lapack. This has two problems:
- For windows we need the clang family of compilers. Thus, it cannot be done with Conan and we need to manually compile the library and distribute it with some extra DLLs and libs.
- For macos & linux Conan doesn't provide a fortran compiler, needed by Lapack, so no binaries will be available from their CI. That would imply that everybody compiling AER, should compile the whole OpenBLAS+Lapack library at least once.

Given these points and the complexity of OpenBLAS compilation, I think we should go back to the previous approach: distribute openblas for windows and use system one for Macos and Linux.
Lapack will come in an upcoming PR, but given that we are going to release a version before that, I think there is no point in changing openblas management to Conan only for this release, deal with the problems that could arise, and then comeback to the original method in the next one.


atilag:
Due to the unfortunate circumstances, I agree this is the way to go :(

PR #775, opened 2020-06-01
Add job to try and build gpu package in CI

mtreinish:
### Summary

To avoid packaging issues we should automate the building the
qiskit-aer-gpu packages like what is done for the non-gpu packages. A
GPU shouldn't be required because the compilation with nvcc doesn't
actually run on a GPU. It is just a matter of figuring out how to get
nvcc installed and configured in the manylinux docker image. This commit
starts that process.

### Details and comments

PR #863, opened 2020-08-06
Only run GPU wheel jobs on tags and publish to pypi

mtreinish:
### Summary

The GPU wheel build job was added in #775, but merged a bit too early.
There was still a test commit merged on the PR branch which was being
used to switch the job to run on every PR and commit and disable the
push to pypi. This commit reverts that test commit so that we're only
running the 2+ hour wheel building job when we tag a release and also
uncomments the section to publish the built wheels to pypi.

### Details and comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,8 +1,8 @@
 name: Build
-on: [push, pull_request]
-#  push:
-#    tags:
-#      - '*'
+on:
+  push:
+    tags:
+      - '*'
 jobs:
   build:
     name: Build qiskit-aer-gpu wheels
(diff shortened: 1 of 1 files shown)


MarcusGDaniels:
There's no mechanism in place now to test GPU builds besides our local builds?

mtreinish:
> There's no mechanism in place now to test GPU builds besides our local builds?

There isn't, unfortunately we're a bit constrained by our CI resources. The builds take ~2 hours each which is too slow for a pre-merge testing job. Also, none of the CI providers provide nodes with GPUs (which is understandable) which limits how effective a test job would be, all we'd really be able to test is compilation. We've been discussing adding a post-merge or cron job that runs a a subset of the GPU build (like 1 or 2 compute capability versions and only 1 python version) to at least provide some compilation coverage without blocking development or eating too much job capacity while it spins.

MarcusGDaniels:
The CUDA compilation issues can be hairy and have obscure symptoms, as I am sure you know.    Especially because CMake intervenes and transforms the compilation rules in non-obvious ways.   Mac, Windows, and Linux all behaving differently.   Makes it hard to modify any of this code without having direct access to all of these systems.

PR #864, opened 2020-08-06
Fix/issue 856 build ppc

### LINKED ISSUE #856: Build failure on Power (ppc64le)
### Informations

- **Qiskit Aer version 0.6:
- **Python version 3.8:
- **Operating system Red Hat Enterprise Linux Server release 7.6:

It seems there are Intel specific codes and compiler options (AVX related things) set without checking the host system.

Intel specific options and codes have to be ignored by cmake or pre-processors on non-Intel systems.

### Suggested solutions
./CMakeLists.txt sets AVX options without checking host system (CMAKE_HOST_SYSTEM_PROCESSOR), so ignore setting if the system is not Intel.
Also ./qiskit/providers/aer/backends/wrappers/CMakeLists.txt has same issue.
And set(SIMD_SOURCE_FILE "${PROJECT_SOURCE_DIR}/src/simulators/statevector/qv_avx2.cpp") should be ignored on non-Intel systems.

In /qiskit/providers/aer/backends/wrappers/bindings.cc, including cpuid.h without setting include path can not find cpuid.h on Power Linux. (also ./src/framework/avx2_detect.hpp has issue on Power)

In ./src/controllers/qasm_controller.hpp using QubitVectorAvx2 class in run_circuit causes error on non-AVX systems.

vvilpas:
### Summary

Avoid compilation of AVX2 instruction in case we are not in x86_64 architecture. This will, hopefully, fix build problems in other architectures as stated in #856

Fixes #856


MarcusGDaniels:
Note that CMAKE_HOST_SYSTEM_PROCESSOR can also take on the value AMD64, such are autodetected on Windows machines, e.g. win-amd64-3.8.

vvilpas:
Hi @MarcusGDaniels!
Not sure if I understand, do you mean it can take other values different than AMD64 on windows?

MarcusGDaniels:
I mean it doesn't always take on "x86_64" when AVX2 is available, it can also be detected as "AMD64".   So those conditionals will disable the AVX2 support for some users.

vvilpas:
yeah. It is already added, maybe you missed the last commit?
In any case, it is work in progress and we still have to test on a ppc machine, so this may not be the final solution. Just take it into account if you are pulling code from here.

qiskit/providers/aer/backends/wrappers/CMakeLists.txt
```diff
@@ -6,14 +6,15 @@ find_package(Pybind11 REQUIRED)
...
-
-# We build SIMD filed separately, because they will be reached only if the
-# machine running the code has SIMD support
-set(SIMD_SOURCE_FILE "../../../../../src/simulators/statevector/qv_avx2.cpp")
-if(APPLE OR UNIX)
-    set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES COMPILE_FLAGS "-mfma -mavx2")
-elseif(MSVC)
-    set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES COMPILE_FLAGS "/arch:AVX2")
+if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64" OR CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "AMD64")
+    # We build SIMD filed separately, because they will be reached only if the
```
atilag, line 10: typo: `filed` -> `file` (was my  fault originally)

src/misc/common_macros.hpp
```diff
@@ -0,0 +1,11 @@
+
+#ifndef QASM_SIMULATOR_COMMON_MACROS_HPP
+#define QASM_SIMULATOR_COMMON_MACROS_HPP
+
+#if defined(__GNUC__) && defined(__x86_64__)
+ #define GNUC_AVX2 1
+#else
+ #define GNUC_AVX2 0
```
atilag, line 8: I think we can just define GNUC_AVX2 without any value, and saving the `#else` block
```suggestion
#define GNUC_AVX2
```
if `GNUC_AVX2` is not defined, then this expression will be evaluated to false.
```c++
#ifdef GNUC_AVX2
// True!
#else
// False!
#endif
```

src/simulators/statevector/qubitvector_avx2.hpp
```diff
@@ -76,6 +77,9 @@ class QubitVectorAvx2 : public QubitVector<data_t, QubitVectorAvx2<data_t>> {
   size_t calculate_num_threads();
 };
 
+// We do not define this functions in case we don't use AVX2
+// so it can compile, as this class won't be used
+#if defined(_MSC_VER) || GNUC_AVX2
```
atilag, line 82:
```suggestion
#if defined(_MSC_VER) || defined(GNUC_AVX2)
```

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -799,8 +799,8 @@ void QubitVectorChunkContainer<data_t>::SetupP2P(int nDev)
       for(i=0;i<nDev;i++){
         if(i != m_iDevice){
           cudaDeviceCanAccessPeer(&m_p2pEnable[i],m_iDevice,i);
-          if(m_p2pEnable[i])
-            cudaDeviceEnablePeerAccess(i,0);
+//          if(m_p2pEnable[i])
```
atilag, line 802: What happened here?... can't see the relation.
If we really want to disable this, just remove the lines.

vvilpas: @doichanj fixed and checked the CUDA compilation on PPC and he found there were some problems with the tests because of these lines, so he disabled them.

src/misc/common_macros.hpp
vvilpas: Absolutely!

PR #865, opened 2020-08-06
Update cibuildwheel version

mtreinish:
### Summary

This PR serves 2 purposes the first is to verify if our build publish jobs still work with the recent changes in packaging. The second is to bump the cibuildwheel version we're using to the latest to keep up to date.


mtreinish:
This should be good to go now

PR #873, opened 2020-08-08
[Stable] Fix bug in StatevectorSimulator preventing AVX2, GPU, and single-precision simulations (#872)

chriseclectic:
### Summary

Backport for 0.6.1 patch


mtreinish:
Before we push out 0.6.1 I'd like to see if we can find a fix for the issue causing the pulse tutorial failure in CI on qiskit/qiskit#998 which is looking like it is a pulse simulator issue

PR #881, opened 2020-08-12
Migrate most of CI to Github Actions

mtreinish:
### Summary

Since the period with an extra 10 concurrent job on travis is ending and
there is already extensive resource contention in the shared azure
pipelines instance for the qiskit organization we need to find an
alternative for CI to maintain job throughput. This commit migrates all
the existing CI jobs to github actions which has a very generous quota
for open source projects. The only jobs we'll still use travis for after
this are on non-x86 platforms because nothing else offers that.

### Details and comments

~Fix build system on windows~
~Add wheel publish jobs for windows~
- [x] Add docs job
- [x] Add lint job


mtreinish:
So I've managed to get the wheel builds working on windows (still need to figure out why the sdist build can't find MSVC) but I was only able to get this working with the latest version of visual studio. Doing some searching I've come across https://github.com/actions/virtual-environments/issues/387 which explains it, there is no vs2015 env installed on the vm. We still advertise vs2015 support so I'm thinking we might want to keep azure pipelines for windows just for now.

chriseclectic (approved):

PR #889, opened 2020-08-13
Add hermitian eigensolver

vvilpas:
### Summary

Add hermitian eigensolver

### Details and comments

This PR implements function to get eigenvalues and eigenvectors for a hermitian matrix with single precision or double precision complex values.
The more specialized function for PSD systems will be implemented in a later PR, as are more specific and need a bit more work. Meanwhile this will unblock #670, as every PSD matrix is hermitian.

This PR continues the work done on #673.


src/framework/linalg/eigensystem.hpp
```diff
@@ -0,0 +1,101 @@
...
+ * @param hermitian_matrix: The Hermitian matrix.
+ * @param eigenvalues: On output: the eignevalues of the matrix (input is overwritten)
+ * @param eigenvectors: On input: a copy of psd_matrix, on output: the eigenvectors of the matrix.
+ *
+ * @returns: void
+ */
+template <class T>
+void eigensystem_hermitian(const matrix<std::complex<T>>& hermitian_matrix,
+               /* out */ std::vector<T>& eigenvalues,
+               /* out */ matrix<std::complex<T>>& eigenvectors);
```
chriseclectic, line 36: Are the eigenvectors stored as the rows, or the columns of this matrix? This should be added to the documentation

src/framework/linalg/almost_equal.hpp
```diff
@@ -19,29 +19,109 @@
 #include <limits>
 #include <type_traits>
 
+#include "framework/matrix.hpp"
 #include "framework/linalg/enable_if_numeric.hpp"
 
+#define MAXDIFF 5.0*AER::Linalg::epsilon<T>::value
```
chriseclectic, line 25: What are the reason for these (and the 5.0 * factor)?

src/framework/linalg/eigensystem.hpp
vvilpas: Columns. I'll add to the documentation.

src/framework/linalg/almost_equal.hpp
vvilpas: These were introduced in https://github.com/Qiskit/qiskit-aer/commit/3589fe9b0d76076d794a2fce8de559a4c1607a53.
It seems they are used as default values for almost_equal functions but now with the 5.0 factor. I'm not sure if the 5.0 factor was introduced for some reason. Maybe @ereastman can shed some light?

chriseclectic (approved):
LGTM

Issue #132, opened 2019-04-05
Add additional test coverage for optimized gates


chriseclectic:
### What is the expected behavior?

Add additional test coverage for optimized gates in qubitvector class.

- [ ] mcx gate (currently only cx, ccx are tested using this)
- [ ] mcy gate
- [ ] mcz gate
- [ ] mcu1 gate
- [ ] mcu2 gate
- [ ] mcu3 gate
- [ ] mcswap
- [ ] apply_matrix for N-qubit gates (using unitary instruction)
- [ ] apply_diagonal_matrix for N-qubit gates (using unitary instruction)

jparr721:
Hi @chriseclectic, 

I am heavily interested in helping out here. I don't necessarily have a heavy background in the subject but I am eager to learn. How might the ideal form of this expanded test coverage look? Is there any literature I can leverage to get up to speed more quickly?

Thanks

chriseclectic:
Hi @jparr721 that would be great! If you're interested in helping out on this I just made some issues that might be a better place to start for the controlled-swap (#274) and controlled-u3 gates (#274), since these are both supported by Terra compiler to unroll into basis gates.

After that you could try these gates, but since there isn't a specific gate object in Terra yet you would have to define your own to build the tests, and you couldn't test the unrolling.

In general we should define class objects for these multi-controlled gates, but you could do a quick version for testing by directly using the gate objects (note that this will be defining them for a fixed number of qubits though). Ie something like

```python
from qiskit import QuantumCircuit, execute
from qiskit.circuit import Gate
from qiskit.providers.aer import QasmSimulator

# Make a 3-qubit controlled X gate
cccx = Gate("mcx", 4, [])   # 4-qubit gate

# Add to a quantum circuit where control bits are all set to 1
circ = QuantumCircuit(4, 4)
circ.x(0)
circ.x(1)
circ.x(2)
circ.append(cccx, [0, 1, 2, 3])
for j in range(4):
    circ.measure(j, j)

# Use custom basis gates because mcx isn't in QasmSimulator basis_gates yet
basis_gates = ['mcx', 'x']

# Get counts (should all be "1111")
result = execute(circ, QasmSimulator(), basis_gates=basis_gates).result()
result.get_counts(0)
```

abhik-99:
Hey @chriseclectic, are the test added? I'd like to help out in any way I can.

clarebirch:
[...]

PR #890, opened 2020-08-14
WIP: add tests for MCX gate (4-qubit)

ivanightingale:
### Summary
To fix one of the items in #132 .

### Details and comments
Adds tests and test reference for MCX gate (4-qubit).

### DIFF
--- a/test/terra/backends/qasm_simulator/qasm_noncliffords.py
+++ b/test/terra/backends/qasm_simulator/qasm_noncliffords.py
@@ -173,6 +173,26 @@ class QasmNonCliffordTestsCGates:
         self.compare_counts(result, circuits, targets, delta=0)
 
 
+class QasmNonCliffordTestsMCGates:
+    """QasmSimulator multi-controlled gates tests in default basis."""
+
+    SIMULATOR = QasmSimulator()
+    BACKEND_OPTS = {}
+    # ---------------------------------------------------------------------
+    # Test mcx-gate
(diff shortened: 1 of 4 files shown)


ivanightingale:
Hi @chriseclectic I've added deterministic qasm tests for MCX gates (4-qubit), basically in the same fashion as CCX tests were implemented. Could you please take a quick look and let me know if I'm going the right direction?

yaelbh:
Hi @ivanightingale. This is definitely the right direction. A few small comments:
1. I think you can add calls to this test also in `test_qasm_simulator_matrix_product_state` and `test_unitary_simulator`.
2. I recommend that in your environment you add debug prints to verify that all the simulation methods that you want to cover really run in the test.
3. I  have a suggestion for a general change in the non-Clifford tests. I'll open an issue for it. Please go ahead and continue with this pull request as it is, since it's independent of the issue that I'll open.

ivanightingale:
Hi @yaelbh , thanks for the feedbacks! I haven't added all the necessary tests yet (like the ones you mentioned) but I'll do them next since now I know I'm on the right track. I'll let you know when the pr is ready for review.

yaelbh:
Tests for multi-control gates were recently added in `qasm_standard_gates.py`. The file `qasm_noncliffords.py` has be removed, and `qasm_cliffords.py` tests only the stabilizer simulator. Therefore this pull request can be closed.

Issue #895, opened 2020-08-18
Expectation value of Y and variances are wrong in SnapshotExpectationValue


ikkoham:
### Informations

- **Qiskit Aer version**: master
- **Python version**: 3.8.5
- **Operating system**: macOS

### What is the current behavior?

```python
from qiskit import QuantumCircuit, execute, Aer
from qiskit.providers.aer.extensions import SnapshotExpectationValue

qc = QuantumCircuit(1)
qc.snapshot_expectation_value('x', [(1, 'X')], qubits=[0], variance=True)
qc.snapshot_expectation_value('y', [(1, 'Y')], qubits=[0], variance=True)
qc.snapshot_expectation_value('z', [(1, 'Z')], qubits=[0], variance=True)

result = execute(qc, Aer.get_backend('qasm_simulator'), shots=1024).result()
print(result.data()['snapshots'])
```

The output is 
```
{'expectation_value': {'x': [{'value': 0j, 'variance': 0j}],
  'y': [{'value': (1+0j), 'variance': 0j}],
  'z': [{'value': (1+0j), 'variance': 0j}]}}
```

Are these behaviors as expected? I'm concerned about the following points:
1. The expectation value of Y is 1, but should be 0.
2. All variances are 0.

ikkoham:
For the first problem, I think the following function is wrong.
https://github.com/Qiskit/qiskit-aer/blob/master/src/simulators/statevector/qubitvector.hpp#L1988

ikkoham:
I just found the expectaiton value of Y is correct for **`statevector_simulator`**.

```python
from qiskit import QuantumCircuit, execute, Aer

qc = QuantumCircuit(1)
qc.snapshot_expectation_value('x', [(1, 'X')], qubits=[0], variance=True)
qc.snapshot_expectation_value('y', [(1, 'Y')], qubits=[0], variance=True)
qc.snapshot_expectation_value('z', [(1, 'Z')], qubits=[0], variance=True)

result = execute(qc, Aer.get_backend('statevector_simulator')).result()
print(result.data())
```
the output is 
```
{'snapshots': {'expectation_value': {
'x': [{'value': 0j, 'variance': 0j}], 
'y': [{'value': 0j, 'variance': 0j}],
'z': [{'value': (1+0j), 'variance': 0j}]}},
'statevector': array([1.+0.j, 0.+0.j])}
```

I have checked this for other input states.

yaelbh:
[...]

PR #898, opened 2020-08-19
Fix evaluation y expectation

ikkoham:
### Summary

Fix first part of #895.

I think this needs more tests.
~~I will write later.~~
I have written it.

### DIFF
--- a/releasenotes/notes/fix-y-expectation-1ad355c31b4c3247.yaml
+++ b/releasenotes/notes/fix-y-expectation-1ad355c31b4c3247.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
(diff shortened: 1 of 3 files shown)


atilag:
I'm having a hard time figuring out what did you actually change, there seems to be a lot of formatting changes but not changes in the logic? (apart from tests).

ikkoham:
Thanks for the review.
The change can be seen in [my first commit](https://github.com/Qiskit/qiskit-aer/pull/898/commits/c9b9d09f0e78a5913d6dc9573f257535b3b15cc6).

The second commit is adding the test, and
the rest are format changes.

yaelbh (approved):
Looks good, thanks a lot for spotting and fixing it

ikkoham:
I've added a release note.

PR #902, opened 2020-08-23
Introduce by_measurements snapshot

### LINKED ISSUE #810: Expectation value with shot noise

For expectation value with `variance=True`, the simulator should perform measurements instead of direct calculation of the expectation value. And calculate the variance of course.

Currently a user who needs expectation value with shot noise has to create an equivalent set of circuits. This makes the simulator simulate the same circuit multiple times (a separate simulation for every term in the expectation value).

The mechanism of appending gates to change basis and measurement gates, simulating, and repeating for each term - should be done through the controller. In this way it will be identical for all the simulators.

yaelbh:
### Summary

I intend to resolve #810 through a series of small pull requests, each one will be created after the previous one has been reviewed and merged.  This first pull request creates a new type of snapshot in the interface (which is not yet supported by the simulator).

### DIFF
--- a/qiskit/providers/aer/extensions/snapshot_expectation_value.py
+++ b/qiskit/providers/aer/extensions/snapshot_expectation_value.py
@@ -26,7 +26,10 @@ from qiskit.providers.aer.extensions import Snapshot
 class SnapshotExpectationValue(Snapshot):
     """Snapshot instruction for supported methods of Qasm simulator."""
 
-    def __init__(self, label, op, single_shot=False, variance=False):
+    def __init__(self, label, op,
+                 single_shot=False,
+                 variance=False,
+                 by_measurements=False):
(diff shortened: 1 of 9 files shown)


yaelbh:
Travis is failing for two reasons:
- Building on Mac - I don't understand why.
- Docs:  
```
2020-08-23T19:47:30.7675396Z /home/vsts/work/1/s/.tox/docs/lib/python3.7/site-packages/qiskit/providers/aer/extensions/snapshot_expectation_value.py:docstring of qiskit.providers.aer.extensions.SnapshotExpectationValue:13:Inline substitution_reference start-string without end-string.
```
I don't understand why, and this line was not modified.

yaelbh:
Closing because this will probably done outside of the simulator, in the experiment class

PR #909, opened 2020-08-27
add qobj dump option for standalone testing

hhorii:
### Summary

This PR adds an option to dump qobj files into a directory for testing standalone qasm_simulator

### Details and comments

Add 'qobj_save_dir' an option of `backend_options`. Generated qobj files can be inputs of standalone qasm simulator.

```
backend_options = { 'qobj_save_dir':'/tmp'}
simulator = Aer.get_backend("qasm_simulator")
result = execute(qft, simulator, backend_options=backend_options).result()
```


chriseclectic (changes requested):
I don't think this is a good way to handle getting a serialized qobj, since you are also executing on a the python simulator.

In #486 I updated the AerBackend so the AerJob will contain the exact run qobj (including all config options). If we want to make this easier to construct without actually running the simulation we could add an `assemble` method to the AerBackends (which we will probably have to do eventually for the V2 provider anyway).

PR #913, opened 2020-08-31
Fix C++ for-loop by const ref rather than copy

chriseclectic:
### Summary

Fixes bug in C++ code where certain for-loops were iterating over copies rather than references of container elements.

### Details and comments

Fixes clang-format of other for loops so they are `const auto &` not `const auto&`


releasenotes/notes/fix-cpp-for-loop-refs-5a6bb879bf58a4fb.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes bug in C++ code where certain for-loops were iterating over copies
```
atilag, line 4: I'd not consider this a bug, more an improvement :)
I'd swear that the compiler is able to optimize this to avoid the copy, after all you cannot change a `const` variable, but agree that we shall use references and be explicit about it.

atilag (approved):

atilag:
Do we really want to create a patched version because of this?

chriseclectic:
> Do we really want to create a patched version because of this?

No this isn't an urgent fix.  I added the backport tag since if we end up releasing a patch for a different reason this could also be included in that patch.

PR #917, opened 2020-09-01
Drop support for Python 3.5

mtreinish:
### Summary

This commit drops support for running with python 3.5. It marks the
minimum supported version of the package as python 3.6, removes python
3.5 package pins, removes the 3.5 CI jobs, and removes the warning on
python 3.5. Looking at the PyPI stats since the deprecation period
started the number of users on Python 3.5 has diminished significantly,
but not disappeared. There were 359 downloads with pip from pypi out of
total of 18,980 total pip downloads in the last 30 days. Compared to the
roughly 10% figure when we deprecated Python 3.5.

Merging this means we can not release until after the documented EoL
date for Python 3.5 support of September 13. This shouldn't be a problem
because with Qiskit/qiskit-terra#4767 we will need to coordinate the
release of all the qiskit elements and are planning to do that after
09/13/2020.

It's worth noting that we should start planning to deprecate python 3.6
support sooner rather than later it goes EoL upstream at the end of
next year [1] and some of our other upstream dependencies (mainly numpy
et al) are going to remove support before the upstream Python EoL date
[2].

### Details and comments

[1] https://devguide.python.org/#branchstatus
[2] https://numpy.org/neps/nep-0029-deprecation_policy.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ sudo: false
 stage_generic: &stage_generic
   install:
     - pip install -U pip
-    - pip install cython
+    - pip install cython packaging
     # Install terra from master on aer master else use the most recent release
     - |
       if [[ $TRAVIS_BRANCH == "master" ]] ; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ sudo: false
 stage_generic: &stage_generic
   install:
     - pip install -U pip
-    - pip install cython packaging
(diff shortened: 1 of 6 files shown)


atilag (approved):
Love to see this finally coming!

mtreinish:
I should add a release note here, but since 3.5 windows is blocking CI we can go ahead and merge this and I'll push a release note in a follow up PR after we unblock CI.

Issue #892, opened 2020-08-17
Parallel thread tests fail on CPU with >64 cores


drewrisinger:
### Informations

- **Qiskit Aer version**: 0.5.2
- **Python version**: 3.8.5
- **Operating system**: Linux (NixOS)

Cross-post from https://github.com/NixOS/nixpkgs/issues/95527

### What is the current behavior?

Parallel tests fail on AMD Threadripper 3990X system with 128 cores, e.g.:
https://github.com/Qiskit/qiskit-aer/blob/25de39f0916290d914251c1f755fba0332deee1c/test/terra/backends/qasm_simulator/qasm_thread_management.py#L99-L102

All failing tests:
```
FAILED test/terra/backends/test_qasm_simulator.py::TestQasmSimulator::test_parallel_experiment_thread_assignment
FAILED test/terra/backends/test_qasm_simulator.py::TestQasmSimulator::test_parallel_shot_thread_assignment
FAILED test/terra/backends/test_qasm_simulator.py::TestQasmSimulator::test_parallel_thread_assignment_priority
FAILED test/terra/backends/test_qasm_simulator.py::TestQasmSimulator::test_parallel_thread_defaults
FAILED test/terra/backends/test_qasm_simulator_statevector.py::TestQasmSimulatorStatevector::test_parallel_experiment_thread_assignment
FAILED test/terra/backends/test_qasm_simulator_statevector.py::TestQasmSimulatorStatevector::test_parallel_shot_thread_assignment
FAILED test/terra/backends/test_qasm_simulator_statevector.py::TestQasmSimulatorStatevector::test_parallel_thread_assignment_priority
FAILED test/terra/backends/test_qasm_simulator_statevector.py::TestQasmSimulatorStatevector::test_parallel_thread_defaults
```

Full build log can be found at: https://gist.githubusercontent.com/jonringer/a86a889f2684ee64bb95732f0b065126/raw/efb78b20ac536cbedbfa490f342374228b1f0b2d/qiskit-aer%2520failure

### Steps to reproduce the problem
On 128 core system: 
```bash
git clone https://github.com/nixos/nixpkgs
git checkout f1efdd2c0b2efb3af9e2822bd490ede184a2bf6f
# install Nix
curl -L https://nixos.org/nix/install | sh
# build qiskit-aer with Nix
nix-build -A python38Packages.qiskit-aer
```
[...]

PR #918, opened 2020-09-02
fix a bug in thread management tests

hhorii:
### Summary

Change a way to get a number of maximum threads in `QasmThreadManagementTests`

### Details and comments

In some environment, `multiprocessing.cpu_count()` does not return a number of maximum OMP threads correctly.
Especially if `OMP_NUM_THREADS` is different from an actual number of hardware threads, tests are failed.

This PR changes a way to get a number of maximum OMP threads by running dummy simulation.

I hope that this PR resolves issues in #892.


test/terra/backends/qasm_simulator/qasm_thread_management.py
```diff
@@ -102,7 +102,13 @@ def test_parallel_thread_defaults(self):
         """Test parallel thread assignment defaults"""
 
         opts = self.BACKEND_OPTS
-        max_threads = multiprocessing.cpu_count()
+        
+        # Run single circuit, no noise to get max_threads
+        result = execute(self.dummy_circuit(1),
```
chriseclectic, line 107: This could be made a class function so you don't have to repeat it
```suggestion
def available_threads(self):
    """"Return the threads reported by the simulator"""
    result = execute(self.dummy_circuit(1),
                             self.SIMULATOR,
                             shots=1,
                             backend_options=self.BACKEND_OPTS).result()
    return self.threads_used(result)[0]['total']
```

test/terra/backends/qasm_simulator/qasm_thread_management.py
```diff
@@ -96,13 +96,21 @@ def test_max_memory_settings(self):
         self.assertEqual(max_mem_result, max_mem_target,
                          msg="Custom 'max_memory_mb' is not being set correctly.")
 
+    def availablve_threads(self):
```
vvilpas, line 99: availablve_threads -> available_threads

chriseclectic (approved):

Issue #715, opened 2020-04-25
Optimize statevector Pauli expectation value and sequences of Pauli gates


yaelbh:
In `State<statevec_t>::snapshot_pauli_expval`, denote by `k` if the number of qubits in `pauli` whose Pauli is different from `I`. Then there are `k` passes on the `2^n` basis states. This can be reduced to only two passes:
- One pass for all the `X`'s, switching amplitudes between pairs of basis gates.
- One pass for all the `Z`'s, flipping the sign of amplitudes of basis states with odd parity in the corresponding qubits.

This can help chemistry VQE simulation, where there are thousands of Pauli terms, and some of the terms have many qubits (i.e., large `k`).

yaelbh:
Can probably be done in a single apply matrix loop, based on an algorithm to construct the Pauli matrix from a string in Terra's `PauliTable` class.

yaelbh:
It is possible to first write a multi-qubit Pauli gate, and use it in the Pauli expectation value.

yaelbh:
Denote: n - the number of qubits, m - the total number of all qubits participating in at least one Pauli term, k - number of Pauli terms. Once we evaluate the complexity of #471, we will be able to identify cases where #471 is more efficient than the solution suggested here, which takes k apply matrix loops, each costs 2^n.

chriseclectic:
I implemented an optimized method for all-qubit expectation values is in #737 (this is the method based on the sympletic representation in the Pauli table class)

yaelbh:
Expectation value is fine since #737. However instead of closing the issue I'm renaming it to apply to all cases where a circuit contains a sequence of Pauli gates.

yaelbh:
There will be a new instruction of a set of Paulis (rather than an automatic detection in the circuit). As long as Terra doesn't support it, we will create our own instruction and add it to the set of basis gates when transpiling.

yaelbh:
Done in #919 and #1012

PR #919, opened 2020-09-03
Multiple Pauli gates operation

gadial:
### Summary
This PR implements the suggestion in #715  to add an operation performing multiple Pauli gates in one pass in the Qasm simulator.


yaelbh:
You have an internal loop that iterates over the qubits, which makes the complexity equal to the current state of applying the Pauli gates sequentially. See in https://github.com/Qiskit/qiskit-aer/blob/a387d1f7da2fcfe270e3a24f9d6d90101233b17a/src/simulators/statevector/qubitvector.hpp#L1988 how to do it really in one pass. Don't be confused from the index `i` being used in two different contexts in three different loops.

yaelbh:
There should be a way to unify the function for multi-Pauli gate and Pauli expectation value. They should somehow share the code that iterates over the amplitudes, with differences in what they do to the amplitudes inside the iteration. Same goes for yet an additional functionality that I want to add, which is a multi-Pauli measurement instruction.

src/framework/operations.hpp
```diff
@@ -571,6 +574,31 @@ Op json_to_op_initialize(const json_t &js) {
...
+  // If label is not specified record the gate name as the label
+  std::string label;
+  JSON::get_value(label, "label", js);
+  if  (label != "")
+    op.string_params.push_back(label);
+  else
+    op.string_params.push_back(op.name);
+
+  // Conditional
+  add_condtional(Allowed::No, op, js);
```
yaelbh, line 594: typo in conditional

src/simulators/statevector/qubitvector.hpp
```diff
@@ -2080,6 +2114,53 @@ double QubitVector<data_t, Derived>::expval_pauli(const reg_t &qubits,
...
+        data_[idxs[j]] *= -1;
+      }
+      data_[idxs[j]] *= phase;
+    }
+    std::swap(data_[idxs[0]], data_[idxs[1]]);
+  };
+  apply_lambda(lambda, (size_t) 0, (data_size_ >> 1));
+}
+
 //------------------------------------------------------------------------------
```
yaelbh, line 2164: There is still quite a lot of duplication between `pauli_expval` and `apply_multipauli`, any way to address it?

gadial: Good question. Currently the duplicated code is mostly bit tricks which do not encapsulate nicely; let's see what happens when we add Pauli measurements.

yaelbh: Pauli measurements are more similar to expectation value than gates. I'm probably going to rewrite `expval_pauli` as `pauli_expval_and_measure`, and it will be essentially the same function as now, with small changes.

chriseclectic: I main difference between Pauli expval and Pauli measure is expval is deterministic and can be computed as a single reduction without storing an intermediate probability vector.

With measure you need to sample an outcome from the Pauli measurement distribution (which could also be computed as a reduction).

chriseclectic:
For this PR it might make sense to input the Pauli's in their symplectic representation (and refactor the expvals to do the same). A lot of the bit tricks are more easier done from the symplectic form, and this is also the internal form of the operators in terra (eg PauliTable / SparsePauliOp).

qiskit/providers/aer/extensions/multi_pauli.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self, pauli_string=None):
+        self.pauli_string = pauli_string
+        super().__init__('multi_pauli', len(pauli_string), [pauli_string[::-1]])
+
+    def _define(self):
+        """
+        gate multi_pauli (p1 a1,...,pn an) { p1 a1; ... ; pn an; }
+        """
+        gates = {'X': XGate, 'Y': YGate, 'Z': ZGate}
```
chriseclectic, line 43: This should support identity Pauli's too.

This could be done by changing the loop to:
```python
rules = [(gates[p](), [q[i]], []) for (i, p) in enumerate(self.pauli_string) if p != 'I']
```
or adding the `IGate` to the gates dict.

qiskit/providers/aer/extensions/multi_pauli.py
```diff
@@ -0,0 +1,75 @@
...
+Simulator command to perform multiple pauli gates in a single pass
+"""
+import numpy as np
+
+from qiskit import QuantumCircuit, QuantumRegister
+from qiskit.circuit.library import XGate, YGate, ZGate
+from qiskit.circuit.gate import Gate
+
+
+class MultiPauliGate(Gate):
```
chriseclectic, line 23: This gate class should be moved to the terra circuit library

chriseclectic, line 23: The `Pauli` op is already a multi-qubit Pauli, so I think this should just be called `PauliGate` not `MultiPauliGate`

qiskit/providers/aer/extensions/multi_pauli.py
```diff
@@ -0,0 +1,75 @@
...
+            'Y': np.array([[0, -1j], [1j, 0]], dtype=complex),
+            'Z': np.array([[1, 0], [0, -1]], dtype=complex)
+        }
+        mat = np.eye(1, dtype=complex)
+        for pauli in self.pauli_string:
+            mat = np.kron(mat, pauli_matrices[pauli])
+        return mat
+
+
+def multi_pauli(self, qubits, pauli_string):
```
chriseclectic, line 69:
```suggestion
def pauli(self, qubits, pauli_string):
```

qiskit/providers/aer/extensions/multi_pauli.py
```diff
@@ -0,0 +1,75 @@
...
+    """Adds a multi pauli instruction to the circuit"""
+    return self.append(MultiPauliGate(pauli_string), qubits)
+
+
+# Add to QuantumCircuit class
+QuantumCircuit.multi_pauli = multi_pauli
```
chriseclectic, line 75:
```suggestion
QuantumCircuit.pauli = pauli
```

src/framework/operations.hpp
```diff
@@ -354,6 +354,7 @@ Op json_to_op_measure(const json_t &js);
 Op json_to_op_reset(const json_t &js);
 Op json_to_op_bfunc(const json_t &js);
 Op json_to_op_initialize(const json_t &js);
+Op json_to_op_multi_pauli(const json_t &js);
```
chriseclectic, line 357:
```suggestion
Op json_to_op_pauli(const json_t &js);
```

src/framework/operations.hpp
```diff
@@ -418,6 +419,8 @@ Op json_to_op(const json_t &js) {
     return json_to_op_kraus(js);
   if (name == "roerror")
     return json_to_op_roerror(js);
+   if (name == "multi_pauli")
```
chriseclectic, line 422:
```suggestion
   if (name == "pauli")
```

src/simulators/statevector/qubitvector.hpp
```diff
@@ -209,6 +210,8 @@ class QubitVector {
   // If N=3 this implements an optimized Fredkin gate
   void apply_mcswap(const reg_t &qubits);
 
+  void apply_multipauli(const reg_t &qubits, const std::string &pauli);
```
chriseclectic, line 213:
```suggestion
  void apply_pauli(const reg_t &qubits, const std::string &pauli);
```

chriseclectic: Also any new methods to QubitVector need to be added to QubitVectorThrust

chriseclectic (changes requested):
This is a good gate optimization to add (assuming circuits actually use it!). A couple of change requests:
* Rename the instruction from `MultiPauliGate` to `PauliGate` and move it to the terra circuit library
* Rename C++ methods from `multi_pauli` to `pauli`. We already define a `Pauli`operator to be a general N-qubit Pauli in most parts of qiskit.
* Support for this gate needs to be added to QubitVectorThrust otherwise the statevector_state class will have issues (it needs same interface for both CPU and GPU simulation).
* This gate should also be added to DensityMatrix / DensityMatrixThrust simulators in a followup.

qiskit/providers/aer/extensions/multi_pauli.py
gadial: I added IGate; it seems better than to omit the gate, since the user might want to put noise on the gate.

qiskit/providers/aer/extensions/multi_pauli.py
gadial: Agreed.

hhorii:
We need to add `pauli` (current `multi_pauli`) for `qubitvector_thrust.hpp`.

doichanj:
I'm still testing and debugging multi-Pauli for multi-GPU parallelization. I will put PR Thrust implementation maybe tomorrow.

src/simulators/statevector/qubitvector.hpp
```diff
@@ -363,6 +366,19 @@ class QubitVector {
...
+  //-----------------------------------------------------------------------
+  // Apply a lambda function to all entries of the statevector
+  // between start and stop
+  // The function signature should be:
+  //
+  // [&](const int_t k)->void
+  //
+  // where k is the index of the vector
+  template <typename Lambda>
+  void apply_lambda(Lambda&& func, size_t start, size_t stop);
```
chriseclectic, line 380: Why not just call `QV::apply_lambda(start, stop, omp_threads_managed(), func)` directly instead of adding this function.

src/simulators/statevector/qubitvector.hpp
```diff
@@ -1766,18 +1786,17 @@ double QubitVector<data_t>::expval_pauli(const reg_t &qubits,
...
+  return std::make_tuple(x_mask, z_mask, num_y, x_max);
+}
 
-  // Special case for only I Paulis
-  if (x_mask + z_mask == 0) {
-    return norm();
-  }
-
+template <typename data_t>
+void compute_phase(uint_t num_y, std::complex<data_t>& phase){
```
chriseclectic, line 1793: The way this function is used it would make more sense to have signature
```suggestion
std::complex<data_t> compute_phase(uint_t num_y)
```

src/simulators/statevector/qubitvector.hpp
```diff
@@ -1792,6 +1811,21 @@ double QubitVector<data_t>::expval_pauli(const reg_t &qubits,
...
+
+  uint_t x_mask, z_mask, num_y, x_max;
+  std::tie(x_mask, z_mask, num_y, x_max) = pauli_masks_and_phase(qubits, pauli);
+
+  // Special case for only I Paulis
+  if (x_mask + z_mask == 0) {
+    return norm();
+  }
+  std::complex<data_t> phase;
+  compute_phase(num_y, phase);
```
chriseclectic, line 1828: With above comment this would be
```suggestion
  auto phase = compute_phase(num_y);
```

chriseclectic (commented):
A couple small suggestions but this basically looks ready.

chriseclectic:
I missed that this gate also needs to be added to the basis gates of the the Python classes for the statevector simulator, unitary simulator, and statevector method of qasm simulator.

test/terra/backends/qasm_simulator/qasm_cliffords.py
```diff
@@ -237,6 +237,26 @@ def test_swap_gate_nondeterministic_default_basis_gates(self):
...
+        """Test multipauli-gate circuits compiling to backend default basis_gates."""
+        backends_to_test = ['statevector', 'statevector_gpu', 'statevector_thrust']
+        shots = 100
+        circuits = ref_1q_clifford.pauli_gate_circuits_deterministic(
+            final_measure=True)
+        targets = ref_1q_clifford.pauli_gate_counts_deterministic(shots)
+        basis_gates = None
+        if 'method' not in self.BACKEND_OPTS or self.BACKEND_OPTS['method'] in backends_to_test:
+            basis_gates = ['pauli', 'h', 'z'] # implemented only for statevector for now
+        job = execute(circuits, self.SIMULATOR, shots=shots,
```
chriseclectic, line 253: I don't think this is testing the gate, since you haven't added it to the basis gates it will be getting unrolled to regular I, X, Y, Z instructions.

src/simulators/statevector/qubitvector.hpp
gadial: Have to admit I'm not sure why it mattered to me... I'll fix as you suggested.

gadial: Purely cosmetic, but in line with the other `apply_lambda` functions defined in this file.

src/simulators/statevector/qubitvector.hpp
gadial: This is what I originally had but it caused a template-related compilation error I didn't know how to solve so I went around it. Something alone this lines:

```/home/gadial/Work/qiskit-aer/src/simulators/statevector/qubitvector.hpp: In member function ‘void AER::QV::QubitVector<data_t>::apply_pauli(const reg_t&, const string&)’:
/home/gadial/Work/qiskit-aer/src/simulators/statevector/qubitvector.hpp:1880:35: error: no matching function for call to ‘compute_phase(AER::QV::uint_t&)’
 1880 |   auto phase = compute_phase(num_y);
      |                                   ^
/home/gadial/Work/qiskit-aer/src/simulators/statevector/qubitvector.hpp:1793:22: note: candidate: ‘template<class data_t> std::complex<_Tp> AER::QV::compute_phase(AER::QV::uint_t&)’
 1793 | std::complex<data_t> compute_phase(uint_t& num_y){
      |                      ^~~~~~~~~~~~~
/home/gadial/Work/qiskit-aer/src/simulators/statevector/qubitvector.hpp:1793:22: note:   template argument deduction/substitution failed:
/home/gadial/Work/qiskit-aer/src/simulators/statevector/qubitvector.hpp:1880:35: note:   couldn’t deduce template parameter ‘data_t’
```

test/terra/backends/qasm_simulator/qasm_cliffords.py
gadial: I actually verified this explicitly when I wrote it - changes to `apply_pauli` in the cpp code casued test failures. Note I pass the basis gates when running the job:

```
job = execute(circuits, self.SIMULATOR, shots=shots,
                      basis_gates=basis_gates,
                      backend_options=self.BACKEND_OPTS)
```

gadial: Anyway, I changed the code the be cleaner - I obtain the basis gates list directly from the backend's `_method_configuration` method.

src/simulators/statevector/qubitvector.hpp
chriseclectic: Ah that makes sense. Yeah because the template param is the return you wouldn't be able to use it with auto, you would have to specify the return type `complex<data_t> phase = compute_phase(num_y)`

chriseclectic (approved):

PR #923, opened 2020-09-07
Enable Appleclang 12 to work with Conan

### LINKED ISSUE #905: Conan doesn't work with Apple Clang 12.0 by default
### Informations

- **Qiskit Aer version**: 0.7
- **Python version**: 3.7.7
- **Operating system**: MacOS 10.15.6

After updating to the latest XCode Commandline tools which includes Apple Clang 12.0.0 Conan can't build by default.

Trying to build gives:
```
ERROR: Invalid setting '12.0' is not a valid 'settings.compiler.version' value.
Possible values are ['10.0', '11.0', '5.0', '5.1', '6.0', '6.1', '7.0', '7.3', '8.0', '8.1', '9.0', '9.1']
Read "http://docs.conan.io/en/latest/faq/troubleshooting.html#error-invalid-setting"
CMake Error at cmake/conan.cmake:402 (message):
  Conan install failed='1'
Call Stack (most recent call first):
  cmake/conan.cmake:497 (conan_cmake_install)
  cmake/conan_utils.cmake:33 (conan_cmake_run)
  CMakeLists.txt:124 (setup_conan)
```

Build works

### Suggested solutions

Probably we need to wait for Conan to release an update, but I was able to build by editing my local conan settings manually to include '12.0' as per the Conan troubleshooting guide.

vvilpas:
### Summary
Conan won't officially support Apple Clang 12.0 as right now is in beta state. This PR updates 'settings.yml' automatically when building AER to support that version.

Closes #905

### DIFF
--- a/cmake/conan_utils.cmake
+++ b/cmake/conan_utils.cmake
@@ -30,6 +30,9 @@ macro(setup_conan)
         set(REQUIREMENTS ${REQUIREMENTS} catch2/2.12.1)
     endif()
 
+    # Add Appleclang-12 until officially supported by Conan
+    conan_config_install(ITEM ${PROJECT_SOURCE_DIR}/conan_settings)
+
     conan_cmake_run(REQUIRES ${REQUIREMENTS}
                     OPTIONS ${CONAN_OPTIONS}
                     ENV CONAN_CMAKE_PROGRAM=${CMAKE_COMMAND}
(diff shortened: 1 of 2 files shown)


chriseclectic (approved):
This works, though we will have to be careful of when to remove/update this settings file

PR #925, opened 2020-09-08
Enable building 32bit binaries in wheel jobs

mtreinish:
### Summary

This commit is an attempt to start building 32bit binaries in CI for
publishing at release time (and the corresponding wheel compile jobs).
This is primarily an issue for windows were a large number of users are
running with a win32 python binary and aer fails to install from wheel
because there is not 32bit binary available.

### Details and comments

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -98,9 +98,9 @@ stages:
             mkdir wheelhouse
             for version in 3.6 3.7 3.8 ; do
                 source activate qiskit-aer-$version
-
                 conda install --yes --quiet --name qiskit-aer-$version python=$version numpy cmake virtualenv pip setuptools pybind11 cython scipy
-                python setup.py bdist_wheel -- -G "Visual Studio 15 2017 Win64"
+                python -m pip install -U setuptools wheel
+                python setup.py bdist_wheel -- -G "Visual Studio 15 2017 Win32"
                 virtualenv test-$version
                 test-$version/Scripts/pip install -c constraints.txt dist/*whl
                 test-$version/Scripts/python tools/verify_wheels.py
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -67,7 +67,61 @@ stages:
...
+            CONDA_FORCE_32BIT: 1
+        - bash: |
+            set -x
+            set -e
+            mkdir wheelhouse
+            for version in 3.6 3.7 3.8 ; do
+                source activate qiskit-aer-$version
+                conda install --yes --quiet --name qiskit-aer-$version python=$version numpy cmake virtualenv pip setuptools pybind11 cython scipy
+                python -m pip install -U setuptools wheel
+                python setup.py bdist_wheel -- -G "Visual Studio 15 2017 Win64" -A "Win32"
```
vvilpas, line 103: I think the syntax is different for VS 2017 and it is just:
`python setup.py bdist_wheel -- -G "Visual Studio 15 2017"`
If it doesn't work with the `-A win32` try with the above syntax

azure-pipelines.yml
```diff
@@ -122,6 +176,64 @@ stages:
...
+              conda install --yes --quiet --name qiskit-aer python=$(python.version) numpy cmake virtualenv pip setuptools pybind11 cython scipy
+            displayName: Create Anaconda environments
+            env:
+              CONDA_FORCE_32BIT: 1
+          - bash: |
+              set -x
+              set -e
+              source activate qiskit-aer
+              pip install -U setuptools wheel
+              python setup.py bdist_wheel -- -G "Visual Studio 15 2017 Win64" -A "Win32"
```
vvilpas, line 221: See previous comment

vvilpas (dismissed):
LGTM!

vvilpas (approved):

PR #927, opened 2020-09-10
Kraus noise gate fusion method

chriseclectic:
### Summary

~Depends on #889~

Adds a Kraus method to NoiseModel and gate fusion that allows applying the gate fusion optimization to noisy statevector simulations with Kraus noise.

### Details and comments

Initial benchmarks show this gives a 3-9x performance improvement for noisy statevector simulations using the device noise model for 10-20 qubits.

kraus_fusion_perf

Benchmarks run on Ubuntu desktop w/ AMD Ryzen 3700X and Geforce 1070.

### DIFF
--- a/releasenotes/notes/kraus-noise-fusion-5577d4a38862e966.yaml
+++ b/releasenotes/notes/kraus-noise-fusion-5577d4a38862e966.yaml
@@ -1,0 +1,6 @@
+---
+features:
+  - |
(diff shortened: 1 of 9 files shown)


vvilpas (commented):
Looks good. 
I have a minor comment which can  make sense in case we use big matrices, otherwise it is ok.

PR #930, opened 2020-09-14
Add tutorials job to CI

mtreinish:
### Summary

This commit adds a new job to run the qiskit tutorials in CI. The
tutorials are used as a form of upgrade testing to ensure that for key
examples Qiskit as a whole is N and N+1 releases. This caused friction
during the qiskit 0.20.0 release because the tutorials were never
updated to stop using deprecated code (or in the case of other elements
by merging backwards incompatible changes). To ensure there aren't any
surprises at the last minute when we run the tutorials with a proposed
new metapackage this commit adds a job to CI to ensure that they always
run with terra changes. It means for PRs that change an api (after a
deprecation cycle) the tutorial will need to be updated first. This will
ensure that users will have an upgrade path because CI in
qiskit/qiskit-tutorials runs with the metapackage.

### Details and comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -206,3 +206,36 @@ jobs:
           stestr run --slowest
         shell: bash
         if: runner.os == 'macOS' && matrix.python-version == '3.8'
+  tutorials:
+    name: Tutorials
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


vvilpas (approved):
LGTM

PR #932, opened 2020-09-15
Add custom Vector class

### LINKED ISSUE #868: Add a custom C++ numeric vector class

Add a custom numeric `vector` class like the `matrix` class that uses a C-style array internally and supports move semantics. The main motivation for this is to allow the use of move semantics with the QubitVector/Statevector simulators and the rest of the code, including the Numpy array of the output and Numpy array of initial states.

For example the UnitarySimulator can move its final state to results to be returned as a Numpy array without a copy, however the StatevectorSimulator requires a memory copy to a std::vector.

This sort of array is already used internally by the QubitVector class so would abstract out its internal container to a separate class, and could also be used in other parts of the code that need a numeric vector instead of std::vector.

chriseclectic:
### Summary

Closes #868 

Adds a custom Vector container to C++ code that allows moving memory to and from the QubitVector class and Numpy arrays through Pybind11.

### Details and comments

At the moment this is just a container for a C-style array similar to the `matrix` class that is useful for results. In the future this should be improved by adding additional numerical methods and better integrating with the matrix class so that it can be used for internal numeric vector computations where we currently use `std::vectors` or `std::arrays`.

#### Memory Profiling

Memory profile of statevector simulator for a 28-qubit circuit running as single-precision:

**Stable**

mprof_sv_28q_old

**This PR**
mprof_sv_28q_new


src/framework/linalg/linops/linops_aer_vector.hpp
```diff
@@ -0,0 +1,142 @@
...
+template <class T, typename = enable_if_numeric_t<T>>
+Vector<T>& isub(Vector<T>& lhs, const Vector<T>& rhs) {
+  lhs -= rhs;
+  return lhs;
+}
+
+//----------------------------------------------------------------------------
+// Affine operations
+//----------------------------------------------------------------------------
+template <class T, class Scalar, typename = enable_if_numeric_t<T>,
```
vvilpas, line 58: For these affine operations, in order to reduce code duplication, it would be better to implement the _op_ function in terms of the _iop_ equivalent one:
```
T op(const T& a, const Scalar& val){
   T tmp(a);
   return iop(a, val);
}
```

src/framework/linalg/linops/linops_aer_vector.hpp
```diff
@@ -0,0 +1,142 @@
...
+  const T cast_val(val);
+  std::for_each(data.data(), data.data() + data.size(),
+                [&cast_val](T& a)->T{ a += cast_val; });
+  return data;
+}
+
+template <class T, class Scalar, typename = enable_if_numeric_t<T>,
+          typename = enable_if_numeric_t<Scalar>>
+Vector<T> sub(const Vector<T>& data, const Scalar& val) {
+  Vector<T> result(data.size(), false);
```
vvilpas, line 80: To avoid code duplication as well we can implement `sub` function in terms of `add`:
```
T sub(const T& vec, const Scalar& val){
    return add(vec, -val);
}
```

vvilpas (changes requested):
Nice improvement in memory footprint!
A few comments below.

vvilpas (approved):

PR #937, opened 2020-09-17
Fix THRUST build error

vvilpas:
### Summary

Fix build error when setting AER_THRUST_BACKEND trough an environment variable

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,11 @@ if(NOT DEFINED AER_BLAS_LIB_PATH AND DEFINED ENV{AER_BLAS_LIB_PATH})
 	set(AER_BLAS_LIB_PATH $ENV{AER_BLAS_LIB_PATH})
 endif()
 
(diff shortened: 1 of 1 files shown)


vvilpas:
Once this is merged #935 should work

mtreinish:
Was this a recent breakage? I know the env var used to work because for the 0.6.0/0.6.1 release we published the gpu binaries using the env var.

vvilpas:
No, actually it has been there since we added the env variable mechanism, but it doesn't affect `CUDA`. It only affects the builds using the ENV var with value `OMP` or `TBB`, because they need  the `Thrust` library. To deliver this dependency we rely on Conan, which only gather it if `AER_THRUST_BACKEND` has the right values (`OMP` or `TBB`).  The problem is that when we use the ENV var, the CMake var `AER_THRUST_BACKEND` was set __after__ `Conan` setup step, so it was not aware that it has to deliver `Thrust`. 
For `CUDA` builds there is no problem because it comes with its own `Thrust` library, which is the one that we use in that case.

chriseclectic (approved):

Issue #240, opened 2019-06-18
[Feature Request] Saving AerJobs to Disk


ajavadia:
### What is the expected behavior?
Currently in Aer, an AerJob with a particular ID is created on every execution. This mimics the IBMQJob and its async execution model. However this AerJob is not as useful, as it only persists during that session (the job id is basically useless for future reference).

Can we treat the user's disk as the "cloud" used in IBMQ jobs, save all the qobj and result and simulation configs associated with each execution, and allow them to be queried later?
The size of this data is unlikely to grow extremely large, so I'd enable this by default, with an option to turn off.

This would make @adcorcol and @dcmckayibm happy.

aditya-giri:
This looks interesting, I'd be happy to get started with a little nudge in the right direction :)

atilag:
Hi @aditya-giri thanks for jumping in, but I'm afraid we are not implementing this in the near future. If for whatever reason this shows up again, I'll be glad to mentor here :)

atilag:
This escalated quickly.. ok @aditya-giri , it's yours :)

aditya-giri:
That's great! Happy to get started! Let me take a day or two to get aer built locally and look through the source a little? I'd appreciate some guidance once I'm ready to go.

atilag:
So before getting our hands into the code, let's bring some more details about what we want here.
This is the use-case:
"Users want to have information (job, results and configuration) about jobs already executed in the simulator."
We could split these up into the following tasks:
- Define the public interface (let's start with this) for storing and loading jobs from disk.
- Choose a format for serializing all the data (probably json)
- Define a file management policy

Constraints:
- Everything happens in the Python side (no C++ changes beforehand), so the standalone won't have this feature (to discuss with @ajavadia)
- As simulation runs locally, there's no point on saving the jobs before the simulation finishes.
- If something fails during simulation, we want to save the status with details about the problem.

@aditya-giri  let's start by proposing a public API :)

aditya-giri:
Got it. Just one clarification re: `Everything happens in the Python side (no C++ changes beforehand), so the standalone won't have this feature` - could you elaborate on this a little?
About the API - how are we expecting users to use this? Are we only going to allow retrieval of jobs from the disk? Or do we want to provide more control over which jobs get written to the disk?

atilag:
> could you elaborate on this a little?

[...]

PR #938, opened 2020-09-17
#240: WIP: Saving AerJobs to Disk

aditya-giri:
### Summary
Draft PR for #240. This adds APIs to serialize Aer jobs to the disk and retrieve them later. This is still a work-in-progress.

### Details and comments
The current implementation of this feature, as proposed by @atilag, is to allow users to configure if and where they want to save results of locally executed jobs using the `backend_options` parameter. The `Result` object is serialized and written to a file using `pickle`. The filename is a combination of the backend's name, the job id, and the UNIX timestamp at the time of serialization. Deserialization is also done using `pickle` and the plan is to add 2 APIs (similar to those available on the cloud) - one to retrieve all jobs run on a backend and one to retrieve a specific job's result using its `job_id`.


qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -131,7 +131,11 @@ def _run_job(self, job_id, qobj, backend_options, noise_model, validate):
             self._validate(qobj, backend_options, noise_model)
         output = self._controller(self._format_qobj(qobj, backend_options, noise_model))
         end = time.time()
-        return Result.from_dict(self._format_results(job_id, output, end - start))
+        result = Result.from_dict(self._format_results(job_id, output, end - start))
+
+        self._save_job(result, backend_options)
```
atilag, line 136: Let's generalize this a bit, because it could be interesting to leave it open for future post processing actions over results:
```python
result = self._post_process(result, backend_options)
```
The implementation of `post_process()` can be something like:
```python
def _post_process(self, result, backend_options):
  if (backend_options is not None) and ("save_jobs" in backend_options):
     self._save_job(result) 
  
  return result
  
```

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -186,6 +190,18 @@ def _validate(self, qobj, backend_options, noise_model):
         """Validate the qobj, backend_options, noise_model for the backend"""
         pass
 
+    def _save_job(self, result, backend_options):
+        if (backend_options is not None) and ("path_to_save" in backend_options):
+            path_to_save = backend_options["path_to_save"]
```
atilag, line 195: Per last [comment](https://github.com/Qiskit/qiskit-aer/issues/240#issuecomment-694115558) , let's just use a hardcoded path directory. A good candidate could be: `$HOME/.qiskit/aer_jobs/`.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -186,6 +190,18 @@ def _validate(self, qobj, backend_options, noise_model):
...
+            if not path_to_save.endswith("/"):
+                path_to_save += "/"
+            filename = path_to_save + "_".join(
```
atilag, line 198: Let's change the name so the date is in first place, this way it's easier to list them with `ls -la`.
Also, a more readable date/time format is encouraged so we can have more info about the job by just looking at the file name:
```
ss-mm-hh-dd-MM-yy-name-jobid.job
```

atilag (commented):
Looking good!
Left some comments.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -186,6 +190,18 @@ def _validate(self, qobj, backend_options, noise_model):
...
+                [self.name(), str(result.job_id), str(int(time.time()))]
+            )
+            os.makedirs(os.path.dirname(filename), exist_ok=True)
+            with open(filename, "wb") as file:
+                pickle.dump(result, file, 0)
```
mtreinish, line 203: Do not use pickle as a general purpose serialization format. As you had already mentioned in the issue this is a big security issue. This kind of use case is the exact thing you should **not** be using pickle for.

It is not hard to serialize a results object there is already a `to_dict()` method on it which gets you pretty close, you can use json, yaml, or any other number of other off the shelf serialization formats that work with python dicts. There are just a couple of types (like numpy arrays) in the dict that you'll have to handle in whatever you pick. If you go json for example it's easy to write a small encoder class to handle that.

qiskit/providers/aer/backends/aerbackend.py
mtreinish: I agree we shouldn't be taking this in as a backend option. But besides a hardcoded path it should override-able with an env variable. You can leverage the settings.conf as mentioned in the issue down the road but for something like this you don't not want a hardcoded path without giving any option to override it easily. For example on some of my systems home directory space is intentionally very limited and always writing to `/home` is not viable.

That being said we still also should make saving results an opt-in feature. It's not something everyone will want and does add overhead (both serialization and disk io).

mtreinish (changes requested):

qiskit/providers/aer/backends/aerbackend.py
aditya-giri: Okay, that makes sense. Will do :+1:

qiskit/providers/aer/backends/aerbackend.py
aditya-giri: @mtreinish @atilag understood :+1: Let me look into leveraging `settings.conf`.
> That being said we still also should make saving results an opt-in feature

Yes, I plan on adding a `backend_option` called `save_job` which takes a boolean value indicating if that specific job needs to be saved. Does that make more sense?

qiskit/providers/aer/backends/aerbackend.py
aditya-giri: Sure!

qiskit/providers/aer/backends/aerbackend.py
aditya-giri: Got it. I also just noticed the `AerJSONEncoder` class that's already been written, so I'll try and leverage that.

chriseclectic:
Why not just add a local save/load method to the BaseJob or Result object in Terra and then it will work for any backend / provider?

qiskit/providers/aer/backends/aerbackend.py
aditya-giri: I see `terra` [has a class to read the config file](https://github.com/Qiskit/qiskit-terra/blob/master/qiskit/user_config.py), but aer doesn't seem to have one at the moment. Should I add this as well? I can modify the [read_config_file](https://github.com/Qiskit/qiskit-terra/blob/master/qiskit/user_config.py#L48) function to just read the directory to be used for saving and leave the terra stuff untouched.

atilag:
I don't think this is a general use case that can be reused in all the backends. For example, IBM-Q provider backends send everything to the cloud and don't need to save anything locally. Thus I'd say this is a very specific use-case for local simulators. Also, a `Job` should  be agnostic to any kind of serialization/transportation method.

I agree with you (in conversations we had before) that adding this functionality to the backend is probably not the best option because we are adding new responsibilities to it, and don't feel comfortable coupling disk serialization to a backend. Here is the problem, if we want to hide as much details as possible to the user (developer) and provide an seamless experience, the backend happens to have all the information we need: backend_config, job_id, results, so implementing this behavior there is not that crazy.
Another option that I like (and it was my initial thought about how to implement this) is, as you said, provide a new component  for serialization mechanism to the user, and force them to explicitly use it, something in the lines of:
```
from aer.serialization import save_job, load_job, load_jobs

...
job = sim.run(qobj, ...)
save_job(job)
...

jobs = load_jobs()
[...]
job = load_job(jobs[1])
```
This could be another option, even cleaner, and moving this PR implementation to this model should be fairly easy.
Thoughts?

chriseclectic:
@atilag I have frequently saved local copies of IBMQ backend results. You want to do this to include any used data as supplemental material for papers for example, or for sharing your results with collaborators.

We also use saved result files (from simulator) in Ignis for CI tests so we don't have to run long executions every time. All you really want to do is be able to dump a Result/Job to some file format (like JSON), and then reload it later back to a Result /Job object.

If you want to add a local database on top of that that mimics the IBMQ provider database for automatically storing and retrieving files locally you could do that too, but i think that is overkill most of the time. You just save the specific results you want to a file and location you pick.

aditya-giri:
Just a heads-up that I've been a little busy and haven't got around to making the proposed changes yet. I will have it done shortly.

aditya-giri:
Hi, I'm closing this PR now and I'll open a new one shortly - the branches have diverged a lot and rebasing is getting a little cumbersome.

PR #941, opened 2020-09-20
add SIMD support for unitary and density matrix simulators

hhorii:
### Summary

Add AVX2 support for density and unitary matrix simulators

### Details and comments

#684 added AVX2 support only to statevector. This PR aims to add it to unitary and density matrix simulators.


hhorii:
```
from time import time
from qiskit import *
from qiskit.circuit.library import *
from qiskit.providers.aer import *

sim=QasmSimulator()
shots = 1024
time_ideal_with_simd = []
time_ideal_without_simd = []
for qubit in range (5, 11):
    circ = QuantumVolume(qubit).decompose()
    qobj = assemble(circ, shots=shots)
    t = %timeit -o -r 3 result = sim.run(qobj, backend_options={"method": "density_matrix"}).result()
    time_ideal_with_simd.append(t.average)
    t = %timeit -o -r 3 result = sim.run(qobj, backend_options={"method": "density_matrix", 'enable_simd': False}).result()
    time_ideal_without_simd.append(t.average)

import matplotlib.pyplot as plt

nqs = list(range(5, 11, 1))
plt.semilogy(nqs, time_ideal_with_simd, 'o-', label='with SIMD (Ideal)')
plt.semilogy(nqs, time_ideal_without_simd, 'o-', label='without SIMD (Ideal)')
plt.grid()
plt.legend()
plt.title('density matrix on Macbook Pro (2.6 GHz 6-Core Intel Core i7)')
plt.xlabel('N qubits')
plt.ylabel('time (s)')
plt.show()
```
image

vvilpas (approved):
Overall it looks good and enables a nice improvement in performance as shown in the plot. However, I think all AVX2 classes share a lot of code in common. I think we can merge this PR and I'll try to address that issue in another PR.

chriseclectic (changes requested):
The gate fusion transpile pass also needs to be updated to switch between the simd/non simd Unitary and SuperOp simulators

src/controllers/qasm_controller.hpp
```diff
@@ -298,6 +298,9 @@ class QasmController : public Base::Controller {
 
   // Controller-level parameter for CH method
   bool extended_stabilizer_measure_sampling_ = false;
+
+  // Disable SIMD optimization for testing
+  bool disable_simd_ = false;
```
chriseclectic, line 303: Is it really necessary to build this into the controller? It greatly complicates the code if it is only for testing. Why would anyone want to disable SIMD?

src/noise/noise_model.hpp
```diff
@@ -48,6 +48,9 @@ class NoiseModel {
   NoiseModel() = default;
   NoiseModel(const json_t &js) {load_from_json(js);}
 
+  // Load config from a JSON
+  void set_config(const json_t &config);
```
chriseclectic, line 52: I don't think this is a good way to deal with this in the noise model class. If you really want to disable simd, you should add a specific method to do so, not use a config. But i think this is not necessary if we remove the "disable_simd" option.

src/noise/quantum_error.hpp
```diff
@@ -38,6 +39,12 @@ class QuantumError {
   // Alias for return type
   using NoiseOps = std::vector<Operations::Op>;
 
+  //-----------------------------------------------------------------------
+  // Configuration
+  //-----------------------------------------------------------------------
+
+  void set_config(const json_t &config);
```
chriseclectic, line 46: See same comment for NoiseModel class

src/noise/quantum_error.hpp
```diff
@@ -350,13 +372,22 @@ const std::vector<cmatrix_t>& QuantumError::kraus() const {
   return canonical_kraus_;
 }
 
-
 void QuantumError::compute_superoperator() {
+#if defined(_MSC_VER) || defined(GNUC_AVX2)
```
chriseclectic, line 376: The ifdefs seem a bit messy here, can't it just be like the controller class with:
```
if (is_avx2_supported()) {
  compute_superoperator_<QubitSuperoperator::State<QV::SuperoperatorAvx2<double>>>();
} else {
  compute_superoperator_<QubitSuperoperator::State<QV::Superoperator<double>>>();
}
```

src/controllers/unitary_controller.hpp
```diff
@@ -111,6 +111,9 @@ class UnitaryController : public Base::Controller {
 
   // Precision of a unitary matrix
   Precision precision_ = Precision::double_precision;
+
+  // Disable SIMD optimization for testing
+  bool disable_simd_ = true;
```
chriseclectic, line 116: Same as for QasmController

src/framework/utils.hpp
```diff
@@ -20,6 +20,9 @@
 #include <cmath>
 #include <limits>
 
+#include <immintrin.h>
+#include <type_traits>
```
chriseclectic, line 24: These don't seem to be used in this header file so I don't think they should be here.

If they are put here for including in another file these seems dangerous since it is not obvious where they will be loaded from.

src/transpile/fusion.hpp
```diff
@@ -105,6 +105,8 @@ class Fusion : public CircuitOptimization {
 
   void add_fusion_qubits(reg_t& fusion_qubits, const op_t& op) const;
 
+  bool disable_simd_ = false;
```
chriseclectic, line 108: Again, I don't think this is needed

src/transpile/fusion.hpp
```diff
@@ -299,28 +304,60 @@ op_t Fusion::generate_fusion_operation(const std::vector<op_t>& fusioned_ops,
   ExperimentData dummy_data;
 
   if (method == Method::unitary) {
-    // Unitary simulation
-    QubitUnitary::State<> unitary_simulator;
-    unitary_simulator.initialize_qreg(qubits.size());
-    unitary_simulator.apply_ops(fusioned_ops, dummy_data, dummy_rng);
-    return Operations::make_unitary(qubits, unitary_simulator.qreg().move_to_matrix(),
-                                    std::string("fusion"));
+#if defined(_MSC_VER) || defined(GNUC_AVX2)
```
chriseclectic, line 307: No need for ifdef

src/transpile/fusion.hpp
```diff
@@ -299,28 +304,60 @@ op_t Fusion::generate_fusion_operation(const std::vector<op_t>& fusioned_ops,
...
+    if (!disable_simd_) {
```
chriseclectic, line 308:
```suggestion
    if (is_avx2_supported()) {
```

src/transpile/fusion.hpp
```diff
@@ -299,28 +304,60 @@ op_t Fusion::generate_fusion_operation(const std::vector<op_t>& fusioned_ops,
...
+      // Unitary simulation
+      QubitUnitary::State<QV::UnitaryMatrixAvx2<>> unitary_simulator;
+      unitary_simulator.initialize_qreg(qubits.size());
+      unitary_simulator.apply_ops(fusioned_ops, dummy_data, dummy_rng);
+      return Operations::make_unitary(qubits, unitary_simulator.qreg().move_to_matrix(),
+                                      std::string("fusion"));
+    } else
+#endif
```
chriseclectic, line 316: This ifdef block is wrong, since if it is not used it will run this simulation twice. But i think it should just be removed for a normal if / else block like the controller classes

src/transpile/fusion.hpp
```diff
@@ -299,28 +304,60 @@ op_t Fusion::generate_fusion_operation(const std::vector<op_t>& fusioned_ops,
...
+    }
   }
 
-  // For both Kraus and SuperOp method we simulate using superoperator
-  // simulator
-  QubitSuperoperator::State<> superop_simulator;
-  superop_simulator.initialize_qreg(qubits.size());
-  superop_simulator.apply_ops(fusioned_ops, dummy_data, dummy_rng);
-  auto superop = superop_simulator.qreg().move_to_matrix();
+#if defined(_MSC_VER) || defined(GNUC_AVX2)
```
chriseclectic, line 327: As above, this ifdef block looks wrong

src/transpile/fusion.hpp
```diff
@@ -299,28 +304,60 @@ op_t Fusion::generate_fusion_operation(const std::vector<op_t>& fusioned_ops,
...
+  if (!disable_simd_) {
```
chriseclectic, line 328:
```suggestion
  if (is_avx2_supported()) {
```

chriseclectic (changes requested):
After reviewing the code it seems the `disable_simd_` option is not really needed and makes the code much more cumbersome. There also appear to be several unnecessary ifdefs that are inconsistently used in different AVX2 locations.

chriseclectic:
I think #961 accomplishes the same as this PR but greatly simplifies the code.

PR #942, opened 2020-09-20
Fix a copy-paste-error in the noise model> it checks the wrong variable.

carstenblank:
### Summary

While working at a new provider and testing the possibility to use the noise model also for other devices, I found a copy-and-paste error in the value creation for the thermal relaxation routine. The problem is quite obvious when spotted and simply fixed too.

### Details and comments

The variable to check for the property `value` should for obvious reasons be `freq_params` instead of `t2_params`. 

I would have changed or added tests, but since this module is not tested at all, starting a test module would be a major change to the code instead of one simple modification of one line of code. I think it could be good to test this part too, but this should be a new issue / PR by itself.


chriseclectic (approved):
Thanks for fixing this @carstenblank. The frequency param is never actually used by the current device noise model, but its good to fix this since someone might use these utility functions for building their own custom noise model.

vvilpas (approved):

PR #945, opened 2020-09-21
Make controller_wrapper functions functors

ereastman:
### Summary

This adjusts the controller_execute functions to be functors rather than free-functions. This allows pickling of these objects - specifically it allows them to be used in multiprocessing.map(...) calls.

### Details and comments

See https://github.com/pybind/pybind11/issues/1261


qiskit/providers/aer/backends/wrappers/bindings.cc
```diff
@@ -20,21 +20,33 @@ DISABLE_WARNING_POP
 #include "controllers/unitary_controller.hpp"
 #include "controllers/controller_execute.hpp"
 
-PYBIND11_MODULE(controller_wrappers, m) {
-
-    m.def("qasm_controller_execute_json", &AER::controller_execute_json<AER::Simulator::QasmController>, "instance of controller_execute for QasmController");
```
vvilpas, line 25: I guess we don't use the `controller_execute_json` functions. Shouldn't we remove then their C++ counterparts: `AER::controller_execute_json<AER::Simulator::QasmController>`, etc?

mtreinish (commented):
Can you add a test case for the pickle/deepcopy of the backend classes? Something like: https://github.com/Qiskit/retworkx/blob/master/tests/test_deepcopy.py (which is what I added when I had to fix a deepcopy/pickle issue in Qiskit/retworkx#15)

ereastman:
> Can you add a test case for the pickle/deepcopy of the backend classes? Something like: https://github.com/Qiskit/retworkx/blob/master/tests/test_deepcopy.py (which is what I added when I had to fix a deepcopy/pickle issue in [Qiskit/retworkx#15](https://github.com/Qiskit/retworkx/pull/15))

I didn't know copy.deepcopy used the functionality I think we're going for .. but it kind of makes sense, cool. 

I might've gone a bit overboard in 16bffd0 but lmk what you think.

qiskit/providers/aer/backends/wrappers/bindings.cc
ereastman: Sure - done in 7bb6843

vvilpas (approved):
Looks good!

chriseclectic (approved):

PR #915, opened 2020-08-31
MPS: Apply kraus

merav-aharoni:

PR #948, opened 2020-09-22
MPS: Support snapshot_density_matrix

merav-aharoni:
### Summary
Added support for `snapshot_density_matrix` in the MPS simulator.

### Details and comments
This involved change in the `density_matrix_internal` function, because it previously did not support reordering of the qubits. 
This also required a change in expectation_value that depends on `density_matrix`: We no longer require reordering of the qubits here, because this is now implemented in `density_matrix`.


chriseclectic (approved):
Looks good. Can you add a release note

chriseclectic:
@merav-aharoni Since we're having trouble with the CI tests taking so long to run I've merged this, so can you add the release note for this PR to #915

PR #949, opened 2020-09-24
Fix test failures due to qiskit-terra #5125

chriseclectic:
### Summary

A recent Terra PR caused an incorrect global_phase value to be set in qobj headers for some circuits. This was causing tests to fail which tested that output unitary and statevectors were phase equal. This PR changes the failing tests to only check that hte output is equal up to global phase until the Terra issue can be fixed.

### Details and comments

Related terra issue: https://github.com/Qiskit/qiskit-terra/issues/5125


mtreinish (approved):
Hopefully there will be a terra-side fix soon but until this should be fine

PR #950, opened 2020-09-24
Allow compilation on Win32

vvilpas:
### Summary

Added Openblas library in 32 bits and changes made to allow compilation in Win32.
Openblas Win64 library has been modified to avoid dependencies on other dlls.


mtreinish (approved):
LGTM, this fixes win32 compilation for me locally. My 32bit CI PR pulled this in and it also works there: https://github.com/Qiskit/qiskit-aer/runs/1162077136?check_suite_focus=true

PR #952, opened 2020-09-25
Revert "Fix test failures due to qiskit-terra #5125 (#949)"

mtreinish:
### Summary

Qiskit/qiskit-terra#5125 has been closed as fixed by
Qiskit/qiskit-terra#5126. So we no longer need to disable checking the
results are phase equal. This commit adds back the phase checking so we
can test that global phase handling is correct again.

### Details and comments

This reverts commit a0274c33e3e5989d78fe67ffb7dd1332f3d71bd1.


chriseclectic (changes requested):
949 included some changes that shouldn't be reverted. Instead you should delete the `ignore_phase=True` flag in the locations where the TODO comments are.

chriseclectic:
Replaced by #958

PR #956, opened 2020-09-26
[WIP] Automatically set optimal OpenMP and fusion thresholds

### LINKED ISSUE #954: Re-evaluate default thresholds for OpenMP and Fusion

With the recent changes to the gate fusion algorithm and including SIMD we should re-evaluate the default threshold values for enabling OpenMP, and for enabling gate fusion for the different qubitvector based simulations States (Statevector, DensityMatrix, Unitary, Superoperator) to ensure good default performance.

chriseclectic:
### Summary

Adds basic profiling code to find suitable threshold values for OpenMP and gate fusion for backend options.

Closes #954 

### Details and comments

The automatic settings can be set by running `AerProvider.optimize_backend_options()` which sets attributes of the AerBackend that are used to populate the backend options at runtime (if no custom values are input).

This is sub-optimal since it needs to be run each session, TODO is to save this profile to an rc file that can be loaded automatically if present so it only needs to be run once per system. @mtreinish or @atilag can you help with how to do this?

### DIFF
--- a/qiskit/providers/aer/aerprovider.py
+++ b/qiskit/providers/aer/aerprovider.py
@@ -48,3 +48,10 @@ class AerProvider(BaseProvider):
 
     def __str__(self):
         return 'AerProvider'
+
+    @staticmethod
+    def optimize_backend_options(min_qubits=10, max_qubits=20, ntrials=10):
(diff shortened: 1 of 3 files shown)


chriseclectic:
Superseded by #1183

PR #958, opened 2020-09-28
Fix test performance

chriseclectic:
### Summary

This fixes certain tests were causing the GHA CI tests to fail to complete. 

### Details and comments

* Removes some redundant CCX and T gate tests that were slow for the extended stabilzier simulator
* Sets tests to run single threaded (except for thread management tests) to not conflict with thread usage of test library
* Reduces shot number of certain tests that don't use measurement sampling
* Reduces the number of qubits in fusion tests now that they must also be run on the density matrix simulator.
* Changes tests to to use terra circuit library QFT and QV functions instead of custom ones.


mtreinish (approved):
LGTM

PR #959, opened 2020-09-28
Refactor GHA config and attempt to fix pickle tests

chriseclectic:
### Summary

Changes the order of test dependencies in GHA to attempt to reduce bottlenecks. A local pip cache is also added to the job config.

### Details and comments

* The `wheel` and `sdist` tests were also set to require the `lint` test to pass
* The `wheel` test was removed as dependencies of the `tests` and `tutorial` actions since these tests takes a long time to run and block other tests running.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -157,6 +157,7 @@ jobs:
     name: tests-python${{ matrix.python-version }}-${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     needs: [standalone, sdist, lint, docs]
+    timeout-minutes: 25
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8]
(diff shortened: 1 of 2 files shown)


mtreinish (approved):

.github/workflows/main.yml
```diff
@@ -154,7 +156,7 @@ jobs:
   tests:
     name: tests-python${{ matrix.python-version }}-${{ matrix.os }}
     runs-on: ${{ matrix.os }}
-    needs: [standalone, wheel, sdist, lint, docs]
+    needs: [standalone, sdist, lint, docs]
```
mtreinish, line 159: We're still having job timeout issues despite the timeout fixtures and test performance improvements. Something else must be going on. But until we get to the bottom of it we should timeout sooner than 360mins after a job starts. Lets add:

```suggestion
    needs: [standalone, sdist, lint, docs]
    timeout-minutes: 25
```

PR #961, opened 2020-09-28
Move AVX2 logic into Transformer class and enable SIMD on Unitary, DensityMatrix, SuperOp

chriseclectic:
### Summary

This moves AVX2 logic into the class where it is used and removes needing to handle checking in the controller class. This also enables AVX2 automatically for unitary, density matrix, and superoperator simulators since they inherit the applied matrix function from Qubitvector.

### Details and comments

* Removes the QubitVectorAVX2 class and moves the logic for selecting the avx2 enabled `apply_matrix` into the base QubitVector class.
* SIMD will automatically be enabled in the unitary, superoperator and density matrix simulators since they call the qubitvector `apply_matrix` function.
* SIMD will also be used in gate fusion and noise model conversion since it they use the unitary and superoperator state classes.


src/simulators/statevector/qubitvector.hpp
```diff
@@ -959,95 +958,21 @@ QubitVector<data_t, Derived>::apply_reduction_lambda(Lambda&& func,
...
-      apply_lambda(lambda, qubits, convert(mat));
-    }
-  } // end switch
+  // TODO: Move transformer initialization somewhere else
+  if (is_avx2_supported()) {
+    TransformerAVX2<std::complex<data_t>*, data_t> transformer(data_, data_size_, omp_threads_managed());
+    transformer.apply_matrix(qubits, mat);
+  } else {
+    Transformer<std::complex<data_t>*, data_t> transformer(data_, data_size_, omp_threads_managed());
+    transformer.apply_matrix(qubits, mat);
```
vvilpas, line 970: As a first step we can make the `apply_matrix` functions static and pass all the additional data they need (`data_, data_size_, omp_threads_managed()`), avoiding the object initialization in each step, right?

chriseclectic: Yeah, if you would prefer i could change the signature to be `static apply_matrix(container, data_size, omp_threads, qubits, mat)`.

chriseclectic: But I was imagining the threads is specific to omp implementation, so maybe don't want it in the function signature for mixin subclasses

vvilpas (approved):
LGTM

PR #969, opened 2020-10-01
Add available memory checks for Statevector and Unitary simulators

### LINKED ISSUE #968: Segmentation fault during randomized benchmarking using statevector_simulator
### Informations

- **Qiskit Aer version**: 
{'qiskit-terra': '0.15.1', 'qiskit-aer': '0.6.1', 'qiskit-ignis': '0.4.0', 'qiskit-ibmq-provider': '0.8.0', 'qiskit-aqua': '0.7.5', 'qiskit': '0.20.0'}
- **Python version**: 3.6.6 
- **Operating system**:  CentOS Linux 7

Benchmarking using circuit gengerated by qiskit.ignis.verification.randomized_benchmarking on 36 qubits and simulate on statevector_simulator with reparations. segmentation fault (core dumped). Scirpt works until 34 qubits

Debuged with gdb and get 

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff0f7e6700 (LWP 313147)]
0x00007fff23237fd0 in AER::QV::QubitVector<double, AER::QV::QubitVectorAvx2<double> >::zero() [clone ._omp_fn.158] ()
```

test script
```
import qiskit.ignis.verification.randomized_benchmarking as rb
import qiskit
import random
from qiskit import  Aer, execute
import re
import os

nQ = 36
rb_opts = {}
rb_opts['length_vector'] = [100]
rb_opts['nseeds'] = 1
id = [i for i in range(nQ)]
random.shuffle(id)
pattern = [[id[i],id[i+int(nQ/2)]] for i in range(int(nQ/2))]
rb_opts['rb_pattern'] = pattern

rb_circs, xdata = rb.randomized_benchmarking_seq(**rb_opts)
circuit = rb_circs[0][0]
backend = Aer.get_backend('statevector_simulator')
results = execute(circuit, backend, shots=2048).result()
```

Scirpt finish without error

### Suggested solutions
no idea

chriseclectic:
### Summary

Closes #968 

Add missing available memory checks for the StatevectorController and UnitaryController in the C++ code. These throw an exception if the memory required to simulate the number of qubits in a circuit exceeds the  available memory of the system.

### Details and comments

This C++ memory check was only run for the QasmSimulator, while the other two simulators did the error checking in Python before running the C++ code. However the python checks are disabled if a qobj is run with `validate=False` which is the default setting.

### DIFF
--- a/releasenotes/notes/validate-memory-statevector-d1b4fc48b002856f.yaml
+++ b/releasenotes/notes/validate-memory-statevector-d1b4fc48b002856f.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    Add missing available memory checks for the
+    :class:`~qiskit.providers.aer.StatevectorSimulator` and
+    :class:`~qiskit.providers.aer.UnitarySimulator`. This throws an exception if
+    the memory required to simulate the number of qubits in a circuit exceeds the
+    available memory of the system.
--- a/src/controllers/statevector_controller.hpp
+++ b/src/controllers/statevector_controller.hpp
@@ -280,6 +280,9 @@ void StatevectorController::run_circuit_helper(
   // Validate circuit and throw exception if invalid operations exist
   validate_state(state, circ, noise, true);
 
+  // Validate memory requirements and throw exception if not enough memory
+  validate_memory_requirements(state, circ, true);
+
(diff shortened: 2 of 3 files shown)


vvilpas (approved):
LGTM

PR #973, opened 2020-10-04
multiplexer is an "opaque instruction"

1ucian0:
Terra is moving towards a model were Gates are reversible and we can check for that. For that we include more checks on a root gate (see  https://github.com/Qiskit/qiskit-terra/pull/4181)

This change raised a warning now on Aer (see https://github.com/Qiskit/qiskit-terra/issues/5099) because `Multiplexer` (introduced in https://github.com/Qiskit/qiskit-aer/pull/192) has parameter that is not definable under the traditional definition of opaque gate (a la QASM).

For now, I suggest this change to make it a subclass of `Instruction` that has no restrictions on the parameter types.


chriseclectic:
Maybe multiplexer should be a circuit in circuit library rather than gate/instruction.

1ucian0:
Agree. It  should not be a root instruction neither.

Issue #903, opened 2020-08-24
Remove parameter variance from snapshots


yaelbh:
### What is the expected behavior?

Computing the variance does not require much resources. In essence this is an accumulation of the squares of the values (see `add_data` and `normalize` in `src/framework/results/data/average_data.hpp`). Therefore it makes sense to just always compute and return the variance.

chriseclectic:
I think we should remove variance entirely from snapshots since it is actually the variance only from sampling over noise (and hence is always zero for ideal simulations). This seems to confuse people who are expecting it to be variance from measurement statistics.

yaelbh:
Ok, makes sense. By the way, the variance can be non-zero also for ideal simulations, if the circuit contains different measurement gates that write to the same classical qubit.

yaelbh:
Done in #974

PR #974, opened 2020-10-05
Deprecate snapshot variance

chriseclectic:
### Summary

The `variance` kwarg of Snapshot instructions has been deprecated. This function computed the sample variance in the snapshot due to noise model
    sampling, not the variance due to measurement statistics so was often
    being used incorrectly. If noise modeling variance is required single shot
    snapshots should be used so variance can be computed manually in
    post-processing.

### Details and comments

Deprecation will allow closing #903 in future release.


vvilpas (approved):
LGTM

PR #975, opened 2020-10-05
Use move semantics when snapshot is final circuit instruction

chriseclectic:
### Summary

Use move semantics for statevector and density matrix snapshots for the `"statevector"` and `"density_matrix"` methods of the QamsSimulator if they are the final instruction in a circuit. This reduces the memory usage of the simulator improves the performance by avoiding copying a large array in  the results.

### Details and comments

This adds a flag to `apply_ops` in the state class that can be set to True if no further ops will be applied to the state. This is currently used only for snapshots to check if they are the last instruction in a circuit, and if so can use move semantics where possible to avoid copying large arrays for statevector and density matrix states.

@hhorii In a future PR this could be used for other instructions such as measure sampling so that the statevector memory can be re-used to store the probabilities.

#### Density matrix snapshot 

This PR
mprof_dm_snap_pr

Master (with fix from #971)
mprof_dm_snap_master

Stable (0.6.1)
mprof_dm_snap_stable

#### Statevector snapshot

This PR
mprof_sv_snap_pr

Master (with fix from #971)
mprof_sv_snap_master

Stable (0.6.1)
mprof_sv_snap_stable


src/controllers/qasm_controller.hpp
```diff
@@ -962,16 +962,19 @@ void QasmController::run_multi_shot(const Circuit& circ,
...
+    bool final_ops = (pos == circ.ops.size());
     initialize_state(circ, state, initial_state);
-    state.apply_ops(ops, data, rng);
+    state.apply_ops(ops, data, rng, final_ops);
 
-    // Get measurement operations and set of measured qubits
-    ops = std::vector<Operations::Op>(circ.ops.begin() + pos,
-                                      circ.ops.end());
-    measure_sampler(ops, shots, state, data, rng);
+    if (!final_ops) {
```
vvilpas, line 969: Is this needed? If I remove this all tests pass.

chriseclectic: Its not needed, the functions in that block shouldn't do anything if `final_ops = false` anyway

vvilpas (approved):
LGTM

PR #979, opened 2020-10-07
Try to fix aer lint failure by installing aer

mtreinish:
### Summary

Since Qiskit/qiskit-terra#5086 merged the aer CI jobs have been failing.
It looks like it's because pylint can't find the pybind modules. I'm not
sure how Qiskit/qiskit-terra#5086 changed things, but this commit tries
to fix it by installing aer in the pylint job.

### Details and comments


chriseclectic:
This works, but it kind of defeats the point of a lint test since it now requires the sdist build to pass first.

PR #980, opened 2020-10-08
Alternate try to fix aer lint failure by installing aer 

chriseclectic:
### Summary

Qiskit/qiskit-terra#5086 seemd to break pylint testing by preventing the Aer package from being found by pylint without building and installing.

This tries to fix the issue by replacing all absolute imports in Aer with relative module imports. This is a work around until general namespace packaging issues are implemented to move all of Aer into its own namespace.

### Details and comments

#979 fixes the issue by building and installing Aer in the lint module, however this means lint passing is dependent on Aer being built via sdist, which is a separate test usually run after linting.


mtreinish (approved):
I guess this is fine, I think the bigger part of why this works is the disabling of no-name-in-module for the pybind imports from `controller_wrappers`. But disabling import checking on compiled extensions is pretty common practice for pylint.

PR #981, opened 2020-10-08
Use ExperimentResult rather than ExperimentData in Controller and State functions

chriseclectic:
### Summary

This changes the source code to pass around an ExperimentResult reference rather than an ExperimentData reference for storing simulator result data. This is so other data in the result can be accessed such as metadata, which was previously being copied from ExperimentData to ExperimentResult. This PR is also in preparation of changing the way snapshot data is handled to make it easier to deprecate the current `ExperimentData` class and add a new class for storing experiment data.


vvilpas (approved):
LGTM!

PR #982, opened 2020-10-08
add SIMD optimization of diagonal matrix multiplication

hhorii:
### Summary

Add AVX2/FMA intrinsics to multiple double/float complex vectors

### Details and comments

This PR provides SIMD instructions for diagonal matrix multiplication.


hhorii:
image

20-30% improvement for 20 or more qubits in Xeon x2 env.
No improvement (but no side-effect) in Macbook env.

src/simulators/statevector/qv_avx2.cpp
```diff
@@ -11,13 +11,15 @@
...
-
+#include <iostream>
 #include "qv_avx2.hpp"
 #include <immintrin.h>
 #include <cstdint>
 #include <cstring>
 #include <type_traits>
 #include <utility>
+#include <complex>
+#include <omp.h>
```
chriseclectic, line 22: I think this needs a an `#ifdef` guard:
```suggestion
#ifdef _OPENMP
#include <omp.h>
#endif
```
same for any part of the code that uses something from this header file

src/simulators/statevector/qv_avx2.cpp
```diff
@@ -1045,12 +1102,150 @@ template Avx apply_matrix_avx<double>(double*,
...
+
+  std::complex<double>** inputs;
+  inputs = reinterpret_cast<std::complex<double>**>(malloc(sizeof(std::complex<double*>) * omp_threads));
+
+#pragma omp parallel for if (omp_threads > 1) num_threads(omp_threads)
+  for (int i = 0; i < omp_threads; ++i) {
+#if !defined(_WIN64) && !defined(_WIN32)
+    void* data;
+    posix_memalign(&data, 64, sizeof(std::complex<double>) * 2);
+    inputs[omp_get_thread_num()] = reinterpret_cast<std::complex<double>*>(data);
```
chriseclectic, line 1132: I think this is the part that needs a safe option for no OpenMP, something like (just guessing here for the non-omp case):

```c++
#ifdef _OPENMP
size_t index = omp_get_gread_num()
#else
size_t index = 0
#endif

inputs[index] = reinterpret_cast<std::complex<double>*>(data);
```

Similarly for other parts of the code that use `omp_get_thread_num`

chriseclectic (commented):
I'll leave the proper code review for @vvilpas, but I think the use of the omp header needs to be put in an `#ifdef _OPENMP` block so that it can work for systems without openmp (this is what is currently done in other parts of code that use the <omp.h> header)

vvilpas (changes requested):
I've left some comments to review.

Another thing that would be great is, if possible, to name those magic numbers, that appears through the implementation, like the capacity of the AVX vectors (being 2 or 4 for lfoat or double)

src/simulators/statevector/qv_avx2.cpp
```diff
@@ -1045,12 +1102,150 @@ template Avx apply_matrix_avx<double>(double*,
...
+
+  size_t q0_mask_ = 0;
+  size_t q1_mask_ = 0;
+  for (size_t i = 0; i < qregs_size; ++i) {
+    if (qregs[i] == 0) {
+      q0_mask_ = 1UL << i;
+    } else if (qregs[i] == 1) {
+      q1_mask_ = 1UL << i;
+    }
+  }
```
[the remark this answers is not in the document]
hhorii: Can I keep the current impl? I guess this is simpler than a case that uses a condition.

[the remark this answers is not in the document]
vvilpas: Yeah, if the change is not worth it and taking the last index is correct in this case.

src/simulators/statevector/qv_avx2.cpp
```diff
@@ -1118,21 +1171,24 @@ Avx apply_diagonal_matrix_avx<double>(double* qv_data_,
...
-  inputs = reinterpret_cast<std::complex<double>**>(malloc(sizeof(std::complex<double*>) * omp_threads));
+  std::complex<double>** tmps;
+  tmps = reinterpret_cast<std::complex<double>**>(malloc(sizeof(std::complex<double>*) * omp_threads));
 
-#pragma omp parallel for if (omp_threads > 1) num_threads(omp_threads)
-  for (int i = 0; i < omp_threads; ++i) {
+#pragma omp parallel if (omp_threads > 1) num_threads(omp_threads)
+  {
+#pragma omp for
+    for (int i = 0; i < omp_threads; ++i) {
```
vvilpas, line 1183: Not sure but, aren't these statements redundant? Aren't we already parallelizing in `num_threads` with the above `#pragma omp parallel if (omp_threads > 1) num_threads(omp_threads)` statement?

src/simulators/statevector/qv_avx2.cpp
```diff
@@ -1186,21 +1236,23 @@ Avx apply_diagonal_matrix_avx<float>(float* qv_data_,
...
-#endif
+  auto qv_data = _to_complex(qv_data_);
+  const auto input_vec = _to_complex(vec_);
+
+  auto tmps = reinterpret_cast<std::complex<float>**>(malloc(sizeof(std::complex<float>*) * omp_threads));
+
+#pragma omp parallel if (omp_threads > 1) num_threads(omp_threads)
+  {
+#pragma omp for
+    for (int i = 0; i < omp_threads; ++i) {
```
vvilpas, line 1247: Same comment as above

vvilpas (commented):
Looks good! I have only a question about OMP (see comments)

src/simulators/statevector/qv_avx2.cpp
hhorii: This is the same with [`indexes.hpp`](https://github.com/Qiskit/qiskit-aer/blob/a8b30d3e07129cbedeb4ded9cb03e18132fb542f/src/simulators/statevector/indexes.hpp#L180-L187). `omp parallel` is applied to `for`-loop with `#pragma omp for` annotation within its scope, in my understanding.

vvilpas: But in `indexes.hpp` we use the index `k` coming from the `for` statement and running from `start` to `stop`. Here we don't use the index from the `for` in any way and we are running the loop from 0 to `num_threads`. Looks like with the first `pragma` we are already creating the threads that we need.

src/simulators/statevector/qv_avx2_test.cpp
```diff
@@ -0,0 +1,259 @@
+#include <iostream>
```
vvilpas, line 1: Any reason to keep this file?

src/simulators/statevector/qv_avx2.cpp
hhorii: Oh, that's true. In this case, we do not need `omp` for `for`. Thanks for your comments!

vvilpas (approved):

PR #986, opened 2020-10-12
[WIP] refactor fusion and add paralellization of fusion

hhorii:
### Summary

Refactor fusion to add variation of fusion techniques in future

### Details and comments

This PR isolates `FusionMethod` from `Fusion` transpilation. This `FusionMethod` will be used in the other from `Fusion`.
In addition, this PR enables multi-threading for transpilation for very-long circuits.


chriseclectic (commented):
Initial review: I'm not sure if this is still intended to be WIP or not so if so maybe some of this is still being worked on. It seems we should break up the fusion analysis (finding the sets of gates to fuse) from the fusion application. For the fusion method classes it looks like some of the config settings are being built into the classes in a way that hides their configuration (like the qubit thresholds), the classes also seem to have some unnecessary methods.

PR #990, opened 2020-10-14
Fix fusion thresholds and defaults

chriseclectic:
### Summary

Changes default gate fusion thresholds for improved performance, and fixes default max qubit size for fusion bugs in density matrix and noise fusion introduced in #927 

This is more of a quick fix for the 0.7 release, and a more complete refactor of fusion logic can be tackled in #986 

### Details and comments

* Changes `"fusion_threshold"` backend option to apply fusion when the number of qubits is above the threshold, not equal or above the threshold, to match the behaviour of the OpenMP qubit threshold parameter.

* Changes the default value of "fusion_threshold" from 20 to 14 for the QasmSimulator (statevector) and StatevectorSimulator, and from 10 to 7 for the UnitarySimulator and QasmSimulator (density matrix).

* Allows fusion for MPS if fusion enabled is specified in config. Default for MPS is still fusion disabled.

* Fixes bug in fusion for density matrix method introduced in #927 which was preventing the max qubit size for fusion being halved for the density matrix method. This was resulting in the fused circuit being slower for density matrix simulations.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,7 +8,7 @@
 
 # Generic rule for the repository. This pattern is actually the one that will
 # apply unless specialized by a later rule
-* @chriseclectic @atilag
+* @chriseclectic @vvilpas @atilag
 
 # Individual folders on root directory
 /qiskit         @chriseclectic @atilag @vvilpas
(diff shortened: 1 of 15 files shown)


vvilpas (approved):
LGTM!

PR #999, opened 2020-10-20
Cmake: make conan optional

### LINKED ISSUE #757: Add option to build without conan and document it

vvilpas:
Quoting original PR creator (#877):
### Summary

This allows Conan to be an optional dependency. Conan requires an internet connection to build, and is meant for replicating development environments and not for downstream packaging. It also does not allow using the existing libraries in your system, e.g. ``muparserx`` (https://aur.archlinux.org/packages/muparserx/).

### Details and comments

This is useful for having a single option for disabling Conan in the build system, instead of patching CMakeLists.txt.

This isn't the prettiest solution, and I'm not super familiar with CMake files, so I'm open to suggestions. The downside of this is that it doesn't do any version checking on the packages that it finds, which is probably not a great practice.

Fixes #757 
Continues work started on #877 by drewrisinger


vvilpas:
Ready for review. I can build now on linux and macos without conan.

cmake/conan_utils.cmake
```diff
@@ -1,33 +1,46 @@
 include(conan)
 
+macro(_rename_conan_lib package)
+    add_library(AER_DEPENDENCY_PKG::${package} INTERFACE IMPORTED)
```
atilag, line 4: oh! nice hack! :)

CMakeLists.txt
```diff
@@ -268,16 +279,16 @@ endif()
...
-	CONAN_PKG::spdlog
 	${DL_LIB}
-	${THRUST_DEPENDANT_LIBS})
+	${THRUST_DEPENDENT_LIBS})
 
 set(AER_COMPILER_DEFINITIONS ${AER_COMPILER_DEFINITIONS} ${CONAN_DEFINES})
 # Cython build is only enabled if building through scikit-build.
 if(SKBUILD) # Terra Addon build
-	set(AER_LIBRARIES ${AER_LIBRARIES} CONAN_PKG::muparserx)
+	set(AER_LIBRARIES ${AER_LIBRARIES} AER_DEPENDENCY_PKG::muparserx)
```
atilag, line 291: This is probably on me, but I'm wondering if we should be using `list()` here instead of `set()` , we are using `list()` in other places for appending libraries (see `AER_CONAN_LIBS` usage), also feels more semantically correct. Thoughts?

CMakeLists.txt
```diff
@@ -320,9 +331,9 @@ else() # Standalone build
 		PRIVATE ${AER_COMPILER_DEFINITIONS})
 	if(WIN32 AND NOT BLAS_LIB_PATH)
```
atilag, line 332: Also wondering if we can simplify the `BLAS_LIB_PATH` logic now that we have a way to disable Conan. 
I'm concerned about the complexity of the build system in general, and having a way to skip over BLAS, and skip over everything else looks confusing, doesn't it?. 
What about removing the `BLAS_LIB_PATH` and if someone wants to build with their own BLAS, force them to go through the disable Conan path?. Certainly it will require them to have not just BLAS but every dependency installed.
Thoughts?

CMakeLists.txt
atilag: BTW, I think I started to use `set()` instead of `list()` due to some misbehavior in CMake, I just can't recall exactly which one. If  `list()` works and covers our use-cases then go to `list()`. Maybe in another PR with a "Good Fist Issue" tag?.

cmake/conan_utils.cmake
```diff
@@ -40,4 +53,25 @@ macro(setup_conan)
...
+    if(AER_THRUST_BACKEND AND NOT AER_THRUST_BACKEND STREQUAL "CUDA")
+        set(AER_SIMULATOR_CPP_EXTERNAL_LIBS ${AER_SIMULATOR_CPP_EXTERNAL_LIBS} ${CONAN_INCLUDE_DIRS_THRUST})
+    endif()
+
+    # Reassign targets from CONAN_PKG to AER_DEPENDENCY_PKG
+    foreach(CONAN_LIB ${AER_CONAN_LIBS})
+        _rename_conan_lib(${CONAN_LIB})
+    endforeach()
+
+    if(APPLE)
```
atilag, line 67: At some point we'll need to review this. Maybe it's been fixed with current versions of Clang?

cmake/dependency_utils.cmake
```diff
@@ -0,0 +1,64 @@
...
+		string(TOLOWER ${AER_THRUST_BACKEND} THRUST_BACKEND)
+		_import_aer_system_dependency(Thrust 1.9.5)
+	endif()
+
+	if(BUILD_TESTS)
+		_import_aer_system_dependency(Catch2 2.12.1)
+	endif()
+
+	if(APPLE)
+		# Fix linking. See https://stackoverflow.com/questions/54068035
```
atilag, line 43: Really? :(

atilag: Sometimes MacOS platform is so unpleasant for system devs :(

atilag (approved):
LGTM!
Haven't seen any blockers, just left some comments that could be addressed later.

CMakeLists.txt
vvilpas: yeah, i think `list` is the way to go. We can check that it works and and replace all those `set`s.

CMakeLists.txt
vvilpas: Not sure, because this was requested in order to use the same `BLAS` library as the one `Numpy` uses. So probably not in the usual paths where CMake will look for. Not sure if this option is very used, though. 
But yeah, I agree, the build system is quite complex.

cmake/dependency_utils.cmake
vvilpas: Maybe there are better ways to do this, but that's the way that I found. :(

CMakeLists.txt
vvilpas: yeah, i think list is the way to go. We can check that it works and and replace all those sets.

CMakeLists.txt
vvilpas: Not sure, because this was requested in order to use the same `BLAS` library as the one `Numpy` uses. So probably not in the usual paths where CMake will look for. Not sure if this option is very used, though. 
But yeah, I agree, the build system is quite complex.

cmake/dependency_utils.cmake
vvilpas: Maybe there are better ways to do this, but that's the way that I found. :(

Issue #991, opened 2020-10-14
Kraus noise crashes MPS simulator


chriseclectic:
### Informations

- **Qiskit Aer version**: 0.7
- **Python version**: 3.7
- **Operating system**: MacOS

### What is the current behavior?

While running some tests I was getting crashes when I tried to use Kraus noise on the MPS simulator.

### Steps to reproduce the problem

Run the following code:

```python
from qiskit import execute
from qiskit.circuit.library import QFT
from qiskit.providers.aer import QasmSimulator
from qiskit.providers.aer.noise import NoiseModel, amplitude_damping_error

error1 = amplitude_damping_error(0.2)
noise_model = NoiseModel()
noise_model.add_all_qubit_quantum_error(error1, ['id', 'u1', 'u2', 'u3'])
noise_model.add_all_qubit_quantum_error(error1.tensor(error1), ['cx'])

backend = QasmSimulator(method='matrix_product_state', noise_model=noise_model)

job = execute(QFT(5), backend)
```

This crashes the python kernel with a blas error:
```
ldb must be >= MAX(K,1): ldb=0 K=0. BLAS error: Parameter number 11 passed to cblas_zgemm had an invalid value
```

### What is the expected behavior?

No crash

### Suggested solutions

The error looks like something is going wrong with a matrix multiplication somewhere in the source code

chriseclectic:
Closed by #1003

chriseclectic:
@merav-aharoni looks like this is still an issue on master. A couple of times when running the above script I get a malloc error instead of the BLAS error:

```
python(56590,0x7000157e6000) malloc: *** error for object 0x7fa91b675a60: pointer being freed was not allocated
python(56590,0x7000187f8000) malloc: Incorrect checksum for freed object 0x7fa91b6ce188: probably modified after being freed.
Corrupt value: 0x20
python(56590,0x7000157e6000) malloc: *** set a breakpoint in malloc_error_break to debug
```

But if i run it several times in a row it seems to return to the BLAS error

chriseclectic:
Hopefully fixed this time by #1188

PR #1003, opened 2020-10-22
Fixed bug in MPS::apply_kraus

merav-aharoni:
### Summary
Fixed bug where MPS::apply_kraus was crashing on QFT(5) - see issue #991. 

### Details and comments
In MPS, for all gates, we need to apply the gate on the internal qubits, because the qubits are moved around during execution of the circuit. In apply_kraus, we forgot to take the internal_qubits, therefore the kraus noise was being applied to the wrong qubits. See #issue991 for the failing test case.


src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -648,6 +648,59 @@ void MPS::apply_diagonal_matrix(const AER::reg_t &qubits, const cvector_t &vmat)
...
 }
 
+void MPS::apply_kraus(const reg_t &qubits,
+                   const std::vector<cmatrix_t> &kmats,
+                   RngEngine &rng) {
+  reg_t internal_qubits = get_internal_qubits(qubits);
+  apply_kraus_internal(qubits, kmats, rng);
+
+}
+void MPS::apply_kraus_internal(const reg_t &qubits,
```
chriseclectic, line 658: I think this would be clearer with just a single function that converts the args to the internal qubits and uses them:

```cpp
void MPS::apply_kraus(const reg_t &qubits,
                   const std::vector<cmatrix_t> &kmats,
                   RngEngine &rng) {

   reg_t internal_qubits = get_internal_qubits(qubits);
  // code from apply_kraus_internal using internal qubits

}
```

chriseclectic (changes requested):
Can you combine the two internal functions int oa single `apply_kraus` function that computes and uses the internal qubits.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I can, of course unite the functions, but I would rather not, because of the following reasons:
1. This is the structure that is defined for all the operations - a short api function that only gets the internal qubits and the internal function that actually does something. There is a logical separation between the 2 levels.
2. My initial bug came from not doing the separation, and forgetting that it was necessary. That is also the reason I moved the functions to be inside matrix_product_state_internal.[ch]pp.
3. Sometimes I may want to access the internal function directly. For example, when I call it from another internal function.

chriseclectic (approved):
LGTM @merav-aharoni Can you add a release note for the bugfix?

chriseclectic (approved):

PR #1005, opened 2020-10-27
Fix bug in unsupported instructions error message

### LINKED ISSUE #1004: New errors about invalid instructions for the MPS simulator
When a circuit contains the `ry` gate (this happens also with many other gates, probably all the `r` ones), like the following circuit:
```
import qiskit
sim = qiskit.Aer.get_backend('qasm_simulator')

qc = qiskit.QuantumCircuit(2)
qc.h(0)
qc.cx(0, 1)
qc.ry(4, 0)

res = qiskit.execute(qc, sim,
                     backend_options={'method': 'matrix_product_state'}).result()

from pprint import pprint
pprint(vars(res))
```
This is the output with Qiskit version 0.23.0:
```
{<some irrelevant data that I deleted>,
 'results': [ExperimentResult(shots=0, success=False, meas_level=MeasLevel.CLASSIFIED, data=ExperimentResultData(), status=ERROR: Circuit contains invalid instructions {"optypes": {gate}, "gates": {cx, ry, h}} for "matrix_product_state" method, seed_simulator=0, time_taken=0.0)],
 'status': 'ERROR:  [Experiment 0] Circuit contains invalid instructions '
           '{"optypes": {gate}, "gates": {cx, ry, h}} for '
           '"matrix_product_state" method',
 'success': False}
```
whereas with Version 0.22.0 I get:
```
{...,
 'results': [<qiskit.result.models.ExperimentResult object at 0x7fdbc2558668>],
 'status': 'COMPLETED',
 'success': True}
```

Questions/comments:
1. Is it true that the MPS simulator does not support gates like `ry` and `rz`? This is not what I remember.
2. If `ry` is invalid, then the error message should not include `h` and `cx`; users are surprised to hear that `cx` apparently is invalid for the MPS simulator...
3. How comes it is not working now, if it used to work until recently?
4. Looks like more tests are required, since I expect coverage to be large enough to capture such a scenario.

chriseclectic:
### Summary

Fixes bug in the error message returned when a circuit contains unsupported simulator instructions. Previously some supported instructions were also being listed in the error message along with the unsupported instructions.

Fixes #1004 

### Details and comments

Fixes bug in how sets of unsupported instructions were constructed. The previous method didn't work properly with `std::unordered_set`.

### DIFF
--- a/src/framework/opset.hpp
+++ b/src/framework/opset.hpp
@@ -36,29 +36,30 @@ private:
     }
   };
 
-
 public:
   // Alias for set of OpTypes
   using optypeset_t = std::unordered_set<Operations::OpType, EnumClassHash>;
 
   // Public data members
-  optypeset_t optypes;     // A set of op types
-  stringset_t gates;      // A set of names for OpType::gates
-  stringset_t snapshots;  // set of types for OpType::snapshot
+  optypeset_t optypes;   // A set of op types
+  stringset_t gates;     // A set of names for OpType::gates
+  stringset_t snapshots; // set of types for OpType::snapshot
 
   OpSet() = default;
 
-  OpSet(const optypeset_t &_optypes,
-        const stringset_t &_gates,
+  OpSet(const optypeset_t &_optypes, const stringset_t &_gates,
         const stringset_t &_snapshots)
-    : optypes(_optypes), gates(_gates), snapshots(_snapshots) {}
-  
-  OpSet(optypeset_t &&_optypes,
-        stringset_t &&_gates,
-        stringset_t &&_snapshots)
-    : optypes(std::move(_optypes)), gates(std::move(_gates)), snapshots(std::move(_snapshots)) {}
+      : optypes(_optypes), gates(_gates), snapshots(_snapshots) {}
 
-  OpSet(const std::vector<Op> &ops) { for (const auto &op : ops) {insert(op);} }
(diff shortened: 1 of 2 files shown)


src/framework/opset.hpp
```diff
@@ -216,27 +221,33 @@ OpSet OpSet::difference(const OpSet &_opset) const {
...
 OpSet::optypeset_t OpSet::difference(const optypeset_t &_optypes) const {
   optypeset_t ret;
-  std::set_difference(_optypes.begin(), _optypes.end(),
-                      optypes.begin(), optypes.end(),
-                      std::inserter(ret, ret.begin()));
+  for (const auto& item : _optypes) {
+    if (optypes.count(item) == 0) {
+      ret.insert(item);
+    }
+  }
```
vvilpas, line 228: as we use it several times, If it is easy maybe we can put this in a function?
```
template <typename T>
std::unordered_set<T> set_difference_unordered(const std::unordered_set<T>& a, const std::unordered_set<T>& b){ ... }
```
(you may need the Hash template parameter as well to make it work with `optypeset_t`s)

and have only:
```
OpSet::optypeset_t OpSet::difference(const optypeset_t &_optypes) const {
   return set_difference_unordered(_optypes, optypes);
}

```

vvilpas (changes requested):
Good catch about unordered_sets and set_difference.

vvilpas (approved):
LGTM!

PR #1006, opened 2020-10-27
Fix Aer provider to return new backend instances with backends/get_backend.

chriseclectic:
### Summary

Updates AerProvider to return a new backend instance each time `backend` or `get_backend` is called. This is so that if the backend instance is modified with `set_options` it won't persist pass the scope of where that backend object is used.


mtreinish (approved):
LGTM, seems straightforward enough

mtreinish (approved):

PR #1007, opened 2020-10-27
Allow `AerProvider.get_backend` to return configured backend

chriseclectic:
### Summary

Allows returning a configured backend using `AerProvider.get_backend` by calling `set_options` on the backend object before returning.

### Details and comments

PR #1006 should be merged first to prevent the backend options from persisting past the lifetime of the returned object.

### DIFF
--- a/releasenotes/notes/aer-provider-config-backend-3a89b7587c34c6ee.yaml
+++ b/releasenotes/notes/aer-provider-config-backend-3a89b7587c34c6ee.yaml
@@ -1,0 +1,43 @@
+---
+upgrade:
+  - |
+    Allows setting configurable backend options when using
+    :meth:`~qiskit.providers.aer.AerProvider.get_backend` to return a backend
+    object. This is done via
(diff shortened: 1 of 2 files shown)


releasenotes/notes/aer-provider-config-backend-3a89b7587c34c6ee.yaml
```diff
@@ -0,0 +1,43 @@
+---
+upgrade:
```
mtreinish, line 2: This note reads like a feature note instead of an upgrade note. 

That being said I actually worry about the backwards compat for this. Previously the kwargs would be used as filters that get passed to `backends`. This is used to set filters on `backend.configuration()` which is useful for real backends with immutable configurations, for aer I'm not sure how this would be used. But if you want to make the breaking change you need to call it out as such

vvilpas:
PR #1006 is already merged. Is this still on hold?

chriseclectic:
I think this can be closed and if people use the provider interface they need to set additional options manually by calling `set_options` on the returned backend.

Issue #997, opened 2020-10-19
Incorrect result in CCX using MPS simulation method


merav-aharoni:
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?
From StackExchange:
"I implemented Grover algorithm with matrix_product_state method in qiskit. But I found something strange. I designed the oracle part come out |1111⟩ but, real output was 0011⟩. I check out my code, but there is nothing wrong with it. Even simulating with the other simulation method (such as statevector, density_matrix etc.) "

### Steps to reproduce the problem

See use code in : ``https://quantumcomputing.stackexchange.com/questions/14236/grovers-algorithm-with-matrix-product-state-method/14248#14248

### What is the expected behavior?
Should give correct result.

chriseclectic:
Can you reproduce this or confirm its an error in gate, not the circuit generation code?

Unit tests should (hopefully) catch an error in the ccx and other basis gates. Also running randomized tests like the following i get the same distributions between MPS and statevector:

```python
import qiskit
import qiskit.quantum_info as qi
from qiskit.providers.aer import QasmSimulator
from qiskit.visualization import plot_histogram
from numpy.random import choice

# Random pre and post unitary
nq = 4
U1 = qi.random_unitary(2 ** nq).data
U2 = qi.random_unitary(2 ** nq).data
qubits = list(range(nq))

# Random CCX qubits
ccx_qubits = choice(qubits, 3, replace=False)

qc = qiskit.QuantumCircuit(nq)
qc.unitary(U1, qubits)
qc.barrier(qubits)
qc.ccx(*ccx_qubits)
qc.barrier(qubits)
qc.unitary(U2, qubits)
qc.measure_all()
circ = qiskit.transpile(qc, basis_gates=['u1', 'u2', 'u3', 'cx', 'ccx'])

sim1 = QasmSimulator(method='statevector')
sim2 = QasmSimulator(method='matrix_product_state')
counts1 = qiskit.execute(circ, sim1, shots=10000).result().get_counts(0)
counts2 = qiskit.execute(circ, sim2, shots=10000).result().get_counts(0)
plot_histogram([counts1, counts2])
```

merav-aharoni:
From what I have seen up to now, I believe the reported problem is in the measurement, not in the ccx gate. But I have just started investigating.

merav-aharoni:
Regarding your example above, the distributions are close enough. There is no problem there.

[...]

PR #1011, opened 2020-10-28
Corrected qubits ordering in MPS::sample_measure_using_probabilities. 

merav-aharoni:
### Summary
Bug was reported in issue #997. The error was in the ordering of the qubits in sample_measure_using_probabilities.

### Details and comments
The correction is to order the qubits before invoking `sample_measure_using_probabilities`, because the input qubits are always ordered in advance in `qasm_controller.py`. See ordering also in `apply_measure`.


vvilpas (dismissed):
LGTM!

chriseclectic:
@merav-aharoni Can you add a release note for the bugfix?

vvilpas (approved):

PR #1012, opened 2020-10-28
Add PauliGate to density matrix simulator

chriseclectic:
### Summary

Extends #919 to work with the density matrix simulators.

### Details and comments

Also removes unnecessary expval_pauli function from densitymatrix class since this is equivalent to calling the base statevector class method


src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -4797,22 +4800,22 @@ double QubitVectorThrust<data_t>::expval_pauli(const reg_t &qubits,
...
       // phase = -1
-      phase = thrust::complex<data_t>(-1, 0);
+      phase = thrust::complex<data_t>(-phase.real(), -phase.imag());
       break;
     case 3:
       // phase = 1j
-      phase = thrust::complex<data_t>(0, 1);
+      phase = thrust::complex<data_t>(-phase.imag(), phase.real());
       break;
   }
```
vvilpas, line 4820: If I'm not wrong we can use `add_y_phase` function defined in `qubitvector` adding a template parameter:

```
template <template<class> typename complex_type>
void add_y_phase(uint_t num_y, complex_type<data_t>& coeff){ .... }
```

src/simulators/statevector/qubitvector_thrust.hpp
```diff
@@ -5038,22 +5043,22 @@ void QubitVectorThrust<data_t>::apply_pauli(const reg_t &qubits, const std::stri
...
       break;
     case 2:
       // phase = -1
-      phase = thrust::complex<data_t>(-1, 0);
+      phase = thrust::complex<data_t>(-phase.real(), -phase.imag());
       break;
     case 3:
       // phase = 1j
-      phase = thrust::complex<data_t>(0, 1);
+      phase = thrust::complex<data_t>(-phase.imag(), phase.real());
```
vvilpas, line 5061: Same as above

vvilpas (changes requested):
I think we can avoid a bit code duplication (although it was already in the code before this PR).
Other than that it is ok!

src/simulators/statevector/qubitvector_thrust.hpp
chriseclectic: Only thing is QVThrust cant access functions of QV (its basically a duplicate class), so it would just be adding that function to this class as well.

vvilpas: right, I wrongly assumed it was a derived class of QubitVector. I guess there are more chunks of code duplicated so I think we can add the function here as well, and at some point try to put the common code in a place accesible to both classes.

vvilpas (approved):

PR #1014, opened 2020-10-29
Use Terra operators instead of Qutip qobjects.

vvilpas:
### Summary

Migrate usage from Qutip Qobjects to Terra operators. Remove qutip and associated cython dependencies as they are not needed anymore.

### Details and comments

Use Terra operators now instead of qutip qobj. 

- The main difference is that Terra operators use a numpy array whereas qobj use a sparse matrix as internal representation. This has an impact in rhs function evaluation (~30% performance decrease), so that function internally converts the problematic matrix to a sparse representation to keep performance as before.

- As we are not using qutip qobjects anymore, qutip and cython dependencies have been removed.


DanPuzzuoli (commented):
Aside from my one comment the python changes look good to me. 

Unfortunately I don't know C++ so am unable to review those pieces.

vvilpas:
@chriseclectic @mtreinish Should I add a release note here? It is an internal change, but we get rid of cython dependency.

chriseclectic (approved):

chriseclectic (approved):

PR #1031, opened 2020-11-11
Fix conan dependency problem with urllib3

vvilpas:
### Summary

`urllib3` 1.26.0 has been released, causing a dependency when installing conan (coming form `requests` lib).
Here I add `urllib3` add as a dependency and constraint it to <1.26 to avoid this problem.

### Details and comments

The issue has been already reported to `Conan` team by some users: https://github.com/conan-io/conan/issues/8041.
Here https://github.com/conan-io/conan/pull/8042 they mention that `requests` lib version 2.25 (to be released ~~today~~ tomorrow: https://github.com/psf/requests/issues/5654) will be compatible with this `urllib3` 1.26. In the meantime we can use this ugly fix.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -81,7 +81,7 @@ jobs:
         with:
           python-version: 3.7
       - name: Install deps
-        run: pip install conan
+        run: pip install urllib3<1.26 conan
       - name: Install openblas
         run: |
           set -e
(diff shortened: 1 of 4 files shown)


mtreinish (approved):
The only thing I'm not sure of is the pyproject.toml change. But it doesn't hurt since this is just a temporary PR to unblock CI. We should revert this once there is a working combination of urllib3, requests, and conan again.

Issue #928, opened 2020-09-13
Snapshot to return a single amplitude on MPS


merav-aharoni:
### What is the expected behavior?
In order to use the `MPS simulation_method`, @gadial requested a snapshot that returns a single amplitude of the `statevector`, without computing the entire `statevector`.
This functionality is useful for MPS, because it saves computing the entire `statevector`, if not needed. 
**Question:** Is this useful for other `simulation_methods`?
The interface can look something like:
`circ.snapshot_amplitude("my amp", ["010001"])`

chriseclectic:
I've been meaning to add this functionality for the statevector simulator for awhile, so it should be added for statevector, statevector_gpu, and MPS.

I would make it specify a list of amplitudes (as either as strings or ints) and then return the array of complex numbers for all specified amplitudes. If the params is the full range `[0, ..., n-1]` then the output should be identical to the statevector snapshot.

It could also be convenient to allow a qubits argument, so that the strings/ints are the values for that qubit mask, with all other qubits assumed to be the 0-state for the full amplitude.

merav-aharoni:
@chriseclectic , a few questions:

1. Should this be a new snapshot or a new parameter of the `statevector` snapshot. We could add a parameter "amplitudes" as
follows:
`qc.snapshot_statevector("sv", amplitudes = ["0011", "0111"])`
2. Should we allow the amplitudes to appear in mask form? For example, the above could also be specified as 
`amplitudes = ["0X11"]`
3. I am not sure what you mean by a "qubits argument." The current statevector snapshot does not support a partial subset of the qubits.

chriseclectic:
A new snapshot. Something like

```python
class SnapshotAmplitudes:
    def __init__(self, label: str, params: List[int], num_qubits=0):
```

And it should assemble to a qobj instruction
````python
QasmQobjInstruction(
    name='snapshot',
    snapshot_type='amplitudes',
    label='label',
    qubits=[0, ..., N-1],
    params=[int0, int1, ..]
)
````
where params are stored as ints rather than bitstrings.

From python side we could also allow bit-string inputs for params and convert them to ints when constructing the python object. 

[...]

PR #1034, opened 2020-11-11
Snapshot amplitudes in MPS

merav-aharoni:


merav-aharoni:
This is the first version of a new snapshot: `snapshot_amplitudes`. Given a list of base values in hexa form, this snapshot returns the amplitudes of those base values. This is a respond to issue #928.
Currently this feature is supported in MPS only.
Remaining tasksL
1. Support in` statevector` simulation method
2, Support the 'qubits' parameter as specified in the issue.
I will handle these in separate PRs in order not to have too much code in this one.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1221,6 +1221,44 @@ void MPS::full_state_vector_internal(cvector_t& statevector,
   statevector = reverse_all_bits(temp_statevector, num_qubits);
 }
 
+void MPS::get_amplitude_vector(const reg_t &base_values, cvector_t &amplitude_vector) {
+  move_all_qubits_to_sorted_ordering();
```
yaelbh, line 1225: Insertion of swaps should be avoided when possible, and can be avoided in this case

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1221,6 +1221,44 @@ void MPS::full_state_vector_internal(cvector_t& statevector,
...
+  uint_t num_values = base_values.size();
+  std::string base_value;
+#pragma omp parallel for if (num_values > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
+  for (int_t i=0; i<static_cast<int_t>(num_values); i++) {
```
yaelbh, line 1229: Nicer:
```
for (const auto &bv_int : base_values)
```

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1221,6 +1221,44 @@ void MPS::full_state_vector_internal(cvector_t& statevector,
...
+  cmatrix_t temp = q_reg_[0].get_data(bit);
+
+  for (int_t qubit=0; qubit<num_qubits_-1; qubit++) {
+    if (pos >=0)
+      bit = base_value[pos]=='0' ? 0 : 1;
+    else
+      bit = 0;
+    for (uint_t row=0; row<temp.GetRows(); row++)
+      for (uint_t col=0; col<temp.GetColumns(); col++){
+	temp(row, col) *= lambda_reg_[qubit][col];
```
yaelbh, line 1253: identation

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1221,6 +1221,44 @@ void MPS::full_state_vector_internal(cvector_t& statevector,
...
+      }
+    temp = temp * q_reg_[qubit+1].get_data(bit);
+    pos--;
+  }
+
+  return temp(0, 0); //-> check that matrix is 1x1 to complex_t
```
yaelbh, line 1259: I don't understand the comment

test/terra/reference/ref_snapshot_amplitudes.py
```diff
@@ -0,0 +1,110 @@
...
+def snapshot_amplitudes_labels_params():
+    """Dictionary of labels and params for 3-qubit amplitude snapshots"""
+    return {
+        "[0]": [0],
+        "[7]": [7],
+        "[0,1]": [0,1],
+        "[7,3,5,1]": [7,3,5,1],
+        "[4,1,5]": [4,1,5],
+        "[6,2]": [6,2],
+        "all": [0,1,2,3,4,5,6,7]
```
yaelbh, line 32: Do you test hexa?

yaelbh (changes requested):

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I agree, but under omp parallel I must use an "int" as the iterator of the loop. Compilation fails when I try your suggestion.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Added '{', '}', because in my editor the indentation looks okay. Maybe the braces will help.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Removed the comment (was a reminder for myself).

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: How do you suggest to avoid the swaps here?

test/terra/reference/ref_snapshot_amplitudes.py
merav-aharoni: Right, added a couple of tests for hexa.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
yaelbh: If the internal order is: qubit 1 in position 0, qubit 2 in position 1, qubit 0 in position 2. And the user asks for the amplitude of base state 011 (1 for qubit 0, 1 for qubit 1, 0 for qubit 2). Then transform the base state to another base state, according to the internal ordering: 101. Now contract without swaps.

merav-aharoni: Done. Thanks for this improvement!

yaelbh (dismissed):

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -594,6 +598,22 @@ void State::snapshot_state(const Operations::Op &op,
...
+  if (op.params_amplitudes.empty()) {
+    throw std::invalid_argument("Invalid amplitudes snapshot (No base value given).");
+  }
+  reg_t base_values;
+  for (const auto &param : op.params_amplitudes) {
+    base_values.push_back(param);
+  }
+  uint_t num_amplitudes = base_values.size();
+  cvector_t amplitude_vector(num_amplitudes);
+  qreg_.get_amplitude_vector(base_values, amplitude_vector);
```
vvilpas, line 613: A small comment here. I think it is better to move the creation of `amplitude_vector` inside the function `get_amplitude_vector` and return it. Like this:
```
auto amplitude_vector = qreg_.get_amplitude_vector(base_values);
```
This way the function `get_amplitude_vector` is fully in charge of creating that vector and someone using the function doesn't need to care about creating the vector, etc.

vvilpas (changes requested):
Looks good!
I left just a small comment.

src/simulators/matrix_product_state/matrix_product_state.hpp
merav-aharoni: I agree. Done.

vvilpas (approved):

PR #1038, opened 2020-11-13
Add explanation for static wheel compilation

### LINKED ISSUE #1033: Unable to compile a statically linked wheel
Please apologize the length of this bug report in advance, but I am trying to provide as much information as possible.

### Informations

- **Qiskit Aer version**: `master`
- **Python version**: `3.6.9`
- **Operating system**: `uname -a` provides:
```
Linux qt00 4.15.0-112-generic #113-Ubuntu SMP Thu Jul 9 23:41:39 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

To provide some context: I am trying to compile Aer from source on a compute server to allow multiple users to access the same master-branch installation from a centrally managed virtualenv. While doing so I ran into the problem of not being able to produce a correctly statically linked wheel.
Specifically this means that I log in onto a login node, create a virtualenv there and expect all computations on the respective compute nodes to work normally.

I am using the following Python script for testing:
```python
import sys
print(sys.version)
from qiskit.providers.aer import Aer
print(Aer)
```
and the following bash script for testing the execution on the compute nodes:
```bash
#!/bin/bash
locate libopenblas.so.0
source ~/path/to/venv/bin/activate
which python
which pip
pip list | grep qiskit
python test.py  # this is the above script
```

In order to explain the problem at hand I will step through a number of different scenarios and explain the results which I have achieved with each:

#### A *stable* installation using `pip install qiskit-terra qiskit-aer`
Starting from a clean virtualenv and running the above pip installation command everything works fine.
On the login node I obtain:
```
3.6.9 (default, Oct  8 2020, 12:12:24)
[GCC 8.4.0]
/home/a/oss/software/qiskit-stable/venv/lib/python3.6/site-packages/qiskit/__init__.py:75: RuntimeWarning: Could not import the IBMQ provider from the qiskit-ibmq-provider package. Install qiskit-ibmq-provider or check your installation.
  RuntimeWarning)
AerProvider
```
And on the compute nodes:
```
/path/to/stable/venv/bin/python
/path/to/stable/venv/bin/pip
qiskit-aer         0.7.0
qiskit-terra       0.16.0
/home/a/oss/software/qiskit-stable/venv/lib/python3.6/site-packages/qiskit/__init__.py:75: RuntimeWarning: Could not import the IBMQ provider from the qiskit-ibmq-provider package. Install qiskit-ibmq-provider or check your installation.
  RuntimeWarning)
3.6.9 (default, Oct  8 2020, 12:12:24)
[GCC 8.4.0]
AerProvider
```

#### A *master* installation using `python setup.py bdist_wheel`
When trying to compile Aer manually from `master` everything works as expected on the login node (i.e. same as above) but on the compute node I obtain the following error:
```python
  File "test.py", line 3, in <module>
    from qiskit.providers.aer import Aer
  File "/home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/__init__.py", line 64, in <module>
    from .aerprovider import AerProvider
  File "/home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/aerprovider.py", line 20, in <module>
    from .backends.qasm_simulator import QasmSimulator
  File "/home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/backends/__init__.py", line 17, in <module>
    from .qasm_simulator import QasmSimulator
  File "/home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/backends/qasm_simulator.py", line 25, in <module>
    from .controller_wrappers import qasm_controller_execute
ImportError: libopenblas.so.0: cannot open shared object file: No such file or directory
```
This was to be expected as the first line of the bash script (`locate libopenblas.so.0`) does not return anything which shows that OpenBLAS cannot be found on the compute nodes (while it is available normally on the login node).
Having talked to the system administrator, we agreed that it would be best if, thus, we can compile Aer with static linking (as this appears to work with the `stable` installation using `pip install` which ships a Python wheel bundled with `libopenblas`).

#### Static linking: `python setup.py bdist_wheel -- -DSTATIC_LINKING=True`
This attempt causes even the installation on the login node to break with the following error:
```python
  File "test.py", line 3, in <module>
    from qiskit.providers.aer import Aer
  File "/home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/__init__.py", line 64, in <module>
    from .aerprovider import AerProvider
  File "/home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/aerprovider.py", line 20, in <module>
    from .backends.qasm_simulator import QasmSimulator
  File "/home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/backends/__init__.py", line 17, in <module>
    from .qasm_simulator import QasmSimulator
  File "/home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/backends/qasm_simulator.py", line 25, in <module>
    from .controller_wrappers import qasm_controller_execute
ImportError: /home/a/oss/software/qiskit-master/venv/lib/python3.6/site-packages/qiskit/providers/aer/backends/controller_wrappers.cpython-36m-x86_64-linux-gnu.so: undefined symbol: _gfortran_concat_string
```

#### Static linking with `FindBLAS.cmake.fix-static-linking`
I then found out about this patch: https://github.com/Qiskit/qiskit-aer/blob/master/cmake/FindBLAS.cmake.fix-static-linking
But after applying that and trying the same `-DSTATIC_LINKING=True` compilation as above, I still obtain the same error.

The following summarizes the steps necessary to reproduce this issue. However, your mileage may vary as I have tried to reproduce this on my Fedora laptop where I am unable to do so. Both of the last two steps do not result in the problem but I am not sure how to verify that this is because the static linking was successful or whether it falls back to dynamic linking.
The Fedora laptop specs are:
```bash
$ uname -a
Linux pulsano 5.8.15-201.fc32.x86_64 #1 SMP Thu Oct 15 15:56:44 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
$ python --version
Python 3.6.12  # the same works with Python 3.8.6
$ gcc --version
gcc (GCC) 9.3.0
$ cmake --version
cmake version 3.18.2  # same as on the server
```

```bash
mkdir tmp
cd tmp
python3 -m virtualenv venv
source venv/bin/activate
git clone https://github.com/Qiskit/qiskit-terra.git
cd qiskit-terra
pip install -c constraints.txt -r requirements.txt -r requirements-dev.txt
pip install -e .
cd ..
git clone https://github.com/Qiskit/qiskit-aer.git
cd qiskit-aer
pip install -c constraints.txt -r requirements-dev.txt
python setup.py bdist_wheel -- -DSTATIC_LINKING=True
cd ..
python
% from qiskit.providers.aer import Aer
```

I would expect to be able to compile a statically linked Python wheel of Aer

### Suggested solutions

I am at a loss to be honest, which is why I came here :slightly_smiling_face:

mrossinek:
### Summary

As discussed in #1033 we should document how to compile a static wheel
of Qiskit Aer. This commit adds such documentation to `CONTRIBUTING.md`.

Closes #1033

### Details and comments

I have attempted to make the explanation easy to follow but also easily
searchable for potential pitfalls (thus, the inclusion of possible error
messages). Hopefully, this will help with the visibility of this section
and guide users through the process more easily.


mrossinek:
I don't really understand why the CI keeps failing here. I only changed some documentation and that really should not affect the CI.

vvilpas:
@mrossinek Don't worry, it's not your fault. There is a problem with CI (a PR is on the way #1081 )

CONTRIBUTING.md
```diff
@@ -630,6 +630,71 @@ Few notes on GPU builds:
...
+
+Should you encounter a runtime error like
+```
+    Inconsistency detected by ld.so: dl-version.c: 205: _dl_check_map_versions: Assertion `needed != NULL' failed!
+```
+this means that your [patchelf](https://github.com/NixOS/patchelf) version (which is used by
+`auditwheel` under the hood) is too old (https://github.com/pypa/auditwheel/issues/103)
+Version `0.9` of `patchelf` is the earliest to include the patch
+https://github.com/NixOS/patchelf/pull/85 which resolves this issue.
+If you are on a Ubuntu system it may happen that the `apt`-image of `patchelf` only ships version
```
mtreinish, line 684: Which version of ubuntu is this on. Looking at https://packages.ubuntu.com/xenial/patchelf shows 0.9 for xenial which is from 2014.

mtreinish (commented):
Sorry for the delay in review this LGTM. It has clear instructions on how to convert a dynamic binary in a wheel into something static and distributable in different environments. Just one inline question/comment other than that I think is ready to go.

CONTRIBUTING.md
mrossinek: You're right! That's kind of odd because this was on a Ubuntu 18.04 (i.e. bionic). I am not sure why that didn't work when I tried it. Should I update this slightly or should be keep it as is nonetheless?

mtreinish: I would probably remove this sentence, especially because looking more closely at the package listing it depends on the CPU architecture (arm64 and power have 0.8, but the xenial-updates adds 0.9 for all arches and adds s390x). Also newer than xenial everything seems to be >=0.9

mrossinek: I updated it to be more general and not state the specifics about Ubuntu. I hope you agree with the changes.

mtreinish (approved):

PR #1039, opened 2020-11-13
Fix tests due to bad input to assemble function

vvilpas:
### Summary

Fix tests due to bad input to assemble function


mtreinish (approved):
LGTM, it looks like this was caused by Qiskit/qiskit-terra#5298I'll make a note to make sure that release note is highlighting a potential upgrade impact besides just the bugfix. This is kind of an edge case, but still good to highlight.

PR #1043, opened 2020-11-17
Remove a redundant function for getting qubits together

yaelbh:
When the MPS simulator has to perform an operation on multiple qubits, it first swaps qubits, to make the relevant qubits adjacent. The code contains two different functions to stick the qubits together. This pull request removes the older function from the code, and directs the calls to this function to the other function.


merav-aharoni:
The function you wrote:
`static uint reverse_number(uint num, uint nqubits) {`
seems to be identical to the function
`uint_t reverse_bits(uint_t num, uint_t len) {`
Can you check if indeed they have the same functionality?
If they do, then it could help to move your comment to the existing function.

merav-aharoni:
In the lines:
image
you reverse the order of the lines/columns in `mat`.
Could the same be done by reversing the order of the qubits and then applying the matrix as is? 
This would be more consistent with other methods, and also better performance-wise.

merav-aharoni:
The code looks fine to me now.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -624,28 +631,36 @@ void MPS::apply_matrix_internal(const reg_t & qubits, const cmatrix_t &mat,
...
+  uint nqubits = qubits.size();
+  uint sidelen = 1 << nqubits;
+  cmatrix_t new_mat(sidelen, sidelen);
+  for (uint i=0; i<sidelen; ++i)
+    for (uint j=0; j<sidelen; ++j) {
+      uint new_i = reverse_bits(i, nqubits);
+      if (i==j)
+	new_mat(new_i, new_i) = mat(i, i);
+      else {
+	uint new_j = reverse_bits(j, nqubits);
```
vvilpas, line 644: I think changing those `uint` to `uint_t` will fix windows compilation errors

yaelbh:
@vvilpas Thank for you advice. All the tests are passing now.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -624,28 +631,36 @@ void MPS::apply_matrix_internal(const reg_t & qubits, const cmatrix_t &mat,
...
+  for (uint_t i=0; i<sidelen; ++i)
+    for (uint_t j=0; j<sidelen; ++j) {
+      uint_t new_i = reverse_bits(i, nqubits);
+      if (i==j)
+	new_mat(new_i, new_i) = mat(i, i);
+      else {
+	uint_t new_j = reverse_bits(j, nqubits);
+	new_mat(new_i, new_j) = mat(i, j);
+      }
+    }
```
vvilpas, line 648: I'd add braces to the outer loop to avoid possible future errors.

```suggestion
  for (uint_t i=0; i<sidelen; ++i) {
    for (uint_t j=0; j<sidelen; ++j) {
      uint_t new_i = reverse_bits(i, nqubits);
      if (i==j)
        new_mat(new_i, new_i) = mat(i, i);
      else {
        uint_t new_j = reverse_bits(j, nqubits);
        new_mat(new_i, new_j) = mat(i, j);
      }
    }
  }
```

vvilpas (changes requested):
Other than a small comment it looks good.

yaelbh:
Thanks @vvilpas, it's a good idea. I've added the braces.

vvilpas (approved):

PR #1046, opened 2020-11-17
[WIP] Create Python Result object directly

vvilpas:
### Summary

Create Python Result object directly instead of python dict and the Result from it


vvilpas:
Closing this as calling `from_dict` is actually cleaner, simpler and safer than creating directly the python `Result` object.

PR #1051, opened 2020-11-20
Run C++ unit tests on CI

vvilpas:
### Summary

Add a step to run C++ unit tests after compiling the standalone version.


mtreinish (approved):
This looks fine to me. I do wonder if down the road we'll want more output processing that just error code, like use something like subunit or xunitxml to generate a more standardized output.

vvilpas:
> This looks fine to me. I do wonder if down the road we'll want more output processing that just error code, like use something like subunit or xunitxml to generate a more standardized output.

Yeah, we should probably add something like that if we really add more and more unit tests.

PR #1055, opened 2020-11-24
Fix bug with nested omp being applied incorrectly

chriseclectic:
### Summary

Fixes bug with nested OpenMP flag getting set to True when it shouldn't have

### Details and comments

* The condition was checking parallel shots or parallel state update, when it should have been parallel shots or parallel experiments.
* Explicitly disable nested openmp when the above condition isn't satisfied

### DIFF
--- a/src/controllers/controller.hpp
+++ b/src/controllers/controller.hpp
@@ -543,18 +543,10 @@ Result Controller::execute(std::vector<Circuit> &circuits,
 
 #ifdef _OPENMP
     if (parallel_shots_ > 1 || parallel_experiments_ > 1) {
-      #ifdef _WIN32
-        omp_set_nested(1);
-      #else:
-        omp_set_max_active_levels(2);
-      #endif
+      omp_set_nested(1);
       result.metadata["omp_nested"] = true;
     } else {
-      #ifdef _WIN32
-        omp_set_nested(0);
-      #else
-        omp_set_max_active_levels(1);
-      #endif
+      omp_set_nested(0);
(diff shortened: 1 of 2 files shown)


hhorii (approved):

src/controllers/controller.hpp
```diff
@@ -645,6 +656,19 @@ void Controller::execute_circuit(Circuit &circ,
       // Vector to store parallel thread output data
       std::vector<ExperimentResult> par_results(parallel_shots_);
       std::vector<std::string> error_msgs(parallel_shots_);
+
+    #ifdef _OPENMP
+    if (!parallel_nested_) {
+      if (parallel_shots_ > 1 && parallel_state_update_ > 1) {
+        // Nested parallel shots + state update
+        omp_set_nested(1);
+        result.metadata["omp_nested"] = true;
```
hhorii, line 665: It may be better to update `parallel_nested_` with `true`. But this is not mandatory.

chriseclectic: I didn't update it here since the variable was just needed to know if the outer loop was already parallelized or not. If it isn't each individual circuit execution has the option to be nested or not nested.

vvilpas (approved):

Zhaoyilunnn:
Hi, when I test qiskit-aer 0.7.0, I run simulations using GPUs (and the memory requirement exceed the memory capacity of GPUs), the computation on CPU is utilizing multiple threads. However, after this revision (since 0.7.3), it will only use 1 thread for the computation on CPU.

PR #1057, opened 2020-11-29
Test gates on all basis states

yaelbh:
### Summary

Currently the gates are tested only on the ground state. This PR extends to checking the gates on all the basis states.

### Details and comments

This is in preparation of removing files like `ref_non_clifford.py`, since all the gates are already contained in `qasm_standard_gates.py`.


chriseclectic (approved):
LGTM

PR #1058, opened 2020-11-29
Added support for partial initialize in MPS

merav-aharoni:
### Summary
Added support for initialize in MPS.

### Details and comments
Initialization of the full statevector existed before. In this PR, I added support for initializing a subset of the qubits and for permutations in the ordering of the qubits.
I added a function `permute_all_qubits` that was needed for this implementation. This function can be used for the implementation of the function `reorder_all_qubits`. I will open a separate PR for this update.


merav-aharoni:
@chriseclectic, @vvilpas, @yaelbh - I'll be happy if you can review this PR.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -173,6 +205,57 @@ uint_t reorder_qubits(const reg_t qubits, uint_t index) {
   return new_index;
 }
 
+template <class T>
+void permute_all_qubits(const std::vector<T>& orig_statevector, 
+			const reg_t &input_qubits,
+			const reg_t &output_qubits,
+			std::vector<T>& new_statevector) {
```
yaelbh, line 212: This function, and other functions (some of them are already existing, like `squeeze_qubits`), are not related to the MPS structure, and as such should be located elsewhere (not necessarily in this PR).

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -173,6 +205,57 @@ uint_t reorder_qubits(const reg_t qubits, uint_t index) {
...
+      if (shift > 0) {
+	new_index += current_val << shift;
+      } else if (shift < 0) {
+	new_index += current_val >> -shift; 
+      } else {
+	new_index += current_val;
+      }      
+    }
+  }
+  return new_index;
```
yaelbh, line 256: Not sure which one is preferable, in terms of code simplicity and performance: this code, or conversion from integer to a binary string -> permute characters of the string -> convert back to integer.

yaelbh, line 256: Consider rewriting `reverse_bits` and `reverse_all_bits`, such that they call new functions like `permute_qubits`, instead of reimplementing them for the special case of reversed bits.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1279,6 +1361,10 @@ uint_t binary_search(const rvector_t &acc_probvector,
 
 double MPS::norm() {
     reg_t qubits(num_qubits_);
+    return norm(qubits);
+}
+
+double MPS::norm(reg_t qubits) {
```
yaelbh, line 1367: Maybe `reg_t&` ? Or `const reg_t&` ?

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -112,11 +112,10 @@ class State : public Base::State<matrixproductstate_t> {
 
   // Initializes an n-qubit state to the all |0> state
   virtual void initialize_qreg(uint_t num_qubits) override;
+  void initialize_qreg(reg_t qubits, const cvector_t &statevector);
```
yaelbh, line 115: Maybe `reg_t&` ?  Or `const reg_t&` ?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1327,7 +1413,7 @@ reg_t MPS::sample_measure_using_probabilities_internal(const rvector_t &rnds,
   #pragma omp parallel if (SHOTS > omp_threshold_ && omp_threads_ > 1) num_threads(omp_threads_)
     {
       #pragma omp for
-  for (int_t i = 0; i < SHOTS; ++i) {
+      for (int_t i = 0; i < static_cast<int_t>(SHOTS); ++i) {
```
yaelbh, line 1416: Here in GitHub the indentation looks wrong

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1479,6 +1596,51 @@ void MPS::initialize_from_matrix(uint_t num_qubits, const cmatrix_t mat) {
...
+//--------------------------------------------------
+// Algorithm for initialize_component:
+// 1. Centralize 'qubits'
+// 2. Compute the norm of 'qubits'
+// 3. Normalize the values in 'statevector' to the norm computed in (3)
+// 4. Create a new MPS consisting of 'qubits'
+// 5. Initialize it to 'statevector'
+// 6. Cut out the old section of 'qubits' in the original MPS
+// 7. Stick the new section of 'qubits' in the original MPS
+//---------------------------------------------------
```
yaelbh, line 1608: This is not what partial initialization is supposed to do. I'd expect this to fail in tests where the initialized qubits are entangled to other qubits. Correct partial initialization is a random instruction, where the initialized qubits are measured before the initialization. This is why the random generator is passed.

yaelbh (changes requested):

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -112,11 +112,10 @@ class State : public Base::State<matrixproductstate_t> {
...
 
   // Initializes to a specific n-qubit state given as a complex std::vector
-  virtual void initialize_qreg(uint_t num_qubits, const matrixproductstate_t &state) override;
-
-  void initialize_qreg(uint_t num_qubits, const cvector_t &statevector);
+  virtual void initialize_qreg(uint_t num_qubits, const matrixproductstate_t &state);
```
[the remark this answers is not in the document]
merav-aharoni: It is intentional, because this method does not override any method of the base class `State`.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -65,19 +65,51 @@ static const cmatrix_t one_measure =
...
+// Function name: permute_all_qubits
+// Description: Given an input ordering of the qubits, an output ordering of the 
+//    qubits and a statevector, create the statevector that represents the same 
+//    state in the output ordering of the qubits.
+// Input: orig_statevector - the ordered vector of probabilities/amplitudes
+//        input_qubits - a list containing the original ordering of the qubits
+//        output_qubits - a list containing the new ordering
+// Returns: new_statevector - the vector in the new ordering
+// Note that the qubits as numbered from left to right, i.e., the msb is 0
+//
```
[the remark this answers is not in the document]
merav-aharoni: indeed

src/simulators/matrix_product_state/matrix_product_state.hpp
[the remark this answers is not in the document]
merav-aharoni: ok, done.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: added '&'

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: fixed the indentation

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I wrote the comment above about rewriting `reorder_all_qubits` as a function of `permute_all_qubits`. Same goes for `reverse_all_bits`, although here I may decide to leave it, for simplicity reasons. Anyway, I will open a separate PR for this.

merav-aharoni: In terms of performance I expect this version to be better. I think  converting to a string would be simpler and easier to understand. But I chose performance over simplicity.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I agree, but since I developed it for this PR, it is here.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I changed the functionality, as you pointed out. This required quite a few changes in reset and related functions. I added tests to `reset` and to `initialize` as I noticed two types of tests were missing: tests where the qubits are entangled before the `reset` or `initialize` and tests where the qubits may have moved in mps simulation.

merav-aharoni: I think the changes require another review - @yaelbh , @vvilpas , @chriseclectic .

test/terra/backends/qasm_simulator/qasm_reset.py
```diff
@@ -75,3 +75,15 @@ def test_repeated_resets(self):
...
+        """Test QasmSimulator reset with for circuits where qubits have moved"""
+        # count output circuits
+        shots = 1000
+        circuits = ref_reset.reset_circuits_with_entangled_and_moving_qubits(final_measure=True)
+        targets = ref_reset.reset_counts_with_entangled_and_moving_qubits(shots)
+        qobj = assemble(circuits, self.SIMULATOR, shots=shots)
+        result = self.SIMULATOR.run(
+            qobj, **self.BACKEND_OPTS).result()
+        self.assertSuccess(result)
+        self.compare_counts(result, circuits, targets, delta=0.05*shots)
```
yaelbh, line 89: I'd expect to see two more tests: `initialize` with moving qubits, and `reset` with entangled qubits. I see in the references files that the test of reset has both, so maybe change its name to something like `test_reset_entangled_and_moving_qubits`.

yaelbh (dismissed):

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1277,11 +1357,18 @@ uint_t binary_search(const rvector_t &acc_probvector,
...
+
+double MPS::norm(const reg_t &qubits) const {
+  reg_t temp_qubits = qubits;
+    std::iota( std::begin(temp_qubits), std::end(temp_qubits), 0);
     double trace = 0;
-    rvector_t vec = diagonal_of_density_matrix(qubits);
+    MPS temp_MPS;
+    temp_MPS.initialize(*this);
+    rvector_t vec = temp_MPS.diagonal_of_density_matrix(temp_qubits);
     for (uint_t i=0; i<vec.size(); i++)
```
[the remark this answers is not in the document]
merav-aharoni: I changed `norm()` to be const. To do this, I couldn't allow norm to move the ordering of the qubits, which it did in the previous version. Therefore I do the operation on a copy of the MPS and not on the main MPS structure. This is costly in performance, but having `norm()` change the qubit ordering is bug prone (I fell on this).

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1345,35 +1431,28 @@ reg_t MPS::apply_measure(const reg_t &qubits,
...
+  return outcome_vector;
 }
 
-uint_t MPS::apply_measure(uint_t qubit, 
-			 RngEngine &rng) {
+uint_t MPS::apply_measure_internal_single_qubit(uint_t qubit, RngEngine &rng) {	
   reg_t qubits_to_update;
   qubits_to_update.push_back(qubit);
 
   // step 1 - measure qubit 0 in Z basis
```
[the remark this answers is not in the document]
merav-aharoni: I didn't understand the question. This method measures a single qubit whose index is `qubit`.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1479,7 +1591,65 @@ void MPS::initialize_from_matrix(uint_t num_qubits, const cmatrix_t mat) {
...
 }
  
+//--------------------------------------------------
+// Algorithm for initialize_component:
+// 1. Centralize 'qubits'
+// 2. Compute the norm of 'qubits'
+// 3. Normalize the values in 'statevector' to the norm computed in (3)
+// 4. Create a new MPS consisting of 'qubits'
+// 5. Initialize it to 'statevector'
+// 6. Reset 'qubits' - note that this stage may affect qubits that are not in 'qubits',
```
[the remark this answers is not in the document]
merav-aharoni: ok

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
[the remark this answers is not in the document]
yaelbh: I think the comment should say "measure the qubit in Z basis" instead of "measure qubit 0 in Z basis", because "qubit 0" can be another qubit, which we don't measure.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1479,7 +1591,65 @@ void MPS::initialize_from_matrix(uint_t num_qubits, const cmatrix_t mat) {
...
+  double qubits_norm = norm(new_qubits);
+
+  cmatrix_t mat(1, num_amplitudes);
+  for (uint_t i=0; i<num_amplitudes; i++) {
+    complex_t normalized_i = statevector[i]/qubits_norm;
+    mat(0, i) = normalized_i;   
+  }
+  reset_internal(new_qubits, rng);
+  MPS qubits_mps;
+  qubits_mps.initialize_from_matrix(num_qubits, mat);
```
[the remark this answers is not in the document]
merav-aharoni: Added documentation.
mat was already const&. Maybe you meant some other place that I missed?

releasenotes/notes/initialize-in-mps-e9f8b70c798e7e25.yaml
```diff
@@ -0,0 +1,6 @@
+---
+
+features:
+  - |
+    Added support for `initialize` in the MPS simulation_method.
```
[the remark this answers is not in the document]
merav-aharoni: ok

test/terra/backends/qasm_simulator/qasm_reset.py
merav-aharoni: You are right that the same tests test both features. I was hesitant about the word `moving` because that aspect is relevant only for mps, and people who run other simulation methods won't understand the name. 
What do you think?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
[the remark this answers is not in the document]
yaelbh: This is the signature that I see in `matrix_product_state_internal.hpp`:
```
void initialize_from_matrix(uint_t num_qubits, cmatrix_t mat);
```
I see the `const` but not the `&`.

test/terra/backends/qasm_simulator/qasm_reset.py
yaelbh: Maybe remove the word `moving` from the test name and functions in the reference file; and add a comment somewhere, that we also test moving, which is relevant only for MPS.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1479,7 +1591,65 @@ void MPS::initialize_from_matrix(uint_t num_qubits, const cmatrix_t mat) {
...
+  reg_t outcome_vector =  apply_measure_internal(qubits, rng);
+  // Apply update to reset state
+  measure_reset_update_internal(qubits, 0, outcome_vector);
+}
+
+void MPS::measure_reset_update_internal(const reg_t &qubits,
+					const uint_t final_state,
+					const reg_t &meas_state) {
+  for (uint_t i=0; i<qubits.size(); i++) {
+    if(meas_state[i] != final_state) {
```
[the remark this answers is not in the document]
merav-aharoni: I agree. I use this method only for reset, so `final_state` is always 0. Will change to 0.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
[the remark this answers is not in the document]
merav-aharoni: ah, now I see the line you are referring to. Missed that. You are right, of course.

yaelbh (approved):

Issue #1032, opened 2020-11-11
Statevector simulator result of PauliGate is incorrect


ikkoham:
### Informations

- **Qiskit Aer version**:0.8.0.dev0+0e5f01c
- **Python version**:3.8.5
- **Operating system**:macOS

### What is the current behavior?

When I use `PauliGate` instruction in Qiskit Aer, TypeError occured. 

### Steps to reproduce the problem

```Python
from qiskit import QuantumCircuit, execute, Aer, BasicAer
qc = QuantumCircuit(3)
qc.pauli('XYZ', [0,1,2])

print(execute(qc, Aer.get_backend('statevector_simulator')).result().get_statevector())
```
TypeError
```
TypeError: statevector_controller_execute(): incompatible function arguments. The following argument types are supported:
    1. (arg0: object) -> object
```

### What is the expected behavior?

When I use BasicAer,
```python
print(execute(qc, BasicAer.get_backend('statevector_simulator')).result().get_statevector())
```
outputs
```
[0.000000e+00+0.j 0.000000e+00+0.j 0.000000e+00+0.j 0.000000e+00+0.j
 0.000000e+00+0.j 0.000000e+00+0.j 6.123234e-17+1.j 0.000000e+00+0.j]
```

hhorii:
I have tested this with the master of Aer (1b6accdb7fbf32dfd668ddb0a753706c63ba7d27) and Terra (9e6a016dbd44fdeedf27241b9e02fbd8d4a617c8),

```
from qiskit import QuantumCircuit, execute, Aer, BasicAer
qc = QuantumCircuit(3)
qc.pauli('XYZ', [0,1,2])
print(execute(qc, Aer.get_backend('statevector_simulator')).result().get_statevector())
```

and then got:
```
[-0.+0.j  0.+0.j  0.+0.j -0.+0.j -0.+0.j  0.+0.j  0.-1.j -0.+0.j]
```

vvilpas:
Closing as https://github.com/Qiskit/qiskit-terra/pull/1059 has been merged into master.

PR #1059, opened 2020-11-30
fix apply_pauli

hhorii:
### Summary

Resolve Issue #1032 

### Details and comments

swap is performed before applying z.


chriseclectic:
Could we add some extra tests that would catch this bug?

chriseclectic (approved):

PR #1061, opened 2020-12-01
Remove deprecated usage of omp_set_nested

chriseclectic:
### Summary

On mac and linux use `omp_set_max_active_levels` instead of `omp_set_nested`

### Details and comments

OpenMP has deprecated `omp_set_nested` for `omp_set_max_active_levels` which was introduced in OMP 3.0. This means it isn't available on Windows (OMP 2.0) so we still need to use the old function call there.


vvilpas (approved):
LGTM

PR #1065, opened 2020-12-02
Compile CUDA on win

### LINKED ISSUE #787: Fix Windows (MSVC) builds for GPU simulators

vvilpas:
### Summary

Allow cuda compilation on windows. We need to update pybind11 to 2.6.

### Details and comments

Fixes #787


releasenotes/notes/allow_cuda_on_win-c3e5c920528eb282.yaml
```diff
@@ -0,0 +1,5 @@
+---
+upgrade:
+  - |
+    It is now possible to build AER with CUDA enabled in windows. To support this,
+    ``pybind11`` minimum version required is 2.6 instead of 2.4.
```
mtreinish, line 5: I would split this up into 2 notes. One under a `features` (or maybe fixes) section advertising the support for building with cuda on windows. The upgrade note should be about just bumping the minimum (and use the windows building support as the reason for the change).

chriseclectic (approved):

PR #1072, opened 2020-12-08
Small cleaning in the MPS simulator

yaelbh:
Especially removing unused variables


merav-aharoni (commented):
Looks good. One small comment - I think the final function should be named `centralize_qubits` and not `centralize_and_sort_qubits`, because the qubits are not actually sorted.

yaelbh:
Thanks @merav-aharoni, done.

chriseclectic (approved):

PR #1074, opened 2020-12-08
Restrict cvxpy testing to python 3.7 and 3.8

mtreinish:
### Summary

The cvxpy packaging (more specifically scs) is causing issues in CI for
us because since it is incorrectly trying to install numpy from a
pre-release package which does not support python 3.6. The nature of
this issue is difficult to pin because it's not using pip to install
numpy. Since cvxpy is only used in a single place, noise
transformations, instead of trying to workaround these scs packaging
issues this commit just restricts the test requirements on cvxpy to only
match pytho 3.7 or 3.8. If cvxpy is not installed the noise
transformation tests will now skip instead of fail.

### Details and comments

Co-authored-by: Daniel Puzzuoli <dan.puzzuoli@gmail.com>

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -80,7 +80,7 @@ jobs:
         run: |
           set -e
           cd qiskit-tutorials
-          rm -rf tutorials/chemistry tutorials/circuits tutorials/circuits_advanced tutorials/finance tutorials/optimization tutorials/algorithms tutorials/operators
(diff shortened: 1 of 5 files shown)


mtreinish:
This is working except for some pulse simulator test failures that look like they were probably caused by Qiskit/qiskit-terra#5481

chriseclectic (approved):

PR #1075, opened 2020-12-08
fixing Delay instruction bug in pulse simulator

DanPuzzuoli:
Delay instructions are no longer present in the pulse library of a qobj, resulting in a key error in the pulse digest when looking up properties of instructions in the internal pulse-simulator representation of the pulse library. The bug was fixed by modifying the code that loops thourhg instructions in an experiment to simply do nothing when a delay instruction is found.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -80,7 +80,7 @@ jobs:
         run: |
           set -e
           cd qiskit-tutorials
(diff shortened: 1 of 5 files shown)


vvilpas:
I'm closing this as it looks like it was already included in https://github.com/Qiskit/qiskit-aer/pull/1074.

PR #1081, opened 2020-12-14
Fix CI problem when updating wheels

vvilpas:
### Summary

Updating wheel with pip fails in azure sdist_builds. Installing it with Conda works.


mtreinish (approved):
LGTM as it fixes the job, but maybe we should remove the pip step below if we're conda installing it before now.

mtreinish (approved):

PR #1083, opened 2020-12-15
Increase delta in a non deterministic test to avoid CI failing.

vvilpas:
### Summary

The test `test_sparse_output_probabilities` is failing in some macos runs. As it has a non zero `delta` I assume is non deterministic, so I incremented delta.


chriseclectic (approved):
Might also be worth checking rng handling of the extended stabilizer simulator to see why this non-deterministically failing with a fixed simulator seed.

Issue #874, opened 2020-08-10
Farm embarrassingly-parallel simulation tasks to a cluster


levbishop:
Many large simulation jobs have embarrassing parallelism (over shots and/or circuit instances) and could pick up a big speedup on an HPC cluster. We need some way to:

1. Break the task up into managable chunks (not too many that the job setup/teardown overhead becomes dominating)
2. Deal with cluster configuration issues: queuing system LSF vs SLURM vs PBS etc. Shared NSF home directory vs other work-distribution methods, etc. Authentication, resource allocation and quotas.
3. Wait for the jobs to complete while providing progress feedback and depending on the reliability of the underlying queuing system maybe providing a reliability layer
4. Reassemble the partial results as though it were executed as one task
5. Handle RNG seeding to ensure the same level of reproducability as the underlying simulation method.

It seems to me that most self-contained way to implement this is as a new aer provider.

For point 2, I think that [dask-distributed](https://docs.dask.org/en/latest/setup/single-distributed.html) is potentially a great option, since it contains [dask-jobqueue](https://jobqueue.dask.org/) who wraps up all the big cluster queueing systems: LSF, PBS, Slurm, HTCondor, Moab, SGE, OAR giving jupyter widgets to monitor and with some well-thought-out example configurations and workflows. In addition there are components that provide the same interface via mpirun, an informal cluster of ssh-able machines, DRMAA-compliant resource managers, kubernetes-based cloud clusters, YARN/Hadoop cloud providers etc. You get a nice API that extends `concurrent.futures`. It can implement the same API locally (ie without a cluster) for testing, CI, etc. They've put some effort into serialize/deserialize that goes beyond pickle. It can do some forms of load scaling and such. And it's part of the [dask project](https://docs.dask.org/en/latest/) which is large, modern and very active. I guess we point users to the dask documentation for them to set up a `dask.Cluster()` instance which they can pass to our new provider.

chriseclectic:
Dask looks like a really nice option for doing this

woodsp-ibm:
Has there been any progress on this with regards to having such a capability in Aer?

levbishop:
@woodsp-ibm see #1084 which is intended for next release

chriseclectic:
Dask executor support added in 0.9.0

PR #946, opened 2020-09-21
Clustermanager

ereastman:
### Summary

Adds AerClusterManager - a thin wrapper around a dask.distributed Client which allows users to use a local or remote cluster for submitting and managing jobs. 

### Details and comments

Example usage (see tests):

```
cman = AerClusterManager()
backend = Aer.get_backend('qasm_simulator')
circs = transpile([random_circuit(num_qubits=3, depth=4, measure=True) for _ in range(5)], backend)
backend_options = None
noise_model = None
jobset = cman.set_assemble_config(shots=shots).run(circs, backend)
result = jobset.results(raises=True).combine_results()
```

PR #1084, opened 2020-12-16
Cluster backend

ereastman:
### Summary

Replaces #946.

Adds a new backend ```ClusterBackend``` which can wrap other backends and allows the user to provide their own executor.
With a properly configured executor the ClusterBackend can be used to run large* jobs faster by making use of distributed compute such as those available on most HPC clusters. 
To the user ClusterBackend behaves exactly like the wrapped backend - inheriting its properties, configuration, defaults, etc.
When a QasmQobj is run, it is split up into one qobj per experiment and each experiment is submitted to the executor independently.
The futures for each job are collected and returned within a JobSet object, which in addition to providing per-future access, also provides the same result() blocking method as an AerJob (JobSets can be treated like Jobs).

Example usage:

```
qbackend = Aer.get_backend('qasm_simulator')
with ThreadPoolExecutor() as exc:
    backend = ClusterBackend(qbackend, exc)
    res = execute(circs, backend).result()
```

Notice, the to-be-wrapped backend (```qbackend```) must be passed as an argument to ClusterBackend's constructor (along with the executor). This makes it untenable to return a ClusterBackend from e.g. Aer.get_backend('cluster_backend').

### Details and comments

Efficiency gains are dependent on many things the two largest being platform and experiment makeup.
Overall efficiency gains are as-expected when parallelizing a previously serial task after overhead.
Splitting up an already formed qobj into its constituent experiments and recombining each job's result into a single result does add some overhead to each ClusterBackend.run call as compared to e.g. QasmSimulator.run(...).
For executors such as Dask there is additional transport overhead when sending each experiment to the worker which will run it.

Here are some tests using the QasmSimulator as control and as wrapped backend.
Platform: Intel(R) Xeon(R) CPU E5-2667 v2 @ 3.30GHz, RHEL 7.6

First, we use the QasmSimulator as-is to run some jobs and get a baseline.
We vary each experiment's size (in number of qubits, each point on a line) and the number of experiments per qobj (each line) and measure the time taken and peak memory.
Only the time taken to complete the ```backend.run(payl, backend_options, noise_model).result()``` call is recorded.
Each data point is the average time over 5 runs.
Memory is measured using memory_profile and is only measured once.
For each data point a new interpreter process is started, the call is run once for 'warm up', then again to measure peak memory, and then 5 more times and the average time is computed over those 5 runs.

mem_vs_qubits_by_circ_ctrl
time_vs_qubits_by_circ_ctrl

** The control job was killed before the 64-experiment, 24-qubit job could be completed so that data point is faked.

Then we create a Dask executor connected to a remote scheduler with 3 workers (each on an approximately equivalent node) and wrap the QasmSimulator in a ClusterBackend.
We only measure the peak memory on the client node.

mem_vs_qubits_by_circ_dask_3w
time_vs_qubits_by_circ_dask_3w

Raw data is contained in the attached txt file.
[all_data.txt](https://github.com/Qiskit/qiskit-aer/files/5704487/all_data.txt)


test/terra/extensions/test_cluster.py
```diff
@@ -0,0 +1,219 @@
...
+    def setUpClass(cls):
+        super(ThreadPoolFixture, cls).setUpClass()
+        cls._test_executor = concurrent.futures.ThreadPoolExecutor(max_workers=1)
+
+class DaskFixture(CBFixture):
+    @classmethod
+    def setUpClass(cls):
+        super(DaskFixture, cls).setUpClass()
+        if DASK_TESTS:
+            cls._test_executor = Client(address=LocalCluster(n_workers=1, processes=True))
```
vvilpas, line 81: Shouldn't we try to tests with more than one worker? Or is there any problem?

qiskit/providers/aer/aerprovider.py
```diff
@@ -21,6 +21,7 @@
 from .backends.statevector_simulator import StatevectorSimulator
 from .backends.unitary_simulator import UnitarySimulator
 from .backends.pulse_simulator import PulseSimulator
+# from .backends.cluster_backend import ClusterBackend
```
vvilpas, line 24: remove

qiskit/providers/aer/aerprovider.py
```diff
@@ -35,6 +36,7 @@ def __init__(self, *args, **kwargs):
             ('statevector_simulator', StatevectorSimulator),
             ('unitary_simulator', UnitarySimulator),
             ('pulse_simulator', PulseSimulator)
+            # ('cluster_backend', ClusterBackend)
```
vvilpas, line 39: remove

qiskit/providers/aer/backends/__init__.py
```diff
@@ -18,3 +18,4 @@
 from .statevector_simulator import StatevectorSimulator
 from .unitary_simulator import UnitarySimulator
 from .pulse_simulator import PulseSimulator
+# from .cluster import ClusterBackend
```
vvilpas, line 21: remove

qiskit/providers/aer/backends/cluster/clusterjob.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def submit(self, executor: Optional[futures.Executor] = None) -> None:
+        """Submit the job and obtain/save future.
+
+        Args:
+            executor: The executor to be used to submit the job.
+        """
+        logger.debug("Submitting job %s in future", self._id)
+        _exec = executor or self._executor
+        self._future = _exec.submit(self._backend._run_job, self._id, self._qobj,
```
vvilpas, line 67: Shouldn't we use '_run' instead of '_run_job' as this last one is deprecated?

vvilpas (commented):
The code looks good, very clean and readable! I've left a few comments.
I think we also discussed to make the interface as close to existing providers as possible. I'm not very familiar with the providers so maybe @chriseclectic can check that side. 
Two other things:
1) The plots you provide look good, but it would be great if you could provide plots with the ratios as well (time_no_cluster/time_cluster vs num_qubits, etc). With those it will be easier to check the performance improvements.
2) I've tried to run the cluster tests with dask enabled and I get the following error: `tornado.application - ERROR - Exception in callback functools.partial(<bound method PeriodicCallback._run of <tornado.ioloop.PeriodicCallback object at 0x115791450>>)`. Have you experienced anything similar?

levbishop:
Fixes #874

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+        self._future = True
+        total_jobs = len(self._experiments)
+        for i, exp in enumerate(self._experiments):
+            cjob = CJob(backend, exp, *run_args, **run_kwargs)
+            cjob.submit(executor=executor)
+            logger.debug("Job %s submitted", i + 1)
+            self._futures.append(cjob)
+
+    @requires_submit
+    def statuses(self) -> List[Union[JobStatus]]:
```
chriseclectic, line 94: There should be a regular `status` method for job that is a single value that combines the subjobs (ie RUNNING if any sub jobs are still running, DONE if all done, etc)

Then you can have a method like this one for returning status of sub-jobs. I don't particularly like the name `statuses`, maybe something like `worker_status` that can take a worker index and return a single status, or if none return the list for all workers

```suggestion
    def worker_status(self, worker: Union[int, Iterable[int]]) -> Union[JobStatus, List[JobStatus]]:
```

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+
+        Raises:
+            AerClusterTimeoutError: if unable to retrieve all job results before the
+                specified timeout.
+
+        """
+        return self.results(timeout, raises).combine_results()
+
+    @requires_submit
+    def results(
```
chriseclectic, line 128: This could be `worker_result` and take a worker index as its argument.
ie `job.worker_result(i)` and each worked should return a regular qiskit `Result` object from this function.

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+        self._future = None
+        self._futures = []
+        self._results = None
+        self._error_msg = None
+
+    def run(self,
+            executor: futures.Executor,
+            backend: AerBackend,
+            *run_args: Any,
+            **run_kwargs: Any) -> None:
```
chriseclectic, line 69:
```suggestion
            **run_kwargs: Any):
```

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+        Returns:
+            A list of job statuses.
+        """
+        return [cjob.status() for cjob in self._futures]
+
+    @requires_submit
+    def result(
+            self,
+            timeout: Optional[float] = None,
+            raises: Optional[bool] = False
```
chriseclectic, line 106: This raises would be a good option to add to the regular `AerJob` too, so simulator exceptions can be raised early rather than giving cryptic error messages on `job.get_counts` etc about missing data, when really `job.status` contains the exception thrown by the c++ code.

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+        Raises:
+            AerClusterTimeoutError: if unable to retrieve all job results before the
+                specified timeout.
+
+        """
+        return self.results(timeout, raises).combine_results()
+
+    @requires_submit
+    def results(
+            self,
```
chriseclectic, line 129:
```suggestion
            self, worker: Union[int, Iterable[int]]
```

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+            timeout: Optional[float] = None,
+            raises: Optional[bool] = False
+    ) -> CResults:
```
chriseclectic, line 132:
```suggestion
    ) -> Union[Result, List[Result]]
```

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+        Returns:
+            A :class:`CResults`
+            instance that can be used to retrieve results
+            for individual experiments.
+
+        Raises:
+            AerClusterTimeoutError: if unable to retrieve all job results before the
+                specified timeout.
+        """
+        if self._results is not None:
```
chriseclectic, line 151: I think this method just needs to call the individual worker `job.result()` something like:

```python
if isinstance(worker, Integral):
    return self._futures[worker].result(timeout=timeout, raises=raises)
else:
    return [self._futures[i].result(timeout=timeout, raises=raises) for i in worker]
```

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+        return self._results
+
+    @requires_submit
+    def cancel(self) -> None:
+        """Cancel all jobs in this job set."""
+        for cjob in self._futures:
+            cjob.cancel()
+
+    @requires_submit
+    def job(self, experiment: Union[str, QuantumCircuit, Schedule]) -> Tuple[CJob, int]:
```
chriseclectic, line 191: Maybe this should be broken up into two methods:
```python
def worker(self, experiment: Union[Integral, str, QuantumCircuit, Schedule]) -> Union[int, List[int]]
```
where you can also pass in the experiment by integer position from the original execution list. The return type could also include a list of integers to allow for spreading a single circuit into multiple workers in the future (eg for shot/noise parallelization).

Then the job method could be
```python
    def worker_job(self, worker: Union[int, Iterable[int]]) -> Union[AerJob, List[AerJob]]:
```

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+                    * Schedule: The name of the schedule instance will be used.
+
+        Returns:
+            A tuple of the job used to submit the experiment and the experiment index.
+
+        Raises:
+            AerClusterJobNotFound: If the job for the experiment could not
+                be found.
+        """
+        if isinstance(experiment, (QuantumCircuit, Schedule)):
```
chriseclectic, line 209: This should also support looking for experiment by integer index

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+        for job in self.jobs():
+            for i, exp in enumerate(job.qobj().experiments):
+                if hasattr(exp.header, 'name') and exp.header.name == experiment:
+                    return job, i
+
+        raise AerClusterJobNotFound(
+            'Unable to find the job for experiment {}.'.format(experiment))
+
+    @requires_submit
+    def jobs(self) -> List[Union[CJob, None]]:
```
chriseclectic, line 220: See above comment about `worker_job` function. If a worked job failed to submit I think it should raise an exception rather than return None.

qiskit/providers/aer/backends/cluster/clusterjobset.py
```diff
@@ -0,0 +1,236 @@
...
+        """Return jobs in this job set.
+
+        Returns:
+            A list of :class:`~qiskit.providers.aer.cluster.CJob`
+            instances that represents the submitted jobs.
+            An entry in the list is ``None`` if the job failed to be submitted.
+        """
+        return self._futures
+
+    def name(self) -> str:
```
chriseclectic, line 230: I don't think this method is necessary, its not part of base Job API

qiskit/providers/aer/backends/cluster/clusterresults.py
```diff
@@ -0,0 +1,235 @@
...
+import numpy  # pylint: disable=unused-import
+from qiskit.result import Result
+from qiskit.circuit import QuantumCircuit
+from qiskit.pulse import Schedule
+
+from qiskit.providers import JobError
+from .exceptions import AerClusterResultDataNotAvailable
+
+
+class CResults:
```
chriseclectic, line 30: We shouldn't need a custom result class.

Handling should be done via the `AerJob` and `AerJobSet` classes, so that each worker is described by an `AerJob` which just returns a regular `Result` object. The AerJobSet class should have functionality for combining all the individual worker job results into a single `Result` object.

qiskit/providers/aer/backends/cluster/exceptions.py
```diff
@@ -0,0 +1,47 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Exception for the Job Manager modules."""
+
+from qiskit import QiskitError
+
+
+class AerClusterError(QiskitError):
```
chriseclectic, line 20: Not sure we need so many additional error classes. Can't we just use `AerError` for all of them?

chriseclectic (changes requested):
Suggestions for improving API and integration with Aer by making use of the existing AerJob and qiskit Result classes rather than adding CJob, and CResult. I think the only new class needed is the `AerJobSet` class, plus modifications to AerBackend to handle optional execution via the job set if an executor is provided, and just regular exeuction with a single AerJob if none is provided.

chriseclectic:
Replaced by #1281

PR #1089, opened 2020-12-28
Fix condtional instructions using `c_if` not working for reg of size>=64 on BasicAer backend

### LINKED ISSUE #1077: c_if does not work for large size classical registers
### Informations

- **Qiskit Aer version**: 0.7.1
- **Qiskit Terra version**: 0.16.1
- **Qiskit version**: 0.23.1
- **Python version**: 3.8.5
- **Operating system**: macOs Big Sur 11.0.1

As far as I understand, a `c_if` should work on a classical register of arbitrary size. However, a test I performed seems to indicate that 
- `c_if` works if the register size is less than 64 bits
- It causes a crash for a register size of exactly 64 bits
- It fails for register size larger than 64 bits

I also noticed that if I run the QASM simulator using BasicAer, `c_if` works as it should in all the three cases mentioned above.

Here is the test code that I ran

```
from qiskit import *

def get_circuit(n):
    creg = ClassicalRegister(n)
    dreg = ClassicalRegister(1)
    
    qreg = QuantumRegister(1)
    mreg = QuantumRegister(1)
    
    circ = QuantumCircuit(qreg, mreg, creg, dreg)
    
    for i in range(n):
        circ.measure(qreg[0], creg[i])
    
    circ.x(mreg[0]).c_if(creg, 0)
    circ.measure(mreg[0], dreg[0])
    
    return circ


sim_basic_aer = BasicAer.get_backend('qasm_simulator')
sim_aer = Aer.get_backend('qasm_simulator')

circ72 = get_circuit(72)
circ48 = get_circuit(48)

print("72 clbits (Aer)       : ", execute(circ72, sim_aer).result().get_counts())
print("72 clbits (Basic Aer) : ", execute(circ72, sim_basic_aer).result().get_counts())
print("")
print("48 clbits (Aer)       : ", execute(circ48, sim_aer).result().get_counts())
print("48 clbits (Basic Aer) : ", execute(circ48, sim_basic_aer).result().get_counts())
```

Here is the result I get after running the above code

```
72 clbits (Aer)       :  {'0 000000000000000000000000000000000000000000000000000000000000000000000000': 1024}
72 clbits (Basic Aer) :  {'1 000000000000000000000000000000000000000000000000000000000000000000000000': 1024}

48 clbits (Aer)       :  {'1 000000000000000000000000000000000000000000000000': 1024}
48 clbits (Basic Aer) :  {'1 000000000000000000000000000000000000000000000000': 1024}
```

When I have a classical register with 72 bits, the leftmost bit is set to 0, while it should be 1. 

### Suggested solutions

Given that the code fails at 64-bits (crashes at n=64, fails for n>64), I would naively think that the code is not handling big integers (bigger than the system word size) well.

b63:
### Summary
Added fix for bug where executing conditional instruction with `c_if` on the `BasicAer` backend  either resulted in a crash or incorrect result for registers larger than or equal to 64-bits. 

Fixes #1077 

### Details and comments

The `hex2bin` routine in `util.hpp` used 8 hex-character instead for 16 for the block size to process in 64-bit chunks. Also added check to avoid invoking `std::stoull` with empty string, which was causing the crashes.

In the implementation of `ClassicalRegister::apply_bfunc,` truncated the leading zeros before comparing the masked register value with the target value for the conditional.

### DIFF
--- a/src/framework/creg.hpp
+++ b/src/framework/creg.hpp
@@ -178,13 +178,20 @@ void ClassicalRegister::apply_bfunc(const Operations::Op &op) {
   } else {
     // We need to use big ints so we implement the bit-mask via the binary string
     // representation rather than using a big integer class
-    std::string mask_bin = Utils::hex2bin(mask); // has 0b prefix while creg_register_ doesn't
-    size_t length = std::min(mask_bin.size() - 2, creg_register_.size()); // -2 to remove 0b
+    std::string mask_bin = Utils::hex2bin(mask, false);
+    size_t length = std::min(mask_bin.size(), creg_register_.size());
     std::string masked_val = std::string(length, '0');
     for (size_t rev_pos = 0; rev_pos < length; rev_pos++) {
       masked_val[length - 1 - rev_pos] = (mask_bin[mask_bin.size() - 1 - rev_pos] 
                                           & creg_register_[creg_register_.size() - 1 - rev_pos]);
     }
+    // remove leading 0's
+    size_t end_i = masked_val.find('1');
+    if (end_i == std::string::npos)
+        masked_val = "0";
+    else
+        masked_val.erase(0, end_i);
+
     masked_val = Utils::bin2hex(masked_val); // convert to hex string
     // Using string comparison to compare to target value
     compared = masked_val.compare(target_val);
--- a/releasenotes/notes/fix-c-if-large-register-79d6649395e57ff5.yaml
+++ b/releasenotes/notes/fix-c-if-large-register-79d6649395e57ff5.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixes a bug that resulted in `c_if` not working when the
+    width of the conditional register was greater than 64. See
+    `#1077 <https://github.com/Qiskit/qiskit-aer/issues/1077>`.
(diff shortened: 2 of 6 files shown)


src/framework/creg.hpp
```diff
@@ -178,13 +178,20 @@ void ClassicalRegister::apply_bfunc(const Operations::Op &op) {
   } else {
     // We need to use big ints so we implement the bit-mask via the binary string
     // representation rather than using a big integer class
-    std::string mask_bin = Utils::hex2bin(mask); // has 0b prefix while creg_register_ doesn't
-    size_t length = std::min(mask_bin.size() - 2, creg_register_.size()); // -2 to remove 0b
+    std::string mask_bin = Utils::hex2bin(mask, false); // has 0b prefix while creg_register_ doesn't
+    size_t length = std::min(mask_bin.size(), creg_register_.size()); // -2 to remove 0b
```
chriseclectic, line 182: I think the // -2 to remove 0b comment is redundant now and can be removed?

chriseclectic (commented):
LGTM, thanks for fixing this bug @b63

src/framework/creg.hpp
b63: Yeah I think so too. I removed it and rebased the commits onto [Qiskit:master](https://github.com/Qiskit/qiskit-aer/commit/33702ebe5c7b3d0cecbac5b7e2ddae7c1dd54dca).

vvilpas (approved):

PR #1095, opened 2021-01-08
Update README.md to mention Linux-only GPU support

amirebrahimi:
### Summary
After installing CUDA on Windows and trying to figure out why pip was complaining:
```
ERROR: Could not find a version that satisfies the requirement qiskit-aer-gpu
ERROR: No matching distribution found for qiskit-aer-gpu
```

I found on [Slack](https://qiskit.slack.com/archives/C88B13092/p1596183091333600?thread_ts=1596151161.324800&cid=C88B13092) that this is only supported on Linux. I figured I'd save others from wasting time by being explicit about what is supported.


chriseclectic:
@amirebrahimi Support for building with CUDA on windows will be added in #1065

mtreinish:
> @amirebrahimi Support for building with CUDA on windows will be added in #1065

This is a bit more nuanced. #1065 adds support for building a package with CUDA on windows, but the precompiled package we ship to pypi on qiskit-aer-gpu is still linux only. We need to come up with a way to install CUDA in the windows CI nodes to start publishing windows wheels for qiskit-aer-gpu. I think it's probably fine to say `pip install qiskit-aer-gpu` is linux only for now. When/if we add windows gpu wheel support and publish that we can remove it.

vvilpas:
We may want to add that for windows you can build from source yourself and refer to the contributing guide.

chriseclectic (approved):

PR #1099, opened 2021-01-14
Migrate to new provider interface

mtreinish:
### Summary

This commit starts the migration to the v2 provider interface. In
qiskit-terra 0.16.0 a new version of the provider interface was added
which is explicitly versioned which enables graceful changes to the
interface over time. For the first version of this new interface the key
changes are the circuits can be passed directly to Backend.run() and to
go along with that a new options object has been added to the backend to
track run time options for a backend.

### Details and comments

Co-authored-by: Christopher J. Wood <cjwood@us.ibm.com>

TODO:

- [x] Finish _default_options class method to cover all potential options (which should fix most of the test failures)
- [x] Add tests for AerBackend.run() from a QuantumCircuit and a Schedule (for the pulse simulator)
- [x] Add release notes


qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -100,21 +104,22 @@ def __init__(self,
...
+                self.set_option(key, val)
 
     # pylint: disable=arguments-differ
+    @deprecate_arguments({'qobj': 'circuits'})
     def run(self,
-            qobj,
+            circuits,
             backend_options=None,  # DEPRECATED
             validate=False,
             **run_options):
```
ajavadia, line 117: are the run_options documented somewhere? I think it should be documented here so the backend.run() user knows what kind of options are valid. Especially now that things like number of shots which used to be in qobj have to be passed here.

qiskit/providers/aer/backends/pulse_simulator.py
```diff
@@ -167,16 +170,18 @@ def __init__(self,
...
 
     # pylint: disable=arguments-differ, missing-param-doc
+    @deprecate_arguments({'qobj': 'schedules'})
     def run(self,
-            qobj,
+            schedules,
             *args,
             backend_options=None,  # DEPRECATED
             validate=True,
             **run_options):
```
ajavadia, line 179: same here for documenting run_options

ajavadia (commented):
thanks for getting this done. i can re-review once it's ready

qiskit/providers/aer/backends/aerbackend.py
merav-aharoni: Here are all the run options for mps:
The following options are unique for matrix_product_state:
`matrix_product_state_truncation_threshold`: double
Discard the smallest coefficients for which the sum of their squares is smaller than this threshold.   (Default: 1e-16).
`matrix_product_state_max_bond_dimension` – uint
Sets a limit on the number of Schmidt coefficients retained at the end of the svd algorithm. Coefficients beyond this limit will be discarded.  (Default: None, i.e., no limit on the bond dimension).
`mps_sample_measure_algorithm` – `mps_probabilities`/ `mps_apply_measure`
Choose which algorithm to use for `sample_measure`. `mps_probabilities` means all state probabilities are computed and measurements are based on them. It is more efficient for a large number of shots, small number of qubits and low entanglement. `mps_apply_measure` creates a copy of the mps structure and makes a measurement on it. It is more efficient for a small number of shots, high number of qubits, and low entanglement. If the user does not specify the algorithm, a heuristic algorithm is used to select between the two algorithms.      (Default: `mps_heuristic`).

The following options are applicable for other simulation methods as well as mps.
`Chop_threshold` – double
Cutoff threshold for precision of computations involving floating point numbers. (Default: 1E-8)
`mps_parallel_threshold` – uint
Threshold for enabling omp. (Default: 14)

`mps_omp_threads` – uint
Number of threads when using omp. (Default: 1)

chriseclectic (approved):

Issue #1062, opened 2020-12-01
Design doc: saving additional simulator data to results


chriseclectic:
# Goal

The goal of this is to refactor simulator output data handling, and in particular snapshot data, to be more user friendly and more flexible.

# Design

## Data snapshot accessor instructions

Changing snapshots to be a new type of `Data` snapshot instruction that can be added to a circuit to record data from the internal state of the simulator, or derived from the state of the simulator.

These instructions should be identified by:

- `datatype` string which identities the type of data being recorded
- `subtype` strings which identities the subtype of the data being recorded
- `key` value to identity the data in the result data

Optionally they may take additional parameters such as qubits, clbits, or other parameters. As a qobj the base Data instruction would look like.

One important point is that these instructions should not attempt to insert handle simulation of noise in recording data. Noise can be added to circuits using its own instructions which modify the simulator state queried by these instructions.

### Data type

The data type refers to the type of data being recorded for example:

- Counts
- Memory
- Probabilities
- Expectation value
- Amplitudes 
- Statevector
- State ket
- Density matrix

### Data subtypes

Each data type may support 1 or more subtypes which determine how the information is recorded. Possibly data subtypes are:

- *oneshot (single) data*: only record a single value of the data (ie the value for the last shot). Example use: final statevector or unitary for the statevector or unitary simulators.
- *pershot (list) data*: record a value of each shot of the simulator. Example use: memory, statevector snapshot, amplitude snapshot, or single-shot version of current average snapshots like probabilities or expectation values.
- *accumulated data*: return a single quantity of the data accumulated over all shots. Example use: counts.
- *average data*: return the accumulated data averaged by the number of shots. Example use: probabilities, expectation values, density matrix. Note that this is averaged over all shots unconditional on any classical register values from intermediate measurements.

Furthermore each data subtype (other than single) could be made conditional on the current classical register values of the simulator (This is the current behavior of average snapshots in the simulator). Rather than store conditional data as a list as is currently done this should simply add an additional layer of nesting to the data.  Eg `key: value` becomes `key: {"0x0": cond_val0, "0x1": cond_val1, ... }`.

This gives 7 total possible data subtypes: `[single, list, accum, ave, c_list, c_accum, c_ave]`. Assuming a single datum for a datatype is an object `T`, the format of the returned subtypes would be:

- *single, accum, ave*: `T`
- *list*: `List[T]`
- *c_accum, c_ave*: `Dict[str, T]`
[...]

PR #1101, opened 2021-01-15
Add new save expectation value instructions

chriseclectic:
### Summary

Adds support for new `SaveExpectationValue` and `SaveExpectationValueVariance` instructions as outlined in #1062. These function can be used with the circuit methods `save_expval`, `save_expval_var` which are automatically monkey patched when importing the Aer provider.

### Details and comments

The `SaveExpectationValuel` instruction saves expval `E[H]` of a Hermitian operator as a float, while the `SaveExpectationValueVariance` saves an array `[E[H], V[H]]` of the expval and variance `V[H] = (E[H], E[H^2] - E[H]^2)`. The operator can be input as any `BaseOperator` subclass from the quantum_info module, though it will be converted to a `SparsePauliOp` internally for the instruction.

For each of these types the result can be returned as one of the following subtypes:

* saving as a list of values for each simulated shot (`pershot=True`)
* saving without normalizing the average by number of shots (`unnormalized=True`)
* saving any of the above (average, pershot, unnormalized) conditional on any classical register values from preceding measurements (`conditional=True`)

#### Basic example

```python
import qiskit

# Generate random Hermitian operator
op = qi.random_hermitian(2 ** 3, seed=10)

# Initial state circuit
qc = qiskit.QuantumCircuit(3)
qc.h(0)
qc.cx(0, 1)
qc.cx(1, 2)

# Save expectation value
qc.save_expectation_value('expval', op, [0, 1, 2])
qc.save_expectation_value_variance('expval_var', op, [0, 1, 2])

# Run
sim = QasmSimulator()
result = qiskit.execute(qc, sim).result()

print(result.data(0))
```
```python
{'expval': -1.0219294356952844, 'expval_var': array([-1.02192944,  1.44909355])}
```

### DIFF
--- a/qiskit/providers/aer/library/save_expectation_value.py
+++ b/qiskit/providers/aer/library/save_expectation_value.py
@@ -60,8 +60,6 @@ class SaveExpectationValue(SaveAverageData):
             operator = SparsePauliOp(operator)
         elif not isinstance(operator, SparsePauliOp):
             operator = SparsePauliOp.from_operator(Operator(operator))
-        if not isinstance(operator, SparsePauliOp):
-            raise ExtensionError("Invalid input operator")
         if not allclose(operator.coeffs.imag, 0):
             raise ExtensionError("Input operator is not Hermitian.")
         params = _expval_params(operator, variance=False)
@@ -114,8 +112,6 @@ class SaveExpectationValueVariance(SaveAverageData):
             operator = SparsePauliOp(operator)
         elif not isinstance(operator, SparsePauliOp):
             operator = SparsePauliOp.from_operator(Operator(operator))
-        if not isinstance(operator, SparsePauliOp):
-            raise ExtensionError("Invalid input operator")
         if not allclose(operator.coeffs.imag, 0):
             raise ExtensionError("Input operator is not Hermitian.")
         params = _expval_params(operator, variance=True)
--- a/qiskit/providers/aer/library/save_expectation_value.py
+++ b/qiskit/providers/aer/library/save_expectation_value.py
@@ -53,7 +53,7 @@ class SaveExpectationValue(SaveAverageData):
         .. note:
 
             This instruction can be directly appended to a circuit using the
-            :func:`save_expval` circuit method.
+            :func:`save_expectation_value` circuit method.
         """
         # Convert O to SparsePauliOp representation
         if isinstance(operator, Pauli):
@@ -107,7 +107,7 @@ class SaveExpectationValueVariance(SaveAverageData):
 
         .. note:
             This instruction can be directly appended to a circuit using
-            the :func:`save_expval_var` circuit method.
+            the :func:`save_expectation_value` circuit method.
         """
         # Convert O to SparsePauliOp representation
         if isinstance(operator, Pauli):
@@ -161,13 +161,13 @@ def _expval_params(operator, variance=False):
     return list(params.items())
 
 
-def save_expval(self,
-                key,
-                operator,
-                qubits,
-                unnormalized=False,
-                conditional=False,
-                pershot=False):
+def save_expectation_value(self,
+                           key,
+                           operator,
+                           qubits,
+                           unnormalized=False,
+                           conditional=False,
+                           pershot=False):
     r"""Save the expectation value of a Hermitian operator.
 
     Args:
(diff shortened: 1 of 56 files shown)


qiskit/providers/aer/library/save_expectation_value.py
```diff
@@ -0,0 +1,246 @@
...
+            This instruction can be directly appended to a circuit using the
+            :func:`save_expectation_value` circuit method.
+        """
+        # Convert O to SparsePauliOp representation
+        if isinstance(operator, Pauli):
+            operator = SparsePauliOp(operator)
+        elif not isinstance(operator, SparsePauliOp):
+            operator = SparsePauliOp.from_operator(Operator(operator))
+        if not isinstance(operator, SparsePauliOp):
+            raise ExtensionError("Invalid input operator")
```
vvilpas, line 64: Can we reach this code? Aren't we always transforming to a SparePauliOp?

qiskit/providers/aer/library/save_expectation_value.py
```diff
@@ -0,0 +1,246 @@
...
+        if not allclose(operator.coeffs.imag, 0):
+            raise ExtensionError("Input operator is not Hermitian.")
+        params = _expval_params(operator, variance=True)
+        super().__init__('save_expval_var',
+                         key,
+                         operator.num_qubits,
+                         conditional=conditional,
+                         pershot=pershot,
+                         unnormalized=unnormalized,
+                         params=params)
```
vvilpas, line 128: This is almost the same as the code in `SaveExpectationValue`, maybe we can make a common parent and pass to its constructor the `name`: `save_expval` or `save_expval_var` and if `variance` is `False` or `True`.

vvilpas (commented):
Looks good! I leave a few comments.

src/simulators/density_matrix/densitymatrix_state_chunk.hpp
```diff
@@ -32,6 +32,32 @@
...
+    {"memory", "register", "probabilities",
+     "probabilities_with_variance", "expectation_value_pauli",
+     "expectation_value_pauli_with_variance"});
+
+// Allowed gates enum class
+enum class Gates {
+  u1, u2, u3, r, rx,ry, rz, id, x, y, z, h, s, sdg, sx, t, tdg,
+  cx, cy, cz, swap, rxx, ryy, rzz, rzx, ccx, cp, pauli
+};
+
```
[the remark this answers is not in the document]
chriseclectic: For now yes, but in future PR they will be different since I think density_matrix_chunk shouldn't support save_density_matrix.

src/simulators/density_matrix/densitymatrix_state_chunk.hpp
```diff
@@ -629,28 +691,9 @@ void State<densmat_t>::snapshot_pauli_expval(const Operations::Op &op,
...
-      icol = (BaseState::global_chunk_index_ + i) - (irow << ((BaseState::num_qubits_ - BaseState::chunk_bits_)/2));
-      if(irow == icol){   //only diagonal chunks are calculated
-        auto exp_tmp = coeff * BaseState::qregs_[i].expval_pauli(op.qubits, pauli);
-        exp_re += exp_tmp.real();
-        exp_im += exp_tmp.imag();
-      }
-    }
-    complex_t t(exp_re,exp_im);
-    expval += t;
+    expval += coeff * pauli_expval(op.qubits, pauli);
```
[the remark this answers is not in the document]
chriseclectic: I think I probably broke it. I'm waiting for @doichanj to make a PR to this branch to fix the chunk implementation of save expval.

src/simulators/statevector/statevector_state_chunk.hpp
```diff
@@ -33,6 +33,39 @@
...
+     "expectation_value_pauli_single_shot"});
+
+// Allowed gates enum class
+enum class Gates {
+  id, h, s, sdg, t, tdg,
+  rxx, ryy, rzz, rzx,
+  mcx, mcy, mcz, mcr, mcrx, mcry,
+  mcrz, mcp, mcu2, mcu3, mcswap, mcsx, pauli
+};
+
```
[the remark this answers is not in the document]
chriseclectic: Statevector chunk wont support save_statevector or save_density_matrix

qiskit/providers/aer/library/save_expectation_value.py
chriseclectic: I think you're right

qiskit/providers/aer/library/save_expectation_value.py
chriseclectic: We could but I think making a common base class adds more lines of code than it removes from the copy pasted parts.

vvilpas (approved):

PR #1106, opened 2021-01-20
Allow v2 provider backends

jyu00:
### Summary
Qiskit/qiskit-terra#5086 introduced v2 provider with new classes `Backend`, `Job`, and `Provider`. `ibmq-provider` is switching to v2 provider interface, and this PR ensures that `Backend` will be accepted in addition to `BaseBackend`.


mtreinish (approved):
LGTM, it'd be good to have a test here, but the easy way to go about that using a fake backend won't work yet because the fake backends are still legacy backends. We can circle back here after we have some local new interface backends that will work.

PR #1109, opened 2021-01-22
Fix bugs in automatic updates of simulator configuration basis gates

chriseclectic:
### Summary

Fixes bugs with automatically setting configuration basis gates when using custom configurations (in from backend), methods, and noise models options.

### Details and comments

Fixes issue with setting :class:`~qiskit.providers.aer.QasmSimulator`
    basis gates when using ``"method"`` and ``"noise_model"`` options
    together, and when using them with a simulator constructed using
    :meth:`~qiskit.providers.aer.QasmSimulator.from_backend`. Now the
    listed basis gates will be the intersection of gates supported by
    the backend configuration, simulation method, and noise model basis
    gates. If this intersection of the noise model basis gates and 
    simulator basis gates is empty a warning will be logged.

### DIFF
--- a/qiskit/providers/aer/backends/qasm_simulator.py
+++ b/qiskit/providers/aer/backends/qasm_simulator.py
@@ -285,15 +285,18 @@ class QasmSimulator(AerBackend):
         'max_shots': int(1e6),
         'description': 'A C++ QasmQobj simulator with noise',
         'coupling_map': None,
-        'basis_gates': [
+        'basis_gates': sorted([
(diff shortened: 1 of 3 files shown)


mtreinish (commented):
LGTM, basically you're treating the aer custom instructions separately and ensuring they're always present regardless of backend creation method and adjusting the supported ones based on the method. Although the selection logic here is involved enough I'd say it warrants extra testing to make sure the basis gates are what we expect for every permutation of backend creation function and simulation method to ensure things are set as we expect. That way we can avoid errors in the future.

mtreinish (approved):

PR #1110, opened 2021-01-25
Add Fusion variations

hhorii:
### Summary

Add fusion variations

### Details and comments

Currently cost-based-fusion is an only method of fusion. This PR adds 1- and 2-Qubit fusion, which greedily fuse operations with consideration of commutativity in gate operations. To add them, this PR introduces `AER::Transpile::Fuser`, which is the base class of three fusion methods.

- [x] Add N-Qubit fusion
- [x] Add tests


hhorii:
@vvilpas Tests were failed because of a issue in recent MPI optimization. @doichanj is sending a PR to fix this issue.

src/transpile/fusion.hpp
```diff
@@ -84,57 +683,52 @@ class Fusion : public CircuitOptimization {
...
-   Operations::OpType::superop,
-   Operations::OpType::reset},
-  {}, {}
-);
-
+Fusion::Fusion() {
+  fusers.push_back(std::make_shared<DiagonalFusion>());
+  fusers.push_back(std::make_shared<NQubitFusion<1>>());
+  fusers.push_back(std::make_shared<NQubitFusion<2>>());
+  fusers.push_back(std::make_shared<CostBasedFusion>());
```
vvilpas, line 730: Is there a usecase where we wouldn't want to use all these `fusers` but only a few them (selected by conf for example)?

vvilpas (commented):
I leave a few comments, while we wait for Doi's PR. Looks like they are many, but a large fraction of them are just to add forgotten `override` keyword.

src/transpile/fusion.hpp
```diff
@@ -32,6 +32,612 @@ using oplist_t = std::vector<op_t>;
...
+
+    auto cx_end = pos - 1;
+
+    bool found = false;
+    // find diagonals
+    for (; pos < ops.size(); ++pos)
+      if (is_diagonal_op(ops[pos]))
+        found = true;
+      else
+        break;
```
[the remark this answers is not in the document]
hhorii: This is good for`found`, but not for `u1_end`. Consecutive diagonal gates may exists after CX as follows:

```
  //      ┌───┐                                   ┌───┐
  // q_0: ┤ X ├───────────────────────────────────┤ X ├
  //      └─┬─┘┌───┐            ┌──────────┐ ┌───┐└─┬─┘
  // q_1: ──■──┤ X ├────────────┤ diagonal ├─┤ X ├──■──
  //           └─┬─┘┌──────────┐└──────────┘ └─┬─┘
  // q_2: ───────■──┤ diagonal ├───────────────■───────
  //                └──────────┘
  //        ■ [from]                           ■ [pos,u1_end]
  //             ■ [cx_end]
```

I will add more comments on this code.

src/transpile/fusion.hpp
```diff
@@ -32,6 +32,612 @@ using oplist_t = std::vector<op_t>;
...
+    std::set<uint_t> checking_qubits_set;
+    auto next_diagonal_end = get_next_diagonal_end(ops, op_idx, checking_qubits_set);
+
+    if (next_diagonal_end < 0)
+      continue;
+
+    if (checking_qubits_set.size() > max_fused_qubits)
+      continue;
+
+    std::set<uint_t> fusing_qubits_set = checking_qubits_set;
```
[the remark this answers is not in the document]
hhorii: True. I will take it. Thanks!

src/transpile/fusion.hpp
hhorii: When performance degradation happens because of fusion, we will disable some of them for diagnostics. All of fusers can be disabled with configuration (such as `fusion_enable.diagonal`).

src/transpile/fusion.hpp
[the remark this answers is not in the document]
vvilpas: Ah, right, you keep increment `pos`. Missed that.

src/transpile/fusion.hpp
vvilpas: ok, you always have the fusers and you can disable by config.

hhorii:
#1149 will resolve fails in test.

chriseclectic:
I think this needs rebasing to remove #1149 now that the bug fixes have been merged into master. Otherwise we need to wait for 1149 to be merged before this can be merged

src/framework/operations.hpp
```diff
@@ -288,6 +288,19 @@ inline Op make_unitary(const reg_t &qubits, cmatrix_t &&mat, std::string label =
   return op;
 }
 
+inline Op make_diagonal(const reg_t &qubits, const cvector_t &vec, std::string label = "") {
```
chriseclectic, line 291: Since vec is usually a temporary in the fusion code that uses this, you should make a move-semantics version as well with
```suggestion
inline Op make_diagonal(const reg_t &qubits, cvector_t &&vec, std::string label = "") {
  // same but with
  op.params = std::move(vec);

File: src/transpile/fusion.hpp [hunk 5]
```diff
@@ -32,6 +32,645 @@ using oplist_t = std::vector<op_t>;
...
 
+class FusionMethod {
+public:
+  // Return name of method
+  virtual std::string name() = 0;
+
+  virtual bool support_diagonal() const = 0;
+
+  // Aggregate a subcircuit of operations into a single operation
+  virtual op_t generate_operation(std::vector<op_t>& fusioned_ops, bool diagonal = false) const {
```
chriseclectic, line 43: Does diagonal mean all ops in fusioned_ops are diagonal, or that the output is diagonal even if the ops are not?
If its the first we should have a specialized simulation method that doesn't require building the full square matrix.

src/transpile/fusion.hpp
```diff
@@ -32,6 +32,645 @@ using oplist_t = std::vector<op_t>;
...
+      for (size_t i = 0; i < op.qubits.size(); i++)
+        op.qubits[i] = orig2remapped[op.qubits[i]];
+
+    auto fusioned_op = generate_operation_internal(fusioned_ops, arg_qubits);
+
+    // Revert qubits
+    for (size_t i = 0; i < fusioned_op.qubits.size(); i++)
+      fusioned_op.qubits[i] = remapped2orig[fusioned_op.qubits[i]];
+
+    if (diagonal) {
```
chriseclectic, line 68: To continue from above comment. If all ops are diagonal, we should have `generate_operation_internal` take the diagonal arg and return the diagonal operation if the method supports it.

src/transpile/fusion.hpp
```diff
@@ -32,6 +32,645 @@ using oplist_t = std::vector<op_t>;
...
+      std::vector<complex_t> vec;
+      vec.assign((1UL << fusioned_op.qubits.size()), 0);
+      for (size_t i = 0; i < vec.size(); ++i)
+        vec[i] = fusioned_op.mats[0](i, i);
+      fusioned_op = Operations::make_diagonal(fusioned_op.qubits, vec, std::string("fusion"));
```
chriseclectic, line 73: If you add the move semantics to `make_diagonal` this should be 
```suggestion
      fusioned_op = Operations::make_diagonal(fusioned_op.qubits, std::move(vec), std::string("fusion"));
```

src/transpile/fusion.hpp
```diff
@@ -32,6 +32,645 @@ using oplist_t = std::vector<op_t>;
...
+        return false;
+    }
+  };
+};
+
+class SuperOpFusion : public UnitaryFusion {
+public:
+  virtual std::string name() override { return "superop"; };
+
+  virtual bool support_diagonal() const override { return false; }
```
chriseclectic, line 161: We could add a diagonal_superop instruction for the density matrix / superop simulators

src/transpile/fusion.hpp
```diff
@@ -32,6 +32,645 @@ using oplist_t = std::vector<op_t>;
...
+      return op.params[0] == std::complex<double>(0.) && op.params[1] == std::complex<double>(0.);
+    else
+      return false;
+  }
+
+  if (op.type == Operations::OpType::diagonal_matrix)
+    return true;
+
+  if (op.type == Operations::OpType::matrix)
+    return op.mats.size() == 1 && Utils::is_diagonal(op.mats[0], 1e-7);
```
chriseclectic, line 518: Won't it be expensive to check every matrix is diagonal?

chriseclectic (commented):
I have a couple questions about diagonal param in fusion. The main one is if we are fusing a set of diagonal matrices, we should not use the unitary simulator since that is inefficient, but directly update the diagonal vector instead.

src/transpile/fusion.hpp
hhorii: It is the second because CNOT-diagonal-CNOT patttern is detected.

I implemented [more efficient version](https://github.com/hhorii/qiskit-aer/blob/9f593798e0a9d9be4bcc7549797d7a3b0349a8fe/src/transpile/fusion_method.hpp#L160-L180) to generate a diagonal matrix but performance improvement was very limited for generation of small diagonal matrix.

src/transpile/fusion.hpp
hhorii: I took it.

vvilpas (approved):

PR #1115, opened 2021-02-02
Fix macos tests in CI

vvilpas:
### Summary

Use constraints file when installing deps in running macos tests CI step.


mtreinish:
we need to add machine learning to this list too I guess: https://github.com/Qiskit/qiskit-aer/blob/master/.github/workflows/docs.yml#L83 we shouldn't be running those tutorials in CI. Although, we'll probably need to remove the `-U` on the pip commands there too.

vvilpas:
Ok, finally it seems to work with `numpy` 1.20 although `scs` fails to install in `macos` and an older version is installed instead of the last one. In addition, we need to avoid latest `cvxpy` version (1.1.8).

mtreinish (approved):

Issue #1102, opened 2021-01-16
Add support for Pauli expval computation to the extended stabilizer method


chriseclectic:
### What is the expected behavior?

Add support for computing single N-qubit Pauli expectation values to the extended stabilizer simulation method.

PR #1101 adds the new instruction for saving expectation values. Currently this is supported by all qasm simulation methods *except* the extended stabilizer simulator. It would be good to make this instruction work uniformly for all simulation methods.

The implementation of this instruction is done by decomposing the operator into the Pauli basis and calculating the expectation value in terms of the weighted sum of single Pauli expectation values, so to support it on the extended stabilizer would only require adding functionality to compute single Pauli expectation values <P>.

PR #1121, opened 2021-02-09
Extended stabilizer simulator expval command

gadial:
### Summary
This PR adds support for the snapshot type "expectation_value_pauli" in the extended stabilizer simulator requested in #1102.


chriseclectic (changes requested):
Can you update this to use the new `SaveExpectationValue` and `SaveExpectationValueVariance` instructions from #1101. There isn't any need to add legacy snapshot support since they are being deprecated next release.

src/simulators/extended_stabilizer/extended_stabilizer_state.hpp
```diff
@@ -777,6 +802,38 @@ void State::statevector_snapshot(const Operations::Op &op, ExperimentResult &res
      BaseState::qreg_.statevector(norm_estimation_samples_, 3, rng));
 }
 
+double State::expval_pauli(const reg_t &qubits,
```
chriseclectic, line 810: Here we should add the `rng` argument to the function, so that the seed_simulator value is respected.

We can either remove `expval_pauli` from base class, or you can just implement a trivial instance of the base class function `expval_pauli(args) {return 0; }` and only use the rng version of the function

chriseclectic (changes requested):
Having the extra rng state will lead to issues with how simulator seeds are used so I think we need to avoid that.

To that end either we can remove `apply_save_expval` and `expval_pauli` from the base class (and copy paste the needed code to other simulator state classes), or just implement trivial one here and add override functions for both that take the rng as an argument.

chriseclectic (commented):
I think only changes now are to add `save_expectation_value`` and `save_expectation_value_variance` to the custom instructions list for the extended_stabilizer in the qasm_simulator.py, and to add `extended_stabilizer` method to the stabilizer circuit qasm_save_expval.py tests.

chriseclectic (approved):

PR #1129, opened 2021-02-11
Add `SaveAmplitudes` and `SaveAmplitudesSquared` instructions.

### LINKED ISSUE #882: Add amplitudes snapshot

Add a pershot snapshot that only records select entries of a statevector rather than the entire statevector. This can be useful for debugging and verifying computations on large numbers of qubits.

In Qiskit the snapshot should be specified by a list of vector positions as either integers or bitstrings, in qobj we can pass the params as a list of integers. The output should be a complex numpy array of those vector amplitudes 
eg input -> `idx = [i0, i1, ...]`, output -> `np.array([statevec[idx[0]], statevec[idx[1]], ...])`

chriseclectic:
### Summary

Adds save amplitudes and save amplitudes squared functions.

Remove snapshot_amplitudes which this replaces, since it hasn't been released in stable yet.

Closes #882

### Details and comments

save amplitudes is supported by MPS and statevector methods, save amplitudes squared is supported by MPS, statevector, and density matrix.


src/simulators/statevector/qubitvector.hpp
```diff
@@ -670,6 +673,17 @@ std::complex<data_t> QubitVector<data_t>::operator[](uint_t element) const {
...
 
+template <typename data_t>
+void QubitVector<data_t>::set_state(uint_t pos, std::complex<data_t>& val) {
+  data_[pos] = val;
+}
+
+template <typename data_t>
+std::complex<data_t> QubitVector<data_t>::get_state(uint_t pos) const {
+  return data_[pos];
+}
```
vvilpas, line 684: Isn't it enough with the `operator[](int i)` functions?

chriseclectic: Issue here was with thrust backend. It's `[]` operator returns raw pointer to thrust compex type and requires `get_state` to cast to std::complex. So rather than mess around with template overloads I just added the same function to the non-thrust qubitvector.

vvilpas (dismissed):

vvilpas (approved):

PR #1130, opened 2021-02-12
Allow disabling of dependency installation in setup.py

### LINKED ISSUE #1128: Allow disabling of pip installations in setup.py

Currently setup.py is quite aggressive in installing dependencies (see https://github.com/Qiskit/qiskit-aer/blob/master/setup.py#L25). This leads to unexpected behaviour on HPC systems where Qiskit is being installed as an environment module (i.e., self-contained) as it installs these dependencies for the current `python`, which in our case is in another environment module. This makes our Python installation non-reproducible, and the Qiskit module broken if another Python environment module is used. It should be possible to disable this behaviour since it can have severe unintended consequences.

ocaisa:
Fixes #1128

### Summary

Allow disabling of automated dependency installation in `setup.py`

### Details and comments

I leave this here as a proof of concept for what we are looking for in #1128

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,29 +14,31 @@ import sys
 
 PACKAGE_NAME = os.getenv('QISKIT_AER_PACKAGE_NAME', 'qiskit-aer')
 _DISABLE_CONAN = distutils.util.strtobool(os.getenv("DISABLE_CONAN", "OFF").lower())
+_DISABLE_DEPENDENCY_INSTALL = distutils.util.strtobool(os.getenv("DISABLE_DEPENDENCY_INSTALL", "OFF").lower())
 
-if not _DISABLE_CONAN:
+if not _DISABLE_DEPENDENCY_INSTALL:
+    if not _DISABLE_CONAN:
+        try:
+            from conans import client
+        except ImportError:
+            subprocess.call([sys.executable, '-m', 'pip', 'install', 'conan>=1.31.2'])
+            from conans import client
+
     try:
-        from conans import client
+        from skbuild import setup
     except ImportError:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -14,29 +14,31 @@
 
 PACKAGE_NAME = os.getenv('QISKIT_AER_PACKAGE_NAME', 'qiskit-aer')
 _DISABLE_CONAN = distutils.util.strtobool(os.getenv("DISABLE_CONAN", "OFF").lower())
+_DISABLE_DEPENDENCY_INSTALL = distutils.util.strtobool(os.getenv("DISABLE_DEPENDENCY_INSTALL", "OFF").lower())
+
+if not _DISABLE_DEPENDENCY_INSTALL:
```
ocaisa, line 19: It is probably better to move this under the `except` in each case with a statement like
```
dependency = 'conan>=1.31.2'
if _DISABLE_DEPENDENCY_INSTALL:
    print("{} is a required dependency. Please provide it and repeat install".format(dependency))
    sys.exit(1)
else:
    subprocess.call([sys.executable, '-m', 'pip', 'install', dependency])
    from conans import client
```

ocaisa:
Signing a contributor agreement is a big hoop for me, what I propose here are just ideas for how you might implement things, feel free to close once it is no longer useful

vvilpas:
> Signing a contributor agreement is a big hoop for me, what I propose here are just ideas for how you might implement things, feel free to close once it is no longer useful

Sure, no problem. I am implementing this, based on your inputs, in #1135.
Thanks!

PR #1132, opened 2021-02-14
Fixed bug in sample_measure_using_probabilities

### LINKED ISSUE #1126: MPS simulator gives incorrect results
### Informations

- **Qiskit Aer version**: stable, master
- **Python version**: 3.8
- **Operating system**: Linux

See code snippet below.

```
from qiskit import execute, QuantumCircuit

sim = QasmSimulator(method='matrix_product_state')

circ = QuantumCircuit(3, 1)
circ.h(0)
circ.x(1)
circ.cx(0, 2)
circ.measure(1, 0)

res = execute(circ, sim, shots=30,
              optimization_level=0).result()
print(res.get_counts(0))
```

Output is:
```
{'0': 30}
```

(should have been `'1'` for all shots, which is indeed the output of the statevector simulation method).

### Suggested solutions

Need to debug. A priori, this seems to be related to the fact that the middle qubit was moved internally.

merav-aharoni:
### Summary
Fixed bug reported in issue #1126 

### Details and comments
When measuring on a subset of the qubits, when converting the sampled measurement to register form, the result was widened to all qubits in the call to int2reg. The gives incorrect results in the final register. Instead, we should only widen to the number of measured qubits.

Closes #1126 #1165

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+---
+notifications:
+  email: false
+
+cache:
+  pip: true
+
+os: linux
+dist: bionic
+language: python
+python: 3.7
+jobs:
+  include:
+    - name: Build aarch64 wheels
+      arch: arm64
+      services:
+        - docker
+      install:
+        - echo ""
+      env:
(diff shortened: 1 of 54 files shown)


vvilpas (changes requested):
Looks good! Can we add a small test reproducing the fixed problem? That way we'll be sure it doesn't happen again.

vvilpas (approved):

PR #1135, opened 2021-02-15
Allow disabling of dependency installation in setup.py

### LINKED ISSUE #1128: Allow disabling of pip installations in setup.py

vvilpas:
### Summary

- Some Python dependencies are installed outside regular `setuptools` installation because are required
  by compilation of C++ code. Now this installation can be disabled setting the environment variable
  _DISABLE_DEPENDENCY_INSTALL to 1 or "On".

- This commit also adds version checks on those packages when `setup.py` checks if they are already available.

### Details and comments

Fixes #1128


mtreinish (approved):
This LGTM, it might be worth pointing out that newer pip versions shouldn't need this because of the pyproject.toml build requirements outlining these. But it comes up enough that people are calling setup.py directly or using an older pip version that this doesn't work.

My only hesitation here is the additional use of distutils which will be deprecated in python 3.10: https://www.python.org/dev/peps/pep-0632/ they specifically call out strtobool as something we need to implement ourselves (https://www.python.org/dev/peps/pep-0632/#migration-advice ). Maybe for future proofing we do that in a follow up and write our own strtobool

vvilpas:
I've added our won `strtobool` function, mimicking  `distutils` one.

setup.py
```diff
@@ -3,36 +3,56 @@
...
-
-try:
-    import pybind11
-except ImportError:
-    subprocess.call([sys.executable, '-m', 'pip', 'install', 'pybind11>=2.6'])
+        mod = importlib.import_module(to_import)
+        mod_ver = parse_version(mod.__version__)
+        if ((min_version and mod_ver < parse_version(min_version))
+                or (max_version and mod_ver >= parse_version(max_version))):
+            raise RuntimeError('{} {} is installed but required version is {}.'.
```
chriseclectic, line 40:
```suggestion
            raise RuntimeError(f'{package_name} {mod_ver} is installed but required version is {install_ver}.'.
```

setup.py
```diff
@@ -3,36 +3,56 @@
...
+                               format(to_install, mod_ver, to_install_ver))
+
+    except ImportError as err:
+        if _DISABLE_DEPENDENCY_INSTALL:
+            raise ImportError(str(err) +
+                              "\n{} is a required dependency. Please provide it and repeat install"
```
chriseclectic, line 46:
```suggestion
                              f"\n{package_name} is a required dependency. Please provide it and repeat install"
```

chriseclectic (commented):
Some suggestions for making the code a little easier to read

chriseclectic (approved):

PR #1137, opened 2021-02-15
Migrate windows CI to all be in github actions

mtreinish:
### Summary

This commit migrates the remaining windows CI away from using
azure-pipelines and unifies the configuration with the rest of the
supported operating systems in github actions. It also switches the
wheel building jobs away from using the custom for loop we used in azure
to leveraging cibuildwheel. This simplifies our CI configuration and
also removes CI resource contetion with terra which uses a shared queue
in azure.


.github/workflows/deploy.yml
```diff
@@ -72,18 +70,18 @@ jobs:
         name: Install Python
         with:
           python-version: '3.7'
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
+        if: runner.os == 'Windows'
```
vvilpas, line 75: I guess we don't need this, as there is no windows gpu build yet.

mtreinish: Ah, yeah it's not strictly required here. I was just adding it to every cibuildwheel job blindly and didn't really think about it.

chriseclectic (approved):

PR #1147, opened 2021-02-19
explicitly avoid use of omp for experiments in serial execution

hhorii:
### Summary
Nested OMP parallel loops produce significant performance regression. Therefore, in Aer, parallelism is enabled only for experiments OR statevector updates with default configuration. Parallelization for experiments is configured in `controller.hpp` and for statevector update is in `simulators/statevector/`.
 
#788 changed `controller.hpp` to guard use of `omp parallel` in `#pragma` description as follows:
```
#pragma omp parallel for if (parallel_experiments_ > 1) num_threads(parallel_experiments_)
```
This can not avoid the issue of nested OMP in some implementation, especially in Linux. Even though OMP does not parallelize the following loop, another OMP in the loop wrongly recognizes that current execution is parallelized. 

### Details and comments
This PR explicitly avoids use of OMP directive if `parallel_experiments_` is more than `1`. 

```
    if (parallel_experiments_ > 1) {
      #pragma omp parallel for num_threads(parallel_experiments_)
      for (int j = 0; j < result.results.size(); ++j) {
      ...
      }
    } else {
      for (int j = 0; j < result.results.size(); ++j) {
          ...
      }
    }
```


src/controllers/controller.hpp
```diff
@@ -808,13 +808,23 @@ Result Controller::execute(std::vector<Circuit> &circuits,
...
 #endif
+      }
+    } else {
+      for (int j = 0; j < result.results.size(); ++j) {
+#ifdef AER_MPI
+        execute_circuit(circuits[j+distributed_experiments_begin_], circ_noise_models[j+distributed_experiments_begin_], config, result.results[j]);
+#else
+        execute_circuit(circuits[j], circ_noise_models[j], config, result.results[j]);
+#endif
+      }
```
vvilpas, line 827: This OMP bug is weird. Is this something they will fix in the future?

To avoid some duplication maybe we can write it as:

```
    uint_t offset = 0;
    #ifdef AER_MPI
    offset = distributed_experiments_begin_;
    #endif

    if (parallel_experiments_ > 1) {
      #pragma omp parallel for num_threads(parallel_experiments_)
      for (int j = 0; j < result.results.size(); ++j) {
        execute_circuit(circuits[j + offset], circ_noise_models[j + offset], config, result.results[j]);
      }
    } else {
      for (int j = 0; j < result.results.size(); ++j) {
        execute_circuit(circuits[j + offset], circ_noise_models[j + offset], config, result.results[j]);
      }
    }

```

chriseclectic: I don't think they will be fixing it. We had this issue before in stable and had to add this conditional block to avoid using the pragma, but it looks like it was accidently removed as part of the large MPI PR.

chriseclectic (approved):

PR #1157, opened 2021-02-25
Bump min terra version

### LINKED ISSUE #996: Error in the reported Boltzmann constant and Planck constant in qiskit.providers.aer.noise.device.models
Comment in the **_excited_population** function **from qiskit.providers.aer.noise.device.models**: there are errors in the reported Boltzmann and Plank constants.
The correct values should be:
kB = 8.617333262e-5 (eV/K)
h = 4.135667696e-15 (eV*s)

chriseclectic:
### Summary

* Bumps minimum terra version to 0.17, which is required for transpiler to handle save instructions correctly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,7 +79,7 @@ jobs:
           CIBW_BEFORE_ALL_LINUX: "yum install -y openblas-devel"
           CIBW_SKIP: "cp27-* cp34-* cp35-* pp*"
           CIBW_TEST_COMMAND: "python {project}/tools/verify_wheels.py"
-          CIBW_TEST_REQUIRES: "git+https://github.com/Qiskit/qiskit-terra.git"
+          CIBW_BEFORE_TEST: "git+https://github.com/Qiskit/qiskit-terra.git"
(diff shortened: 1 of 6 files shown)


qiskit/providers/aer/noise/device/models.py
```diff
@@ -249,11 +249,11 @@ def _excited_population(freq, temperature):
     if freq != inf and temperature != 0:
         # Compute the excited state population from qubit
         # frequency and temperature
-        # Boltzman constant  kB = 6.62607015e-34 (eV/K)
-        # Planck constant h =  6.62607015e-34 (eV.s)
+        # Boltzman constant  kB = 8.617333262-5 (eV/K)
+        # Planck constant h = 4.135667696e-15 (eV.s)
```
mtreinish, line 253: didn't want to change the units to J.S?

mtreinish (dismissed):

qiskit/providers/aer/noise/device/models.py
chriseclectic: That would have also fixed it, but this way they are in compatible units for computing h/kb

PR #1160, opened 2021-02-25
Remove previously deprecated methods

chriseclectic:
### Summary

Remove deprecated old-style conditional support
Remove deprecated old-style x90 noise model decomposition

### DIFF
--- a/releasenotes/notes/remove-deprecated-ba6eb1060207af99.yaml
+++ b/releasenotes/notes/remove-deprecated-ba6eb1060207af99.yaml
@@ -1,0 +1,6 @@
+---
+upgrade:
+  - |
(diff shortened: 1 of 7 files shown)


releasenotes/notes/remove-deprecated-ba6eb1060207af99.yaml
```diff
@@ -0,0 +1,11 @@
+---
+upgrade:
+  - |
```
mtreinish, line 3: For these release notes can you add 2 pieces of information to each note, when the feature was deprecated (for example something like "This was previously deprecated as part of the 0.7.0 release"), and what users should use instead (if there is another option). It's good to put both that so a user who's code breaks when upgrading to 0.8 can read the release notes and figure out how to change their code to make things work again.

PR #1161, opened 2021-02-25
Add default save instruction labels

chriseclectic:
Changes order of save instruction args to make the first positional arg `key` into a kwarg `label` with a default string value specific to each save instruction type.

### Summary

This PR changes the format of the recently added save instructions to make the positional `key` arg a `label` kwarg. It also gives a default label value to each instruction so that they can be used without specifying a label.

### Details and comments

This allows use such as `circ.save_statevector()`, which will return result `{"statevector": state_vec}` and `circ.save_statevector(label='psi')` which will return result as `{"psi": state_vec}`.

Note that duplicate labels in a circuit raise an exception on execution, so if two of the same save instruction are used in a circuit non-default labels must be specified in the additional instructions. A note on this was added to the API docs.

### DIFF
--- a/qiskit/providers/aer/library/save_amplitudes.py
+++ b/qiskit/providers/aer/library/save_amplitudes.py
@@ -21,17 +21,17 @@ from .save_data import SaveSingleData, SaveAverageData, default_qubits
 class SaveAmplitudes(SaveSingleData):
     """Save complex statevector amplitudes."""
     def __init__(self,
-                 key,
                  num_qubits,
                  params,
+                 label="amplitudes",
                  pershot=False,
                  conditional=False):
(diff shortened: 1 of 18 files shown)


qiskit/providers/aer/library/save_amplitudes.py
```diff
@@ -42,10 +42,10 @@ def __init__(self,
         Raises:
             ExtensionError: if params is invalid for the specified number of qubits.
         """
+        if label is None:
+            label = "amplitudes"
```
vvilpas, line 46: I guess you forgot to remove this, as `label` already has a default value?

vvilpas (dismissed):
looks good. Only a minor comment, otherwise good to go.

PR #1162, opened 2021-02-26
Add arm64 release wheel jobs

mtreinish:
### Summary

This commit adds a new CI job to publish precompiled binary wheels for
linux on aarch64 (arm64). Since the latest numpy and scipy releases now
publish wheels for aarch64 we can run CI jobs to do the same since we
won't be spending all our CI time budget compiling upstream dependencies
from source anymore (terra 0.17.0 will be publishing wheels aarch64
wheels too). This enables users running on linux aarch64 systems
to install qiskit without having to compile everything from source
anymore. To build these wheels this commit uses travis, which is the
only CI system that offers non-x86 nodes.

### Details and comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ jobs:
         - CIBW_SKIP="cp27-* cp34-* cp35-* pp*"
         - TWINE_USERNAME=qiskit
         - CIBW_TEST_COMMAND="python {project}/tools/verify_wheels.py"
+      if: tag IS present
       script:
         - pip install -U twine importlib-metadata keyring cibuildwheel==1.9.0
         - cibuildwheel --output-dir wheelhouse
+        - twine upload wheelhouse/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,30 @@
+---
+notifications:
+  email: false
+
+cache:
+  pip: true
+
+os: linux
+dist: bionic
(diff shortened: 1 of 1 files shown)


mtreinish:
The wheel jobs succeeded, you can see the logs here: https://travis-ci.com/github/Qiskit/qiskit-aer/builds/218354639 I'm going to revert the DNM test patch, this should be ready for review now.

chriseclectic:
@mtreinish Can you make this part of the GHA build / deploy workflow instead of travis?

mtreinish:
> @mtreinish Can you make this part of the GHA build / deploy workflow instead of travis?

Not really, travis is the only CI service with non-x86 CI nodes (they offer arm64, ppc64le, and s390x nodes). It's possible to build an arm64 binary on an x86 node using qemu emulation but it's painfully slow and I expect for a project as big as aer we likely wouldn't have the time budget even on GHA to finish all 4 wheel builds in emulation.

This PR is only adding jobs that run at releases. It's intended to build release aarch64 wheels for PyPI only and does not add jobs that run on PRs or post-merge. Just publishing wheels at release time is enough for now. Having arm pre-merge testing (to catch potential incompatibilities before release time) would be nice some day, but I agree we don't want to rely on travis for that.

chriseclectic (approved):

PR #1163, opened 2021-02-26
Add instructions to set simulator state

chriseclectic:
### Summary

Add instructions for setting the simulator state to a custom value.

Note: performance for these instructions for large numbers of qubits will be quite bad since the instruction is being serialized via JSON which is very inefficient for matrices and vectors.

TODO:
 - [x] Tests
 - [x] Release note

### Details and comments

Adds instructions for setting the simulator state for statevector, density matrix, unitary, and stabilizer methods. These are simulator directive instructions so cannot be unrolled to gates like `initialize` instructions. They also must be applied to all qubits in a circuit, unlike initialize.

#### Example

```python
from qiskit import QuantumCircuit, execute

nq = 2
rho = qi.random_density_matrix(2 ** nq)

qc = QuantumCircuit(nq)
qc.set_density_matrix(rho)
qc.save_density_matrix()

qiskit.execute(qc, QasmSimulator()).result().data(0)
```

 
#### For follow up PRs
- Chunk state classes
- Set matrix product state
- Set extended stabilizer state


src/simulators/statevector/statevector_state.hpp
```diff
@@ -580,6 +580,9 @@ void State<statevec_t>::apply_ops(const std::vector<Operations::Op> &ops,
           else if(op.name == "end_register_blocking"){
             BaseState::qreg_.leave_register_blocking();
           }
+        case OpType::set_statevec:
```
vvilpas, line 583: Are we missing a `break` statement for the `sim_op` case?

vvilpas (commented):
Looks good to me. I only have a small doubt regarding a switch block.

PR #1149, opened 2021-02-19
Fixes of multi-chunk State implementation

doichanj:
### Summary
Here are several fixes for parallel implementation by using cache blocking transpiler and multiple chunks.

### Details and comments
- Implementations of save state functions for multi-chunk
New save functions are implemented after #788 is merged, so I implemented these functions in multi chunk State classes.

- Fix for diagonal matrices
Cache blocking transpiler blocked diagonal matrix even though they do not need data exchange.
So now diagonal matrices are not blocked and masked in StateChunk class. The masked diagonal matrices are passed to qubit register classes.
This fixes issue in #1110 
**This fix is moved to separate PR** #1155 

- Splitting pauli gates
Cache blocking transpiler splits pauli gates when target qubits includes qubit out of chunk.

- Automatic cache blocking setting
Now controller set the cache block setting s if there is not enough memory without using multiple chunks. 
(e.g. multiple GPUs)

- Fix DensityMatrixChunk::State::move_to_matrix
 It was implemented with wrong ordering.  

- Added test cases for multi-chunk QasmSimulator
I changed the naming for the test from MPI test to chunk test.
I included save function tests and pauli gate test for multi-chunk.

PR #1167, opened 2021-03-01
Fix out of bounds array access.

vvilpas:
### Summary

Fixes a bug in CopyIn/CopyOut operations where out of bounds array elemets
were accessed.
`QubitvectorThrust` has a number of qubits `num_qubits`. When `move_to_vector` or
`copy_to_vector` functions are called a temporary array or vector of size
(1 << `num_qubits`) is created and passed to chunk_ CopyOut function. There, a number
of elements (1 << `chunk_qubits_`) of the underlying structure are copied to
this array/vector. The problem is that `chunk_qubits_` can be larger that `num_qubits`,
accessing values in the array or vector out of bounds. This is what was making
PR #1132 failing in CI.

### Details and comments

`chunk_qubits_` is initialized from maximum number of qubits in a list of circuit whereas `num_qubits` is initilaized as the number of qubits of a given circuit.

### DIFF
--- a/src/simulators/statevector/chunk/chunk.hpp
+++ b/src/simulators/statevector/chunk/chunk.hpp
@@ -155,13 +155,13 @@ public:
   {
     chunk_container_->CopyOut(dest,chunk_pos_);
   }
-  void CopyIn(thrust::complex<data_t>* src)
+  void CopyIn(thrust::complex<data_t>* src, uint_t size)
   {
-    chunk_container_->CopyIn(src,chunk_pos_);
+    chunk_container_->CopyIn(src,chunk_pos_, size);
   }
-  void CopyOut(thrust::complex<data_t>* dest)
+  void CopyOut(thrust::complex<data_t>* dest, uint_t size)
   {
-    chunk_container_->CopyOut(dest,chunk_pos_);
+    chunk_container_->CopyOut(dest,chunk_pos_, size);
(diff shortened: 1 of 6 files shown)


vvilpas:
@doichanj This a patch to fix temporarily the issue, as I got a bit lost in the structure of chunks, etc. Could you take a look to check if you see a better way to solve it?

doichanj:
I agree with checking size to be copied.
However the problem of chunk bits should be also fixed and this problem is already fixed in PR #1149 (https://github.com/Qiskit/qiskit-aer/pull/1149/commits/424ae67349f47c4a2572afaa7bce38487788e717)

chriseclectic (approved):

chriseclectic:
I accidentally merged this one when I meant to merge #1168. Should I revert or is it ok as is? @doichanj

vvilpas:
> I accidentally merged this one when I meant to merge #1168. Should I revert or is it ok as is? @doichanj

Both are needed. This one fixes a possible out of bounds write in the array. That was causing the memory corruption we saw in Merav's PR. The other one fixes a memory leak, some memory was not freed when it should have been. It doesn't cause wrong behavior but after some time it could let the program without memory.

doichanj:
I agree with @vvilpas, and I think there is no problem to merge both.

PR #1168, opened 2021-03-01
Fix memory_leak due to shared_ptr circular references.

vvilpas:
### Summary

- The chunk objects hold a shared_ptr to a chunk_container and the chunk_container
holds shared_ptrs to the chunks. This creates a circular reference problem which
avoids the destructions of chunks and chunk_containers.

- Some override warnings fixed

- Set other AER::Vector size to zero when performing moving operations.

### DIFF
--- a/src/framework/linalg/vector.hpp
+++ b/src/framework/linalg/vector.hpp
@@ -203,6 +203,7 @@ template <class T>
 Vector<T>::Vector(Vector<T> &&other) noexcept
     : size_(other.size_), data_(other.data_) {
   other.data_ = nullptr;
+  other.size_ = 0;
 }
 
 //-----------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


doichanj:
I think this PR solves circular reference issue by using weak_ptr, I think we can merge this.

chriseclectic:
Was this superseded by #1167, is there still some extra fixes in this PR?

chriseclectic (approved):

src/framework/linalg/vector.hpp
```diff
@@ -203,6 +203,7 @@ template <class T>
 Vector<T>::Vector(Vector<T> &&other) noexcept
     : size_(other.size_), data_(other.data_) {
   other.data_ = nullptr;
+  other.size_ = 0;
```
chriseclectic, line 206: This fix for `AER::Vector` should probably be backported to stable

PR #1170, opened 2021-03-03
Prepare 0.7.6 release

mtreinish:
### Summary

This commit is to prepare the 0.7.6 release. It backports the relevant fixes from master to include in the release and bumps the version string to 0.7.6. After this merges we can tag the merge commit as 0.7.6.

### Details and comments

This needs to be merged with a merge commit to preserve the backport history.


qiskit/providers/aer/noise/noise_model.py
```diff
@@ -19,6 +19,10 @@
 
 from qiskit.circuit import Instruction
 from qiskit.providers import BaseBackend
+try:
+    from qiskit.providers import Backend
```
chriseclectic, line 23: Could do something like:
```suggestion
try:
    from qiskit.providers import Backend as _Backend
    Backend = (BaseBackend, _Backend)
 except ImportError:
    Backend = BaseBackend

...
if isinstance(backend, Backend)
```

mtreinish: Yeah that works too and would be a bit simpler, I'll update it now

mtreinish: Done in: https://github.com/Qiskit/qiskit-aer/pull/1170/commits/e92de0d92d826539fb4dafe3c194690872308c9f

chriseclectic (approved):

PR #1171, opened 2021-03-03
Fix density matrix expval_pauli

chriseclectic:
### Summary

Somewhere in the refactors since the 0.7 release he way the `expval_pauli` function was removed from the density matrix method leading to incorrect computation of expectation values for noisy circuits. This restores the density matrix specific implementation of the expval_pauli function.

### Details and comments

On master expval_pauli was actually computing `Tr[O.rho^2]` instead of `Tr[O.rho]`, so was incorrect unless the simulation was ideal.

This PR also adds some extra tests including noise for testing expvals on the density matrix simulators.

@doichanj can you check the chunk versions still work in the way I updated them?

### DIFF
--- a/src/simulators/density_matrix/densitymatrix.hpp
+++ b/src/simulators/density_matrix/densitymatrix.hpp
@@ -125,6 +125,20 @@ public:
   // outcome in [0, 2^num_qubits - 1]
   virtual double probability(const uint_t outcome) const override;
 
+  //-----------------------------------------------------------------------
+  // Expectation Values
+  //-----------------------------------------------------------------------
+
+  // Return Pauli expectation value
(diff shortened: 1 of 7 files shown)


chriseclectic:
Looks like the Thrust version still isn't working. @doichanj I think the issue might be with it iterating over the full vector range, when it should only be over the range (0, nrows).

doichanj:
Now issue in Thrust implementation is fixed. https://github.com/chriseclectic/qiskit-aer/pull/14

chriseclectic:
> Looks like the Thrust version still isn't working. @doichanj I think the issue might be with it iterating over the full vector range, when it should only be over the range (0, nrows).

Thanks @doichanj !

vvilpas (approved):

PR #1176, opened 2021-03-09
From python to C++ structures

vvilpas:
### Summary

This PR enables the direct conversion from python objects on the C++ side instead of transform the python object to a dict, then to a json structure and then to a C++ structure. Some C++ objects, like `config` and `header`, are still `json` structures though.

### Details and comments

The standalone version of the simulator keeps using json structures as inputs, so we need to keep both input formats. To achieve that without duplicating code I've created a set of templatized functions which homogenises the access to the different structures. For the python objects the function considers that the object is either a python dict or an object whose attributes would correspond to the dict keys.

We observe the following performance gain when setting a density_matrix:

py_bind_times
py_bind_times_comp

Testing with the following code:

```
import qiskit
from qiskit import QuantumCircuit, execute

results = []
for nq in range(2,14):
    rho = qi.random_density_matrix(2 ** nq)

    qc = QuantumCircuit(nq)
    qc.set_density_matrix(rho)
    qc.save_density_matrix()

    results.append(qiskit.execute(qc, QasmSimulator()).result())
```

averaging for 100 runs when num_qubit < 12


src/framework/python_parser.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+    static py::list get_as_list(const py::handle& po){
+        if(!is_list_like(po)){
+            throw std::runtime_error("Object is not list like!");
+        }
+        return  py::cast<py::list>(po);
+    }
+
+    static py::list get_list(const std::string& key, const py::handle& po){
+        py::object the_list = get_py_value(key, po);
```
chriseclectic, line 109: Does this result in a copy or a reference? Ideally it should be a const reference if its still a py object.

src/framework/python_parser.hpp
```diff
@@ -0,0 +1,154 @@
...
+        if(!is_array(the_list)){
+            throw std::runtime_error("Object " + key + "is not a list!");
+        }
+        return py::list(the_list);
```
chriseclectic, line 113: Same here is `py::list(the_list)` a copy or a reference?

src/framework/pybind_casts.hpp
```diff
@@ -0,0 +1,70 @@
...
+#ifndef _aer_framework_pybind_casts_hpp_
+#define _aer_framework_pybind_casts_hpp_
+
+#include "../simulators/stabilizer/clifford.hpp"
+
+namespace py = pybind11;
+
+namespace pybind11 {
+namespace detail {
+template <typename T> struct type_caster<matrix<T>>{
```
chriseclectic, line 24: I think we might also need type casters for numpy to std::vector or Vector to avoid going via a python list right?

src/framework/pybind_casts.hpp
```diff
@@ -0,0 +1,70 @@
...
+  using base = type_caster_base<matrix<T>>;
+public:
+  PYBIND11_TYPE_CASTER(matrix<T>, _("matrix_complex_t"));
+  // Conversion part 1 (Python->C++):
+  bool load(py::handle src, bool convert){
+      // TODO: Check if make sense have to flavors of matrix: F-style and C-style
+      auto py_matrix = py::cast<py::array_t<T>>(src);
```
chriseclectic, line 31: Does this make a copy of src if its already an array?

src/framework/pybind_casts.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace pybind11 {
+namespace detail {
+template <typename T> struct type_caster<matrix<T>>{
+  using base = type_caster_base<matrix<T>>;
+public:
+  PYBIND11_TYPE_CASTER(matrix<T>, _("matrix_complex_t"));
+  // Conversion part 1 (Python->C++):
+  bool load(py::handle src, bool convert){
+      // TODO: Check if make sense have to flavors of matrix: F-style and C-style
```
chriseclectic, line 30: If src if `F` order we should be able to copy the array more efficiently since the underlying array is ordered the same way as the internal `matrix` array. Then we can use `matrix<T>::copy_from_buffer`.

If the matrix F-ordered maybe we can cast to an F-ordered (which probably makes a copy) and then move the copied buffer

src/framework/pybind_casts.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include "../simulators/stabilizer/clifford.hpp"
+
+namespace py = pybind11;
+
+namespace pybind11 {
+namespace detail {
+template <typename T> struct type_caster<matrix<T>>{
+  using base = type_caster_base<matrix<T>>;
+public:
+  PYBIND11_TYPE_CASTER(matrix<T>, _("matrix_complex_t"));
```
chriseclectic, line 27: Should this be
```suggestion
  PYBIND11_TYPE_CASTER(matrix<T>, _("matrix_t"));
```
Since matrix doesnt need to be complex

src/framework/pybind_casts.hpp
```diff
@@ -0,0 +1,70 @@
...
+  // Conversion part 1 (Python->C++):
+  bool load(py::handle src, bool convert){
```
chriseclectic, line 29: what does `convert` do?

src/framework/pybind_casts.hpp
```diff
@@ -0,0 +1,70 @@
...
+      value = matrix<T>(nrows, ncols, false);
+      for (size_t r = 0; r < nrows; r++) {
+          for (size_t c = 0; c < ncols; c++) {
+              value(r, c) = raw_mat(r, c);
+          }
+      }
+      return true;
+  }
+  // Conversion part 2 (C++ -> Python):
+  static py::handle cast(matrix<T>, py::return_value_policy policy, py::handle parent){
```
chriseclectic, line 48: This could probably replace our pybind11 code for matrix class (and I suppose is the proper way to do this)

chriseclectic (commented):
Looks good. I have some questions about the actual pybind11 casting code since it isn't clear how many copies are getting made.

src/framework/python_parser.hpp
vvilpas: I think there is no copy , but I'll change to `py::cast<py::list>(the_list)`

src/framework/pybind_casts.hpp
vvilpas: I think for std::vector the caster is already in place. For Vector we would need one but there wasn't any error so I guess we don't have any direct conversion from input to Vector. I'll check, because sometimes it's confusing working with pybind11.

src/framework/pybind_casts.hpp
vvilpas: No, as far as I know this just creates the appropriate wrapper to let you manipulate it as an array based object.

src/framework/pybind_casts.hpp
vvilpas: We can check the order and use `copy_from_buffer`. Casting to an F ordered makes a copy so probably not worth it.

src/framework/pybind_casts.hpp
vvilpas: It is to indicate if implicit conversions are accepted.

src/framework/pybind_casts.hpp
vvilpas: Yep, we could replace all those C++ --> python conversions with casters implementing this function. But I would leave that to another PR as I think it implies changing our conversion system from C++ to python.

src/framework/pybind_casts.hpp
chriseclectic: Iirc think currently all inputs are converted to `std::vector` anyway, `AER::Vector` is only used in results

src/framework/pybind_casts.hpp
chriseclectic: Yeah, lets come back to this after the 0.8 release

src/framework/pybind_casts.hpp
```diff
@@ -0,0 +1,75 @@
...
+  PYBIND11_TYPE_CASTER(matrix<T>, _("matrix_t"));
+  // Conversion part 1 (Python->C++):
+  bool load(py::handle src, bool convert){
+      // TODO: Check if make sense have to flavors of matrix: F-style and C-style
+      auto py_matrix = py::cast<py::array_t<T>>(src);
+      auto c_order = py_matrix.attr("flags").attr("carray").template cast<bool>();
+      if(py_matrix.ndim() != 2){
+          throw std::invalid_argument(std::string("Python: invalid matrix (empty array)."));
+      }
+      size_t ncols = py_matrix.shape(0);
```
chriseclectic, line 36: Shouldn't shape(0) be rows, not cols? I guess it doesn't actually matter since we are only using this for square matrices.

chriseclectic (approved):

PR #1178, opened 2021-03-12
Unify hamming weight and parity functions

mtreinish:
### Summary

Modern CPUs have built-in special instructions for calculating the
hamming weight. For example, in x86_64 the `POPCNT` instruction (which was
included with SSE4.2 on Intel CPUs and SSE4a on AMD CPUs) or the `VCNT`
instruction on ARM Neon. These enable us to calculate the hamming weight
in a single instruction instead of multiple operations. In Aer there
were previously were 2 implementations of the hamming weight function,
one lived in the extended stabilizer core module the other lived in
`AER::utils`. The `AER::utils` version used an efficient algorithm for
compute the hamming weight but would not leverage special instructions.
While the extended stabilizer version used compiler builtin functions
(if available) that when compiled with the proper flags would leverage
the special instructions and fell back to a different implementation.

This commit unifies the 2 implementations so we only have a single copy
of the hamming weight function (the hamming pairty function from
the extended stabilizer is moved too, although it is only used in the
extended stabilizer). The compiler builtin function usage was copied
from the extended stabilizer version, but the fallback is the more
efficient version used from `AER::utils`.

This commit also adds `-mpopcnt` flag is added to the avx2 flag as this is
required to tell GCC that the `POPCNT` instruction is available. SSE4.2
(and SSE4a for AMD) is present in most modern x86_64 CPUs but not on
pre-nehalem Intel CPUs (released 2008) or pre-barcelona AMD CPUs
(released 2007). So we can assume that when we're building with AVX2 those
instructions are available (since AVX2 is much newer), but we don't want
to make a general assumption it's always present, because that would
create binaries that are incompatible for any users currently running on
older x86_64 CPUs. For other non-x86_64 CPUs we likely will need to add
additional compiler flags to enable use of the special instructions, but
that can happen in a followup as needed.

### Details and comments
Co-authored-by: Victor Villar <59838221+vvilpas@users.noreply.github.com>


mtreinish:
The one thing I don't think is working right now (and never worked before with the extended stabilizer version) is that we're not building binaries that actually use `POPCNT`. At least when I look at the assembly from the built standalone binary I don't see any `popcnt` instructions listed. There are avx2 instructions like `vinserti128` present but I didn't see any popcnt. I expect I'm missing something in the cmake chages.

CMakeLists.txt
```diff
@@ -238,6 +238,7 @@ endif()
 if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64" OR CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "AMD64")
 	if(APPLE OR UNIX)
 		set(SIMD_FLAGS_LIST "-mfma;-mavx2")
+		enable_cxx_compiler_flag_if_supported("-mpopcnt")
```
mtreinish, line 241: Will this be problematic for users on pre-2007 hardware (where popcount x85_64 instructions don't exist) in the precompiled binaries? Or do you think dropping support for things that old is an OK tradeoff?

vvilpas: I think you're right and anything without popcnt instruction will crash. Maybe we can compile both flavours (with and without intrinsic popcnt) and select on startup or first call to functions which one to use. For example we can use existing `is_avx2_supported()` function and in case avx2 is supported use the intrinsic popcnt. I guess we'd be leaving some CPUs (where popcnt is supported) out but we would be on the safe side, right?

chriseclectic (approved):

PR #1181, opened 2021-03-16
[WIP] Assist with moving to the new provider interface

yaelbh:
This pull request builds on #1099. There are now tests that are failing, the purpose is to make them pass.


yaelbh:
I deleted one strange test, that did nothing, from `qasm_method`. I moved another test from `qasm_method` to `qasm_noise`.

yaelbh:
Ok I understand why the strange test was there. I'm going to keep it removed, and here is what I'm going to do:
1. `AerBackend.__init__` will call `set_options`, instead of a separate call to `set_option` for each option (which de-facto implements the `set_options` function of the super class, without applying the special handling of the basis gates that's in `set_options` of `QasmSimulator`).
2. I'll rewrite `QasmSimulator.set_options` and make it cleaner. Then we can safely remove the strange test.

One thing to consider is the difference between the following cases:
1. The user manually sets the basis gates to something, then changes the simulation method.
2. The user sets the simulation method to `stabilizer`, then changes it to `statevector`.

In the current code, when a method is changed, the new set of basis gates is the intersection of the old one with the basis gates of the new method. This behavior seems to address well the first case, because we don't want the new simulation method to add gates that the user explicitly excluded. On the other hand, this triggers a bug for the second case, where the statevector simulation method will have the same set of basis gates as the stabilizer one. So we need to store user basis gates in the configuration, and work a bit harder when the method is changed.

@chriseclectic @mtreinish Does all this make sense?
One more question: will all this be relevant when Aer 8.0 is out?

chriseclectic:
These removed tests were explicitly added in #1109 as test cases for that bug fix so I don't think we should remove them unless we test the equivalent functionality somewhere else. The basis gates is very complicated and i'm not really happy with the implementation I came up with when fixing it.

In any case the use case for the intersection was if someone builds a simulator with a noise model (or from a device backend), that defines a set of allowed  "device" basis gates. Then if you also set a simulation methods, you want the actual simulator basis gates to be the intersection of what the noise model (device) gates are, and what is supported by the simulation method itself. 

Part of how this was implemented is that the simulator stores two configurations (and properties/defaults). The actual configuration at construction which is fixed, and then the custom one that can be modified by set options. of the backend is fixed and is defined by the simulation method alone (or a custom one passed in during init). The custom configuration is stored separately if some parameter of config is changed via set options (such as basis gates). The idea is if you change noise model or method, you can recompute intersection from original one, but the code that implements this is pretty convoluted and not particularly nice. It probably still has bugs.

Maybe we need something more like you suggest where we store the method basis gates separately as well (rather than a full method config)

- original configuration (from init) <-- stored without being changed either is default or user provided
- method basis gates / custom instructions <-- stored separately changes when method changes
- noise model basis gates <-- stored in noise model
- option value for basis gates <-- stores specific value set directly by user

Then if you do `configuration()` we would need to populate `config.basis_gates` with (in order of precedence)
1. directly specified option value
2. intersection of original config basis gates, method basis gates, noise model basis gates (if set)

yaelbh:
I was hoping that at least for the failing tests, it would suffice to replace the many calls to `set_option` with the call to `set_options`, and postpone the rewrite of `set_options` to another time. However when I do it, `set_options` crashes:
```
File "/home/yaelbh/work/terra/System/qiskit/providers/aer/backends/qasm_simulator.py", line 462, in set_options
    noise_gates = out_options['noise_model'].basis_gates
KeyError: 'noise_model'
```
Which means that we will have to do the rewrite already for the v2 provider.

yaelbh:
As part of the rewrite, the options can be sorted before they are parsed. In this case by construction there will not be a difference between function calls with a different order of parameters.

yaelbh:
Since I moved the test to `qasm_noise`, the check that the correct set of basis gates is obtained when noise and method are set has been removed for the extended stabilizer method. I think it's OK, considering that this method does not support noise.

chriseclectic:
@yaelbh The extended stabilizer supports noise (just mixed clifford noise like stabilizer simulator, not kraus)

yaelbh:
Ok, got it, will take care of it

PR #1183, opened 2021-03-17
[WIP] Automatically set and save optimal OpenMP and fusion thresholds

hhorii:
### Summary

Adds basic profiling code to find suitable threshold values for OpenMP and gate fusion for backend options.
This is a revised version of #956.

### Details and comments

This PR will add following into to #956.

- [ ] Save optimized parameters into `settings.conf`
- [ ] Optimize `fusion_cost.N`


hhorii:
Once https://github.com/Qiskit/qiskit-terra/pull/5943 is merged, optimized parameters will be able to be store in user config around here:
https://github.com/Qiskit/qiskit-aer/blob/9f7b832abf8c74ed6d04d62072ff000deae882ab/qiskit/providers/aer/profile.py#L68-L69
https://github.com/Qiskit/qiskit-aer/blob/9f7b832abf8c74ed6d04d62072ff000deae882ab/qiskit/providers/aer/profile.py#L82-L84

qiskit/providers/aer/aerprovider.py
```diff
@@ -49,3 +50,9 @@ def backends(self, name=None, filters=None, **kwargs):
 
     def __str__(self):
         return 'AerProvider'
+
+    @staticmethod
+    def optimize_backend_options(min_qubits=10, max_qubits=20, ntrials=10):
```
chriseclectic, line 55: Lets remove this method from the provider for now and make it have to be run by importing the profiling functions

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -159,9 +160,17 @@ def run(self,
                 DeprecationWarning,
                 stacklevel=3)
 
+        # Add default OpenMP options
+        gpu = backend_options is not None and 'gpu' in backend_options.get('method', '')
```
chriseclectic, line 164: Backend options is a deprecated argument. Typically the method will already be in the qobj config

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -159,9 +160,17 @@ def run(self,
...
+        profiled_options = get_performance_options(gpu)
```
chriseclectic, line 165: This should be handled as pat of the format qobj, and only add the profile option to the qobj config if they aren't present in options or run options

qiskit/providers/aer/profile.py
```diff
@@ -0,0 +1,387 @@
...
+from socket import gethostname
+import numpy as np
+
+from qiskit import transpile, assemble, execute, user_config, QuantumRegister, QuantumCircuit
+from qiskit.circuit.library import QuantumVolume
+from qiskit.quantum_info import random_unitary
+from .aererror import AerError
+
+
+def profile_performance_options(min_qubits=10, max_qubits=25, ntrials=5,
```
chriseclectic, line 26: This `profile.py` file should probably be moved into `aer/utils/` and the user facing functions added to `aer/utils/__init__.py`

chriseclectic (changes requested):
Some changers needed here. I think it might be better to wait until after #1194 is merged and just use profiling for `AerSimulator` which provides a common interface to all simulation methods and makes it easier to check if using GPU or not since it has a separate `device` option.

hhorii:
Here are performance graphs (`time_taken` includes only transpilation + simulation and excludes interface overheads) with default and tuned configurations.

**Configurations:**
- Left: MacBook (2.6 GHz 6-Core Intel Core i7)
  - statevector_parallel_threshold=11
  - fusion_threshold=19
- Right: Xeon (Intel Xeon Gold 6140 CPU )
  - statevector_parallel_threshold=16
  - fusion_threshold=11

**QuantumVolume:**
image image

*IntegerComparator: *
image image

*WeightedAdder:*
image image

**QuadraticForm**:
image image

**QFT:**
image image

**RealAmplitudes with full entanglements**
image image

**RealAmplitudes with linear entanglements**
image image

Red lines are expected the lowest. Because this profiling function uses QuantumVolume, tuned options do not work for others. However, I believe, these side-effects are acceptables.

PR #1185, opened 2021-03-18
Add reading of qubit_lo_freq from pulse Schedule config

### LINKED ISSUE #1179: Pulse Qobj parser doesn't take experimental config

If we run below experiment, i.e. qubit spectroscopy, with the pulse simulator, the simulator doesn't sweep qubit frequency.
```python
frequencies = np.linspace(4.9e9, 5.1e9, 100)
qobj = assemble(schedule, backend, schedule_los=[{pulse.DriveChannel(0): freq} for freq in frequencies])
```
This code internally generates multiple copies of `schedule`, and binds custom frequency `pulse.DriveChannel(0): freq` to each `QobjExperimentalConfig`, i.e. `qobj.experiments[idx].config`. However the parser doesn't check this field and only checks the global experiment configuration `qobj.config`.

https://github.com/Qiskit/qiskit-aer/blob/90248c0938a1ee4217259028b35463c7f90cadd6/qiskit/providers/aer/pulse/controllers/pulse_controller.py#L132-L135

If we provide only _single frequency_ with multiple schedules, this setting is passed to the global configuration `qobj.config` to reduce data volume of qobj, i.e. copying this dictionary to each experiment configuration is just an overhead.

### Suggested solutions

Frequency initialization protocol should be:
- Initialize channel frequency with backend default, i.e. `backend.defaults().qubit_freq_est`/`meas_freq_est` or device Hamiltonian model
- Check `qobj.config` and overwrite initialized frequency with this frequency (if provided)
- Copy above frequency vector to each experiment, and overwrite this with `qobj.experiments[idx].config` (if provided)

DanPuzzuoli:
### Summary

Closes #1179

Follows logic as suggested in #1179 by @nkanazawa1989 : a global `qubit_lo_freq` field is populated as before, either from the qobj config or from the Hamiltonian. When simulating a specific schedule, if no `qubit_lo_freq` is present in the schedule config, the global default is used, otherwise the schedule's `qubit_lo_freq` is used.

### Details and comments

When each schedule is parsed (in dict format in `experiment_to_struct`) it checks if it has a `'config'` key, and if so, will check for a `qubit_lo_freq` field. If it's there, this will be stored in the "experiment struct" dictionary, otherwise, the `struct['qubit_lo_freq']` is set to `None`. When setting up the RHS function before solving, if `struct['qubit_lo_freq'] is not None`, this value will be subbed in, otherwise the default is maintained.

A test has been added to `test_pulse_simulator` with the same syntax as the example given by @nkanazawa1989  in #1179. A schedule which should do a perfect pi pulse (if on resonance) is run on a model with 3 different `qubit_lo_freq` values: the first below resonance, the second on resonance, and the third above resonance. The test verifies that the two off-resonant schedules fail to excite the system by checking that all measurement counts are in the ground state, whereas the on-resonance schedule returns all counts in the excited state.

### Question

Pending CI tests, this PR should be done. One potential issue that could arise with the current code is to do with formatting of the `qubit_lo_freq` field in the schedule config. The pulse simulator stores the default frequencies as a list of floats, which I believe are assumed to be canonically ordered, first with all drive frequencies in ascending order, then with all control channels. The `qubit_lo_freq` in the schedule config seems to just be a list (i.e. it doesn't contain channel name information) - is it safe to assume that it is also ordered in this way, or is there some other way to figure out the order in this list?  @nkanazawa1989

### DIFF
--- a/qiskit/providers/aer/pulse/controllers/pulse_controller.py
+++ b/qiskit/providers/aer/pulse/controllers/pulse_controller.py
@@ -19,6 +19,7 @@ Entry/exit point for pulse simulation specified through PulseSimulator backend
 
 from warnings import warn
 from copy import copy
+from typing import Callable
 import numpy as np
 from qiskit.quantum_info.operators.operator import Operator
 from ..system_models.string_model_parser.string_model_parser import NoiseParser
@@ -142,6 +143,7 @@ def pulse_controller(qobj):
              'so it is beign automatically determined from the drift Hamiltonian.')
 
     pulse_de_model.freqs = system_model.calculate_channel_frequencies(qubit_lo_freq=qubit_lo_freq)
+    pulse_de_model.calculate_channel_frequencies = system_model.calculate_channel_frequencies
 
     # ###############################
     # ### Parse backend_options
@@ -338,6 +340,8 @@ class PulseInternalDEModel:
         self.dt = None
         # holds default frequencies for the channels
         self.freqs = {}
+        # frequency calculation function for overriding defaults
+        self.calculate_channel_frequencies = None
         # diagonal elements of the hamiltonian
         self.h_diag = None
         # eigenvalues of the time-independent hamiltonian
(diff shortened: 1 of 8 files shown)


qiskit/providers/aer/pulse/controllers/pulse_controller.py
```diff
@@ -422,7 +423,15 @@ def init_rhs(self, exp):
         # Init register
         register = np.ones(self.n_registers, dtype=np.uint8)
 
-        ode_rhs_obj = get_ode_rhs_functor(self._rhs_dict, exp, self.system, channels, register)
+        rhs_dict = self._rhs_dict
+
+        # if the experiment overrides qubit_lo_freq, use this one
+        if exp.get('qubit_lo_freq', None) is not None:
```
nkanazawa1989, line 429: why not just `if 'qubit_lo_freq' in exp:`?

nkanazawa1989 (approved):
Thanks @DanPuzzuoli ! This looks good to me. Suffice to say that I'm bit worried about how much is the new unittest robust to numerical error because it looks like an integration test. I think this is still ok, but I would separate a function to generate the rhs dict and add a specific test for that.

As you mentioned qubit lo frequencies are provided by a list without any mapping to channel. We just assume straightforward mapping of qubit index to drive channel index. More precisely the actual map may be provided by the backend, i.e. `backend.configuration().qubit_channel_mapping`, but in the openpulse specification it is not allowed to have multiple drive channels per qubit. So the mapping is obvious. However in the simulator we can define arbitrary Hamiltonian, so this constraint may limit the capability of simulation. However I think this is outside the scope of this PR.

qiskit/providers/aer/pulse/controllers/pulse_controller.py
DanPuzzuoli: True :)

DanPuzzuoli: Actually I'm still going to keep a check for whether it `is not None`

DanPuzzuoli:
@nkanazawa1989 

I've changed:
- Moved the frequency overriding in the RHS dict to its own function `setup_rhs_dict_freqs`. I realized as well that with the previous commit I had completely forgotten that the frequencies in schedule config would just be specifying the `DriveChannel` frequencies (so ordering isn't an issue), and that I needed to call the `calculate_channel_frequencies` method of the `system_model` to compute the control channel frequencies as well (which will put them in whatever order is required by the model, so that's good!). So, I've made `setup_rhs_dict_freqs` also take in a third argument `calculate_channel_frequencies` so that the logic can have access to this function.
- I've added a direct test of this function in `test.pulse.controllers.test_pulse_controller` - making sure that it works when there is no `qubit_lo_freq` key in the exp dictionary, and making sure it works when there is.
- Totally unrelated: I noticed there were some test files without copyright strings, so I added those.

nkanazawa1989 (approved):
Thanks for reviewing guide! I think this is good to go.

vvilpas (approved):

PR #1186, opened 2021-03-18
Improve handling of basis gates and configuration options

chriseclectic:
### Summary

* Updates how simulator options deal with option values belonging to the simulator configuration, properties and defaults.
* Improves how basis gates are reported in `backend.configuration()`

This is to address some of the issues discussed in #1181 

### Details and comments

*UPDATED:* Updates simulators to store separate option dictionaries for set options belonging to configuration, properties, and defaults, rather than a full copy of those objects with new values inserted. Now when the accessor methods are called (ie. `backend.configuration()`) a copy of these structures is generated and any custom option values are inserted when it is returned. In the specific case of basis gates, the value inserted will depend on any custom set values, the init config, the simulation method, and the noise model.

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -93,9 +93,13 @@ class AerBackend(BaseBackend, ABC):
 
         # Custom configuration, properties, and pulse defaults which will store
         # any configured modifications to the base simulator values.
+        # Since this copies the whole config/properties/defaults object when
+        # setting a custom value the `_custom_options` set is used to store
+        # which keys have custom values.
         self._custom_configuration = None
         self._custom_properties = None
         self._custom_defaults = None
+        self._custom_options = set()
 
         # Set available methods
         self._available_methods = [] if available_methods is None else available_methods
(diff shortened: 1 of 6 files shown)


yaelbh:
The code is still too difficult to maintain. I recommend to stick with the clean description of the logic in https://github.com/Qiskit/qiskit-aer/pull/1181#issuecomment-801331150. To this end:
- The final set of basis gates should not be stored, but recalculated every time is it needed, in a public function named `calculate_basis_gates`. `AerBackend.configuration` will call `calculate_basis_gates` to obtain the basis gates of the returned configuration (the current logic that's now inside `AerBackend.configuration`, where custom instructions are added to the set of basis gates, will exist only in `calculate_basis_gates`). `_set_option` will only set options, without applying any logic (if possible).
- I'm not sure that `options` and `custom_options` have to be separate (well, one of them is a dictionary and the other one is only a list of keys, but still). Their co-existence is confusing and may be inconsistent. For example, regular options can be removed by setting them to `None` (`AerBacked._set_option`), while setting basis gates to `None` does not seem to remove the basis gates key from `custom_options`.
- `set_options` should be abstract and called from `AerBackend.__init__`.

chriseclectic:
I don't want to add more public functions to API since that API is specified by base backend. In this case the way to access basis gates is `backend.configuration().basis_gates`, and this is needed for the transpiler and other parts of qiskit. Hence `.configuration()` would need to calculate the basis gates each time its called, which I partially did in this PR. I could make the other function I added here to set basis gates also only get called when calling `.configuration` rather than when options are set. But we would still need something to store a set basis gates option separately to the stored configuration so you don't override any backend basis gates. I'll have to think about that last point some more.

chriseclectic:
Closing since this has been added to #1099

PR #1188, opened 2021-03-18
Fixed bug in apply_kraus  in matrix_product_state

merav-aharoni:
### Summary
Fixed a bug in apply_kraus in matrix_product_state. The bug was seen in QFT - see issue #991.

### Details and comments
There was a mistaken call to an internal function using `qubits` instead of `internal_qubits`. Also added a test with Kraus noise on QFT.


merav-aharoni:
Ready for review - very small fix + test, so review should be very short.

test/terra/backends/qasm_simulator/qasm_noise.py
```diff
@@ -148,3 +150,26 @@ def test_kraus_gate_noise(self):
...
+    BACKEND_OPTS = {}
+
+    def test_kraus_gate_noise_on_QFT(self):
+        """Test Kraus noise on a QFT circuit"""
+
+        shots = 1000
+        noise_models = ref_kraus_noise.kraus_gate_error_noise_models_full()
+
+        for noise_model in noise_models:
+            circuit = QFT(3)
```
yaelbh, line 167: It's better not to depend on the external implementation of QFT, which may change. A small circuit that manifests the bug is also more readable to the developer.

test/terra/backends/test_qasm_simulator_matrix_product_state.py
```diff
@@ -71,6 +72,7 @@ class TestQasmMatrixProductStateSimulator(
         QasmPauliNoiseTests,
         QasmResetNoiseTests,
         QasmKrausNoiseTests,
+        QasmKrausNoiseOnQFTTests,
```
yaelbh, line 75: Why not test the other simulation methods with the new test?

test/terra/reference/ref_kraus_noise.py
```diff
@@ -45,7 +45,7 @@ def kraus_gate_error_circuits():
     circuit.barrier(qr)
     circuit.measure(qr, cr)
     circuits.append(circuit)
-
+
```
yaelbh, line 48: Extra space (for some reason we don't run lint on the test directory in Aer)

test/terra/backends/qasm_simulator/qasm_noise.py
```diff
@@ -148,3 +150,26 @@ def test_kraus_gate_noise(self):
...
+    SIMULATOR = QasmSimulator()
+    BACKEND_OPTS = {}
+
+    def test_kraus_gate_noise_on_QFT(self):
+        """Test Kraus noise on a QFT circuit"""
+
+        shots = 1000
+        noise_models = ref_kraus_noise.kraus_gate_error_noise_models_full()
+
+        for noise_model in noise_models:
```
yaelbh, line 166: Only one model is returned by `kraus_gate_error_noise_models_full`

test/terra/backends/qasm_simulator/qasm_noise.py
```diff
@@ -148,3 +150,26 @@ def test_kraus_gate_noise(self):
...
+            circuit = QFT(3)
+            ref_target = qi.Statevector(circuit).data
+            circuit.save_statevector()
+            job = execute([circuit], self.SIMULATOR, shots=shots,
```
yaelbh, line 170: The last two lines are done again and again for every noise model (which is currently in fact one model)

test/terra/backends/qasm_simulator/qasm_noise.py
```diff
@@ -148,3 +150,26 @@ def test_kraus_gate_noise(self):
...
+                          noise_model=noise_model,
+                          **self.BACKEND_OPTS)
+            result = job.result()
+            self.assertSuccess(result)
+            self.compare_statevector(result, [circuit], [ref_target])
```
yaelbh, line 175: I don't understand.
1. How comes that the MPS simulator, which is a counts simulator, has a statevector as a result? (the function `compare_statevector` calls `result.get_statevector`, I expect this call to fail).
2. Is the statevector at the end of a noisy simulation expected to be the same as in a noiseless one?

yaelbh (changes requested):

test/terra/backends/qasm_simulator/qasm_noise.py
merav-aharoni: I actually liked the idea of using QFT to reproduce the bug. I saw other tests that use QFT, so thought it was legitimate. In addition, I didn't find it easy to construct a small circuit that failed.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -820,7 +820,7 @@ void MPS::apply_kraus(const reg_t &qubits,
                    const std::vector<cmatrix_t> &kmats,
                    RngEngine &rng) {
   reg_t internal_qubits = get_internal_qubits(qubits);
-  apply_kraus_internal(qubits, kmats, rng);
+  apply_kraus_internal(internal_qubits, kmats, rng);
```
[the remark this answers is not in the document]
merav-aharoni: In principle, I agree with you. Curiously, I could not see a warning for this case. I will use this opportunity to address a couple of other warnings I see in mps.

test/terra/backends/test_qasm_simulator_matrix_product_state.py
merav-aharoni: I wasn't sure it is worthwhile to check the other methods. I will check which methods I can add.

test/terra/backends/qasm_simulator/qasm_noise.py
yaelbh: ok

test/terra/backends/qasm_simulator/qasm_noise.py
merav-aharoni: You are correct, I just did it this way to be consistent with the other tests. Do you think I should turn it into a single noise_model?

test/terra/backends/qasm_simulator/qasm_noise.py
merav-aharoni: I changed the structure of the test - see comment in the next box.

test/terra/backends/qasm_simulator/qasm_noise.py
merav-aharoni: This comment was very helpful - I actually was running a model without noise in this test. I now corrected the model and used hard-coded reference values. Thanks!

test/terra/backends/qasm_simulator/qasm_noise.py
yaelbh: I think so

test/terra/backends/qasm_simulator/qasm_noise.py
yaelbh: Great, it looks fine now

chriseclectic (approved):

PR #1189, opened 2021-03-20
remove aqua from benchmark

hhorii:
### Summary

This PR removes aqua from benchmark in test

### Details and comments

Aqua is now being refactored and VQE and UCCSD will be in qiskit-nature.
However, their new versions are under development.
Until they are released, benchmarks do not include VQE applications.


mtreinish:
I would say lets keep the benchmarks in place until qiskit-nature is released, aqua will continue to work for several months as is, and we can replace the benchmarks at some point after the release with a qiskit-nature based versions (and just bump the version of the benchmark then). That way we don't lose any coverage, especially right up until the release.

hhorii:
@mtreinish OK. Please let me leave this as a placeholder for future change after qiskit-nature is released.

hhorii:
https://hhorii.github.io/qiskit-aer/html/#/

test/asv.linux.conf.json
```diff
@@ -34,20 +34,20 @@
...
         "python -c \"import shutil; shutil.rmtree('{build_dir}/qiskit_aer.egg-info', True)\"",
         "pip install git+https://github.com/Qiskit/qiskit-terra",
-        "pip install git+https://github.com/Qiskit/qiskit-aqua",
+        "pip install qiskit_nature",
         "python -mpip install {wheel_file}"
     ],
     "uninstall_command": [
         "return-code=any python -mpip uninstall -y qiskit-terra",
-        "return-code=any python -mpip uninstall -y qiskit-aqua",
+        "pip uninstall qiskit_nature",
```
mtreinish, line 42: maybe add the `-y` here to not fail if it's not installed for some reason. Also maybe instead of doing this we can use the `matrix` kwarg and fix the version: https://asv.readthedocs.io/en/stable/asv.conf.json.html?highlight=matrix#matrix

test/asv.linux.cuda.conf.json
```diff
@@ -34,20 +34,21 @@
         "python -c \"import shutil; shutil.rmtree('{build_dir}/qiskit', True)\"",
         "python -c \"import shutil; shutil.rmtree('{build_dir}/qiskit_aer.egg-info', True)\"",
         "pip install git+https://github.com/Qiskit/qiskit-terra",
-        "pip install git+https://github.com/Qiskit/qiskit-aqua",
+        "pip install qiskit_nature",
```
mtreinish, line 37: Same here as in the cpu version

test/benchmark/simulator_benchmark.py
```diff
@@ -167,49 +163,12 @@ def add_expval(base, num_terms):
...
-                return QOBJS[(runtime, app, measure, measure_count, qubit)]
-            else:
-                runtime_circuit = circuit
-                if (simulator, app, measure, measure_count, qubit) not in QOBJS:
-                    QOBJS[(simulator, app, measure, measure_count, qubit)] = assemble(runtime_circuit, simulator, shots=1)
-                return QOBJS[(simulator, app, measure, measure_count, qubit)]
+        shots = measure_count if measure == self.MEASUREMENT_SAMPLING else 1
+        
+        if (simulator, app, measure, measure_count, qubit) not in QOBJS:
+            QOBJS[(runtime, app, measure, measure_count, qubit)] = assemble(transpile(circuit, simulator), simulator, shots=shots)
```
mtreinish, line 169: Do we want to be passing in qobjs manually to anything here as passing a qobj to `backend.run()` will emit a `DeprecationWarning`

test/benchmark/vqe_application.py
```diff
@@ -40,34 +33,82 @@ def __init__(self,
...
-        result = mgse.compute_energy(cb_create_solver)
-            
+    def _run_uccsd_vqe(self, mol_string, method, simulator, threads):
+        
+        from qiskit_nature.circuit.library import HartreeFock, UCCSD
+        from qiskit_nature.converters.second_quantization import QubitConverter
+        from qiskit_nature.drivers import UnitsType, Molecule
+        from qiskit_nature.drivers.second_quantization.pyscfd import PySCFDriver
+        from qiskit_nature.mappers.second_quantization import JordanWignerMapper, ParityMapper
+        from qiskit_nature.problems.second_quantization import ElectronicStructureProblem
```
mtreinish, line 43: Is there a reason these are runtime imports?

test/benchmark/vqe_application.py
```diff
@@ -40,34 +33,82 @@ def __init__(self,
...
+        optimizer = SLSQP(maxiter=5000)
+        
+        from qiskit.utils import QuantumInstance
+        
+        quantum_instance = QuantumInstance(backend=simulator)
+        
+        # import logging
+        # logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s',
+        #                     level=logging.DEBUG,
+        #                     datefmt='%Y-%m-%d %H:%M:%S')
```
mtreinish, line 60: Maybe delete this?

test/benchmark/vqe_application.py
```diff
@@ -40,34 +33,82 @@ def __init__(self,
...
+                                   max_evals_grouped=max_evals_grouped)
+        
+        from qiskit_nature.algorithms import GroundStateEigensolver
+        
+        calc = GroundStateEigensolver(qubit_converter, vqe_solver)
+        
+        res = calc.solve(es_problem)
+        
     def time_statevector(self, mol_name):
+        from qiskit.providers.aer import AerSimulator
```
mtreinish, line 76: We definitely don't want to be running imports in measured methods since it will skew the runtime especially if multiple trials are done.

test/benchmark/vqe_application.py
```diff
@@ -40,34 +33,82 @@ def __init__(self,
...
         threads = multiprocessing.cpu_count()
         mol_string = self.mol_strings[mol_name][0]
         qubit = self.mol_strings[mol_name][1]
-        self._run_uccsd_vqe(mol_string, 'statevector', threads)
+        self._run_uccsd_vqe(mol_string, 'statevector_cpu', AerSimulator(device='CPU'), threads)
 
     def time_statevector_gpu(self, mol_name):
+        from qiskit.providers.aer import AerSimulator
```
mtreinish, line 83: Same thing here about runtime imports

test/benchmark/vqe_application.py
```diff
@@ -40,34 +33,82 @@ def __init__(self,
...
+        
+        res = calc.solve(es_problem)
+        
     def time_statevector(self, mol_name):
+        from qiskit.providers.aer import AerSimulator
         threads = multiprocessing.cpu_count()
         mol_string = self.mol_strings[mol_name][0]
         qubit = self.mol_strings[mol_name][1]
-        self._run_uccsd_vqe(mol_string, 'statevector', threads)
+        self._run_uccsd_vqe(mol_string, 'statevector_cpu', AerSimulator(device='CPU'), threads)
```
mtreinish, line 80: Do we want to be constructing backend objects in the timed portion here? It feels like we could do that in `setup()` and leave this to just collect the runtime measurement. We could also add a dedicated construction microbenchmark.

test/benchmark/vqe_application.py
```diff
@@ -40,34 +33,82 @@ def __init__(self,
...
+                    print ('{0}'.format(e))
+                except:
+                    import traceback
+                    traceback.print_exc(file=sys.stderr)
+                    print ('unknown error')
+
+if __name__ == "__main__":
+    benrhmarks = [ UCCSDBenchmarkSuite() ]
+    for benrhmark in benrhmarks:
+        benrhmark.run_manual()
```
mtreinish, line 114:
```suggestion
    benchmarks = [ UCCSDBenchmarkSuite() ]
    for benchmark in benchmarks:
        benchmark.run_manual()
```

test/benchmark/vqe_application.py
```diff
@@ -40,34 +33,82 @@ def __init__(self,
...
+                print (f'{self.__name__},uccsd,{method},{mol_string},', end="")
+                try:
+                    elapsed = run_with_timeout(self, method, mol_string)
+                    print ('{0}'.format(elapsed))
+                except ValueError as e:
+                    print ('{0}'.format(e))
+                except:
+                    import traceback
+                    traceback.print_exc(file=sys.stderr)
+                    print ('unknown error')
```
mtreinish, line 109: Do we actually need this? asv should give us the tools to run benchmarks this seems a bit unnecessary.

mtreinish (changes requested):

test/benchmark/vqe_application.py
hhorii: Let me keep this to run benchmark casually and get a csv format.

test/benchmark/vqe_application.py
hhorii: I'm thinking that vqe tests are end-to-end scenarios as possible, which include import, solver construction, and simulator generation.

test/asv.linux.conf.json
```diff
@@ -34,20 +34,20 @@
...
         "pip install git+https://github.com/Qiskit/qiskit-terra",
-        "pip install git+https://github.com/Qiskit/qiskit-aqua",
+        "pip install qiskit_nature",
         "python -mpip install {wheel_file}"
     ],
     "uninstall_command": [
         "return-code=any python -mpip uninstall -y qiskit-terra",
-        "return-code=any python -mpip uninstall -y qiskit-aqua",
+        "return-code=any python -mpip uninstall -y qiskit_nature",
         "return-code=any python -mpip uninstall -y {project}"
```
jakelishman, line 43: Just for consistency, it may be easier as `qiskit-nature` (with a hyphen, not an underscore).

test/asv.linux.conf.json
```diff
@@ -34,20 +34,20 @@
         "python -c \"import shutil; shutil.rmtree('{build_dir}/qiskit', True)\"",
         "python -c \"import shutil; shutil.rmtree('{build_dir}/qiskit_aer.egg-info', True)\"",
         "pip install git+https://github.com/Qiskit/qiskit-terra",
-        "pip install git+https://github.com/Qiskit/qiskit-aqua",
+        "pip install qiskit_nature",
```
jakelishman, line 37: How come we install Terra (and previously Aqua) from GitHub, but Nature from pip?

test/benchmark/simulator_benchmark.py
```diff
@@ -118,28 +114,28 @@ def __init__(self,
             self.noise_models[self.NOISE_DEPOLARIZING] = noise_model
 
         if self.RUNTIME_STATEVECTOR_CPU in runtime_names:
-            self.simulators[self.RUNTIME_STATEVECTOR_CPU] = SIMULATOR
-            self.backend_options_list[self.RUNTIME_STATEVECTOR_CPU] = { 'method': self.RUNTIME_STATEVECTOR_CPU }
+            self.simulators[self.RUNTIME_STATEVECTOR_CPU] = AerSimulator(method='statevector', device='CPU')
+            self.backend_options_list[self.RUNTIME_STATEVECTOR_CPU] = { 'enable_truncation': False}
```
jakelishman, line 118: I think with all these removals, the `SIMULATOR` module constant can now be removed, since it's only used by `all_simulators` which also isn't accessed?

test/benchmark/simulator_benchmark.py
```diff
@@ -225,7 +184,7 @@ def _run(self, runtime, app, measure, measure_count, noise_name, qubit):
             raise ValueError('no qobj: measure={0}:{1}, qubit={2}'.format(measure, measure_count, qubit))
 
         start = time()
-        result = simulator.run(qobj, noise_model=noise_model, **backend_options).result()
+        result = simulator._run(qobj, noise_model=noise_model, **backend_options)
```
jakelishman, line 187: How come we need to change this?  It seems like it's skipping some normal set-up and wrapping code from the timed portion.

test/asv.linux.conf.json
hhorii: sometimes qiskit-aer has dependency from the main of terra and will not work well with the pip-latest version.

test/benchmark/simulator_benchmark.py
hhorii: I wanted to evaluate only C++ implementation as possible by generating `qobj` before running simulation.

test/benchmark/default_simulator.py
```diff
@@ -0,0 +1,281 @@
...
+        start_ts = time()
+        result = self._run(circuit)
+        end_ts = time()
+        # benchmark end
+
+        if result.success:
+            return end_ts - start_ts
+        else:
+            raise ValueError(result.status)
+
```
jakelishman, line 80: This is a pretty large subversion of how `asv` operates, and performance timing with `time.time` is not reliable at all; it's subject to the system clock, and there's no guarantee that it'll have a granularity of better than a second (though in practice it usually will).  We should use proper `time_*` methods and setup functions - you could separate out each one into its own class, and inherit a `setup` function that calls a method `circuit` which you define in each class separately.

I don't know if you can tag individual class methods with `setup` functions in asv, but if you can, we may be able to do it in a single class compactly too.  Docs: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#setup-and-teardown-functions

test/asv.linux.conf.json
```diff
@@ -155,15 +147,15 @@
...
+    "html_dir": ".asv/html",
 
     // The number of characters to retain in the commit hashes.
     // "hash_length": 8,
 
     // `asv` will cache results of the recent builds in each
     // environment, making them faster to install next time.  This is
     // the number of builds to keep, per environment.
-    // "build_cache_size": 2,
+    "build_cache_size": 100000
```
mtreinish, line 158: Heh, that's a lot of builds

test/asv.linux.conf.json
```diff
@@ -33,27 +33,19 @@
     "install_command": [
         "python -c \"import shutil; shutil.rmtree('{build_dir}/qiskit', True)\"",
         "python -c \"import shutil; shutil.rmtree('{build_dir}/qiskit_aer.egg-info', True)\"",
-        "pip install git+https://github.com/Qiskit/qiskit-terra",
-        "pip install git+https://github.com/Qiskit/qiskit-aqua",
+        "pip install -U qiskit-terra",
```
mtreinish, line 36: I'm wondering if we want to use `matrix` here to list qiskit-terra at a specific version so the benchmark results are tied to the qiskit-terra version (so if we bump the terra version in the matrix it shows as different from earlier runs).

test/benchmark/default_simulator.py
```diff
@@ -0,0 +1,228 @@
...
+    def simulator(self):
+        if self._simulator:
+            return self._simulator
+        try:
+            from qiskit.providers.aer import AerSimulator
+            self._simulator = AerSimulator()
+        except ImportError:
+            from qiskit.providers.aer import QasmSimulator
+            self._simulator = QasmSimulator()
+        return self._simulator
```
mtreinish, line 40: Are you still supporting comparisons from the releases prior to `AerSimulator`?

test/benchmark/default_simulator.py
```diff
@@ -0,0 +1,228 @@
...
+
+class _Base(ABC):
+
+    def __init__(self,
+                 qubits=[5, 15, 25]):
+
+        self._simulator = None
+        self.qubits = qubits
+        self.params = (qubits)
+        self.param_names = ["qubits"]
```
mtreinish, line 29:
```suggestion
        self.param_names = ["Number of qubits"]
```

mtreinish (commented):
This LGTM, just a few small inline questions/comments but otherwise I'm fine with this.

test/asv.linux.conf.json
hhorii: good catch :-)

PR #1194, opened 2021-03-24
Add AerSimulator backend

chriseclectic:
### Summary

Adds `AerSimulator` backend that in the future will replace the QasmSimulator, StatevectorSimulator and UnitarySimulator backends.

The `AerProvider` has been updated to include AerSimulator backends for all available methods and devices.

#### C++ Controller

In the C++ code this adds a new `aer_controller` class which is a combination of the previous base controller and qasm controller. The idea is in the future we do not need subclasses as all methods go through the same controller and the qasm_controller, unitary_controller and statevector_controller can be deprecated.

#### New Backend

The main differences between the new `AerSimulator` backend and the `QasmSimulator` include:

* GPU simulation is enabled by using `device='GPU'` option for supported methods. eg `AerSimulator(method='statevector', device='GPU')`
* "unitary" and "superop" are added as simulation methods for simulating circuits without measurements.
* Getting the final state of the simulator must be done using the `save_state` or another `save` instruction, no final statevector or unitary will be returned automatically without the save instruction. 

#### Tests

Started refactoring tests to improve running common tests on all supported methods and devices.

Currently some of these tests are failing
- Unitary method with cache blocking on multi-controlled gates and save unitary / save state


vvilpas (commented):
Looks good. Found a typo that can be dangerous.

Issue #1193, opened 2021-03-24
Incorrect results when using cache_blocking statevector simulator


chriseclectic:
### Informations

- **Qiskit Aer version**: 0.7
- **Python version**: 3.7
- **Operating system**: MacOS

### What is the current behavior?

Running the test teleport circuit on the statevector simulator with cache blocking enabled gives the wrong result.

### Steps to reproduce the problem

```python
import qiskit
from qiskit.providers.aer import QasmSimulator
from qiskit.visualization import plot_histogram

def teleport_circuit():
    """Testing a circuit originated in the teleportation algorithm"""

    # Classic 3-qubit teleportation
    qr = QuantumRegister(3)
    c0 = ClassicalRegister(1)
    c1 = ClassicalRegister(1)
    c2 = ClassicalRegister(1)
    # Compiles to creg order [c2, c1, c0]
    circuit = QuantumCircuit(qr, c0, c1, c2)

    # Teleport the |0> state from qr[0] to qr[2]
    circuit.h(qr[1])
    circuit.cx(qr[1], qr[2])
    circuit.barrier(qr)
    circuit.cx(qr[0], qr[1])
    circuit.h(qr[0])
    circuit.measure(qr[0], c0[0])
    circuit.measure(qr[1], c1[0])
    circuit.z(qr[2]).c_if(c0, 1)
    circuit.x(qr[2]).c_if(c1, 1)
    circuit.measure(qr[2], c2[0])
    
    return circuit

method = 'statevector'
sim1 = QasmSimulator(method=method)
sim2 = QasmSimulator(method=method, blocking_qubits=2)

qc = teleport_circuit()
a = qiskit.execute(qc, sim1).result().get_counts(0)
b = qiskit.execute(qc, sim2).result().get_counts(0)

plot_histogram([a, b])
```

Returns

![Screen Shot 2021-03-24 at 3 48 15 PM](https://user-images.githubusercontent.com/2235104/112374362-61f13180-8cb8-11eb-8747-de8683b71e67.png)

### What is the expected behavior?

The returned counts should be equal (up to sampling fluctuations)

### Suggested solutions

I'm not sure, this might only be an issue with circuits containing conditional instructions. The above example returns the correct result on the density matrix simulator with cache blocking enabled.

doichanj:
[...]

PR #1197, opened 2021-03-29
Fixing cache blocking transpiler for multi-chunk measure, save and kraus operations

doichanj:
### Summary
This PR fixes cache blocking transpiler bug if the circuit has measure operations in the middle of operation sequence (in case sampling measure can not be applied) 
https://github.com/Qiskit/qiskit-aer/issues/1193

### Details and comments
The measure operations (that can not be merged in sampling) is now applied without restoring qubit map caused by chunk swaps. 
Also this PR includes fix for save amplitudes and kraus operator. 
Kraus operator is transpiled when target qubits can be fit in chunk otherwise returns error.


chriseclectic:
This looks to have fixed the test failures with statevector, but after applying these changes in #1194 it looks like there are still some failures for the unitary chunk simulator for save unitary and multi-controlled gates.

chriseclectic (approved):
Rather than fix the unitary cache blocking I think it makes sense to just disable it for unitary simulation, since MPI based distributed simulation isn't practical for that method anyway. I will open a separate PR to do this

PR #1201, opened 2021-03-31
Fix various compiler warnings

danbev:
### Summary
This commit removes unused variables to avoid some compiler warnings.

### Details and comments
Reduces the number of compiler warnings generated.


chriseclectic:
Thanks for making the PRs to fix many of the compiler warnings @danbev. I've combined several of the commits from your other PRs into this one.

danbev:
> I've combined several of the commits from your other PRs into this one.

Great, thanks for doing that!

src/simulators/state_chunk.hpp
```diff
@@ -876,7 +875,10 @@ uint_t StateChunk<state_t>::mapped_index(const uint_t idx)
 template <class state_t>
 void StateChunk<state_t>::apply_chunk_swap(const reg_t &qubits)
 {
-  uint_t q0,q1,nLarge;
+#pragma GCC diagnostic ignored "-Wunused-but-set-variable"
+  uint_t nLarge;
+#pragma GCC diagnostic pop
```
vvilpas, line 880: Cannot we just initialise `nLarge` to 1? Otherwise, as we support other compilers maybe we would need here some sort of ifdefs to address them as well.

vvilpas (commented):
@danbev Thanks for taking the time to remove some warnings!

src/simulators/state_chunk.hpp
danbev: That would be much better, I'll update shortly. Thanks

vvilpas (approved):

PR #1204, opened 2021-03-31
Fix AerSimulator options

chriseclectic:
### Summary

* Adds missing `seed_simulator` option to `AerSimulator` default options
* Fixes method name returned in AerController metadata to not include the device as is done with QasmController.
* Adds some basic tests for method, device, and seed_simulator options

### DIFF
--- a/test/terra/backends/aer_simulator/test_options.py
+++ b/test/terra/backends/aer_simulator/test_options.py
@@ -1,0 +1,92 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2018, 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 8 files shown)


test/terra/backends/aer_simulator/test_options.py
```diff
@@ -0,0 +1,92 @@
...
+"""
+
+from ddt import ddt
+from qiskit import QuantumCircuit, transpile
+from .aer_simulator_test_case import (
+    AerSimulatorTestCase, supported_methods)
+
+
+@ddt
+class TestOptions(AerSimulatorTestCase):
```
mtreinish, line 23: Do we have variants of this test for `set_options` and kwargs on `run()`?

mtreinish (dismissed):

test/terra/backends/aer_simulator/test_options.py
chriseclectic: I would just add those as separate test functions. I added class and run tests for seed_simulator, and we can add more to it later.

Issue #1205, opened 2021-04-01
MPS: invariant that sum of probabilities = 1 is not preserved 


merav-aharoni:
### Informations

- **Qiskit Aer version**: 0.8.0
- **Python version**: 3.8
- **Operating system**: Linux

### What is the current behavior?
In some circuits with Kraus noise, if we print the sum of the probabilities of the qubits during execution, we sometimes get !=1.

### Steps to reproduce the problem
```
from qiskit import *
from qiskit import execute
from qiskit.circuit.library import QFT
from qiskit.providers.aer import QasmSimulator
from qiskit.providers.aer.noise import NoiseModel, amplitude_damping_error
import random
error1 = amplitude_damping_error(0.2)
noise_model = NoiseModel()
noise_model.add_all_qubit_quantum_error(error1, ['u2'])
noise_model.add_all_qubit_quantum_error(error1.tensor(error1), ['cx'])
backend = QasmSimulator(method='matrix_product_state', noise_model=noise_model)
n = 4
qr = QuantumRegister(n)
cr = ClassicalRegister(n)
circ = QuantumCircuit(qr, cr)
circ.cx(0, 3)
circ.u1(0.5, 2)
circ.cx(0, 3)
circ.u2(0.5, 0.1, 0)
circ.cx(0, 1)
circ.cx(1, 2)
circ.measure_all()
job = execute(circ, backend, shots=1)
print(job.result().results)

```
If I print the sum of the probabilities after every operation, sometimes I don't get 1.

### What is the expected behavior?
Should always get 1. Note that this bug is not view externally, but I expect it eventually will lead to failures.

Issue #1184, opened 2021-03-18
MPS segfaults with noise model and measurement


1ucian0:
### Informations

- **Qiskit Aer version**: 0.7.6
- **Python version**: 3.8
- **Operating system**: MacOs

### Steps to reproduce the problem
```python
from qiskit import execute
from qiskit.test.mock import FakeManhattan
from qiskit.providers.aer import QasmSimulator
from qiskit.circuit.library import GraphState
from numpy import zeros 


n_qubits = 12
A = zeros((n_qubits, n_qubits))
rows = [0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9, 10, 10, 11, 11]
cols = [1, 5, 0, 2, 1, 3, 2, 4, 3, 6, 0, 7, 4, 11, 5, 8, 7, 9, 8, 10, 9, 11, 6, 10]
A[rows, cols] = 1

qc = GraphState(A)

qc.measure_all()

simulator = QasmSimulator.from_backend(FakeManhattan(), method='matrix_product_state')

execute(qc, simulator, shots=1024, optimization_level=0, seed_simulator=0, seed_transpiler=0)
```
```
lda must be >= MAX(M,1): lda=0 M=0.ldc must be >= MAX(M,1): ldc=0 M=0. BLAS error: Parameter number 9 passed to cblas_zgemm had an invalid value
Segmentation fault: 11
/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 5 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

Notice that does not segfaults if the `measure_all` is commented out.

yaelbh:
Looks related to #991

merav-aharoni:
@1ucian0, I made a fix for issue #991  (the bug you found with MPS on QFT) in PR #1188 . I would like to check if the fix is relevant for this bug as well. However, I am unable to run this example on master (only on stable). Can you update the test case to run on master?

1ucian0:
Could it be that is already solved in master?

chriseclectic:
@1ucian0 This should be fixed by #1188 which was just merged into master. You can reopen if you find more bugs

merav-aharoni:
[...]

PR #1206, opened 2021-04-01
Bug fix for MPS::apply_kraus

merav-aharoni:
### Summary
After applying kraus noise, we forgot to propagate the effect to the neighbors. This caused a bug that was observed internally - see issue #1205 .


merav-aharoni:
I didn't add a test for this bug, because it only manifests itself internally, when I add a sanity check. I don't have a test case where it eventually fails.

merav-aharoni:
This fix also corrects issue #1184.
Ready for review.

vvilpas (approved):

PR #1207, opened 2021-04-01
Prepare 0.8 release

chriseclectic:


releasenotes/notes/0.8-release-179e83619783737f.yaml
```diff
@@ -0,0 +1,66 @@
...
+    It is now possible to build qiskit-aer with CUDA enabled in Windows.
+    See the
+    `contributing guide<https://github.com/Qiskit/qiskit-aer/blob/master/CONTRIBUTING.md#building-with-gpu-support>`_
+    for instructions on building from source with GPU support.
+  - |
+    Some *Python* packages needed at C++ compilaton step were installed
+    outside the regular *setuptools* circuit, and it was not possible to
+    avoid their installation by any option. Now it is possible to avoid
+    automatic installation of those packages by setting the environment
+    variable *DISABLE_DEPENDENCY_INSTALL* to *1* or *ON*.
```
mtreinish, line 27:
```suggestion
    When building the qiskit-aer Python extension from source several build
    dependencies need to be pre-installed to enable C++ compilation. As a
    user convenience when building the extension any of these build
    dependencies which were missing would be automatically installed using
    ``pip`` prior to the normal ``setuptools`` installation steps, however it was
    previously was not possible to avoid this automatic installation. To solve
    this issue a new environment variable ``DISABLE_DEPENDENCY_INSTALL``
    has been added. If it is set to ``1`` or ``ON`` when building the python
    extension from source this will disable the automatic installation of these
    missing build dependencies.
```

mtreinish (commented):
Mostly LGTM, just some small issues inline

mtreinish (approved):

PR #1213, opened 2021-04-07
Disable GPU methods if no device is available

vvilpas:
### Summary

GPU methods are disabled if no device is available. An exception is thrown early in the process if no device is available.

Fixes the deploy failure we had in latest 0.8.0 release.


vvilpas:
I have disabled the GPU methods, when no device is available, throwing an exception in the execute try/catch block. 
Given that deploy test passes, @hhorii @doichanj do you see any issues/problems with this approach?

doichanj:
No, I agree with this fix.

vvilpas:
Ok. Seems it is working. Reverting `deploy.yml` to allow merging.

chriseclectic (approved):

PR #1216, opened 2021-04-07
Suppress warnings from build_tests target

danbev:
### Summary
Suppress warnings from build_tests target

### Details and comments
Currently, there are a number of warnings generated by the build_tests
target.

This commit removes two std::move calls where these would prevent Named
Return Value Optimization (NRVO) taking place (where the compiler can
omit any copy/move and construct the object directly on the caller
stack). Also, two index types are changed from int to size_t.


chriseclectic (approved):

chriseclectic:
Closing this PR since I have added the commit to #1201

PR #1220, opened 2021-04-11
Support set_matrix_product_state

merav-aharoni:
### Summary
Add support for the instruction set_matrix_product_state.

### Details and comments
Also supports set_statevector for the mps simulation method.


src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1705,13 +1705,39 @@ mps_container_t MPS::move_to_mps_container() {
...
+  std::iota(qubit_ordering_.order_.begin(), qubit_ordering_.order_.end(), 0);
+
+  qubit_ordering_.location_.clear();
+  qubit_ordering_.location_.resize(num_qubits);
+  std::iota(qubit_ordering_.location_.begin(), qubit_ordering_.location_.end(), 0);
+
+  // initialize values from mps_container_t
+  for (uint_t i=0; i<num_qubits; i++) {
+    MPS_Tensor next_tensor(mps.first[i].first, mps.first[i].second);
+    q_reg_[i] = next_tensor;
```
vvilpas, line 1734: If `data_t` in `MPS_Tensor` holds large `cmatrix_t`s, it is worth it to implement movement constructor and movement assignment operator for `MPS_Tensor`, to avoid extra copies of large matrices.

vvilpas (commented):
I have a comment that may be important depending on size of matrices involved. Otherwise looks good to me.

qiskit/providers/aer/backends/qasm_simulator.py
```diff
@@ -290,7 +290,7 @@ class QasmSimulator(AerBackend):
         'save_amplitudes', 'save_amplitudes_sq', 'save_state',
         'save_density_matrix', 'save_statevector', 'save_statevector_dict',
         'save_stabilizer', 'set_statevector', 'set_density_matrix',
-        'set_stabilizer'
+        'set_stabilizer', 'set_matrix_product_state'
```
chriseclectic, line 293: This shouldn't be added to the default instructions, since it is only supported on qasm simulator if you set the mps method.

The instruction also needs to be added to the mps method instructions in `aer_simulator.py`

test/terra/backends/qasm_simulator/qasm_set_state.py
```diff
@@ -120,3 +121,51 @@ def test_set_stabilizer(self, num_qubits):
             self.assertIn(save_label, data)
             value = qi.Clifford.from_dict(result.data(0)[save_label])
             self.assertEqual(value, target)
+
+class QasmSetMPSTests:
```
chriseclectic, line 125: Can you also add a test for the AerSimulator backend, you can see the other set state tests for how these should be run. It should looks something like:

```python
@supported_methods(['automatic', 'matrix_product_state'])
    def test_set_matrix_product_state(self, method, device):
        # test...
```

qiskit/providers/aer/library/set_instructions/set_matrix_product_state.py
```diff
@@ -0,0 +1,76 @@
...
+class SetMatrixProductState(Instruction):
+    """Set the matrix product state of the simulator"""
+
+    _directive = True
+
+    def __init__(self, mps_state):
+        """Create new instruction to set the matrix product state of the simulator.
+
+        Args:
+            mps_state (mps_container_t): a matrix product state.
```
chriseclectic, line 30: This is referring to a C++ type. It should probably just be `matrix_product_state`, and in the docstring describe the format or reference the save instruction documentation if it is described there

qiskit/providers/aer/library/set_instructions/set_matrix_product_state.py
```diff
@@ -0,0 +1,76 @@
...
+        .. note::
+
+            This set instruction must always be performed on the full width of
+            qubits in a circuit, otherwise an exception will be raised during
+            simulation.
+        """
+        super().__init__('set_matrix_product_state', len(mps_state[0]), 0, [mps_state])
+
+
+def set_matrix_product_state(self, mps_state):
```
chriseclectic, line 41:
```suggestion
def set_matrix_product_state(self, state):
```
We should keep arg names consistent across all these set instructions

qiskit/providers/aer/library/set_instructions/set_matrix_product_state.py
```diff
@@ -0,0 +1,76 @@
...
+from qiskit.extensions.exceptions import ExtensionError
+from ..default_qubits import default_qubits
+
+
+class SetMatrixProductState(Instruction):
+    """Set the matrix product state of the simulator"""
+
+    _directive = True
+
+    def __init__(self, mps_state):
```
chriseclectic, line 26:
```suggestion
    def __init__(self, state):
```

chriseclectic (changes requested):

qiskit/providers/aer/backends/qasm_simulator.py
merav-aharoni: Done

qiskit/providers/aer/library/set_instructions/set_matrix_product_state.py
merav-aharoni: Done

qiskit/providers/aer/library/set_instructions/set_matrix_product_state.py
merav-aharoni: Done.

qiskit/providers/aer/library/set_instructions/set_matrix_product_state.py
merav-aharoni: Done.

qiskit/providers/aer/library/set_instructions/set_matrix_product_state.py
merav-aharoni: Done.

test/terra/backends/qasm_simulator/qasm_set_state.py
merav-aharoni: @chriseclectic , I am not sure what to do here. Should I copy the test from qasm to `test/terra/backends/aer_simulator/instructions/test_set_state.py`?
Isn't this redundant?

chriseclectic (approved):
Test and python file updates LGTM
@vvilpas I'll leave any further comments on the C++ to you

test/terra/backends/qasm_simulator/qasm_set_state.py
merav-aharoni: Done.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: @vvilpas , please review the changes I made in `initialize_from_mps`.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1705,13 +1705,39 @@ mps_container_t MPS::move_to_mps_container() {
...
+  std::iota(qubit_ordering_.order_.begin(), qubit_ordering_.order_.end(), 0);
+
+  qubit_ordering_.location_.clear();
+  qubit_ordering_.location_.resize(num_qubits);
+  std::iota(qubit_ordering_.location_.begin(), qubit_ordering_.location_.end(), 0);
+
+  // initialize values from mps_container_t
+  for (uint_t i=0; i<num_qubits; i++) {
+    MPS_Tensor next_tensor(std::move(mps.first[i].first), std::move(mps.first[i].second));
+    q_reg_[i] = next_tensor;
```
vvilpas, line 1734: Good catch with also moving `mps` internals. The problem is that you cannot move those because `mps` is a `const mps_container_t &`. You would need to get rid of the `const`, assuming it is not used elsewhere. Not sure if that's possible as it comes from a const operation, etc. In addition you would need to create a constructor taking rvalues as params. If it is not possible to pass a `non-const mps_container_t ref`you can remove the `std::move`s in lines 1733 and 1737.

What we can avoid is copying in this line `q_reg_[i] = next_tensor;`
We need to add these constructors accepting rvalues (in matrix_product_state_tensor.hpp) :

```
  MPS_Tensor(MPS_Tensor&& rhs){
    data_ = std::move(rhs.data_);
  }

  MPS_Tensor& operator=(MPS_Tensor&& rhs){
    if (this != &rhs){
      data_ = std::move(rhs.data_);
    }
    return *this;
  }

// Next would only bee needed if we can pass a non-const mps _container_t ref
  MPS_Tensor(cmatrix_t&& data0, cmatrix_t&& data1){ 
    if (!data_.empty())
      data_.clear();
    data_.push_back(data0);
    data_.push_back(data1);
  }
```

And change:
```suggestion
    q_reg_[i] = std::move(next_tensor);
```

vvilpas (changes requested):

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: @vvilpas , thanks for explaining. I was not familiar with the `&&`. A few questions:
1. I still think it is better with the `std::move` on line 1737, since the lambda vectors may also be very big. Why don't we need this?
2. In the code you wrote for the constructors, should we have 
```rhs = nullptr``` after the assignment? as well as for data0, data1.
3. As you said,indeed we cannot change the mps_container_t to be non-const because it is part of the `op` structure. So I didn't include the third cot'r you suggested.

Comment (vvilpas) [reply src/simulators/matrix_product_state/matrix_product_state_internal.cpp]: No problem!
Regarding questions:
1) The problem is that `mps` is a `const &` so there won't be any actual movement. More or less it's like you mark it as movable by doing `std::move(mps.second[i])`, but as it is `const` the compiler refuse to move it and copies it instead.

2) When an object is moved basically you steal its resources and leave it in some undefined state. In `rhs` we "steal" `data_` which is a `std::vector`, which has its appropriate move constructor to handle the final state of `rhs.data_` (cleared vector), so there is no need to do anything with `rhs.data_`. If `rhs` would have a naked pointer to an array as `data_` member, we could do as you mention: `rhs.data_ = nullptr` so that `rhs` no longer controls `data_`. 
In the case of `data0` and `data1`, they are moved to the vector and the `matrix` move constructor takes care of handling their data pointers, etc.

File: src/simulators/matrix_product_state/matrix_product_state_internal.cpp [hunk 8]
```diff
@@ -1705,13 +1705,39 @@ mps_container_t MPS::move_to_mps_container() {
...
+  qubit_ordering_.location_.resize(num_qubits);
+  std::iota(qubit_ordering_.location_.begin(), qubit_ordering_.location_.end(), 0);
+
+  // initialize values from mps_container_t
+  for (uint_t i=0; i<num_qubits; i++) {
+    MPS_Tensor next_tensor(mps.first[i].first, mps.first[i].second);
+    q_reg_[i] = std::move(next_tensor);
+  }
+  for (uint_t i=0; i<num_qubits-1; i++) {
+    lambda_reg_[i] = std::move(mps.second[i]);
```
chriseclectic, line 1737:
```suggestion
    lambda_reg_[i] = mps.second[i];
```

chriseclectic (approved):

vvilpas (approved):

Issue #1196, opened 2021-03-26
Statevector save and snapshot not normalized for extended stabilzier method


chriseclectic:
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?

The snapshot and save statevector functionality of the extended stabilizer simulator returns an unnormalized statevector.

### Steps to reproduce the problem

```python
from qiskit import QuantumCircuit
from qiskit.providers.aer.extensions import *

qc = QuantumCircuit(3)
qc.h(0)
qc.cx(0, 1)
qc.t(0)
qc.cx(0, 2)
qc.snapshot_statevector('final')

method = 'extended_stabilizer'
sim = AerSimulator(method=method)
result = sim.run(qc, shots=1).result()
psi = result.data(0)['snapshots']['statevector']['final'][0]
norm = sum(abs(psi) ** 2)
print(norm)
```
The returned norm is not 1

### What is the expected behavior?

Even though this is an approximate method and involves norm sampling, it should still return a normalized state.

### Suggested solutions

I don't really understand the norm sampling function used to compute the statevector for the simulator, but maybe there is something that can be done to fix this there (Any ideas @gadial?)

paniash:
@chriseclectic When running the above snippet, it looks like there is no function `AerSimulator()`. Could you clarify?

or1426:
Given that we're already computing the overlaps with all 2^n computational basis states when computing the 'statevector' output I can't see much reason not to use this information to compute the norm of the state and then normalize it.

This would require a trivial change to the `Runner::statevector` method in ch_runner.hpp. I don't think this method is used anywhere else in the code so it shouldn't be a problem. Currently this method takes a parameter `AER::RngEngine &rng` which would not be needed if we remove the norm estimation so we can also get rid of that. I expect changing the parameters to remove this will require some changes elsewhere as well.

PR #1221, opened 2021-04-13
Normalize statevector save and snapshot for extended stabilizer method 

or1426:
### Summary
Fix issue #1196. Since the statevector method of CHSimulator::Runner already computes the inner product with all 2^n computational basis states we can use this information to exactly compute the norm rather than using the fast norm estimation algorithm.

### Details and comments

Since the CHSimulator::Runner::statevector does not use the fast norm estimation algorithm anymore it does not need to take the parameters relating to it. This required changes everywhere this method is called which seems to just be in extended_stabilizer_state.hpp.

I don't think this change requires any additional tests unless we want to add a test that that confirms the output of this method is normalized.

### DIFF
--- a/releasenotes/notes/fix-normalize-statevector-extended-stabilizer-3811f44e003099f3.yaml
+++ b/releasenotes/notes/fix-normalize-statevector-extended-stabilizer-3811f44e003099f3.yaml
@@ -1,0 +1,4 @@
+---
+fixes:
+  - |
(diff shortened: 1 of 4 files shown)


chriseclectic:
Thanks for making this PR @or1426. I added a commit to add the extended stabilizer method to the save statevector tests (since these tests are for clifford circuits, it should be exact for these states).

chriseclectic (approved):

Issue #1224, opened 2021-04-16
Memory leak when running GPU simulation on Aer 0.8


chriseclectic:
### Informations

- **Qiskit Aer version**: 0.8.1
- **Python version**: 3.7
- **Operating system**: RHEL8

### What is the current behavior?

The GPU simulator appears to be using excessive system memory since the 0.8 release. I noticed this running on a HPC cluster and having the jobs killed when they should have been well below the requested memory. 

### Steps to reproduce the problem

I wrote the following test script and ran it with `mprof python test.py`

```python
from qiskit import transpile, assemble, QuantumCircuit
from qiskit.circuit.library import QuantumVolume
from qiskit.providers.aer import QasmSimulator, __version__

print(f'TEST GPU: {__version__}')
NQS = [25, 26, 27, 28, 29, 30]
sim = QasmSimulator(method='statevector_gpu',
                    fusion_enable=False)

circuits = []
for nq in NQS:
    circ = QuantumCircuit(nq)
    circ.h(range(nq))
    circ.measure_all()
    circuits.append(circ)
qobj = assemble(circuits, shots=1, seed_simulator=1000)
result = sim.run(qobj).result()

if not result.success:
    print(f'FAILED!')
else:
    print('FINISHED, Simulation Time: {:.3f}'.format(result.time_taken))
```

At 30 qubits the maximum needed GPU memory should be ~16 GB.

The reported memory usage on 0.7.6 was very low as one would expect for a GPU simulation:

![mprof-gpu-076](https://user-images.githubusercontent.com/2235104/115085845-376e4f00-9ed9-11eb-8c59-98d7b2b6da81.png)

Running on 0.8.1 with the GPU simulation memory you can see that memory usage is increasing linearly through the computation, maxing out at around ~66GB which is over 4x the memory that would be required for the CPU simulation.

![mprof-gpu-081](https://user-images.githubusercontent.com/2235104/115085964-6684c080-9ed9-11eb-8a56-5a1e0565cca8.png)

For comparison running this script on the CPU statevector simulator maxes out at the expected 16GB:

![mprof-cpu-081](https://user-images.githubusercontent.com/2235104/115086319-03dff480-9eda-11eb-8a6f-36bcf1359b97.png)

### What is the expected behavior?

The memory usage of 0.8.1 should look approximately the same as 0.7.6 when running a GPU simulation.

### Suggested solutions

[...]

PR #1200, opened 2021-03-31
Fixes of multi-chunk simulators

doichanj:
### Summary
This PR fixes bugs of multi-chunk simulators with cache blocking transpiler.

### Details and comments
This fix includes:

- Fix of cache blocking transpiler
- Fix of initialize/reset operations for statevector/density-matrix simulators
- Fix of unitary simulator
- Implementation of set_statevector/set_densitymatrix operators.

PR #1226, opened 2021-04-19
Remove unnecessary buffer allocation at chunk manager initialization (Thrust)

doichanj:
### Summary
This is the fix for issue #1224 

### Details and comments
Chunk manager allocated buffers for MPI and buffers for checkpoint at the initialization.
However these are unnecessary for most cases. (MPI host buffer is only required for environments without GPUDirectRDMA)
So I removed these allocations and allocate checkpoint buffer when checkpoint is used.


doichanj:
This fix is merged to PR #1200 , so I would like to close this one.

Issue #1227, opened 2021-04-20
Extended stabilizer give wrong expval for Bell state and Pauli 'YY'


gadial:
### Informations

- **Qiskit Aer version**: 0.9.0
- **Python version**: 3.8.3
- **Operating system**: Ubuntu Linux

### What is the current behavior?
Computing expected value for the Pauli operator 'YY' on a Bell state generated in the standard way results in `1` instead of `-1` on both the stabilizer simulator and the extended stabilizer simulator (which use different algorithms for this computation).

### Steps to reproduce the problem
```
from qiskit import QuantumCircuit
from qiskit import execute
from qiskit.providers.aer import AerSimulator
from qiskit.quantum_info import Pauli, Statevector

op = Pauli('YY')
qc = QuantumCircuit(2,2)
qc.h(0)
qc.cx(0,1)
s = Statevector(qc)
print(s._expectation_value_pauli(op))

qc.save_expectation_value(op, [0, 1], label='expval')
for method in ['stabilizer', 'extended_stabilizer']:
    job = execute(qc, AerSimulator(method=method))
    print(method, job.result().data(0)['expval'])
```

### What is the expected behavior?
Both simulators should return `-1`.

chriseclectic:
It looks like the stabilizer algorithm is wrong (probably makes assumptions about measurement that aren't valid for entangled states).

I tested a new algorithm for stabilizer simulator for computing the Pauli expval that works for the quantum_info operators, but to implement it in Aer we would need a function for computing the conjugation of a Pauli by a Clifford: P -> C^\dagger . P. C:

- Assume Pauli `P` is +1 group element, stabilizer state `|S> = C|0>` for Clifford `C`.
- Expval `<P> = <S|P|S> = <0| C^\dagger.P.C |0>`
- Compute `C^\dagger.P.C = a Q` where `a` = +1 or -1, and `Q` output Pauil
- If `Q` has any X or Y elements, expval = 0, else expval  = `a`

Here is a function to do this in Qiskit:

```python
import numpy as np
import qiskit.quantum_info as qi

def stabilizer_expval(clifford, pauli):
    # Evolve Pauli by clifford
    p_evo = pauli.evolve(clifford)
    
    # Check if zero
    if np.any(p_evo.x):
        return 0.0
    
    # Get expval
    expval = (-1j) ** p_evo.phase
    if expval.imag == 0:
        return expval.real
    return expval
```

chriseclectic:
[...]

PR #1238, opened 2021-04-28
Disable save expectation value for ext stabilizer

chriseclectic:
### Summary

Disables the save expectation value instructions for the extended stabilizer method until Issue #1227 can be resolved.

### Details and comments


mtreinish (approved):
LGTM

Issue #1234, opened 2021-04-27
Undeclared variable in state_chunk.hpp when compiling with AER_MPI


otbrown:
### Informations

- **Qiskit Aer version**: 0.9.0 (master branch)
- **Python version**: Python 3.8.5
- **Operating system**: HPE Cray Linux Environment (based on SLES 15)

### What is the current behavior?
Unable to compile with MPI support, due to an undeclared identifier compilation error. Error thrown in [state_chunk.hpp](https://github.com/Qiskit/qiskit-aer/blob/master/src/simulators/state_chunk.hpp).
`qiskit-aer/src/simulators/state_chunk.hpp:1263:7: error: use of undeclared identifier 'nLarge'`

### Steps to reproduce the problem
Compile from source with `AER_MPI` enabled.
`python ./setup.py bdist_wheel -- -DAER_MPI=True`

### What is the expected behavior?
Compile successfully with no errors.

### Suggested solutions
Add `uint_t nLarge;` somewhere between line 1256 and 1261 in state_chunk.hpp. qiskir-aer compiles successfully for me with this fix.

chriseclectic:
@doichanj can you check this

danbev:
@chriseclectic Sorry about this, this was my fault. I've opened #1241 with a suggestion to fix this.

doichanj:
`nLarge` in `apply_chunk_x` function is not used so we can remove this variable (sorry I copied codes from `apply_chunk_swap` function and forgot removing this variable)

otbrown:
Thanks for the prompt fix :)

PR #1241, opened 2021-04-29
Fix compiler errors when using MPI

danbev:
### Summary
Fix compiler errors when using MPI

### Details and comments
This commit fixes compiler warnings that were caused by Commit
1a51d273dec6e00f62495d2dca97d7f09e72755f ("Fix various compiler warnings (#1201)").

Fixes: https://github.com/Qiskit/qiskit-aer/issues/1234

----
Sorry for causing this. I'll make sure to configure using `-DAER_MPI=True` and check other options in the future to avoid this.

### DIFF
--- a/contrib/standalone/qasm_simulator.cpp
+++ b/contrib/standalone/qasm_simulator.cpp
@@ -99,6 +99,7 @@ int main(int argc, char **argv) {
 
 #ifdef AER_MPI
   int prov;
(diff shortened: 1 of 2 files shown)


chriseclectic:
@vvilpas/@mtreinish It would be good to add a CI build job to check building standalone with MPI enabled to catch these kinds of compiler errors.

PR #1242, opened 2021-04-29
Add MPI standalone CI build job

mtreinish:
### Summary

This commit adds a CI job that builds aer in standalone mode with MPI to
ensure that we can build with MPI support. This should hopefully prevent
regressions like #1234 in the future.

### Details and comments

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,7 +93,7 @@ jobs:
           cd out/bin
           for test in test*
           do echo $test
-            if ! /usr/bin/mprirun -host localhost:2 -np 2 ./$test
+            if ! /usr/bin/mpirun.openmpi -host localhost:2 -np 2 ./$test
             then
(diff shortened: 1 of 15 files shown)


mtreinish:
This is blocked until #1241 merges

mtreinish:
Ok, after merging #1241 into the PR branch so we can actually build, the CI job looks like it's working now. It's running mpi on the local ci node and reflecting that in the output: https://github.com/Qiskit/qiskit-aer/runs/2470401773?check_suite_focus=true#step:8:73

doichanj:
I think it is enough to test standalone MPI execution.
But I think we need some test cases for Python execution. There are already existing test cases for multi-chunk parallelization, but I think some more new test cases are needed to test MPI implementation (e.g. snapshots, expectation values)

chriseclectic (approved):

Issue #1247, opened 2021-05-04
MPS: Add information about approximation in metadata


merav-aharoni:
### What is the expected behavior?
Add prints of bond dimension and of truncation values when using approximation in MPS.

PR #1248, opened 2021-05-10
MPS print debug data

merav-aharoni:
### Summary
If the environment variable `MPS_OUTPUT_DATA == 1`, after every instruction, the bond dimensions and the discarded values (from approximation) are printed to the circuit metadata.

### Details and comments
This implements the enhancement requested in Issue #1247.

### DIFF
--- a/qiskit/providers/aer/backends/aer_simulator.py
+++ b/qiskit/providers/aer/backends/aer_simulator.py
@@ -400,8 +400,7 @@ class AerSimulator(AerBackend):
             'set_stabilizer'
         ]),
         'extended_stabilizer': sorted([
-            'roerror', 'snapshot', 'save_statevector',
-            'save_expval', 'save_expval_var'
+            'roerror', 'snapshot', 'save_statevector'
         ]),
         'unitary': sorted([
             'snapshot', 'save_state', 'save_unitary', 'set_unitary'
--- a/qiskit/providers/aer/backends/aer_simulator.py
+++ b/qiskit/providers/aer/backends/aer_simulator.py
@@ -303,6 +303,9 @@ class AerSimulator(AerBackend):
       number of qubits, and low entanglement. If the user does not specify the algorithm,
       a heuristic algorithm is used to select between the two algorithms.
       (Default: "mps_heuristic").
+    * ``mps_log_data`` (str): if True, output logging data of the MPS
+      structure: bond dimensions and values discarded during approximation.
+      (Default: False)
 
     These backend options apply in circuit optimization passes:
 
@@ -523,7 +526,7 @@ class AerSimulator(AerBackend):
             matrix_product_state_truncation_threshold=1e-16,
             matrix_product_state_max_bond_dimension=None,
             mps_sample_measure_algorithm='mps_heuristic',
-            MPS_log_data=None,
+            mps_log_data=False,
             chop_threshold=1e-8,
             mps_parallel_threshold=14,
             mps_omp_threads=1)
--- a/qiskit/providers/aer/backends/aer_simulator.py
+++ b/qiskit/providers/aer/backends/aer_simulator.py
@@ -337,17 +337,17 @@ class AerSimulator(AerBackend):
         ]),
         'matrix_product_state': sorted([
             'u1', 'u2', 'u3', 'u', 'p', 'cp', 'cx', 'cy', 'cz', 'id', 'x', 'y', 'z', 'h', 's',
-            'sdg', 'sx', 't', 'tdg', 'swap', 'ccx', 'unitary', 'roerror', 'delay',
+            'sdg', 'sx', 't', 'tdg', 'swap', 'ccx', 'unitary', 'roerror', 'delay', 'pauli',
             'r', 'rx', 'ry', 'rz', 'rxx', 'ryy', 'rzz', 'rzx', 'csx', 'cswap', 'diagonal',
             'initialize'
         ]),
         'stabilizer': sorted([
             'id', 'x', 'y', 'z', 'h', 's', 'sdg', 'sx', 'cx', 'cy', 'cz',
-            'swap', 'delay',
+            'swap', 'delay', 'pauli'
         ]),
         'extended_stabilizer': sorted([
             'cx', 'cz', 'id', 'x', 'y', 'z', 'h', 's', 'sdg', 'sx',
-            'swap', 'u0', 't', 'tdg', 'u1', 'p', 'ccx', 'ccz', 'delay'
+            'swap', 'u0', 't', 'tdg', 'u1', 'p', 'ccx', 'ccz', 'delay', 'pauli'
         ]),
         'unitary': sorted([
             'u1', 'u2', 'u3', 'u', 'p', 'r', 'rx', 'ry', 'rz', 'id', 'x',
@@ -361,7 +361,7 @@ class AerSimulator(AerBackend):
             'u1', 'u2', 'u3', 'u', 'p', 'r', 'rx', 'ry', 'rz', 'id', 'x',
             'y', 'z', 'h', 's', 'sdg', 'sx', 't', 'tdg', 'swap', 'cx',
             'cy', 'cz', 'cp', 'cu1', 'rxx', 'ryy',
-            'rzz', 'rzx', 'ccx', 'unitary', 'diagonal', 'delay',
+            'rzz', 'rzx', 'ccx', 'unitary', 'diagonal', 'delay', 'pauli'
         ])
     }
     # Automatic method basis gates are the union of statevector,
--- a/qiskit/providers/aer/backends/aer_simulator.py
+++ b/qiskit/providers/aer/backends/aer_simulator.py
@@ -471,6 +471,10 @@ class AerSimulator(AerBackend):
             configuration = QasmBackendConfiguration.from_dict(
                 AerSimulator._DEFAULT_CONFIGURATION)
 
+        # Cache basis gates since computing the intersection
+        # of noise model, method, and config gates is expensive.
+        self._cached_basis_gates = self._BASIS_GATES['automatic']
+
         super().__init__(configuration,
                          properties=properties,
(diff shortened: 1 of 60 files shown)


merav-aharoni:
A few comments on the implementation:
* MPS logging data is printed to the `result.results[0].metadata`
* MPS logging data includes bond dimensions and discarded values (values discarded by approximation)
* Logging data will only be printed if the environment variable `MPS_OUTPUT_DATA` is turned on
* Bond dimensions will be printed only following instructions that may have changed their values
* discarded values will only be printed if greater than 0
* The prints are not thread save - parallelization will probably mess them up
Here is a sample print:
 `metadata = {'parallel_shots': 1, 'matrix_product_state_truncation_threshold': 1e-16, 'parallel_state_update': 8, 'matrix_product_state_sample_measure_algorithm': 2, 'method': 'matrix_product_state', 'MPS_log_data': '{I0:unitary on qubits 0,4, BD= [2 1 1 1 1 1 1 1 1],  I1:unitary on qubits 4,0, BD= [2 1 1 1 1 1 1 1 1],  I2:unitary on qubits 0,4, BD= [2 1 1 1 1 1 1 1 1],  I3:unitary on qubits 0,4, BD= [2 1 1 1 1 1 1 1 1],  I4:unitary on qubits 6,2, BD= [2 1 1 1 1 2 1 1 1],  I5:unitary on qubits 2,6, BD= [2 1 1 1 1 2 1 1 1],  I6:unitary on qubits 1,7, BD= [2 1 2 1 1 1 2 1 1],  I7:unitary on qubits 3,8, BD= [2 1 2 1 2 1 1 2 1],  I8:unitary on qubits 8,3, BD= [2 1 2 1 2 1 1 2 1],  I9:unitary on qubits 5,9, BD= [2 1 2 1 2 1 2 1 2],  I10:unitary on qubits 1,5, BD= [2 1 2 4 2 4 2 1 2],  I11:unitary on qubits 1,8, BD= [2 1 2 4 8 4 2 1 2],  I12:unitary on qubits 6,5, BD= [2 1 2 4 8 4 2 4 2],  I13:unitary on qubits 5,8, BD= [2 1 2 4 8 4 8 4 2],  discarded_value = 0.00101696, I14:unitary on qubits 0,8, BD= [2 4 2 4 8 8 8 4 2],  discarded_value = 0.0036741,...`

merav-aharoni:
Ready for review.

releasenotes/notes/mps_print_bond_dimension-6dc6f6cf0365c2ae.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    If the environment variable `MPS_OUTPUT_DATA` is set to 1, then MPS
+    log data is printed to the metadata. This includes, after every relevant
+    instruction, the bond dimensions and the values discarded during approximation.
```
yaelbh, line 6: I suggest to add a couple of sentences explaining what you mean by "discarded values". Personally I misunderstood it, and discovered the correct meaning only when I read the code.

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -489,8 +497,30 @@ void State::add_metadata(ExperimentResult &result) const {
...
+void State::add_reporting_metadata(ExperimentResult &result) const {
+  result.metadata.add("{" + MPS::output_log() + "}", "MPS_log_data");
+}
+
+void State::output_bond_dimensions(const Operations::Op &op) const {
+  MPS::print_to_log("I", instruction_number, ":", op.name, " on qubits ", op.qubits[0]);
+  for (uint_t index=1; index<op.qubits.size(); index++) {
+    MPS::print_to_log(",", op.qubits[index]);
+  }
+  MPS::print_to_log(", BD= [");
```
yaelbh, line 512: The spacing could be nicer, currently it looks like this:
```
I0:unitary on qubits 0,4, BD= [2 1 1 1 1 1 1 1 1]`
```
You could insert a space between "I0:" and "unitary", or between "BD" and "=", or alternatively for consistency remove the space between "BD=" and "[".

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -574,6 +604,14 @@ void State::apply_ops(const std::vector<Operations::Op> &ops,
       }
     }
     //qreg_.print(std::cout);
+    // print out bond dimensions only if they may have changed since previous print
+    if (getenv("MPS_OUTPUT_DATA") &&
```
yaelbh, line 608: `getenv` is expensive in run time. Maybe better to read it once at the beginning of execution, and store it in a static variable. Here and in additional locations.

yaelbh (changes requested):

src/simulators/matrix_product_state/matrix_product_state.hpp
merav-aharoni: ok, I prefer to reduce the spaces to make this printout more compact.

src/simulators/matrix_product_state/matrix_product_state.hpp
merav-aharoni: Good comment!

releasenotes/notes/mps_print_bond_dimension-6dc6f6cf0365c2ae.yaml
merav-aharoni: Added a sentence both in release notes and in svd.cpp. Let me know if it makes more sense now.

releasenotes/notes/mps_print_bond_dimension-6dc6f6cf0365c2ae.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    If the environment variable `MPS_OUTPUT_DATA` is set to 1, then MPS
+    log data is printed to the metadata. This includes, after every relevant
+    instruction, the bond dimensions and the discarded_value i.e.,
```
yaelbh, line 6: Remove the underscore

yaelbh (approved):

releasenotes/notes/mps_print_bond_dimension-6dc6f6cf0365c2ae.yaml
merav-aharoni: done

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -523,6 +523,7 @@ def _default_options(cls):
             matrix_product_state_truncation_threshold=1e-16,
             matrix_product_state_max_bond_dimension=None,
             mps_sample_measure_algorithm='mps_heuristic',
+            MPS_log_data=None,
```
chriseclectic, line 526: The doc string of the class and QasmSimulator should be updated to add a descirption of this option in the list of of options.  Looking here its not clear to me if this is a boolean value or not. If bool the default should probably be `False` instead of `None`.

src/controllers/aer_controller.hpp
```diff
@@ -1937,12 +1937,14 @@ void Controller::run_circuit_helper(const Circuit &circ,
   else {
     run_circuit_with_sampled_noise(circ, noise, config, shots, state, method,
                                    cache_blocking, result, rng);
+    state.add_reporting_metadata(result);
```
chriseclectic, line 1940: Can this just be the existing `add_metadata` function call, and we remove earlier fall to `state.add_metadata(result)` and make sure its done at the end of execution rather than at the start?

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -140,6 +143,12 @@ class State : public Base::State<matrixproductstate_t> {
 
   virtual void add_metadata(ExperimentResult &result) const override;
 
+  // Similar to add_metadata, but used to add data at the end of circuit execution
+  virtual void add_reporting_metadata(ExperimentResult &result) const override;
```
chriseclectic, line 147: I think we can remove this and just use add_metadata, but update the controller code to only add metadata at the end of simulation, rather than at the start.

releasenotes/notes/mps_print_bond_dimension-6dc6f6cf0365c2ae.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    If the environment variable `MPS_OUTPUT_DATA` is set to 1, then MPS
```
chriseclectic, line 4: Is the environment variable still used? I think this should say how to use this feature by using the `set_options(MPS_log_data=True)` for the AerSimulator and QasmSimulator backend classes.

chriseclectic (changes requested):
Looks good, main question/suggestion is can we just include this in the existing `add_metadata` state function by changing where it is called in the controller code.

qiskit/providers/aer/backends/aer_simulator.py
merav-aharoni: Changed to `False` and add description in `aer_simulator.py` and in `qasm_simulator.py`.

releasenotes/notes/mps_print_bond_dimension-6dc6f6cf0365c2ae.yaml
merav-aharoni: Indeed I removed the environment variable. You are right that it looks better simply as an option.

src/controllers/aer_controller.hpp
merav-aharoni: I see that the only state that actually implements `add_metadata` is `matrix_product_state` so indeed I can move the call to `add_metadata` to the end of execution and then we don't need the method `add_recording_metadata`.

src/simulators/matrix_product_state/matrix_product_state.hpp
merav-aharoni: done, see above.

chriseclectic (approved):

Issue #1253, opened 2021-05-17
MPS simulator produces incorrect result


liyaochong:
*BUG TEMPLATE* 

### Informations

- **Qiskit AER version**: 0.8.2
- **Operating system**: Windows

### What is the current behavior?
The output of quantum ciruict with 'matrix_product_state' and 'statevector' are different.
{'111000101001': 1024} for 'matrix_product_state' simulator;
{'100100011001': 1023, '100100110001': 1} for 'statevector' simulator.

### Steps to reproduce the problem
import numpy as np
import qiskit
from qiskit import QuantumCircuit, transpile
from qiskit.providers.aer import AerSimulator
import networkx as nx
import matplotlib.pyplot as plt

n = 12
G = nx.Graph()
G.add_nodes_from(range(0,n))
G.add_edges_from([(0,1), (0,2), (1,2), (1,4), (2,4), (2,3), (3,5), (4,6), (5,6), (4,7), (6,7),
             (7,8), (8,9), (7,9), (6,8), (6,9), (8,10), (9,11), (10,11)])
colors = ['b' for node in G.nodes()]
pos = nx.spring_layout(G)
default_axes = plt.axes(frameon=True)
nx.draw_networkx(G, node_color=colors, node_size=600, alpha=.8, ax=default_axes, pos=pos)  

def produce_QAOA_state(gammas,betas,p):
    circuit = QuantumCircuit(n+3, n)
    for pk in range(p):
        for node in G.nodes():
            circuit.rz(-gammas[pk], node)
        for k, k_nbrs in G.adjacency():
            k_nbrs = list(k_nbrs.keys())
            k_nbrs = [int(x) for x in k_nbrs]
            circuit.x(k_nbrs)
            if len(k_nbrs)==1:
                circuit.crx(2*betas[pk],k_nbrs[0],k)
            elif len(k_nbrs)==2:
                circuit.ccx(k_nbrs[0],k_nbrs[1],n)
                circuit.crx(2*betas[pk],n,k)
                circuit.ccx(k_nbrs[0],k_nbrs[1],n)
            elif len(k_nbrs)==3:
                circuit.ccx(k_nbrs[0],k_nbrs[1],n)
                circuit.ccx(k_nbrs[2],n,n+1)
                circuit.crx(2*betas[pk],n+1,k)
                circuit.ccx(k_nbrs[2],n,n+1)
                circuit.ccx(k_nbrs[0],k_nbrs[1],n)
[...]

PR #1255, opened 2021-05-23
Disable Fusion when running on MPS

merav-aharoni:
### Summary
Fusion optimization does not work correctly on the MPS simulation method. Therefore should be disabled, even when set in the config.

### Details and comments
See failure example in issue #1253.

### DIFF
--- a/releasenotes/notes/disable_fusion_with_mps-6757cbf8edb510a5.yaml
+++ b/releasenotes/notes/disable_fusion_with_mps-6757cbf8edb510a5.yaml
@@ -1,0 +1,8 @@
+---
+
+fixes:
+  - |
+    Disabled fusion when using the MPS simulation method, even when fusion is 
+    enabled in the config. Fusion does not run correctly on MPS so should
+    always be disabled. See issue #1253 for an example.
+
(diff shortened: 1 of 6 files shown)


yaelbh:
If a user explicitly asked "MPS and enable fusion" then I'm not sure that you can just override the user's request. Maybe invoke an error, or at least print a warning that you override an explicit request.

merav-aharoni:
There are two problems with writing a warning:
1. There is no mechanism for warnings in the C++ code.
2. The default set-up for the config is to set `fusion_enable = True`, so that even when the use does not set this field explicitly, they will get the warning. 
So overall, I prefer not to have a warning.

mtreinish:
> There are two problems with writing a warning:
> 
>     1. There is no mechanism for warnings in the C++ code.

You can generate a python warning with the Python C api. Something like:

```c++
PyErr_WarnEx(PyExc_RuntimeWarning, "This is my warning text", 1);
```

https://docs.python.org/3/c-api/exceptions.html#c.PyErr_WarnEx
https://docs.python.org/3/c-api/exceptions.html#standard-warning-categories

I'm not sure if their is a pybind native way to do that or not, but under the covers pybind is just calling that anyway.

merav-aharoni:
@yaelbh , please let me know if you agree that we are better off without a comment. Except for this point, the PR is ready for merge.

yaelbh:
Yes, it's fine

chriseclectic (approved):

PR #1259, opened 2021-05-28
fix docstring for error_term

peendebak:
### Summary

Fix mistake in documentation of `error_term`

### Details and comments

The method returns `return self.circuits[position], self.probabilities[position]`, but the docstring said `(p, circuit)`.


mtreinish (approved):
LGTM, thanks for fixing this

chriseclectic (approved):

PR #1262, opened 2021-05-31
Fixing pulse simulator measurement bug

### LINKED ISSUE #1257: PulseSimulator measurement probability bug
### Informations

- **Qiskit Aer version**: stable/0.8 and later
- **Python version**:
- **Operating system**:

The following code produces an incorrect output:

```
import qiskit
from qiskit import *
from qiskit.pulse import *
from qiskit.providers.aer.pulse.system_models.pulse_system_model import PulseSystemModel
from qiskit.test.mock import FakeValencia
from qiskit.providers.aer import PulseSimulator

backend = FakeValencia()
# backend = qiskit.test.mock.FakeArmonk()
config = backend.configuration()
N = config.n_qubits
bases = {0:'X'}
# CONVERT ABSTRACT CIRCUIT            
qc = QuantumCircuit(N, N)
for iq in bases:
    if bases[iq]=='X':
        qc.h(iq)
    if bases[iq]=='Y':
        qc.sdg(iq)
        qc.h(iq)
# for iq in bases:
#     qc.measure(iq,iq)
for iq in range(N):
    qc.measure(iq,iq)
qc = transpile(qc, backend)
measurements = qiskit.schedule(qc, backend)    
sched = Schedule()
sched = sched.append(measurements)
model = PulseSystemModel.from_backend(backend)
backend_sim = PulseSimulator()
qobj = assemble(sched,
              backend=backend_sim,
              meas_level=2,
              meas_return='single',
              shots=1024)
sim_result = backend_sim.run(qobj, system_model=model).result()
counts = sim_result.get_counts()
counts
```

produces output:
```
{'00000': 1024}
```

But the correct output should be roughly
```
{'00000': 386, '00001': 638}.
```

as the correct probabilities are about `[0.38, 0.62]`.

The branch `stable/0.7` produces correct outputs, but `stable/0.8` does not. This appears to correspond to the move from the qutip `Qobj` to using `qiskit.quantum_info` operators. The simulation shot outcomes are computed in line `61`-`63` in `unitary_controller.py`. I've visually inspected the state `psi` when running in both branches and it looks to be the same in both cases, so the real divergence seems to come from line `61` when the probabilities are computed. In `stable/0.7` they are correct, but in `stable/0.8` they are all approximately `0`. This appears to either be the result of an error in the construction of `pulse_sim_desc.measurement_ops`, or in the function `occ_probabilities`, or both. 

@vvilpas if I remember correctly you did the final migration from qutip `Qobj` to the `Operator` - do you have any thoughts/insights? I think there may also be potential issues with differences in dunder methods between `Qobj` and `operator` in the measurement operator construction, but I'm not certain.

DanPuzzuoli:
### Summary

Closes #1257

### Details and comments

As described in the comments in #1257, there was an issue with the way measurement operators were being constructed, which was introduced when switching from using the qutip `Qobj` to using the `qiskit.quantum_info.Operator` class. In particular, the error came from a line in the measurement operator construction function in which a tensor product of `Operator`s is taken. These operators represent "states", but were being stored as "column vectors" (i.e. 2d arrays with shape `(n, 1)`), but the tensor product code produced incorrect results for this case. The problem was solved just be reshaping the initial vectors to be properly 1 dimensional before inputting to the tensor code. 

The existing tests didn't catch this because it seemed to be outputting the correct values when everything started out 2 dimensional, and we didn't have any tests directly testing measurement outcomes for higher than 2 dimensional subsystems (for higher dimensional systems, we only checked the final output state, not the sampled probabilities).

Additionally:
- A test was added that would have caught this.
- A few additional lines were changed to eliminate deprecation warnings being raised by `Operator` relating to the use of `*` and `@` for multiplication.

### DIFF
--- a/qiskit/providers/aer/pulse/controllers/digest_pulse_qobj.py
+++ b/qiskit/providers/aer/pulse/controllers/digest_pulse_qobj.py
@@ -24,8 +24,7 @@ from .pulse_utils import oplist_to_array
 
 
 class DigestedPulseQobj:
-    """Container class for information extracted from PulseQobj
-    """
+    """Container class for information extracted from PulseQobj."""
 
     def __init__(self):
 
--- a/qiskit/providers/aer/pulse/controllers/pulse_controller.py
+++ b/qiskit/providers/aer/pulse/controllers/pulse_controller.py
@@ -390,7 +390,7 @@ class PulseInternalDEModel:
             H_noise = Operator(np.zeros(self.noise[0].data.shape))
             for kk in range(self.c_num):
                 c_op = self.noise[kk]
-                n_op = c_op.adjoint() @ c_op
+                n_op = c_op.adjoint() & c_op
                 # collapse ops
                 self.c_ops_data.append(c_op.data)
                 # norm ops
--- a/qiskit/providers/aer/pulse/controllers/pulse_controller.py
+++ b/qiskit/providers/aer/pulse/controllers/pulse_controller.py
@@ -140,7 +140,7 @@ def pulse_controller(qobj):
     if qubit_lo_freq is None:
         qubit_lo_freq = system_model.hamiltonian.get_qubit_lo_from_drift()
         warn('Warning: qubit_lo_freq was not specified in PulseQobj and there is no default, '
-             'so it is beign automatically determined from the drift Hamiltonian.')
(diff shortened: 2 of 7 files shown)


vvilpas (approved):
LGTM

PR #1273, opened 2021-06-17
Implementation of compute algorithm from arxiv:2101.12223

or1426:
### Summary
The paper https://arxiv.org/abs/2101.12223 contains two algorithms for finding Born rule probabilities of Clifford+phase gate circuits. This pull request contains the  compute algorithm (and everything it depends on). 

This is not suitable to be merged currently. Documentation, tests and are missing, and there are some problems (see below), but I open it to get some feedback.

### Details and comments

In addition to the compute algorithm a new save instruction is added to the quantumcircuit class. The instruction
`save_specific_probability(outcomes, measured_qubits)` to save the Born rule probability of a particular measurement outcome of a computational basis measurement applied to a subset of the qubits. 

##Example usage

`qc.save_specific_probability([0,1,0], [0,1,2]) # save the probability of the obtaining the outcome (0,1,0) when measuring qubits (0,1,0)`

`qc.save_specific_probability([0,1], [5,1]) # save the probability of the obtaining the outcome (0,1) when measuring qubits 5 and 1`

It would also be reasonable to add support for this instruction to other simulators - e.g. the statevector simulator.

##Known problems
1. Currently the `CliffPhaseCompute::State` class does not correctly report it's memory usage
2. Currently the Compute algorithm implementation is not parallel

### DIFF
--- a/qiskit/providers/aer/backends/aer_simulator.py
+++ b/qiskit/providers/aer/backends/aer_simulator.py
@@ -346,6 +346,10 @@ class AerSimulator(AerBackend):
             'cx', 'cz', 'id', 'x', 'y', 'z', 'h', 's', 'sdg', 'sx',
             'swap', 'u0', 't', 'tdg', 'u1', 'p', 'ccx', 'ccz', 'delay'
         ]),
+        'clifford_phase_compute': sorted([
+            'cx', 'cz', 'id', 'x', 'y', 'z', 'h', 's', 'sdg', 
+            'swap', 't', 'tdg',  'delay'
+        ]),
         'unitary': sorted([
             'u1', 'u2', 'u3', 'u', 'p', 'r', 'rx', 'ry', 'rz', 'id', 'x',
             'y', 'z', 'h', 's', 'sdg', 'sx', 't', 'tdg', 'swap', 'cx',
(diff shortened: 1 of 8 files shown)


hhorii:
I think extending `save_probability()` (by adding new keyword argument `states`) is better than adding a new function `save_specific_probability()`.

or1426:
> I think extending `save_probability()` (by adding new keyword argument `states`) is better than adding a new function `save_specific_probability()`.

I don’t disagree with this but it requires a bit of thinking about. One of the functionalities of the new `save_specific_probability()` method is the ability to specify a measurement on an arbitrary subset of qubits. For example you can ask it for the probability of outcome 01 when you measure the first and third qubits of a five qubit state.

We want to add this for two reasons, firstly it might be useful to people, and secondly one of the nice features of the algorithms from this paper is that they can do this without excessive overhead (compute doesn’t need to marginalise the probability distribution or anything like that). If you try to implement this functionality for (for example) the statevector simulator you have to do an exponential (in the number of qubits) amount of work adding up the probabilities to get the outcome you want on the qubits you want.

hhorii:
> the statevector simulator you have to do an exponential (in the number of qubits) amount of work adding up the probabilities to get the outcome you want on the qubits you want.

That's true. Maybe, with a naive implementation, statevector gets all the probability amplitudes and then take specified amplitudes. Though its cost can be huge, we will implement it in statevector, too.

I'm thinking API. `save_probabilities` and `save_probabilities_dict` have an argument `qubits` to filter outcome probabilities. I guess that `states` is a kind of similar argument.

Any comments? > @chriseclectic

Issue #1269, opened 2021-06-15
Numerical error in the Extended Stabilizer simulator


tnguyen-ornl:
### Informations

- **Qiskit Aer version**: 0.8.2
- **Python version**: 3.8.10 
- **Operating system**: Linux

### What is the current behavior?
State vector results from the Extended Stabilizer simulator have a significant level of error compared to the true results from, e.g., the state vector simulator.

Note that this is not related to the norm issue (https://github.com/Qiskit/qiskit-aer/issues/1196), which I think has been fixed.
The option `extended_stabilizer_approximation_error` doesn't seem to help at all.

For example: for a `H - U1(-2.48) - H` circuit, with `extended_stabilizer_approximation_error=0.001`, the probability vector (square norm, see the code below) from the extended stabilizer is `[0.15351169 0.84648831]`, which is quite different from the expected result of `[0.10542975 0.89457025]`

### Steps to reproduce the problem

See the code snippet below. It sweeps the angle and runs an `H - U1(theta) - H` circuit with both the extended stabilizer and state-vector simulators.

Since the resulting state vector may have a different global phase, we look at the probability distribution (square norm) instead.

```
import numpy as np
# Importing standard Qiskit libraries
from qiskit import QuantumCircuit, transpile, Aer, IBMQ
from qiskit.providers.aer import AerSimulator
from qiskit.tools.jupyter import *
from qiskit.visualization import *
from ibm_quantum_widgets import *

# Loading your IBM Quantum account(s)
provider = IBMQ.load_account()

method = 'extended_stabilizer'
sim = AerSimulator(method=method, 
                   extended_stabilizer_approximation_error=0.001)
sim_perfect = AerSimulator()
angles = np.linspace(-np.pi, np.pi, 20)
for angle in angles:
    print('Theta =', angle)
    qc = QuantumCircuit(1)
    qc.h(0)
    qc.u1(angle, 0)
    qc.h(0)
    qc.snapshot('final')
    result = sim.run(qc, shots=1).result()
    psi = result.data(0)['snapshots']['statevector']['final'][0]
    prob_dist = abs(psi) ** 2
    print('Extended stabilizer probability dist:', prob_dist)
    ref_result = sim_perfect.run(qc, shots=1).result()
    psi_ref = ref_result.data(0)['snapshots']['statevector']['final'][0]
    prob_dist_ref = abs(psi_ref) ** 2
    print('Reference probability dist:', prob_dist_ref)
```

[...]

PR #1276, opened 2021-06-18
fix phase normalisation issue leading to #1269

or1426:
### Summary
Previously the phase was unnormalised (indeed had norm sqrt(2)) for odd e. This lead to incorrect outcomes from the Runner::amplitude method 

### Details and comments

As far as I can tell this commit fixes issue #1269


or1426:
I'm not sure what is going on with the checks that failed.

chriseclectic (approved):
Thanks for making this fix

PR #1281, opened 2021-06-24
Cluster backend

hitomitak:
### Summary
Take over #1084 

Adds a new option of the backend to provide the user's executor. When user gives dask client as executor, Aer can execute a simulation on the distributed machines like HPC clusters. 

When the executor is set, AerJobSet object is returned instead of a normal AerJob object. 
AerJobSet divides multiple experiments in one qobj into each experiment and submits each qobj to the executor as AerJob. 
After simulations, AerJobSet collects each result and combines them into one result object. 

Example Usage:
Threadpool execution:
```
exc = ThreadPoolExecutor(max_workers=2)
qbackend = Aer.get_backend('qasm_simulator')
result_ideal = qiskit.execute(circ_list, qbackend, executor=exc).result()
```
Dask execution:
```
exc = Client(address=LocalCluster(n_workers=1, processes=True))
qbackend = Aer.get_backend('qasm_simulator')
result_ideal = qiskit.execute(circ_list, qbackend, executor=exc).result()
```
### Details and comments
I executed QV (10 depth, 64 circuits)and measured the execution time on 3VMs with Dask executor. 
Env: 3VMs, OS: CentOS8.0, Memory 64 GB, CPU: 32 x 2.0 GHz or higher Cores.
1 Node : a normal Aer simulator
1 Node (DASK) : only one node with dask executor
2 Node (DASK) : two nodes with dask executor
3 Node (DASK) : three nodes with dask executor

fig3


qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -146,28 +154,68 @@ def run(self,
...
-                          stacklevel=2)
-            qobj = circuits
+        executor = None
+        if backend_options and "executor" in backend_options:
+            executor = backend_options["executor"]
+            del backend_options["executor"]
+
+        if "executor" in run_options:
+            executor = run_options["executor"]
+            del run_options["executor"]
```
mtreinish, line 164: Instead of doing this why not just add `executor` to the default options? That will also enable users to something like:

```
sim_backend = AerSimulator(executor=dask_magic())
sim_backend.run([qc]*10**5)
sim.backend.run([new_qc]*10**4)
```

and have it all run on dask.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -146,28 +154,68 @@ def run(self,
...
-        if isinstance(circuits, (QasmQobj, PulseQobj)):
-            warnings.warn('Using a qobj for run() is deprecated and will be '
-                          'removed in a future release.',
-                          PendingDeprecationWarning,
-                          stacklevel=2)
-            qobj = circuits
+        executor = None
+        if backend_options and "executor" in backend_options:
+            executor = backend_options["executor"]
+            del backend_options["executor"]
```
mtreinish, line 160: Do we really want to support doing this in `backend_options` even though it's deprecated.

mtreinish (commented):
Overall this looks good. I did a quick initial skim over the code and have a couple inline comments. Also this definitely needs a release note (see: https://github.com/Qiskit/qiskit-aer/blob/main/CONTRIBUTING.md#release-notes ) to document the new feature and how to use it.

Also I'm wondering do you think we want to ask dask as an optional extra dependency in the setup.py. The code relies on the futures interface so it's not required, but it might be nice to let users do something like `pip install 'qiskit-aer[dask]'`. What do you think?

qiskit/providers/aer/backends/aerbackend.py
hitomitak: At first, I added the executor option as your code but DASK client variable is needed to be a local variable for the serialization. So I moved the executor option as a run option.

hitomitak:
Thank you for the comment. I will add a release note soon. 

> Also I'm wondering do you think we want to ask dask as an optional extra dependency in the setup.py. The code relies on the futures interface so it's not required, but it might be nice to let users do something like `pip install 'qiskit-aer[dask]'`. What do you think?

I agree. I will update setup.py to install dask if users add the option.

chriseclectic (commented):
Initial review. I think we can still simply this a little more so there is little difference between running a single job or batch of jobs. I would also like control over the number of circuits per sub-job for the job set rather than hard coding it to always be 1.

One thing I would like is to have `executor` moved from the job to backend configuration with the default being the current `ThreadPoolExecutor(max_workers=1)`, then it can be updated using `set_options`.

I'd also like to add a separate simulator option for `max_job_size` that controls splitting jobs and submitting a job set that applies to any executor (not just Dask). The default option value for this should be `None` and should only run a single job on the executor, but if `max_job_size` is specified it will split the circuits into batches for the set of jobs (with `max_job_size=1` reproducing the current behavior).

qiskit/providers/aer/backends/cluster/aerjobset.py
```diff
@@ -0,0 +1,366 @@
...
+from qiskit.result import Result
+
+from ..aerbackend import AerError
+from .utils import requires_submit
+from ...aerjob import AerJob
+
+logger = logging.getLogger(__name__)
+
+
+class AerJobSet(Job):
```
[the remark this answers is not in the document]
chriseclectic: On second thought maybe not...

chriseclectic:
@hitomitak I accidently pushed directly to your branch when making some changes so I reverted those commits and then added them to a PR to your branch: https://github.com/hitomitak/qiskit-aer/pull/2

qiskit/providers/aer/backends/aerbackend.py
hitomitak:
```
class AerBackend(Backend, ABC):
    def __init__():
        self.executor = None:
   /* ... */
    def run():
        self.executor = getattr(self._options, 'executor')
```
This code occurs a cloudpickle error because the executor is not a local variable. The same error is mentioned in this [website](https://stackoverflow.com/questions/45976880/pickle-error-when-submitting-task-using-dask)

```
class AerBackend(Backend, ABC):
   /* ... */
    def run():
       executor = None
       executor = getattr(self._options, 'executor')
       delattr(self._options, 'executor')
```
I changed an instance variable to a local variable of `executor` before. This code can be executed when the first call of `run` function but the second call is failed because `delattr` is called. If `delattr` is not called, `Failed to load qobj: to_json not implemented for this type of object: <class 'concurrent.futures.thread.ThreadPoolExecutor'>` occurs. 

So I changed the `executor` option as the run option because `executor` is passed every calling `run` function. Are there any good idea to realize the `executor` options as a config option?

hitomitak: I found that the following code can be executed. I don't know why this code is ok but I can move `executor` option to default options. 
```
class AerBackend(Backend, ABC):
    def __init__():
        self.executor = None:
   /* ... */
    def run():
         executor = None
        if self._executor:
           executor = self._executor
        else:
           executor = getattr(self._options, 'executor')
           delattr(self._options, 'executor')
       /* ....*/     
       aer_job_set = AerJobSet(self, job_id, self._run, experiments, executor)
       aer_job_set.submit()
       self._executor = executor
       return aer_job_set
```

hitomitak:
@chriseclectic  Thank you for the refactoring. I changed the code to get `executor` option from `qob.config` to `options` because DASK serialization is failed.

CONTRIBUTING.md
```diff
@@ -738,6 +738,45 @@ meta = dict['metadata']
 myrank = meta['mpi_rank']
 ```
 
+### Running with Threadpool and DASK
```
chriseclectic, line 741: We should move this and the MPI docs in contributing to a page in the API docs, this can be done as a follow up though

CONTRIBUTING.md
```diff
@@ -738,6 +738,45 @@ meta = dict['metadata']
...
+
+Qiskit Aer can parallelize the simulation on a circuit bases with ThreadpoolExecutor or DASK as a custom executor. When user gives dask client as the executor, Aer can execute a simulation on the distributed machines like HPC clusters.
```
chriseclectic, line 743:
```suggestion
Qiskit Aer runs simulation jobs on a single-worker Python multiprocessing ThreadPool executor so that all parallelization is handled by low-level OpenMP and CUDA code. However to customize job-level parallel execution of multiple circuits a user can specify a custom multiprocessing executor and control the splitting of circuits using the `executor`and `max_job_size` backend options. For large scale job parallelization on HPC clusters Qiskit Aer executors support the distributed Clients from the [Dask](https://dask.org/) library.
```

chriseclectic (approved):
Hiding the Dask (and running MPI) documentation in the contributing guide, which is intended for documenting additional details for developers building from source, seems wrong. We should make a follow up PR to move this documentation to its own page in the API docs. Other than that LGTM

chriseclectic (approved):

test/terra/backends/aer_simulator/test_job_splitting.py
```diff
@@ -0,0 +1,82 @@
...
+import logging
+from ddt import ddt, data
+
+from qiskit import transpile, assemble
+from qiskit.circuit.random import random_circuit
+from qiskit.providers.aer.jobs import split_qobj
+from test.terra.reference.ref_snapshot_expval import (
+    snapshot_expval_circuits, snapshot_expval_circuit_parameterized)
+
+from .aer_simulator_test_case import AerSimulatorTestCase
```
chriseclectic, line 25:
```suggestion
from test.terra.backends.simulator_test_case import SimulatorTestCase
```

test/terra/backends/aer_simulator/test_job_splitting.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@ddt
+class TestJobSplitting(AerSimulatorTestCase):
```
chriseclectic, line 29:
```suggestion
class TestJobSplitting(SimulatorTestCase):
```

test/terra/backends/aer_simulator/test_executors.py
```diff
@@ -0,0 +1,120 @@
...
+import logging
+from math import ceil
+import concurrent.futures
+
+from ddt import ddt
+from qiskit import QuantumCircuit, transpile
+from qiskit.circuit.random import random_circuit
+from qiskit.quantum_info import Statevector
+from qiskit.providers.aer.jobs import AerJob, AerJobSet
+from .aer_simulator_test_case import (
```
chriseclectic, line 24:
```suggestion
from test.terra.backends.simulator_test_case import (
```

test/terra/backends/aer_simulator/test_executors.py
```diff
@@ -0,0 +1,120 @@
...
+    AerSimulatorTestCase, supported_methods)
```
chriseclectic, line 25:
```suggestion
    SimulatorTestCase, supported_methods)
```

test/terra/backends/aer_simulator/test_executors.py
```diff
@@ -0,0 +1,120 @@
...
+    for circ in circuits:
+        circ.measure_all()
+
+    circuits = transpile(circuits, backend)
+    job = backend.run(circuits, shots=shots, **run_options)
+    result = job.result()
+    return result, circuits, targets
+
+
+class CBFixture(AerSimulatorTestCase):
```
chriseclectic, line 59:
```suggestion
class CBFixture(SimulatorTestCase):
```

chriseclectic (changes requested):
Changes for renamed file

iotaisolutions:
**Getting serialization errors while trying to submit Qiskit Circuit List to Qiskit Aer Simulator** 

With respect to project: https://github.com/qiskit-advocate/qamp-fall-21/issues/39, I am trying to simulate few simple Qiskit circuit lists on a Qiskit Aer Simulator, with compute engine as Dask on Kubernetes Cluster, but continuously facing serialization (pickling) error(s). 

I am even getting these errors while testing a DASK  code  example available on Qiskit Documentation Portal :  **https://qiskit.org/documentation/apidoc/parallel.html** 



** Execution Environment**

- dask-kubernetes :  2021.3.1
- distributed :   2021.9.0
- minikube v1.23.1  
- Docker version 20.10.7
- Python 3.8
- OS : Ubuntu 20.04

**Error Messages**

  File "/home/ubuntu/.local/lib/python3.8/site-packages/distributed/worker.py", line 3809, in dumps_function
    result = cache_dumps[func]
  File "/home/ubuntu/.local/lib/python3.8/site-packages/distributed/utils.py", line 1366, in __getitem__
    value = super().__getitem__(key)
  File "/usr/lib/python3.8/collections/__init__.py", line 1010, in __getitem__
    raise KeyError(key)
KeyError: <bound method AerBackend._run of AerSimulator('aer_simulator')>

During handling of the above exception, another exception occurred:

  File "/home/ubuntu/.local/lib/python3.8/site-packages/distributed/protocol/pickle.py", line 49, in dumps
    result = pickle.dumps(x, **dump_kwargs)
AttributeError: Can't pickle local object 'Client.__init__.<locals>.<lambda>'

During handling of the above exception, another exception occurred:

  File "dask_test_1.py", line 34, in <module>
    q_exec()
  File "dask_test_1.py", line 30, in q_exec
    result = qbackend.run(circ_list).result()
  File "/home/ubuntu/.local/lib/python3.8/site-packages/qiskit/utils/deprecation.py", line 62, in wrapper
    return func(*args, **kwargs)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/qiskit/providers/aer/backends/aerbackend.py", line 222, in run
    aer_job.submit()
  File "/home/ubuntu/.local/lib/python3.8/site-packages/qiskit/providers/aer/jobs/aerjobset.py", line 90, in submit
    aer_job.submit()
  File "/home/ubuntu/.local/lib/python3.8/site-packages/qiskit/providers/aer/jobs/aerjob.py", line 58, in submit
    self._future = self._executor.submit(self._fn, self._qobj, self._job_id)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/distributed/client.py", line 1550, in submit
    futures = self._graph_to_futures(
  File "/home/ubuntu/.local/lib/python3.8/site-packages/distributed/client.py", line 2571, in _graph_to_futures
    dsk = dsk.__dask_distributed_pack__(self, keyset, annotations)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/dask/highlevelgraph.py", line 1063, in __dask_distributed_pack__
    "state": layer.__dask_distributed_pack__(
  File "/home/ubuntu/.local/lib/python3.8/site-packages/dask/highlevelgraph.py", line 421, in __dask_distributed_pack__
    dsk = toolz.valmap(dumps_task, dsk)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/toolz/dicttoolz.py", line 83, in valmap
    rv.update(zip(d.keys(), map(func, d.values())))
  File "/home/ubuntu/.local/lib/python3.8/site-packages/distributed/worker.py", line 3847, in dumps_task
    return {"function": dumps_function(task[0]), "args": warn_dumps(task[1:])}
  File "/home/ubuntu/.local/lib/python3.8/site-packages/distributed/worker.py", line 3811, in dumps_function
    result = pickle.dumps(func, protocol=4)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/distributed/protocol/pickle.py", line 60, in dumps
    result = cloudpickle.dumps(x, **dump_kwargs)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 602, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle '_asyncio.Task' object

Issue #1260, opened 2021-05-31
Unitary simulation segmentation fault with 16 qubits


mhinkie:
### Informations

- **Qiskit Aer version**: 0.8.2
- **Python version**: 3.8.5
- **Operating system**: KDE neon 5.21

### What is the current behavior?
aer_simulator with save_unitary() instruction gives a segmentation fault when using 16 qubits, while working as expected with 15 qubits. It does not indicate if this is a problem with resource availability (not enough RAM to simulate), or some other problem.

### Steps to reproduce the problem
```
from qiskit import QuantumCircuit, Aer, transpile

circ = QuantumCircuit(15)

simulator = Aer.get_backend('aer_simulator')
circ = transpile(circ, simulator)
circ.save_unitary()
result = simulator.run(circ).result()
unitary = result.get_unitary(circ)

print(unitary)
```

gives: 
```
$ python3 test.py 
[[1.+0.j 0.+0.j 0.+0.j ... 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 1.+0.j 0.+0.j ... 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 1.+0.j ... 0.+0.j 0.+0.j 0.+0.j]
 ...
 [0.+0.j 0.+0.j 0.+0.j ... 1.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 0.+0.j ... 0.+0.j 1.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 0.+0.j ... 0.+0.j 0.+0.j 1.+0.j]]
```

But:
```
from qiskit import QuantumCircuit, Aer, transpile

circ = QuantumCircuit(16)

simulator = Aer.get_backend('aer_simulator')
circ = transpile(circ, simulator)
circ.save_unitary()
result = simulator.run(circ).result()
unitary = result.get_unitary(circ)

print(unitary)
```

gives:
```
$ python3 test.py 
Segmentation fault
```

### What is the expected behavior?
If the segmentation fault comes from an error in the simulator, the expected result should be the identity matrix (for this particular circuit) for 16 qubits.
If the segmentation fault stems from missing memory, the expected result would be an error message similar to:
```
from qiskit import QuantumCircuit, QuantumRegister, Aer, transpile

reg = QuantumRegister(36)
circ = QuantumCircuit(reg)

circ.h(reg)
circ.x(reg)
circ.measure_all()

[...]

PR #1282, opened 2021-06-24
fix a bug in multiple_chunk_required

hhorii:
### Summary

Throw an error if memory is short

### Details and comments

Since multi-chunk was introduced, mulit-chunk is enabled if memory is short because of following code:

```
bool Controller::multiple_chunk_required(const Circuit &circ,
                                const Noise::NoiseModel &noise) const
{
  if(circ.num_qubits < 3)
    return false;

  if(num_process_per_experiment_ > 1 || Controller::get_min_memory_mb() < required_memory_mb(circ, noise))
    return true;

  if(cache_block_qubit_ >= 2 && cache_block_qubit_ < circ.num_qubits)
    return true;

  return false;
}
```

The second branch condition enable multi-chunk if a number of processes is two or more **OR** local memory is smaller than required memory for state.
This PR changes this **OR** with **AND**.

This PR will resolve https://github.com/Qiskit/qiskit-aer/issues/1260.

### DIFF
--- a/.github/workflows/tests_windows.yml
+++ b/.github/workflows/tests_windows.yml
@@ -119,8 +119,11 @@ jobs:
       - name: Run Tests
         env:
           QISKIT_SUPPRESS_PACKAGING_WARNINGS: Y
+          LANG: 'C.UTF-8'
+          PYTHONIOENCODING: 'utf-8:backslashreplace'
         run: |
           set -e
+          chcp.com 65001
           pip check
           stestr run --slowest
         shell: bash
@@ -164,8 +167,11 @@ jobs:
       - name: Run Tests
         env:
           QISKIT_SUPPRESS_PACKAGING_WARNINGS: Y
+          LANG: 'C.UTF-8'
+          PYTHONIOENCODING: 'utf-8:backslashreplace'
         run: |
           set -e
           pip check
+          chcp.com 65001
           stestr run --slowest
         shell: bash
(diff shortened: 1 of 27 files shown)


chriseclectic (commented):
LGTM. Can you add a short release note for the the bug fix?

doichanj:
Second branch should be separated to select multi-chunk mode when using GPU method. 
The code is like this. 

```
bool Controller::multiple_chunk_required(const Circuit &circ,
                                         const Noise::NoiseModel &noise) const 
{
  if (circ.num_qubits < 3)
    return false;
  if (cache_block_qubit_ >= 2 && cache_block_qubit_ < circ.num_qubits)
    return true;

  if(num_process_per_experiment_ == 1 && sim_device_ == Device::GPU){
    return (max_gpu_memory_mb_ / num_gpus_ < required_memory_mb(circ, noise));
  }
  if(num_process_per_experiment_ > 1){
    size_t total_mem = max_memory_mb_;
    if(sim_device_ == Device::GPU)
      total_mem += max_gpu_memory_mb_;
    if(total_mem*num_process_per_experiment_ > required_memory_mb(circ, noise))
      return true;
  }

  return false;
}
```

hhorii:
@doichanj The above code works in aer_controller.hpp but not in controller.hpp that does not have `sim_device_ `.  Instead of `sim_device_ == Device::GPU`, 3bfe7b5 uses `cache_block_qubit_ >= 2`.

doichanj:
I think it is enough for controller.hpp

chriseclectic (approved):

PR #1287, opened 2021-07-06
GPU optimization of multi-shots simulations

doichanj:
### Summary
This is the performance upgrade of GPU simulators (statevector and density matrix simulators) of multiple shots with noise model / conditional branches ~~and multiple experiments (circuits) simulations~~. Before this PR, GPU simulators are slower than CPU simulators with multi-shots ~~/experiments~~ (about 10 times slower than CPU)

### Details and comments
This PR accelerates simulations of noisy circuits / circuits with conditional branches that requires multiple-shots ~~, and multiple circuits simulations~~ by using GPUs and multiple GPU nodes over MPI.

For noisy simulation, noise models are applied at runtime (Kraus noise and Pauli noise), so new GPU simulator applies the same operations to multiple states in a batch, that reduces kernel overhead of CUDA. And conditional branches are now implemented on GPU so each states are parallelly simulated with different conditions. 

Multiple circuit parallelization is dropped from this PR because source codes are too complex.


src/controllers/aer_controller.hpp
```diff
@@ -819,27 +841,67 @@ Result Controller::execute(const inputdata_t &input_qobj) {
...
+      if(max_required == 0){
+        max_batched_states_ = 1;
+      }
+      else{
+        if(sim_device_ == Device::GPU){
+          max_batched_states_ = ((max_gpu_memory_mb_/num_gpus_*8/10) / max_required)*num_gpus_;
+        }
+        else{
+          max_batched_states_ = (max_memory_mb_*8/10) / max_required;
+        }
```
hhorii, line 902: All the above check codes should be in `set_parallelization_experiments()`, which determines experiments-level configuration.

src/controllers/aer_controller.hpp
```diff
@@ -887,20 +949,61 @@ Result Controller::execute(std::vector<Circuit> &circuits,
...
     }
 #endif
+
     // then- and else-blocks have intentionally duplication.
     // Nested omp has significant overheads even though a guard condition
     // exists.
     const int NUM_RESULTS = result.results.size();
-    if (parallel_experiments_ > 1) {
-#pragma omp parallel for num_threads(parallel_experiments_)
+    if (parallel_experiments_ > 1 && sim_device_ != Device::GPU && !multi_chunk_req) {
```
hhorii, line 957: can we set `parallel_experiments = 1` in the extended `set_parallelization_experiments()`? Then this change is not necessary.

src/controllers/aer_controller.hpp
```diff
@@ -887,20 +949,61 @@ Result Controller::execute(std::vector<Circuit> &circuits,
...
+              batch_enable = false;
+              break;
+            }
+            if(!check_measure_sampling_opt(circuits[j], methods[j])){
+              batch_enable = false;
+              break;
+            }
+          }
+        }
+      }
```
hhorii, line 986: Can we move the above change to `set_parallelization_experiments()`? (In that case `batch_enable` will be a class field.)

src/controllers/aer_controller.hpp
```diff
@@ -189,6 +187,12 @@ class Controller {
                           const Method method, bool cache_block,
                           ExperimentResult &result) const;
 
+  template <class State_t>
+  void run_batched_circuits_helper(const std::vector<Circuit> &circs, const Noise::NoiseModel &noise,
+                          const json_t &config, 
+                          const Method method, 
+                          Result &result) const;
+
```
hhorii, line 195: `run_batched_circuits()` and `run_batched_circuits_helper()` are duplicated.  We can integrate them with `run_circuit()` and two `run_circuit_helper()` with templating two arguments: `Simulatable_t` and `Result_t`. `shots` and `seed` can be retrieved from each `circuit`. Though a flag for multi-cunk is not necessary for batched execution, setting a dummy flag (as `false`) does not produce any confusion, I believe.
```
  template <class Simulatable_t, class Result_t>
  void run_circuit(const Simulatable_t &circ, const Noise::NoiseModel &noise,
                   const Method method,const json_t &config, Result_t &result, bool multi_chunk) const;

  template <class State_t>
  void run_circuit_helper(const Circuit &circ, const Noise::NoiseModel &noise,
                          const json_t &config, const Method method, const bool cache_block,
                          ExperimentResult &result) const;

  template <class State_t>
  void run_circuit_helper(const std::vector<Circuit> &circs, const Noise::NoiseModel &noise,
                          const json_t &config, const Method method, const bool cache_block,
                          Result &result) const;
```

src/controllers/aer_controller.hpp
```diff
@@ -1404,85 +1525,262 @@ Transpile::Fusion Controller::transpile_fusion(Method method,
...
-  // General circuit noise sampling
-  else {
-    run_circuit_with_sampled_noise(circ, noise, config, shots, state, method,
-                                   cache_blocking, result, rng);
-    state.add_metadata(result);
-    result.metadata.add("circuit", "noise");
-    return;
+      if(noise_sampling){
+        run_circuit_with_sampled_noise<State_t>(circ, noise, config, shots, method,
+                                       cache_blocking, result, rng_seed);
```
hhorii, line 1611: I guess that `rng_seed` should be replaced with `rng` to guarantee randomness.

src/controllers/aer_controller.hpp
```diff
@@ -1404,85 +1525,262 @@ Transpile::Fusion Controller::transpile_fusion(Method method,
...
+      }
+      else{
+        // Run multishot simulation without noise sampling
+        run_circuit_without_sampled_noise<State_t>(opt_circ, noise, config, shots, 
+                                          method, cache_blocking, result, rng_seed);
```
hhorii, line 1616: Same with the above comment.

src/controllers/aer_controller.hpp
```diff
@@ -1404,85 +1525,262 @@ Transpile::Fusion Controller::transpile_fusion(Method method,
...
+                                    const json_t &config,
+                                    const Method method,
+                                    Result &result) const
+{
+  auto timer_start = myclock_t::now(); // state circuit timer
+
+#pragma omp parallel for
+  for (int j = 0; j < circs.size(); ++j) {
+    // Initialize circuit json return
+    result.results[j].legacy_data.set_config(config);
```
hhorii, line 1655: Can `auto& result = result.results[j];` simplify codes in this loop?

src/controllers/aer_controller.hpp
```diff
@@ -1293,6 +1293,127 @@ void Controller::run_circuit(const Circuit &circ,
   }
 }
 
+void Controller::run_batched_circuits(const std::vector<Circuit> &circs,
+                             const Noise::NoiseModel &noise,
+                             const Method method,const json_t &config,
+                             Result &result) const
```
hhorii, line 1299: I believe that this method can be reduced.

src/controllers/aer_controller.hpp
```diff
@@ -1404,85 +1525,262 @@ Transpile::Fusion Controller::transpile_fusion(Method method,
...
+        opt_circ = noise.sample_noise(circ, rng, Noise::NoiseModel::Method::kraus);
+        result.metadata.add("kraus", "noise");
+      }
+      // General circuit noise sampling
+      else {
+        if(batched_shots_optimization_ && batched_shots_optimization_threshold_ >= max_qubits_ && 
+           sim_device_ == Device::GPU && max_batched_states_ > 1 && max_batched_states_ >= num_gpus_){
+          //for GPU noise sampling is done at runtime
+          opt_circ = noise.sample_noise(circ, rng, Noise::NoiseModel::Method::circuit, true);
+          opt_circ.can_sample = false;
```
hhorii, line 1601: `opt_circ.can_sample = false` should be set in `noise.sample_noise()`.

src/controllers/aer_controller.hpp
```diff
@@ -1404,85 +1525,262 @@ Transpile::Fusion Controller::transpile_fusion(Method method,
...
+    result.metadata.add(circ.remapped_qubits, "remapped_qubits");
+
+    // Add measure sampling to metadata
+    // Note: this will set to `true` if sampling is enabled for the circuit
+    result.metadata.add(false, "measure_sampling");
+    result.metadata.add(false, "batched_shots_optimization");
+
+    if(circ.num_qubits > 0){  //do nothing for query steps
+      // Choose execution method based on noise and method
+      Circuit opt_circ;
```
hhorii, line 1570: We can reduce duplication to generate a sampling circuit in `run_circuit_helper` and `run_batched_circuits_helper` by adding a new method:

```
void Controller::get_sampling_circuit(const Circuit &circ,
                                      const Noise::NoiseModel &noise,
                                      RngEngine &rng,
                                      const Method method,
                                      Circuit &sampling_circuit,
                                      std::string &sampling_method) const;
```

src/controllers/aer_controller.hpp
```diff
@@ -1404,85 +1525,262 @@ Transpile::Fusion Controller::transpile_fusion(Method method,
...
+
+    states.set_parallelization(max_batched_states_);
+    states.allocate(max_qubits_, max_qubits_,circs.size());
+    states.set_config(config);
+
+#pragma omp parallel for if(circs.size() > 1)
+    for (i_circ=0;i_circ< circs.size(); i_circ++) {
+      rng[i_circ].set_seed(circs[i_circ].seed);
+
+      Circuit circ;
```
hhorii, line 1705: We can reduce duplication to generate a sampling circuit.

src/controllers/aer_controller.hpp
```diff
@@ -1404,85 +1525,262 @@ Transpile::Fusion Controller::transpile_fusion(Method method,
...
+      // Kraus noise sampling
+      else if (noise.opset().contains(Operations::OpType::kraus) ||
+               noise.opset().contains(Operations::OpType::superop)) {
+        opt_circ = noise.sample_noise(circ, rng, Noise::NoiseModel::Method::kraus);
+        result.metadata.add("kraus", "noise");
+      }
+      // General circuit noise sampling
+      else {
+        if(batched_shots_optimization_ && batched_shots_optimization_threshold_ >= max_qubits_ && 
+           sim_device_ == Device::GPU && max_batched_states_ > 1 && max_batched_states_ >= num_gpus_){
```
hhorii, line 1598: We can simplify this test expression by making `batch_enable` class field.

src/controllers/aer_controller.hpp
```diff
@@ -1524,65 +1832,161 @@ void Controller::run_circuit_without_sampled_noise(Circuit &circ,
...
-  } else {
+  }
+  else{
     // Perform standard execution if we cannot apply the
     // measurement sampling optimization
-    while (shots-- > 0) {
-      run_single_shot(circ, state, result, rng);
+
+    if(batched_shots_optimization_ && batched_shots_optimization_threshold_ >= max_qubits_ 
+          && sim_device_ == Device::GPU && !cache_blocking && shots > 1 && max_batched_states_ >= num_gpus_){
```
hhorii, line 1869: We can simplify this test expression by making `batch_enable` class field.

src/controllers/aer_controller.hpp
```diff
@@ -887,20 +949,61 @@ Result Controller::execute(std::vector<Circuit> &circuits,
...
       }
-    } else {
-      for (int j = 0; j < result.results.size(); ++j) {
-        execute_circuit(circuits[j], noise_model, methods[j],
-                        config, result.results[j]);
+    }
+    else{
+      bool batch_enable = false;
+      if(batched_shots_optimization_ && batched_shots_optimization_threshold_ >= max_qubits_){
+        if(one_method && sim_device_ == Device::GPU && !multi_chunk_req && circuits.size() > num_gpus_ &&
```
hhorii, line 968: We can simplify this test expression by making batch_enable class field.

src/controllers/aer_controller.hpp
doichanj: Moved to `execute` function to remove duplicate codes

src/controllers/aer_controller.hpp
doichanj: integrated these functions

src/controllers/aer_controller.hpp
doichanj: integrated and removed

src/controllers/aer_controller.hpp
doichanj: removed from controller

src/controllers/aer_controller.hpp
doichanj: moved to `set_parallelization_experiments()`

src/controllers/aer_controller.hpp
doichanj: moved to `set_parallelization_experiments()`

src/controllers/aer_controller.hpp
doichanj: moved to `set_parallelization_experiments()`

src/controllers/aer_controller.hpp
doichanj: moved to `set_parallelization_experiments()`

src/controllers/aer_controller.hpp
doichanj: moved to `set_parallelization_circuit()`

src/controllers/aer_controller.hpp
doichanj: created a function for noise sampling

src/controllers/aer_controller.hpp
doichanj: created a function for noise sampling

src/controllers/aer_controller.hpp
```diff
@@ -887,20 +963,65 @@ Result Controller::execute(std::vector<Circuit> &circuits,
...
-    // Nested omp has significant overheads even though a guard condition
-    // exists.
+
     const int NUM_RESULTS = result.results.size();
-    if (parallel_experiments_ > 1) {
-#pragma omp parallel for num_threads(parallel_experiments_)
-      for (int j = 0; j < result.results.size(); ++j) {
-        execute_circuit(circuits[j], noise_model, methods[j],
-                        config, result.results[j]);
+    bool batch_enable = true;
```
hhorii, line 968: `batch_enable` is not used.

chriseclectic (commented):
I had quite a difficult time understanding what is going on with these changes. While I am not particular concerned with the technical details (which I can't really hope to understand!) I am worried that there are a lot of changes to common base class APIs that are not documented.

Particularly adding a lot of new methods without a description of what they do, so it's hard to tell if they are all necessary to be in the base classes or not. There also now appears to be a lot of duplication in classes that are supposed to be "base" but are not directly related to each other, such as State, StateChunk, MultiStates. This makes the interface quite confusing.

My main sources of concern are:

1. It is no longer clear to me how the basic class hierarchy is supposed to work. With regards to controler -> State -> qreg (this was already muddied when StateChunk was added, but was not a subclass of State). I'm not entirely sure where `MultiStates` fits in relative to `State` since it is not documented. Can it work with any State class? Is it another alternative state like StateChunk was?

2. This appears to change the responsibilities of controller and state classes, so that some things that were formerly handled by the controller are now handled by States. I think the API for this should be thought about a little more to make it clearer and better documented for developers who have to look at this code and possible build their own simulators for it. 

3. A lot of the functionality here seems very specific to GPU statevector simulation, but creeps into the base classes that effect everything else. How much is general to multi-shot optimization and could be implemented by an arbitrary simulation method (eg if I wanted to add batched execution to stabilizer method). Is this isn't possible then it probably shouldn't have so many changes to the base state and controller classes since it is a method specific implementation. If it is possible it should be clear from teh API and documentation (related to previous points).

4. The way noise sampling is handled by modifying op to store circuits is a bit too hacky. If we are going to make this change ALL noise should be modified to be handled this way. But I think a better approach would be to re-do how this is handled by having circuits contain quantum error locations (as was recently added), which can be looked up uniquely in a noise model. Then sampling can be done at runtime as the method for how these quantum error location ops are applied. This might require a bit more refactoring of the C++ noise models which I had planned to do later, but maybe need to do a little earlier for this...

src/noise/quantum_error.hpp
```diff
@@ -160,23 +160,42 @@ QuantumError::NoiseOps QuantumError::sample_noise(const reg_t &qubits,
...
-      auto r = rng.rand_int(probabilities_);
-      // Check for invalid arguments
-      if (r + 1 > circuits_.size()) {
-        throw std::invalid_argument(
-          "QuantumError: probability outcome (" + std::to_string(r) + ")"
-          " is greater than number of circuits (" + std::to_string(circuits_.size()) + ")."
-        );
+      if(sample_at_runtime){
+        NoiseOps noise_op(1);
+        noise_op[0].type = Operations::OpType::runtime_error;
```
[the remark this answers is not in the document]
hhorii: How about `runtime_noise_sampling`?

hhorii (changes requested):

src/noise/quantum_error.hpp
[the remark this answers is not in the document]
doichanj: I agree with Chris that `OpType::runtime_error` was bad naming. I removed this operation.

src/noise/quantum_error.hpp
```diff
@@ -160,23 +160,42 @@ QuantumError::NoiseOps QuantumError::sample_noise(const reg_t &qubits,
...
+        noise_op[0].name = "runtime_error";
```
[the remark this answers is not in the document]
doichanj: So I replaced `OpType::runtime_error` to `OpType::qerror_loc` to sample noise at runtime. Noise model returns as is if the operation is `qerror_loc` and if conventional noise sampling method is used, `qerror_loc` is generated before of after the operation that causes noise.

src/framework/operations.hpp
```diff
@@ -214,6 +218,9 @@ struct Op {
   Clifford::Clifford clifford;
   mps_container_t mps;
 
+  //circuits selected by probability (used for runtime_error)
+  std::vector<std::vector<Op>> circs;
```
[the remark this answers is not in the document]
doichanj: I removed this one, and I added reference to noise model to sample noise at runtime for batched multi-shots optimization.

src/framework/operations.hpp
```diff
@@ -162,6 +162,9 @@ inline std::ostream& operator<<(std::ostream& stream, const OpType& type) {
   case OpType::noise_switch:
     stream << "noise_switch";
     break;
+  case OpType::runtime_error:
```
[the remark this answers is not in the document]
doichanj: I removed this operation

src/noise/noise_model.hpp
```diff
@@ -56,7 +56,9 @@ class NoiseModel {
   // superop: each noisy gate or reset will be returned as a single superop 
   Circuit sample_noise(const Circuit &circ,
                        RngEngine &rng,
-                       const Method method = Method::circuit) const;
+                       const Method method = Method::circuit,bool sample_at_runtime = false) const;
```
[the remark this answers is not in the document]
doichanj: I think runtime sampling flags in noise model is needed when circuit does not have `qerror_loc` but conventional sampling method is used.
If the input circuit is converted to use `qerror_loc` outside of Aer, we can remove this flag for runtime noise sampling.

src/simulators/density_matrix/densitymatrix.hpp
```diff
@@ -95,6 +95,9 @@ class DensityMatrix : public UnitaryMatrix<data_t> {
   // The matrix is input as vector of the matrix diagonal.
   void apply_diagonal_superop_matrix(const reg_t &qubits, const cvector_t<double> &mat);
 
+  //batched matrix is used for GPU optimization
+  void apply_batched_matrix(std::vector<batched_matrix_params>& params,reg_t& qubits,std::vector<std::complex<double>>& matrices){}
```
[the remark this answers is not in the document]
doichanj: Currently most of functions needed for batched multi-shots optimization is implemented for GPU only. So there are many empty functions to prevent compile errors.

src/simulators/statevector/batched_matrix.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+#include "framework/operations.hpp"
+
+namespace AER {
+namespace QV {
+
+class alignas(128) batched_matrix_params 
+{
+public:
+  std::complex<double> matrix2x2_[4];
```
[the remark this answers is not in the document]
doichanj: Most of operations can be calculated by 2x2 so this is used to avoid storing matrix to GPU. Larger matrices (e.g. 4x4) are stored in the array and `offset_matrix_` points the top of stored matrix.

src/simulators/statevector/batched_matrix.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+#ifndef _qv_batched_matrix_hpp_
+#define _qv_batched_matrix_hpp_
+
+#include "framework/operations.hpp"
+
+namespace AER {
+namespace QV {
+
+class alignas(128) batched_matrix_params
```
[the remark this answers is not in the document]
doichanj: I will add comments in the header file

src/simulators/state.hpp
```diff
@@ -187,19 +220,52 @@ class State {
...
 
   // Initialize classical memory and register to specific values
-  void initialize_creg(uint_t num_memory,
+  virtual void initialize_creg(uint_t num_memory,
                        uint_t num_register,
                        const std::string &memory_hex,
                        const std::string &register_hex);
 
+  //set conditional regisiter (if op is conditional)
+  virtual void set_conditional(const Operations::Op &op){}
```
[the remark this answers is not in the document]
doichanj: This is not used now, I think I forgot to remove from the experimental codes.

src/simulators/statevector/statevector_state.hpp
```diff
@@ -529,18 +653,39 @@ void State<statevec_t>::set_config(const json_t &config) {
...
+{
+  if(BaseState::qreg_.batched_optimization_supported()){
+    BaseState::qreg_.apply_bfunc(op);
+  }
+  else
+    BaseState::creg_.apply_bfunc(op);
+}
+
+template <class statevec_t>
+void State<statevec_t>::set_conditional(const Operations::Op &op)
```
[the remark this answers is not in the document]
doichanj: I removed

src/simulators/statevector/qubitvector.hpp
```diff
@@ -269,6 +292,34 @@ class QubitVector {
...
 
+
+  //-----------------------------------------------------------------------
+  // for batched optimization (these are not used for CPU)
+  //-----------------------------------------------------------------------
+  virtual bool batched_optimization_supported(void)
+  {
+    return false;
+  }
+  virtual void apply_bfunc(const Operations::Op &op){}
```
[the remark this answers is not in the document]
doichanj: GPU kernel should refer the classical registers to handle conditional branch for each shot (each shot has different condition). This is needed to set classical registers on GPU's memory. 
This is only used for multi-shots optimization.

src/controllers/aer_controller.hpp
```diff
@@ -298,6 +289,17 @@ class Controller {
                            const Noise::NoiseModel &noise,
                            const json_t &config) const;
 
+  //return maximum number of qubits for matrix
+  int_t get_max_matrix_bits(const Circuit &circ) const;
```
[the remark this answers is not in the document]
doichanj: replaced

src/controllers/aer_controller.hpp
```diff
@@ -832,24 +872,29 @@ Result Controller::execute(std::vector<Circuit> &circuits,
...
-        // catch exception raised by required_memory_mb because of invalid
-        // simulation method
-        set_parallelization_experiments(circuits, noise_model, methods);
-      } catch (std::exception &e) {
-        save_exception_to_results(result, e);
-      }
+    if(multi_chunk_req)
+      num_process_per_experiment_ = num_processes_;
+    else
+      num_process_per_experiment_ = 1;
```
hhorii, line 889: I think above 14 lines can be following 4 lines.
```
    std::vector<bool> multi_chunk;
    for (int i = 0; i < circuits.size(); ++i)
      multi_chunk.push_back(multiple_chunk_required(circuits[i], noise_model, methods[i]));
    bool multi_chunk_req = std::accumulate(multi_chunk.begin(), multi_chunk.end(), false);
    num_process_per_experiment_ = multi_chunk_req? num_processes_: 1;
```

src/framework/operations.hpp
```diff
@@ -33,14 +33,14 @@ namespace Operations {
 // Comparisons enum class used for Boolean function operation.
 // these are used to compare two hexadecimal strings and return a bool
 // for now we only have one comparison Equal, but others will be added
-enum class RegComparison {Equal, NotEqual, Less, LessEqual, Greater, GreaterEqual};
+enum class RegComparison {Equal, NotEqual, Less, LessEqual, Greater, GreaterEqual, Nop};
```
hhorii, line 36: `RegComparison::Nop` is not used anywhere, I think.

src/controllers/aer_controller.hpp
```diff
@@ -1497,92 +1438,206 @@ void Controller::run_single_shot(const Circuit &circ, State_t &state,
...
+
+    if(enable_batch_multi_shots_ && !cache_block_pass.enabled()){
+      //apply batched multi-shots optimization on GPU
+      Base::MultiStates<State_t> states;
+
+      states.set_parallelization(max_batched_states_);
+
+      states.set_max_matrix_qubits(get_max_matrix_qubits(circ) );
+
+      states.allocate(circ.num_qubits, circ.num_qubits,circ.shots);
```
hhorii, line 1513: This second argument (`circ.num_qubits`) is different from others (`block_bits`). Is this intentional?

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -516,6 +516,9 @@ def _default_options(cls):
             accept_distributed_results=None,
             blocking_qubits=None,
             blocking_enable=False,
+            memory_blocking_bits=0,
+            batched_shots_optimization=True,
```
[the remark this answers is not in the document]
chriseclectic: This still needs a description added to the AerSimulator docs

src/framework/operations.hpp
chriseclectic: Is this necessary?

src/framework/operations.hpp
```diff
@@ -33,14 +33,14 @@ namespace Operations {
...
 
 // Enum class for operation types
 enum class OpType {
   gate, measure, reset, bfunc, barrier, qerror_loc, snapshot,
   matrix, diagonal_matrix, multiplexer, initialize, sim_op, nop,
   // Noise instructions
-  kraus, superop, roerror, noise_switch,
+  kraus, superop, roerror, noise_switch,
```
chriseclectic, line 43:
```suggestion
  kraus, superop, roerror, noise_switch,
```

src/framework/rng.hpp
```diff
@@ -53,8 +53,17 @@ class RngEngine {
...
   // Set a fixed seed for the RNG engine
-  void set_seed(size_t seed) { rng.seed(seed); }
-  
+  void set_seed(size_t seed)
+  {
+    rng.seed(seed); 
+    initial_seed_ = seed;
+  }
+
+  size_t seed(void)
```
chriseclectic, line 62:
```suggestion
  size_t initial_seed(void)
```

src/framework/operations.hpp
```diff
@@ -184,6 +184,7 @@ inline std::ostream& operator<<(std::ostream& stream, const OpType& type) {
 //------------------------------------------------------------------------------
 // Op Class
 //------------------------------------------------------------------------------
+struct Op;
```
chriseclectic, line 187: I dont think this is needed anymore
```suggestion
```

src/simulators/density_matrix/densitymatrix.hpp
```diff
@@ -128,6 +131,9 @@ class DensityMatrix : public UnitaryMatrix<data_t> {
   // outcome in [0, 2^num_qubits - 1]
   virtual double probability(const uint_t outcome) const override;
 
+
+  virtual void apply_reset(const reg_t& qubits){}
```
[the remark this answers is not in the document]
chriseclectic: Why is this an empty method?

src/simulators/density_matrix/densitymatrix_state.hpp
```diff
@@ -297,6 +299,18 @@ class State : public Base::State<densmat_t> {
 
   // Table of allowed snapshot types to enum class members
   const static stringmap_t<Snapshots> snapshotset_;
+
+  //scale for density matrix = 2
+  virtual int qubit_scale(void)
```
chriseclectic, line 304: Is this actually used anywhere?

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -472,6 +472,9 @@ def _default_options(cls):
             accept_distributed_results=None,
             blocking_qubits=None,
             blocking_enable=False,
+            memory_blocking_bits=0,
```
chriseclectic, line 475: What does this do?

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -472,6 +472,9 @@ def _default_options(cls):
...
+            batched_shots_optimization=True,
+            batched_shots_optimization_threshold=16,
```
chriseclectic, line 477: What does this mean, you need to add docs that explain what htis threshold value is.
If its num qubits  why so high? I thought the point of this to speed up low number of qubits?

chriseclectic (changes requested):
This looks a lot cleaner now. I think there might be a few left over lines that should have been removed.

My biggest concern now is that there is no documentation for the API docs or release notes explaining what was added and how to use it.

There also don't appear to be any new unittests for using these options. Is it possible to add tests for basic gates / instructions etc that run when setting this option on and off to check it is enabled/disabled, and returns the correct results in both cases for the methods that use it? (both Thrust and CPU)

qiskit/providers/aer/backends/aer_simulator.py
doichanj: memory_blocking_bits is an experimental flag so I removed from aer_simulator.py and this is disabled by default.

This option is used to enable GPU shared memory blocking to decrease memory access (GPU can simulate up to 10 qubits on shared memory without accessing main memory)

qiskit/providers/aer/backends/aer_simulator.py
doichanj: I added documents of batched_shots_optimization options in CONTRIBUTING.md 
batched shots optimization will be enabled when number of qubits of circuit is less than threshold. Because this technique does not improve performance for large number of qubits.

doichanj:
CI fails because terra does not support Python 3.6. So I dropped 3.6 from Linux tests.

src/simulators/density_matrix/densitymatrix_state.hpp
doichanj: Currently multi-shots optimization is only supported if device=GPU. And Thrust(CPU) is not supported. So we can not test batched_shots_optimization on CI. 
batched_shots_optimization is enabled by default for smaller number of qubits, so when running tests on local machine with GPU, all the test cases that require multiple shots use this option. (all the test is passed on local machine.)

qiskit/providers/aer/backends/aer_simulator.py
doichanj: I added documentation for batched_shots_optimization options in aer_simulator.py

chriseclectic (approved):
LGTM. I changed the name of the new options to `batched_shots_gpu` since I think that is clearer to users.

PR #1292, opened 2021-07-13
fix test errors due to recent terra updates

hhorii:
### Summary

Fix test fails from recent Terra's updates

### Details and comments

This PR will fix some test errors due to recent updates of Terra.

### DIFF
--- a/test/terra/backends/qasm_simulator/qasm_noise.py
+++ b/test/terra/backends/qasm_simulator/qasm_noise.py
@@ -163,7 +163,7 @@ class QasmKrausNoiseOnQFTTests:
         shots = 10000
         noise_model = ref_kraus_noise.kraus_gate_error_noise_models_full()
 
-        circuit = QFT(3)
+        circuit = QFT(3).decompose()
         circuit.measure_all()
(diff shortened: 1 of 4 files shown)


test/terra/backends/qasm_simulator/qasm_noise.py
```diff
@@ -163,7 +163,7 @@ def test_kraus_gate_noise_on_QFT(self):
         shots = 10000
         noise_model = ref_kraus_noise.kraus_gate_error_noise_models_full()
 
-        circuit = QFT(3)
+        circuit = QFT(3).decompose()
```
mtreinish, line 166: Is there a reason we skip transpile here?

mtreinish (approved):
One inline question, but it's not critical and getting things unblocked is higher priority.

test/terra/backends/qasm_simulator/qasm_noise.py
hhorii: I just wanted to fix this issue with minimum changes. I think we can add transpile here with basis gates of a noise model.

PR #1293, opened 2021-07-13
Set utf8 character encoding in windows CI

mtreinish:
### Summary

We've recently started seeing a failure in windows CI jobs recently
while the test runner is trying to print some text qiskit is emitting
that is outside the default windows code page. This commit configures
the windows ci test jobs to use utf8 instead of the default windows code
page as that is the default encoding for qiskit. This should hopefully
fix the ci failures (or at least be able to print any errors emitted).


mtreinish:
Baked into #1292 as both are needed to fix ci

Issue #1289, opened 2021-07-12
Statevector simulator results in a segmentation fault for simple 16 qubit test


Greken79:
### Informations

- **Qiskit Aer version**:
{'qiskit-terra': '0.17.4', 'qiskit-aer': '0.8.2', 'qiskit-ignis': '0.6.0', 'qiskit-ibmq-provider': '0.14.0', 'qiskit-aqua': '0.9.2', 'qiskit': '0.27.0', 'qiskit-nature': None, 'qiskit-finance': None, 'qiskit-optimization': None, 'qiskit-machine-learning': None}
- **Python version**:3.8.11
- **Operating system**:Windows 10

### What is the current behavior?
When running a simple test of the state vector simulator with 16 qubits I get a segmentation fault. The script works until 15 qubits. I am using ssh to run the code on a data cluster with a container with the same packages as on my own computer. This code works on my own computer up until 28 qubits when I run out of memory.

When running the code I get:
`Fatal Python error: Segmentation fault

Current thread 0x00007f1ef3462700 (most recent call first):
  File "/usr/local/lib/python3.8/site-packages/qiskit/providers/aer/backends/backend_utils.py", line 42 in cpp_execute
  File "/usr/local/lib/python3.8/site-packages/qiskit/providers/aer/backends/statevector_simulator.py", line 211 in _execute
  File "/usr/local/lib/python3.8/site-packages/qiskit/providers/aer/backends/aerbackend.py", line 265 in _run
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57 in run
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 80 in _worker
  File "/usr/local/lib/python3.8/threading.py", line 870 in run
  File "/usr/local/lib/python3.8/threading.py", line 932 in _bootstrap_inner
  File "/usr/local/lib/python3.8/threading.py", line 890 in _bootstrap

Thread 0x00007f1fb9804740 (most recent call first):
  File "/usr/local/lib/python3.8/threading.py", line 302 in wait
  File "/usr/local/lib/python3.8/threading.py", line 558 in wait
  File "/usr/local/lib/python3.8/threading.py", line 857 in start
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 213 in _adjust_thread_count
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 188 in submit
[...]

PR #1294, opened 2021-07-15
fix a bug to generate a diagonal gate in the end of circuit

hhorii:
### Summary

This PR fixes the issue #1289 

### Details and comments

Fusion to generate diagonal gates has a bug to look up operations in a circuit and this PR fix this bug.

### DIFF
--- a/src/transpile/fusion.hpp
+++ b/src/transpile/fusion.hpp
@@ -658,8 +658,10 @@ bool DiagonalFusion::aggregate_operations(oplist_t& ops,
       continue;
 
     std::vector<uint_t> fusing_op_idxs;
-    for (; op_idx < next_diagonal_start; ++op_idx)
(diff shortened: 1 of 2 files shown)


src/transpile/fusion.hpp
```diff
@@ -658,7 +658,7 @@ bool DiagonalFusion::aggregate_operations(oplist_t& ops,
       continue;
 
     std::vector<uint_t> fusing_op_idxs;
-    for (; op_idx < next_diagonal_start; ++op_idx)
+    for (; op_idx < next_diagonal_start && op_idx < fusion_end; ++op_idx)
```
chriseclectic, line 661: This looks like odd syntax for a `for` loop. Should it be changed to a `while` loop instead?

chriseclectic (approved):

PR #1303, opened 2021-08-04
Fix test deprecation warnings and call parent's setUpClass

mtreinish:
### Summary

In Qiskit/qiskit-terra#6753 the base test classes were fixed to assert
that setUp and setUpClass in the base test class are always called. This
is necessary to ensure that all the warning assertions always work.
However this change had the unintended side effect of causing Aer's CI
to fail because it wasn't calling super().setUpClass(). This commit
makes this change to fix that failure. However, because now we're
running enforcement that tests can't raise unhandled
DeprecationWarnings a few spots in the tests were emitting deprecation
warnings. The deprecated syntax is either fixed or if it's testing
something in Aer that's deprecated an assertion on that warning is added
to fix the failure.

### Details and comments

### DIFF
--- a/test/terra/backends/qasm_simulator/qasm_snapshot.py
+++ b/test/terra/backends/qasm_simulator/qasm_snapshot.py
@@ -206,12 +206,12 @@ class QasmSnapshotStabilizerTests:
         # Get stabilizer and destabilizers and convert to sets
         for stab in stabilizer:
             if stab[0] == '-':
-                pauli_mat = -1 * Pauli.from_label(stab[1:]).to_matrix()
+                pauli_mat = -1 * Pauli(stab[1:]).to_matrix()
             else:
-                pauli_mat = Pauli.from_label(stab).to_matrix()
-            val = statevector.conj().dot(pauli_mat.dot(statevector))
-            if not np.isclose(val, 1):
-                return False
+                pauli_mat = Pauli(stab).to_matrix()
+        val = statevector.conj().dot(pauli_mat.dot(statevector))
+        if not np.isclose(val, 1):
+            return False
         return True
 
     def test_snapshot_stabilizer_pre_measure_det(self):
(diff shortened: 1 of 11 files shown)


test/terra/backends/qasm_simulator/qasm_snapshot.py
```diff
@@ -206,12 +206,12 @@ def stabilizes_statevector(stabilizer, statevector):
...
+                pauli_mat = -1 * Pauli(stab[1:]).to_matrix()
             else:
-                pauli_mat = Pauli.from_label(stab).to_matrix()
-            val = statevector.conj().dot(pauli_mat.dot(statevector))
-            if not np.isclose(val, 1):
-                return False
+                pauli_mat = Pauli(stab).to_matrix()
+        val = statevector.conj().dot(pauli_mat.dot(statevector))
+        if not np.isclose(val, 1):
+            return False
```
jakelishman, line 214: This change brings some logic outside a loop, which looks like it belongs inside?  `pauli_mat` depends on the loop variable `stab`.

jakelishman (commented):
Ideally we should work out the split properly, so Aer doesn't need to derive its test class from Terra-specific functionality, but that's something to work out on the Terra side first.

One question about a test that's dropped out of a loop, though it looks like it still belongs in it?

test/terra/backends/qasm_simulator/qasm_snapshot.py
mtreinish: Oops, good catch I messed up the tab indents here

mtreinish: Done in: https://github.com/Qiskit/qiskit-aer/pull/1303/commits/bc9951211434634964784c0fd076782126a94425

jakelishman (approved):

jakelishman:
Oh wait, I have no power on Aer.  Still, you can have a symbolic tick, if not a meaningful one.

chriseclectic (approved):

PR #1290, opened 2021-07-12
[WIP] MPS: Improve performance of apply_measure for a subset of qubits

merav-aharoni:
### Summary
This is an extension of PR #1277 to measuring a subset of the qubits.

PR #1304, opened 2021-08-05
Performance improvement in MPS::sample_measure_using_apply_measure

merav-aharoni:
### Summary
Added a call to `move_all_qubits_to_sorted_ordering` in `sample_measure_using_apply_measure` so that it is only called once for all shots and not once for every shot.

### Details and comments
I added a flag `is_sorted` to the interface of `apply_measure` and `apply_measure_internal` to indicated whether they qubits were already sorted outside these functions. The reason is that the methods `apply_reset` and `apply_initialize` do not agree with sorting the qubits in advance.

### DIFF
--- a/src/simulators/matrix_product_state/matrix_product_state_internal.cpp
+++ b/src/simulators/matrix_product_state/matrix_product_state_internal.cpp
@@ -1472,14 +1472,16 @@ reg_t MPS::sample_measure_using_probabilities_internal(const rvector_t &rnds,
     return samples;
 }
 
-reg_t MPS::apply_measure(const reg_t &sorted_qubits, RngEngine &rng) {
-  // qubits in MPS structure have been moved to their original location
-  // input 'sorted_qubits' have been sorted
-  return apply_measure_internal(sorted_qubits, rng);
+reg_t MPS::apply_measure(const reg_t &qubits, RngEngine &rng, bool is_sorted) {
+  // Move qubits in the MPS structure to their original location
+  // unless we know they were already sorted
+  if (!is_sorted)
+    move_all_qubits_to_sorted_ordering();
+  return apply_measure_internal(qubits, rng, is_sorted);
 }
 
-reg_t MPS::apply_measure_internal(const reg_t &sorted_qubits, 
-				  RngEngine &rng) {
+reg_t MPS::apply_measure_internal(const reg_t &qubits, 
+				  RngEngine &rng, bool is_sorted) {
   // For every qubit, q,  that is measured, we must propagate the effect of its
   // measurement to its neigbors, l and r, and then to their neighbors, and 
   // so on. If r (or l) is measured next, then there is no need to propagate to
@@ -1494,8 +1496,12 @@ reg_t MPS::apply_measure_internal(const reg_t &sorted_qubits,
   // In both cases, we propagate the effect all the way to the left, because 
   // no more qubits will be measured on the left
   reg_t qubits_to_update;
-  uint_t size = sorted_qubits.size();
+  uint_t size = qubits.size();
   reg_t outcome_vector(size);
+
+  reg_t sorted_qubits = qubits;
+  if (!is_sorted)
+    std::sort(sorted_qubits.begin(), sorted_qubits.end());
 
   uint_t next_measured_qubit = num_qubits_-1;
   for (uint_t i=0; i<size; i++) {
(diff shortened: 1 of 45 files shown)


merav-aharoni:
This PR is built on top of #1290 , so best to merge that one before this one. I will make the review easier.

mishmash:
Thanks Merav. This solution seems nice. One comment: I think this is a potentially *big* performance improvement as needing to reorder via swaps at every shot can be very expensive. I've included some relevant runtime data in https://github.com/Qiskit/qiskit-aer/issues/1305. Hopefully this will push the scaling back down to O(n).

merav-aharoni:
@yaelbh , @chriseclectic , can you please review?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1708,7 +1709,6 @@ void MPS::initialize_component_internal(const reg_t &qubits,
 }
 
 void MPS::reset(const reg_t &qubits, RngEngine &rng) {
-  move_all_qubits_to_sorted_ordering();
```
yaelbh, line 1711: Just verifying that you're removing this line because you've reached the conclusion that `reset` will work properly without it, and not by mistake

test/terra/backends/test_qasm_simulator_matrix_product_state.py
```diff
@@ -83,7 +82,6 @@ class TestQasmMatrixProductStateSimulator(
         QasmSnapshotStatevectorTests,
         QasmSnapshotDensityMatrixTests,
         QasmSnapshotProbabilitiesTests,
-        QasmSnapshotStabilizerTests,
```
yaelbh, line 86: Why? MPS is supposed to run well on these tests

yaelbh (commented):
>  The methods apply_reset and apply_initialize do not agree with sorting the qubits in advance.

I don't see why, can you please provide more details?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I thought it could be removed, but now I am having second thoughts. I will revisit this point.

test/terra/backends/test_qasm_simulator_matrix_product_state.py
merav-aharoni: I was getting multiple warnings that MPS does not support save_stabilizer. I did not check what exactly is giving these warnings. It is possible that some of the tests pass and only some give the warning. I can remove from this PR and check more carefully in a separate PR.

yaelbh: Do you see these warnings also when you run the tests in the main branch?
- If yes (my guess): then it's a good idea to check it, and this can be done independently of this PR (either before or after this PR is merged), but there's no reason to remove it from this PR.
- If no, then it's something that's specific to this PR, and so has to be investigated before this PR is merged.

I expect the same warnings to appear also for the statevector tests, unless it has already been removed there.

merav-aharoni: @yaelbh , I made a changes following your comments:
1. I put back the stabilizer tests in mps. I will check this in a separate issue/PR.
2. I put back the sort in `reset`.
3. I added a few comments regarding the sort. `apply_measure_internal` requires the qubits to be sorted by the caller function. I prefer not to put a validity check in the method for performance considerations. So instead, I put comments in all relevant places.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1472,15 +1472,19 @@ reg_t MPS::sample_measure_using_probabilities_internal(const rvector_t &rnds,
...
+    move_all_qubits_to_sorted_ordering();
+  
+  return apply_measure_internal(qubits, rng, is_sorted);
 }
 
+// The caller to apply_measure_internal is responsible to call 
+// move_all_qubits_to_sorted_ordering before calling this function
 reg_t MPS::apply_measure_internal(const reg_t &qubits, 
-				  RngEngine &rng) {
+				  RngEngine &rng, bool is_sorted) {
```
yaelbh, line 1487: It will be clearer and more consistent to remove the `is_sorted` parameter, and make it the caller's responsibility to sort `qubits` as well.

merav-aharoni: Okay, I did as you suggested.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1473,14 +1473,16 @@ reg_t MPS::sample_measure_using_probabilities_internal(const rvector_t &rnds,
...
+  reg_t sorted_qubits = qubits;
+  std::sort(sorted_qubits.begin(), sorted_qubits.end());
+  return apply_measure_internal(sorted_qubits, rng);
 }
 
-reg_t MPS::apply_measure_internal(const reg_t &qubits, 
-				  RngEngine &rng) {
+// The caller to apply_measure_internal is responsible to call 
+// move_all_qubits_to_sorted_ordering before calling this function
+reg_t MPS::apply_measure_internal(const reg_t &sorted_qubits, RngEngine &rng) {
```
yaelbh, line 1485: Add a comment that `sorted_qubits` must be sorted

yaelbh: Is it really expensive to verify that `sorted_qubits` is sorted, and that the qubits are sorted?

yaelbh (approved):

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: I added a comment that `sorted_qubits` must be sorted.
Regarding adding a verification for the above two conditions, the problem is this will be called for every shot, so I think it is a waste to spend time on this.

PR #1310, opened 2021-08-11
Refactor State apply_ops to use iterators

chriseclectic:
### Summary

This refactors the simulator `State` C++ classes `apply_ops` method:
* Changes signature of `apply_ops` to use iterators instead of a vector of Ops. This will allow applying pieces of a circuit without first copying all the operations (eg for measure sampling).
* Adds an abstract `apply_op` method class to base State
* Adds a concrete implementation of `apply_ops` that loops over calls to `apply_op`, since all but the extended stabilizer state perform this same loop to apply a sequence of ops.


hhorii:
LGTM.

PR #1311, opened 2021-08-11
Remove code deprecated in 0.7.0

chriseclectic:
### Summary

- Removes deprecated `backend_options` kwarg from `backend.run`. All run options must now be passed as individual kwargs.
- This removes unused code from C++ noise model that was used by a previously deprecated feature.
- Remove deprecated internal `_run_job` method from `AerBackend`.
-

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -110,7 +110,6 @@ class AerBackend(Backend, ABC):
     @deprecate_arguments({'qobj': 'circuits'})
     def run(self,
             circuits,
-            backend_options=None,  # DEPRECATED
             validate=False,
             **run_options):
         """Run a qobj on the backend.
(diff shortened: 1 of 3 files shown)


qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -110,42 +110,23 @@ def __init__(self,
...
             validate=False,
             **run_options):
         """Run a qobj on the backend.
 
         Args:
             circuits (QuantumCircuit or list): The QuantumCircuit (or list
                 of QuantumCircuit objects) to run
-            backend_options (dict or None): DEPRECATED dictionary of backend options
-                                            for the execution (default: None).
             validate (bool): validate the Qobj before running (default: False).
```
mtreinish, line 120: is this used anywhere now, the deprecation warning about validate in `_run_job` was removed below I didn't think it was used anymore.

mtreinish (commented):
This LGTM, just one inline question about the `validate` kwarg

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: We opted to turn validation off by default, but I don't think we have deprecated it. We could do so though since I don't think its used.

PR #1313, opened 2021-08-12
Fix shots option

chriseclectic:
### Summary

Fixes issue where setting a `shots` backend option was not working the same as setting it as a run option.

### Details and comments

Updated behavior
```python
backend = AerSimulator(shots=99)
counts = backend.run(circuit).result().get_counts(0)
# Total counts should be 99
```

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -339,7 +339,7 @@ class AerBackend(Backend, ABC):
 
         # Add options
         for key, val in self.options.__dict__.items():
-            if val is not None and not hasattr(config, key):
+            if val is not None:
                 setattr(config, key, val)
 
         # Override with run-time options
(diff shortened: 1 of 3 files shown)


qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -133,6 +133,15 @@ def run(self,
...
             qobj = circuits
+            # Qobj values take precendence over run_options
+            # otherwise values from legacy assembly won't be preserved.
+            if not run_options:
+                run_options = qobj.config.__dict__
+            else:
+                run_options = copy.copy(run_options)
+                for key, value in qobj.config.__dict__.items():
+                    if key not in run_options and value is not None:
+                        run_options[key] = value
```
jakelishman, line 144: The code doesn't seem to match the comment, here - things are only copied in from `qobj.config` if they don't exist in `run_options`, so they're not taking precedence.  Does the comment mean it's trying to take precedence over defaults?

chriseclectic: Ah good catch, I had to rework this a couple of times so it was backwards compatible with qobj options and tests passed, but forgot to update comment along with it.

jakelishman (approved):
I have no power here, but here's a tick anyway.  Is this related to Qiskit/qiskit-terra#6902?

PR #1270, opened 2021-06-15
Deprecate snapshot instructions

chriseclectic:
### Summary

Updates `PendingDeprecationWarning`s on snapshot instructions to `DeprecationWarnings`.

PR #1314, opened 2021-08-12
Deprecate non-local quantum errors

### LINKED ISSUE #1302: Deprecate non-local noise in NoiseModel

Non-local noise should be deprecated as this does not work well within the circuit simulation. It is overly sensitive to the order of instructions in a circuit and these sorts of errors should be applied to scheduled circuits via a transpiler pass before being passed to the simulator. It also prevents or complicates certain optimizations we would like to implement in the simulator that depend on sampled noise being local to the corresponding gates.

chriseclectic:
### Summary

Adding non-local quantum errors to a `NoiseModel` has been deprecated due to inconsistencies in how this noise is applied to the optimized circuit. Non-local noise should be manually added to a scheduled circuit in Qiskit using a custom transpiler pass before being run on the simulator.

### Details and comments

Closes #1302

### DIFF
--- a/qiskit/providers/aer/noise/noise_model.py
+++ b/qiskit/providers/aer/noise/noise_model.py
@@ -15,6 +15,7 @@ Noise model class for Qiskit Aer simulators.
 
 import json
 import logging
+from warnings import warn
 
 from qiskit.circuit import Instruction
 from qiskit.providers import BaseBackend, Backend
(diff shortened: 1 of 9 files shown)


jakelishman (commented):
Retrospective approval, since the `.. warning::` directive used here is replaced with `.. deprecated:: 0.9.0` as part of #1270 now.

PR #1266, opened 2021-06-04
add more pybind for direct use of controller classes

hhorii:
### Summary

This PR will exploit `AER::Simulator::*Controller` to python with classes of their arguments.

### Details and comments

This PR aims to reduce overheads of small qubit simulation.

First, `AER::Circuit` and `AER::Operations::Op` are exploited to python.
`AER::Circuit` object will be generated by parsing transplied `QuantumCircuit` for Aer while calling `make_*` to generate operation.
Then, `AER::Simulator::*Controller.execute(std::vector<Circuit>)` can be directly called.

All the configurations will be set with setter methods of `AER::Simulator::*Controller`.

- [x] `AER::Circuit`
- [x] `AER::Operations::Op`
- [ ] ~~`AER::Simulator::QasmController`~~
- [ ] ~~`AER::Simulator::StatevectorController`~~
- [ ] ~~`AER::Simulator::UnitaryController`~~
- [x] `AER::AerController`
- [x] `AER::Noise::NoiseModel`
- [x] Configuration
- [ ] `AER::Statevector::State`

PR #1317, opened 2021-08-13
WIP: Add parameter_binds to aerbackend.run

### LINKED ISSUE #1249: Option for running parameterized circuits with circuit input

In 0.8.x we started the process of migrating the python interface to only work natively with `QuantumCircuit` objects. However, one thing we did not expose as part of that was a way to pass a parameter bind table with the a parameterized circuit. Previously the qobj interface enabled users to pass a parameterizations table in the qobj header and let aer handle the binding. That interface doesn't exist when passing a circuit directly. We need to fix this so a user doesn't lose this functionality when qobj as an input is deprecated and removed.

mtreinish:
### Details and comments

Fixes #1249 

TODO:

- [ ]  More tests
- [ ]  Docs
- [ ]  release notes


hhorii:
#1266 needs this change because it takes `QuantumCircuit` instead of `Qobj`.

Issue #1323, opened 2021-08-23
Move MPI and Dask documentation to API docs


chriseclectic:
Currently documentation for running HPC cluster simulations using MPI or Dask is hidden in the contributing guide. This documentation should be expanded and moved to its own page in the API docs so that it can be viewed from the qiskit website.

PR #1329, opened 2021-08-30
Move Dask documentation to API docs

hitomitak:
### Summary
Move executor description from contributing to API doc as mentioned #1323


qiskit/providers/aer/jobs/__init__.py
```diff
@@ -12,7 +12,134 @@
...
+.. currentmodule:: qiskit.providers.aer.jobs
+
+This module contains classes and functions to manage Aer jobs.
+
+Running with Threadpool and DASK
+================================
+
+Qiskit Aer runs simulation jobs on a single-worker Python multiprocessing ThreadPool executor
+so that all parallelization is handled by low-level OpenMP and CUDA code.
+However to customize job-level parallel execution of multiple circuits a user can specif
```
jakelishman, line 28:
```suggestion
However to customize job-level parallel execution of multiple circuits a user can specify
```

qiskit/providers/aer/jobs/__init__.py
```diff
@@ -12,7 +12,134 @@
...
+a custom multiprocessing executor and control the splitting of circuits using
+the ``executor`` and ``max_job_size`` backend options.
+For large scale job parallelization on HPC clusters Qiskit Aer executors support
+the distributed Clients from the `DASK <http://dask.org>`_.
+
+
+Installation of DASK packages with Aer
+---------------------------------------
+If you want to install dask client at the same time as Qiskit Aer,
+please add `dask` option as follows.
```
jakelishman, line 38:
```suggestion
please add the ``dask`` extra as follows.
```

qiskit/providers/aer/jobs/__init__.py
```diff
@@ -12,7 +12,134 @@
...
+
+.. code-block:: sh
+
+    pip install .[dask]
+
+Usage of executor
+-----------------
+To use Threadpool or DASK as an executor, you need to set
+``executor`` and ``max_job_size`` by ``set_options`` function.
+If both ``executor`` (default None) and `max_job_size` (default None) are set,
```
jakelishman, line 49:
```suggestion
If both ``executor`` (default None) and ``max_job_size`` (default None) are set,
```

qiskit/providers/aer/jobs/__init__.py
```diff
@@ -12,7 +12,134 @@
...
+Aer splits the multiple circuits to some chunk of circuits and submits them to the executor.
+``max_job_size`` can control the number of splitting circuits.
+When ``max_job_size`` is set to 1, multiple circuits are split into
+one circuit and distributed to the executor.
+If user executes 60 circuits with the executor and `max_job_size=1`,
+Aer splits it to 1 circuit x 60 jobs.
```
jakelishman, line 55:
```suggestion
If a user executes 60 circuits with the executor and ``max_job_size=1``,
Aer splits it as 60 jobs each of 1 circuit.
```

qiskit/providers/aer/jobs/__init__.py
```diff
@@ -12,7 +12,134 @@
...
+If 60 circuits and `max_job_size=2`, Aer splits it to 2 circuits x 30 jobs.
```
jakelishman, line 56:
```suggestion
If there are 60 circuits and ``max_job_size=2``, Aer splits it as 30 jobs, each with 2 circuits.
```

qiskit/providers/aer/jobs/__init__.py
```diff
@@ -12,7 +12,134 @@
...
+    exc = ThreadPoolExecutor(max_workers=2)
+    qbackend.set_options(executor=exc)
+    qbackend.set_options(max_job_size=1)
+
+    result = qbackend.run(circ_list).result()
+
+**Example: Dask execution**
+
+Dask client creates multi-processes so you need to
+guard it by ``if __name__ == "__main__":`` block.
```
jakelishman, line 94:
```suggestion
The Dask client uses ``multiprocessing`` so you need to
guard it by an ``if __name__ == "__main__":`` block.
```

qiskit/providers/aer/jobs/__init__.py
```diff
@@ -12,7 +12,134 @@
...
+
+        result = qbackend.run(circ_list).result()
+
+    if __name__ == '__main__':
+        q_exec()
+
+Classes
+=======
+
+The following are the classes used to management job submitting.
```
jakelishman, line 135:
```suggestion
The following are the classes used to manage job submissions.
```

qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -37,27 +37,25 @@
 
 
 class AerJobSet(Job):
-    """A set of cluster jobs.
+    """A set of :class:`AerJob` class for Qiskit Aer simulators.
```
jakelishman, line 40:
```suggestion
    """A set of :class:`~AerJob` classes for Qiskit Aer simulators.
```

qiskit/providers/aer/jobs/aerjob.py
```diff
@@ -24,13 +24,19 @@
...
-    """
+    """AerJob class for Qiskit Aer Simulators."""
 
     def __init__(self, backend, job_id, fn, qobj, executor=None):
+        """ Initializes the asynchronous job.
+
+        Args:
+            backend(AerBackend): the backend used to run the job.
+            job_id(str): a unique id in the context of the backend used to run the job.
+            fn(function): callabled function.
```
jakelishman, line 35:
```suggestion
            fn(function): callable function.
```

We should probably describe what this function is for, and what inputs/outputs it should expect.

qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -37,27 +37,25 @@
...
 
-    def __init__(self, backend, job_id, func, experiments: List[QasmQobj], executor=None):
+    def __init__(self, backend, job_id, fn, experiments: List[QasmQobj], executor=None):
         """AerJobSet constructor.
 
         Args:
             backend(Aerbackend): Aerbackend.
             job_id(int): Job Id.
-            func(fun): Callabled function.
+            fn(function): Callabled function.
```
jakelishman, line 55:
```suggestion
            fn(function): Callable function.
```

We should probably describe what this function is for, and what inputs/outputs it should expect.

hitomitak:
@jakelishman Thank you for fixing the documentation. I committed your suggestions.

jakelishman:
Thanks!  I can't "unresolve" a comment because I don't have write access to Aer, but my last and second-last suggestions had an extra bit to them as well - at the moment, there's an argument just called `fn` with the description "callable function" - should we explain in the documentation what the function is for, and what arguments it takes?

qiskit/providers/aer/jobs/aerjobset.py
hitomitak: I intend to add the following comments. What do you think? 
fn(function): a callabled function to execute qobj on backend. :class:`AerBackend` sets :meth:`AerBackend._run()` function that takes qobj(QasmQobj) and job_id(string) as an argument.

hitomitak:
@jakelishman Sorry, I missed it and unresolved the comment. I added a detailed description of the function to the conversation. Would you check it? I intend to set the same comments to both second-last descriptions of `fn`.

qiskit/providers/aer/jobs/aerjob.py
```diff
@@ -24,13 +24,21 @@
...
 
     def __init__(self, backend, job_id, fn, qobj, executor=None):
+        """ Initializes the asynchronous job.
+
+        Args:
+            backend(AerBackend): the backend used to run the job.
+            job_id(str): a unique id in the context of the backend used to run the job.
+            fn(function): a callable function to execute qobj on backend.
+                :class:`AerBackend` sets :meth:`AerBackend._run()` function that takes
+                qobj(QasmQobj) and job_id(string) as an argument.
```
jakelishman, line 37: I'm not quite sure what the new description means exactly, sorry, but I'm also not familiar enough with Aer to really suggest a proper alternative.  I think a lot of my confusion is that I don't quite understand the separation of responsibilities between `AerBackend` and `AerJob` here; in the whole library, `AerJob` only seems to be instantiated within `AerBackend`, so `fn` _cannot_ be anything but `backend._run`.  Is `AerJob` meant to be instantiable by users/other things?  If not, then I'm not really sure why the `fn` argument exists at all (`AerJob` could just do it internally); it seems to be very very dependent on `AerBackend` agreeing with the `executor` argument.

That said, this really isn't the point of the PR, so I'm not 100% sure what I'm suggesting here.  Mostly I think I'm just saying that I can't suggest better documentation because I don't understand what's happening!

I _think_ we're saying this:
```suggestion
            fn(function): a callable function to execute qobj on backend.
                This should usually be a bound :meth:`AerBackend._run()` method,
                with the signature `(qobj: QasmQobj, job_id: str) -> Result`.
```
?

hitomitak: Thank you for the update! I think it is better than my description and correct. 

> Is `AerJob` meant to be instantiable by users/other things? 

No. this instance is created by only `AerBackend` as you mentioned. 

> If not, then I'm not really sure why the `fn` argument exists at all (`AerJob` could just do it internally); it seems to be very very dependent on `AerBackend` agreeing with the `executor` argument.

That is certainly true.  AerJob can work well without `fn` argument. If this argument is removed, AerJob will become simpler.  This argument exists for some time so I just use it without thinking.

jakelishman (approved):
This looks good to me now.  The interaction of `AerJob` and `AerBackend` is strange to me coming in fresh, but that's not in scope of this PR; here we've documented the actual behaviour, which is just what we were aiming to do.

(My ticks don't have actual power on Aer at the moment, sorry.)

chriseclectic (approved):

PR #1335, opened 2021-09-02
Fix normalization following approximation in MPS

merav-aharoni:
### Summary
Fixed a bug in the computation of the normalization after truncation of the Schmidt coefficients in MPS.

### Details and comments
In MPS, approximation is implemented by truncating the smallest Schmidt coefficients following SVD. This is followed by normalization, so that the sum of the Schmidt coefficients remains 1. There was a bug in this computation that was fixed in this PR.


merav-aharoni:
@yaelbh , I would appreciate a review on this PR. I still need to add a test for the approximation, but it is taking me time to find something useful, and it would be helpful to at least have the code reviewed.

yaelbh:
1. It can be nice if you can clean the code from previous attempts, such that the differences reflect the essential change of this PR.
2. Can you please explain what was the bug? It seems that it was in the line `S[i] = sqrt(square_i);`, that should have been `S[i] /= sum;`. Is this correct? If yes, then do we need more changes?

merav-aharoni:
2) The bug was in these two lines:
`double square_i = std::norm(S[i])/sum;
 S[i] = sqrt(square_i);
`
because by taking the square of S[i], I turned it into a positive number rather than keeping the sign.
1) The remaining changes I made for better readability, not a previous attempt. Except for the call to `apply_measure_internal` instead of `apply_measure` which is for performance (not significant).

yaelbh (dismissed):

merav-aharoni:
I added a test that compares the results of the exact version with the approximated version. @yaelbh , can you please review?

merav-aharoni:
@yaelbh , I modified the test as you suggested. Please review once more.

test/terra/backends/aer_simulator/test_options.py
```diff
@@ -145,3 +146,28 @@ def test_shots_run_option(self, method, device):
...
+        """Test MPS approximation"""
+        shots = 4000
+        method="matrix_product_state"
+        backend_exact = self.backend(method=method)
+        backend_approx = self.backend(method=method,
+                                     matrix_product_state_max_bond_dimension=8)
+        # The test must be large enough and entangled enough so that
+        # approximation actually does something
+        n = 10
+        circuit = QuantumCircuit(n, n)
```
yaelbh, line 160: You don't need clbits anymore (change to `QuantumCircuit(n)` )

test/terra/backends/aer_simulator/test_options.py
```diff
@@ -145,3 +146,28 @@ def test_shots_run_option(self, method, device):
...
+                circuit.unitary(random_unitary(4), [i, i+1])
+            for i in range(1, n-1):
+                circuit.cx(0, i)
+        circuit.save_statevector('sv')
+
+        result_exact = backend_exact.run(circuit, shots=shots).result()
+        sv_exact = result_exact.data(0)['sv']
+        result_approx = backend_approx.run(circuit, shots=shots).result()
+        sv_approx = result_approx.data(0)['sv']
+        self.assertGreaterEqual(state_fidelity(sv_exact, sv_approx), 0.88)
```
yaelbh, line 172: Do you know if 0.88 is a reasonable fidelity?

test/terra/backends/aer_simulator/test_options.py
```diff
@@ -145,3 +146,28 @@ def test_shots_run_option(self, method, device):
...
+        self.assertLessEqual(state_fidelity(sv_exact, sv_approx), 0.99)
```
yaelbh, line 173: I indeed asked for this check, to verify that there's no bug in the test that compares exact to exact or approximated to approximated, with fidelity of 100%. But I don't think a reader of this code will understand what's going on here, so please add an explanation.

yaelbh (approved):
Approving, please see the three comments that I've left

test/terra/backends/aer_simulator/test_options.py
merav-aharoni: okay.

test/terra/backends/aer_simulator/test_options.py
merav-aharoni: I have no idea, it depends on the circuit and the amount of approximation. I ran this circuit several times, and the lowest fidelity I got was 0.89, so I used that as a bound.

test/terra/backends/aer_simulator/test_options.py
merav-aharoni: okay, added explanation.

PR #1337, opened 2021-09-02
MPS: Added print to log of bond dimension after internal swaps

merav-aharoni:
### Summary
Previously, bond dimensions were printed only for user-defined operations.  Here, we add similar prints after internal swap operations.

### DIFF
--- a/src/simulators/matrix_product_state/matrix_product_state_internal.cpp
+++ b/src/simulators/matrix_product_state/matrix_product_state_internal.cpp
@@ -551,6 +551,13 @@ void MPS::apply_swap_internal(uint_t index_A, uint_t index_B, bool swap_gate) {
     // and the qubit at index_B (or index_A+1) is moved one position 
     //to the left
     std::swap(qubit_ordering_.order_[index_A], qubit_ordering_.order_[index_B]);    
+    // For logging purposes:
+    if (MPS::get_mps_log_data()) {
+      if (swap_gate)
+	print_to_log_internal_swap(index_A, index_B);
+      else
+	print_to_log_internal_swap(index_B, index_A);
+    }
 
     // update qubit locations after all the swaps
     for (uint_t i=0; i<num_qubits_; i++)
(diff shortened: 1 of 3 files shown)


merav-aharoni:
Ready for review. The failure is not related to this PR. @yaelbh , can you please review?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -551,13 +551,35 @@ void MPS::apply_swap_internal(uint_t index_A, uint_t index_B, bool swap_gate) {
     // and the qubit at index_B (or index_A+1) is moved one position 
     //to the left
     std::swap(qubit_ordering_.order_[index_A], qubit_ordering_.order_[index_B]);    
+    // For logging purposes:
+    if (MPS::get_mps_log_data()) {
+      if (swap_gate)
```
yaelbh, line 556: We're already in an `if (!swap_gate)`. So one of the two options:
1. Get out of the scope of the `if (!swap_gate)`. In this case, why do you distinguish between `swap_gate` true or false, and why would it affect the order in the call to `print_to_log_internal_swap` ?
2. Remove the `if (swap_gate)` here.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -551,13 +551,35 @@ void MPS::apply_swap_internal(uint_t index_A, uint_t index_B, bool swap_gate) {
     // and the qubit at index_B (or index_A+1) is moved one position 
     //to the left
     std::swap(qubit_ordering_.order_[index_A], qubit_ordering_.order_[index_B]);    
+    // For logging purposes:
+    if (MPS::get_mps_log_data()) {
```
yaelbh, line 555: How about checking the log level only inside the function, like you do in the previous diff

yaelbh (changes requested):

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: You are right, the internal `if (swap_gate)` is redundant. I removed it.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Did as you suggested.

merav-aharoni:
@yaelbh , please re-review and approve if all looks ok.

yaelbh (approved):

merav-aharoni:
@mtreinish , @hhorii , @chriseclectic  - Yael reviewed and approved. Can one of you please merge?

chriseclectic (approved):

PR #1339, opened 2021-09-09
Cluster backend for shot parallelization 

hitomitak:
### Summary
Recreate #1298 
Enhancement of  #1281

When the noise simulator executes with the `executor` option and `max_shot_size` (default None), AerJobSet object copies experiments in Qobj and distributes them to the workers. After simulations, AerJobSet collects each result, sums it up with each count, and combines them into one result object. 

If the number of shots with a noise model exceeds `max_job_set`, Aer will split one experiment into sub experiments in the qobj. And `max_job_size` is set, Aer splits the qobj into the multiple qobj and distributes them to the multiple hosts. 

Example Usage:
Threadpool execution:
```
exc = ThreadPoolExecutor(max_workers=2)
qbackend = AerSimulator(noise_model=noise_model)
qbackend.set_options(executor=exc)
qbackend.set_options(max_job_size=2)
qbackend.set_options(max_shot_size=100)
result = qbackend.run(circ_list).result()
```
Dask execution (Local only):
```
exc = Client(address=LocalCluster(n_workers=1, processes=True))
qbackend = AerSimulator(noise_model=noise_model)
qbackend.set_options(executor=exc)
qbackend.set_options(max_job_size=2)
qbackend.set_options(max_shot_size=100)
result = qbackend.run(circ_list).result()
```

Dask execution (Multiple Nodes):
After setup dask environment in all nodes, set up a cluster in a master node as a following command. 
```
% dask-ssh dask1 dask2 dask3 &
```
And simulation executes. 
```
exc = Client('dask1:8786')
qbackend = AerSimulator(noise_model=noise_model)
qbackend.set_options(executor=exc)
qbackend.set_options(max_job_size=2)
qbackend.set_options(max_shot_size=100)
result = qbackend.run(circ_list).result()
```
### Details and comments
I executed QV (10 depth, 100 shots ) with a noise model and measured the execution time on 3VMs with Dask executor. 
Env: 3VMs, OS: CentOS8.0, Memory 64 GB, CPU: 32 x 2.0 GHz or higher Cores.
1 Node : a normal Aer simulator
1 Node (DASK) : only one node with dask executor
2 Node (DASK) : two nodes with dask executor
3 Node (DASK) : three nodes with dask executor

qv_noise


qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -218,6 +232,63 @@ def _get_worker_result(self, worker: int, timeout: Optional[float] = None):
...
+            return result_list[0]
+
+        master_result = result_list[0].to_dict()
+        sub_result = result_list[1].to_dict()
+        result_list = []
+
+        for (_master_result, _sub_result) in zip(master_result["results"], sub_result["results"]):
+            result_list.append(self._merge_exp(_master_result, _sub_result))
+        master_result["results"] = result_list
+        return Result.from_dict(master_result)
```
chriseclectic, line 246: If you are modifying an existing master result, you should just be updating it, not constructing a new one I think?

qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -218,6 +232,63 @@ def _get_worker_result(self, worker: int, timeout: Optional[float] = None):
...
+            list: Result list
+
+        """
+        results_list = []
+        for each_result in results:
+            _merge_results = []
+            master_id = None
+            master_result = None
+
+            _dict_result = each_result.to_dict()
```
chriseclectic, line 268: Again, don't convert to dicts. Either apply this to the dicts before they are converted to results, or use the Result attrbiutes.

qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -218,6 +232,63 @@ def _get_worker_result(self, worker: int, timeout: Optional[float] = None):
...
+        master_result = result_list[0].to_dict()
+        sub_result = result_list[1].to_dict()
+        result_list = []
+
+        for (_master_result, _sub_result) in zip(master_result["results"], sub_result["results"]):
+            result_list.append(self._merge_exp(_master_result, _sub_result))
+        master_result["results"] = result_list
+        return Result.from_dict(master_result)
+
+    def _merge_experiments_result(self, results: List[Result]):
```
chriseclectic, line 248: This is a bit confusing with there being two methods for merging results. If one is for combining multiple job results via concatenation, and the other for split circuits into single result via accumulation the names should be clearer, like `_combine_job_results` and `_accumulate_experiment_results` or something like that

qiskit/providers/aer/jobs/utils.py
```diff
@@ -55,35 +56,139 @@ def wrapper(*args, **kw):
...
+    if shot_mod > 0:
+        setattr(qobj.config, "shots", shot_mod)
+        for experiment in qobj.experiments:
+            _id = str(uuid.uuid4())
+            setattr(experiment.header, "metadata", {"id": _id})
+        qobj_list.append(qobj)
+
+    if num_shot_jobs > 1:
+        _qid = qobj_id or str(uuid.uuid4())
+        _config = copy.deepcopy(qobj.config)
```
chriseclectic, line 76: Does this need to be a deep copy? Shouldn't a shallow copy be enough?

qiskit/providers/aer/jobs/utils.py
```diff
@@ -55,35 +56,139 @@ def wrapper(*args, **kw):
...
+        setattr(_config, "shots", max_shot_size)
+        experiment_list = []
+        for experiment in qobj.experiments:
+            _id = str(uuid.uuid4())
+            for _ in range(num_shot_jobs):
+                cpy_exp = copy.deepcopy(experiment)
```
chriseclectic, line 82: Same here, avoid deep copies at all cost (they are extremely slow). Manually write your own copy method if necessary so you can do someting like:
```
cpy_exp = copy.copy(experiment)
cpy_exp.header = copy.copy(experiment.header)
```

qiskit/providers/aer/jobs/utils.py
```diff
@@ -55,35 +56,139 @@ def wrapper(*args, **kw):
...
+                setattr(cpy_exp.header, "metadata", {"id": _id})
```
chriseclectic, line 83: What if there are other entries in the metadata? This will erase them. Is this a problem?
Also if you know the attribute name you should be able to do just do `cpy_exp.header.metadata = value`

qiskit/providers/aer/jobs/utils.py
```diff
@@ -55,35 +56,139 @@ def wrapper(*args, **kw):
...
+    qobj_list = []
 
-    Returns:
-        List: A list of qobjs.
-    """
+    if shot_mod == 0 and num_shot_jobs == 1:
+        return qobj
+
+    if shot_mod > 0:
+        setattr(qobj.config, "shots", shot_mod)
```
chriseclectic, line 68:
```suggestion
        qobj.config.shots = shot_mod
```

qiskit/providers/aer/jobs/utils.py
```diff
@@ -55,35 +56,139 @@ def wrapper(*args, **kw):
...
+        for experiment in qobj.experiments:
+            _id = str(uuid.uuid4())
+            setattr(experiment.header, "metadata", {"id": _id})
```
chriseclectic, line 71: Mentioned below, but is this erasing any other metadata? shouldn't you just update the field like:
```
experiment.header.metadata["id"] = _id
```

qiskit/providers/aer/jobs/utils.py
```diff
@@ -55,35 +56,139 @@ def wrapper(*args, **kw):
...
+        return qobj, seed
 
+    qobjs = []
     # Check for parameterizations
     params = getattr(qobj.config, 'parameterizations', None)
-    qobjs = []
+
+    exp_id = None
+    shift_index = 0
+    seed_shift = 256
```
chriseclectic, line 105: Can this be made same as in C++ so RNG is same for manually splitting and running using:
`run(k * [circuit], shots=shots/k)` as with doing this automatic splitting.

qiskit/providers/aer/jobs/utils.py
```diff
@@ -55,35 +56,139 @@ def wrapper(*args, **kw):
...
         qobjs.append(type(qobj)(sub_id, sub_config, sub_exp, qobj.header))
-    return qobjs
+
+    if seed > 0:
+        seed = _seed + seed_shift
+
+    return qobjs, seed
+
+
+def _check_custom_instruction(experiments, custom_instructions):
```
chriseclectic, line 139: This is a slow check since it has to iterate over all circuits. We should really build this instruction map as part of `assemble` which already iterates over all instructions in a circuit.

qiskit/providers/aer/jobs/utils.py
```diff
@@ -55,35 +56,139 @@ def wrapper(*args, **kw):
...
+        JobError : If max_job_size > 1 and seed is set.
+        JobError : If custom instructions exist.
+
+    Returns:
+        List: A list of qobjs.
+    """
+    split_qobj_list = []
+    if (max_shot_size is not None and max_shot_size > 0):
+        if _check_custom_instruction(qobj.experiments, config.custom_instructions):
+            raise JobError("cluster backend does not support custom instructions.")
```
chriseclectic, line 170: This is a confusing message because there is no "cluster backend". This should refer to the experiment option value that doenst support this:
```suggestion
            raise JobError("`max_shot_size` option cannot be used with circuits containing save instructions.")
```
Also custom instructions like `Kraus` and `SuperOp` and `ReadoutError` support splitting. It should only be save instructions that prevent splitting so you should check for them.

chriseclectic (changes requested):
This looks to have a couple inefficiencies that should be addressed.
- All `to_dict` methods should be removed and replaced with proper attribute based access of the appropriate classes
- `deepcopy` should be removed and replaced with `copy` (and appropriate additional `copy` of specific attributes that are being updated like header/metdata/config)
- The custom instruction checks should only check for save instructions, not all custom instructions. These should also be removed in the future when support is added for them.

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -170,10 +170,15 @@ class AerSimulator(AerBackend):
       on the executor. If ``None`` simulation of all circuits are submitted
       to the executor as a single job (Default: None).
 
+    * ``max_shot_size`` (int or None): If the number of shots with
```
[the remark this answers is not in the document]
hitomitak: Thank you. I changed this description.

qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -218,6 +232,63 @@ def _get_worker_result(self, worker: int, timeout: Optional[float] = None):
...
+                        _merge_results.append(master_result)
+                else:
+                    _merge_results.append(_result)
+            _dict_result["results"] = _merge_results
+            results_list.append(Result.from_dict(_dict_result))
+        return results_list
+
+    def _merge_exp(self, master: Result, sub: Result):
+        master["shots"] = master["shots"] + sub["shots"]
+        if "counts" in master["data"]:
```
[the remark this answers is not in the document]
hitomitak: I added the code to combine "memory" data.

qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -218,6 +232,63 @@ def _get_worker_result(self, worker: int, timeout: Optional[float] = None):
                     "Timeout while waiting for JobSet results")
         return result
 
+    def _merge_results(self, result_list: List[Result]):
+        if len(result_list) == 1:
+            return result_list[0]
+
+        master_result = result_list[0].to_dict()
```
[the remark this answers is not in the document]
hitomitak: I removed all to_dict functions and use the attributes.

qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -218,6 +232,63 @@ def _get_worker_result(self, worker: int, timeout: Optional[float] = None):
...
+        sub_result = result_list[1].to_dict()
```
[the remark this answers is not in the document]
hitomitak: I remove this line and add a general operation.

qiskit/providers/aer/jobs/aerjobset.py
hitomitak: Ok. I changed the code to just return master result.

qiskit/providers/aer/jobs/aerjobset.py
hitomitak: Remove dict function.

qiskit/providers/aer/jobs/aerjobset.py
hitomitak: Change the name of these methods to `_combine_job_results` and `_accumulate_experiment_results`.

qiskit/providers/aer/jobs/utils.py
hitomitak: Change to copy function from deepcopy.

qiskit/providers/aer/jobs/utils.py
hitomitak: Change to copy, too.

qiskit/providers/aer/jobs/utils.py
hitomitak: I changed this function not to overwrite metadata and use the attribute.

qiskit/providers/aer/jobs/utils.py
hhorii: Agree, but I believe that no function to return an instruction map in `assemble`. So, the AerBackend needs to iterate instructions at least once.

qiskit/providers/aer/jobs/utils.py
hitomitak: I moved the operation to set seed and enable it as the c++ operation.

chriseclectic (changes requested):
This looks good. I think the only thing missing now is a release note.

PR #1343, opened 2021-09-14
Update cibuildwheel version and use qemu emulation for aarch64 builds

mtreinish:
### Summary

This commit bumps the cibuildwheel version we run in ci for both testing
and publishing wheels at release. The version we were using was a bit
stale and maybe the source of some ci failures. At the same time this
commit adds a config for running the aarch64 wheel builds in github
actions under qemu emulation. We previously had a config for this in
travis which provides native arm64 ci nodes. However, due to changes in
travis quota system we haven't had ci jobs run in several months so we
can't rely on that for building arm wheels. While emulation is painfully
slow the aer builds should hopefully finish in <6 hours (which is the
gha job time limit).

### Details and comments


mtreinish:
I've marked this as on hold because I pushed: https://github.com/Qiskit/qiskit-aer/pull/1343/commits/5f4debf297061f463cc8cc60a4f37b61ad8e39b8 to validate the aarch64 build config. We do not want to merge it with this applied as the build will take hours. Once it's confirmed to work I'll revert that and remove the on hold label

mtreinish:
Ok I removed the on hold since it looks like the emulated build worked fine: https://github.com/Qiskit/qiskit-aer/runs/3599011638?check_suite_focus=true it only took ~2 hours not too bad it's faster than gpu I think. So I reverted the dnm test pr and this is ready to go

.github/workflows/build.yml
```diff
@@ -129,11 +129,11 @@ jobs:
         uses: microsoft/setup-msbuild@v1.0.2
         if: runner.os == 'Windows'
       - name: Install deps
-        run: python -m pip install -U cibuildwheel==1.9.0
+        run: python -m pip install -U cibuildwheel==2.1.2
       - name: Build Wheels
         env:
           CIBW_BEFORE_ALL_LINUX: "yum install -y openblas-devel"
-          CIBW_SKIP: "cp27-* cp34-* cp35-* pp*"
```
chriseclectic, line 136: How come we don't need to skip these older versions anymore? Are they automatically skipped?

mtreinish: Yeah cibuildwheel dropped support for building wheels for python < 3.6 in 2.0.0: https://github.com/pypa/cibuildwheel/releases/tag/v2.0.0

chriseclectic (approved):

PR #1345, opened 2021-09-15
Fix issue 1258

### LINKED ISSUE #1258: QuantumError.ideal gives incorrect result for 2-qubit error
### Informations

- **Qiskit Aer version**: 0.8.2
- **Python version**: 3.8
- **Operating system**: Windows 10

Running the `ideal()` method from `QuantumError` gives incorrect results for 2-qubit errors.

A minimal example is:
```
from qiskit.providers.aer.noise.errors.quantum_error import QuantumError
from qiskit.providers.aer.noise.errors import pauli_error

p=0
error_gate1 = pauli_error([('X', p), ('I', 1 - p)])
print(error_gate1.ideal())
error_gate2 = error_gate1.tensor(error_gate1)
print(error_gate2.ideal()) # returns False, should be True
```

The `ideal()` method should return `True` for all identity objects

### Suggested solutions

On line https://github.com/Qiskit/qiskit-aer/blob/main/qiskit/providers/aer/noise/errors/quantum_error.py#L247 the index of the qubit seems to be hard-coded to 0.

chriseclectic:
### Summary

Fixes #1258

### DIFF
--- a/qiskit/providers/aer/noise/errors/quantum_error.py
+++ b/qiskit/providers/aer/noise/errors/quantum_error.py
@@ -242,10 +242,8 @@ class QuantumError:
     def ideal(self):
         """Return True if current error object is an identity"""
         instructions, probability = self.error_term(0)
-        if probability == 1 and instructions == [{
-                "name": "id",
-                "qubits": [0]
-        }]:
+        if (probability == 1 and len(instructions) == 1 and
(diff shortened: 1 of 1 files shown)


mtreinish (approved):
A release note would be nice, but this is simple enough it's ok without it. Same for a test

PR #1348, opened 2021-09-20
Update openmp to 12.0.1. 

vvilpas:
### Summary

Update llvm-OpenMP version to 12.0.1 in an attempt to be able to build on Macos M1.


jakelishman:
Is it worth having the tight version requirement only when `CMAKE_SYSTEM_PROCESSOR` indicates we're building for M1 Macs, so we don't need to have such stringent requirements on older systems?

PR #1350, opened 2021-09-21
Add support for adding QuantumError to circuits

chriseclectic:
### Summary

This PR add the ability to add `QuantumError` instructions directly to a circuit without converting them to a less general Kraus instruction first. The intent is that this can be used in the future for noise transpilation passes that add errors directly to circuits.

The main caveat is that since these instructions cannot be loaded as a C++ `Op` (circuits contain ops, so ops can't contain circuits), they must be dealt with on the python side before being serialized for execution by the Aer controller. 

This is handled by a circuit transformation that removes all instances of QuantumError instructions from circuits to be run and replaces them with labelled place holder instructions that reference the original error in a modified noise model. These circuits and noise model can then be run to implement the equivalent simulation of the original circuit containing the errors.

### Details and comments

- Add ~`QuantumErrorInstruction`~ `QuantumChannelInstruction` class that contains a `QuantumError` op. This is the instruction type converted to when adding a QuantumError to a circuit (since QuantumError is intended to be a BaseOperator subclass, not an instruction subclass).
- Adds a `QuantumErrorLocation` instruction which is effectively an N-qubit null-op/identity with a custom label that can be identified in a noise model to insert an N-qubit quantum error into a specific location through C++ noise sampling, and can be run on all simulation methods that support noise.
- ~Adds a `NoiseModelPass` transpiler transformation pass that removes `QuantumErrorInstructions` from a quantum circuit, replaces them with `QuantumErrorLocation` instructions, and adds the original `QuantumError` to the noise model referencing the location.~ The overhead of converting to/from DAG for a transpiler pass isn't necessary since this modification can easily be done directly to `QuantumCircuit.data`.
- Updates `AerBackend` to automatically replace circuit quantum errors and add them to the run noise model as part of its standard assembly routine to build the C++ qobj payload if the input circuits contain any quantum error location instructions (the performance of this needs to be profiled).

### TODO
- [x] Fix currently failing tests.
- [x] Add test coverage.

### Example

This change allows building and running a noisy circuit as follows:

```python
from qiskit.providers.aer.noise import depolarizing_error

# Pauli-error channels
error1 = depolarizing_error(0.1, 1)
error2 = depolarizing_error(0.2, 2)

# Noisy circuit
qc = QuantumCircuit(3)
qc.h(0)
qc.append(error1, [0])
qc.cx(0, 1)
qc.append(error2, [0, 1])
qc.cx(0, 2)
qc.append(error2, [0, 2])
qc.measure_all()
qc.draw()

"""
        ┌───┐┌─────────────────┐     ┌──────────────────┐     »
   q_0: ┤ H ├┤ Quantum_channel ├──■──┤0                 ├──■──»
        └───┘└─────────────────┘┌─┴─┐│  Quantum_channel │  │  »
   q_1: ────────────────────────┤ X ├┤1                 ├──┼──»
                                └───┘└──────────────────┘┌─┴─┐»
   q_2: ─────────────────────────────────────────────────┤ X ├»
                                                         └───┘»
meas: 3/══════════════════════════════════════════════════════»
                                                              »
«        ┌──────────────────┐ ░ ┌─┐      
«   q_0: ┤0                 ├─░─┤M├──────
«        │                  │ ░ └╥┘┌─┐   
«   q_1: ┤  Quantum_channel ├─░──╫─┤M├───
«        │                  │ ░  ║ └╥┘┌─┐
«   q_2: ┤1                 ├─░──╫──╫─┤M├
«        └──────────────────┘ ░  ║  ║ └╥┘
«meas: 3/════════════════════════╩══╩══╩═
«                                0  1  2 
"""

# Run on backend (this example should work on stabilizer simulator for example)
backend = AerSimulator(method='stabilizer')
result = backend.run(qc).result()
result.get_counts(0)

{'100': 24,
 '110': 26,
 '001': 23,
 '101': 76,
 '010': 68,
 '000': 378,
 '011': 23,
 '111': 406}
```


qiskit/providers/aer/noise/noise_model_pass.py
```diff
@@ -0,0 +1,75 @@
...
+        Args:
+            dag (DAGCircuit): input dag
+
+        Returns:
+            DAGCircuit: translated circuit.
+        """
+        for node in dag.op_nodes():
+            if node.name == "qerror":
+                error = node.op._quantum_error
+                qubits = list(dag.qubits.index(q) for q in node.qargs)
```
mtreinish, line 68: This will be pretty awful performance because it will traverse the qubit list each time to find the index. Instead you can just build a map and do this in a dict lookup. Outside of the for loop do something like:

```
qubit_index_map = {bit: index for index, bit in enumerate(dag.qubits)}
```

Then here do:
```suggestion
                qubits = [qubit_index_map[q] for q in node.qargs]
```

qiskit/providers/aer/noise/noise_model_pass.py
```diff
@@ -0,0 +1,75 @@
...
+                label = str(uuid.uuid4())
```
mtreinish, line 69: Unless you need the formatting with the `-` you can just use:

```suggestion
                label = uuid.uuid4().hex
```

For example:
```python
import uuid
print(str(id))
print(id.hex)
```

returns something like:

```
0974d537-41d4-4816-b764-7fb78643ba3c
0974d53741d44816b7647fb78643ba3c
```

qiskit/providers/aer/noise/noise_model_pass.py
```diff
@@ -0,0 +1,75 @@
...
+                error_loc = QuantumErrorLocation(error.num_qubits, label=label)
+                error_loc.condition = node.op.condition
+                self._noise_model.add_quantum_error(error, error_loc, qubits)
+                dag.substitute_node(node, error_loc)
+                self._updated = True
```
mtreinish, line 74: How is this expected to be used? Are you signalling to other passes that you've updated the noise model, if so you probably want to use the property set here instead of a pass property. Something like:

```suggestion
                self._updated = True
                self.property_set["noise_model_updated"] = True
```

qiskit/providers/aer/noise/errors/quantum_error.py
```diff
@@ -30,6 +31,28 @@
 logger = logging.getLogger(__name__)
 
 
+class QuantumErrorInstruction(Gate):
```
itoko, line 34:
```suggestion
class QuantumErrorInstruction(Instruction):
```

ajavadia (approved):
this is great and I'm particularly happy that the same circuit can be evaluated by aer and quantum_info

itoko:
Introduction of `QuantumChannelInstruction` looks good to me but I don't see why we need `QuantumErrorLocation` instruction. Even without it, your example code seems to work fine. Is it for performance?

chriseclectic:
@itoko that instruction is needed for qobj assembly and running simulation. Without it you can't actually run any of these noisy circuits on the simulator. A user should never use it directly though since its only intended for assembly of simulator payload.

itoko:
@chriseclectic Why don't we assemble `QuantumChannelInstruction` directly?
I think it's possible (in a bit hacky way, though). Since `QasmObjInstruction` accepts list of strings for parameters, we can push a list of tuple of probability and circuit as a list of json (representing tuple of prob and circ) into parameters entry, assuming the circuits does not include `QuantumChannelInstruction`.
What do you think of this "direct assembling approach"?

chriseclectic:
That approach won't work in the C++ code for loading circuits. It also goes against our goal of moving to full pybinding for simulator inputs to avoid any circuit/instruction conversion to json / dict for performance improvements.

jakelishman (commented):
Aside from a couple of minor comments inline, I'm happy with this.

The `QuantumErrorLocation` class is necessary as a placeholder which a unique `label` attribute, which allows the noise model builder to exactly associate a point in the circuit with some noise; in the C++ at the moment, the reason `OpType::qerror_loc` isn't generally treated specially is because Aer still has to support the old methods, which mean looking things up on the label.  So `qerror_loc` is absolutely required for this new form to work (because there's no present way to attach the noise directly onto the noisy gate), even if it doesn't _immediately_ look like that from the diff.

---

A point entirely unrelated to the main discussion here:

> The main caveat is that since these instructions cannot be loaded as a C++ `Op` (circuits contain ops, so ops can't contain circuits), they must be dealt with on the python side before being serialized for execution by the Aer controller.

With pointers/references, this isn't a language limitation, but an Aer one for the time being.  It's something we'll need to keep in mind when we're thinking about how we're going to handle conversion of scoped circuits (`for` loops, etc) for the Aer simulation - the current model in Terra is likely to treat these as `Instruction` instances that contain a circuit, or at least some other `Instruction` instances, and with things like `while`, you can't guarantee that you'll be able to unroll the loops before entering the simulator.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -356,6 +339,61 @@ def _assemble(self, circuits, parameter_binds=None, **run_options):
 
         return qobj
 
+    def _assemble_noise_model(self, circuits, **run_options):
+        """Move quantum error instructions from circuits to noise model"""
+        if isinstance(circuits, (QasmQobj, PulseQobj)):
+            return circuits, run_options
+
+        if isinstance(circuits, (QuantumCircuit, Schedule)):
+            circuits = [circuits]
```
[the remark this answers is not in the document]
chriseclectic: I don't think so, unless `ScheduleBlock` can be returned from `schedule(circuits)`?

qiskit/providers/aer/backends/backend_utils.py
```diff
@@ -16,7 +16,7 @@
 """
 import os
 from math import log2
-from qiskit.utils import local_hardware_info
+from qiskit.util import local_hardware_info
```
jakelishman, line 19: This is reverting #1382 by accident.
```suggestion
from qiskit.utils import local_hardware_info
```

jakelishman (approved):
Reticking after rebase.

PR #1354, opened 2021-09-23
Fix issue 1351

### LINKED ISSUE #1351: Segfault running on `AerSimulator` with empty circuit
### Informations

- **Qiskit Aer version**: 0.9.0
- **Qiskit version**: 0.30.0
- **Python version**: 3.8.10
- **Operating system**: Linux (Ubuntu 20.04)

Attempting to run a circuit containing no gates on an `AerSimulator` causes a segmentation fault.

```python
from qiskit import IBMQ, QuantumCircuit

c = QuantumCircuit()
IBMQ.load_account()
provider = IBMQ.get_provider(hub="ibm-q", group="open", project="main")
b = provider.get_backend("ibmq_santiago")
aersim = AerSimulator.from_backend(b)
job = aersim.run(c, n_shots=1)
```

`AerSimulator.run()` should return.

### Suggested solutions

?

chriseclectic:
### Summary

Fixes #1351

### Details and comments

When loading a circuit in C++ the size of an empty circuit was being set to 1 and being populated with a default initialized Op. This caused a crash when this circuit was passed to the noise sampler. This is fixed by correctly initializing an empty to be size 0.

### DIFF
--- a/src/framework/circuit.hpp
+++ b/src/framework/circuit.hpp
@@ -380,7 +380,12 @@ void Circuit::set_params(bool truncation) {
 
   // Counter for current position in ops as we shuffle ops
   size_t op_idx = 0;
-  const size_t head_end = (has_measure && can_sample) ? first_measure_pos : last_ancestor_pos + 1;
+  size_t head_end = 0;
+  if (has_measure && can_sample) {
+    head_end = first_measure_pos;
+  } else if (num_ancestors > 0) {
+    head_end = last_ancestor_pos + 1;
+  }
   for (size_t pos = 0; pos < head_end; ++pos) {
     if (ops_to_remove && !ancestor[pos]) {
       // Skip if not ancestor
--- a/src/framework/circuit.hpp
+++ b/src/framework/circuit.hpp
@@ -252,7 +252,8 @@ void Circuit::add_op_metadata(const Op& op) {
 void Circuit::set_params(bool truncation) {
   // Clear current circuit metadata  
   reset_metadata();
-  
+  if (ops.empty()) return;
+
   // Analyze input ops from tail to head to get locations of ancestor,
   // first measurement position and last initialize position
   const auto size = ops.size();
--- a/src/framework/circuit.hpp
+++ b/src/framework/circuit.hpp
@@ -265,27 +265,25 @@ void Circuit::set_params(bool truncation) {
   size_t last_initialize_pos = 0;
   bool ops_to_remove = false;
 
-  if (!ops.empty()) {
-    std::unordered_set<uint_t> ancestor_qubits;
-    for (size_t i = 0; i < size; ++ i) {
-      const size_t rpos = size - i - 1;
-      const auto& op = ops[rpos];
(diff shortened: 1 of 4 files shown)


src/framework/circuit.hpp
```diff
@@ -252,7 +252,8 @@ void Circuit::add_op_metadata(const Op& op) {
 void Circuit::set_params(bool truncation) {
   // Clear current circuit metadata  
   reset_metadata();
-  
+  if (ops.empty()) return;
+
```
jakelishman, line 256: This early return means that the check a few lines down `if (!ops.empty())` is now guaranteed to be true.

src/noise/noise_model.hpp
```diff
@@ -260,6 +260,10 @@ NoiseModel::param_gate_table_ = {
 Circuit NoiseModel::sample_noise(const Circuit &circ,
                                  RngEngine &rng,
                                  const Method method) const {
+  // Check edge case of empty circuit
+  if (circ.ops.empty()) {
+    return circ;
+  }
```
jakelishman, line 266: If `Circuit::set_params` is doing its job correctly, then this shouldn't be necessary, right?  The auto loop over `circ.ops` further down should just operate zero times.

jakelishman (commented):
The code change as written does appear to correctly handle the case of empty ops in the noise model, but solving it purely by early return in `Circuit::set_params` seems to suggest that there's some logic error in the rest of that function, I think - by the time it hits `ops.resize(op_idx)`, `op_idx` should still be 0.

I think this happens in line 383 of `Circuit::set_params` - `has_measure` will certainly be `false`, so `head_end` will get set to 1, causing the next loop to be entered once, and `op_idx` to be incremented, even though there weren't any ops.  So then at the end, the vector gets expanded to be size 1, even though I think the expansion is only meant to be a truncation.

I'm slightly worried that there may be more off-by-one errors lurking in there, because I can't convince myself that the current off-by-one is absolutely specific to an empty `ops`.

*edit*: also, early return from `set_params` feels like a potential future bug in itself - are we sure that all the parameters that need to be set are always going to be before that return check?  Or can we absolutely rely on the defaults being correct?  Neither seems obviously safe to me - it feels better if the looping logic in the rest of the function correctly handles empty lists.

src/noise/noise_model.hpp
chriseclectic: I pushed another fix for this and tested it worked for the previous segfault error by commenting out the early return statement.

jakelishman (approved):
This looks much better to me.  I think the correction to the resize amount should now mean that the early returns from both `Circuit::set_params` and from `NoiseModel::sample_noise` are unnecessary now, and only serve as speed optimisations.  For clarity, it may not hurt to remove them since they're superfluous, but are in functions that have some side-effects that aren't (immediately) obviously nulled if the ops are empty.  I'm also not too bothered if you want to keep them, though.

I don't know what frameworks we have in place for testing the C++ code, but if we've got them, it's probably worth sticking in a regression test that empty `ops` remain size 0 after `Circuit::set_params` for the future.

PR #1357, opened 2021-09-26
[WIP] MPS: Normalize gamma tensor after approximation

merav-aharoni:
### Summary
The Pauli expectation value by 'I' of every gamma tensor in the MPS structure should be 1. Approximation may cause this invariant to not be satisfied. This PR renormalizes the gamma tensors.

### Details and comments
One of the invariants of the MPS structure is that the Pauli expectation value by 'I' of every gamma tensor  should be 1. 
Up to this point, when approximating an MPS, we normalized the lambda vector following truncation. However, during measurement, the gamma tensors may not collapse correctly to 1. The underlying reason is that the truncation and normalization of the lambda vectors, caused the above invariant to no longer hold. Therefore, we renormalize the gamma tensors before measurement.
I addition, we added an option for the user to determine whether the internal swaps inserted for 2-qubit gates should be toward left or toward right. The user may have knowledge, depending on the circuit, of which direction is better performance-wise.


chriseclectic (changes requested):
Are you able to remove the commits from #1358 which is adding a new feature so that this can be back-ported as a bug fix for the next patch release?

merav-aharoni:
This is still WIP. I will remove the commits from #1358 before review.

merav-aharoni:
I am replacing this by #1369. I think that is a better solution for this problem.

PR #1358, opened 2021-09-26
MPS: Give the user control over the direction of the internal swaps

merav-aharoni:
### Summary
We added an option for the user to determine the overall directions of the internal swaps inserted for 2-qubit gates.

### Details and comments
We added an option for the user to determine whether the internal swaps inserted for 2-qubit gates should be toward left or toward right. The user may have knowledge, depending on the circuit, of which direction is better performance-wise.
The default direction is to the left.
Usage example:
```
backend = AerSimulator(
    method='matrix_product_state',
    mps_swap_direction='mps_swap_right'
)
```


merav-aharoni:
Ready for review.

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -323,10 +323,15 @@ class AerSimulator(AerBackend):
         shots, and a large number of qubits, with complexity around
         O(n * D^2) per shot.
 
-    * ``mps_log_data`` (str): if True, output logging data of the MPS
+    * ``"mps_log_data"`` (str): if True, output logging data of the MPS
```
yaelbh, line 326: I don't see the quotes in `fusion_enable` here below

test/terra/backends/aer_simulator/test_options.py
```diff
@@ -165,13 +168,21 @@ def test_mps_approximation(self):
...
+        result_swap_right = backend_swap_right.run(circuit, shots=shots).result()
+        sv_right = result_swap_right.data(0)['sv']
+        
         result_approx = backend_approx.run(circuit, shots=shots).result()
         sv_approx = result_approx.data(0)['sv']
-        # Check that the fidelity is reasonable
-        self.assertGreaterEqual(state_fidelity(sv_exact, sv_approx), 0.80)
+
+        # swap_left and swap_right should give the same state vector
+        self.assertAlmostEqual(state_fidelity(sv_left, sv_right), 1.0)
```
yaelbh, line 181: It's strange to test this feature in a test called `test_mps_approximation`

yaelbh (approved):

test/terra/backends/aer_simulator/test_options.py
merav-aharoni: Would it be better if I named it `test_mps_options`? or do you think I should separate into 2 tests?

yaelbh: I prefer two tests but `test_mps_options` is also ok

qiskit/providers/aer/backends/aer_simulator.py
merav-aharoni: It seems in the output that the extra quotes are not needed. I removed them here and in several other places.

merav-aharoni:
@yaelbh , can you please approve?

yaelbh (approved):

hhorii (approved):

PR #1360, opened 2021-09-28
Remove references to unused parameters

### LINKED ISSUE #1359: Documentation error for `optimize_noise_threshold`
I wonder if `optimize_noise_threshold` applies for noisy circuit simulation, or just:

```
    * ``optimize_noise_threshold`` (int): Sets the qubit threshold for
      applying circuit optimization passes on ideal circuits.
      Passes include gate fusion and truncation of unused qubits
      (Default: 12).
```
as shown here:

https://github.com/Qiskit/qiskit-aer/blob/ea2ad477d9fe796c14a1b8c7a1d40d636ec30d4c/qiskit/providers/aer/backends/aer_simulator.py#L211

and here:

https://github.com/Qiskit/qiskit-aer/blob/ea2ad477d9fe796c14a1b8c7a1d40d636ec30d4c/qiskit/providers/aer/backends/qasm_simulator.py#L174

jakelishman:
### Summary

The last meaningful uses of the optimize_noise_threshold and
optimize_ideal_threshold parameters were removed in #698 (commit
5828c51).  This just removes references to them, since they are no
longer accessed, but had remained in the documentation.

### Details and comments

Fix #1359.

### DIFF
--- a/qiskit/providers/aer/backends/aer_simulator.py
+++ b/qiskit/providers/aer/backends/aer_simulator.py
@@ -203,16 +203,6 @@ class AerSimulator(AerBackend):
       values (16 Bytes). If set to 0, the maximum will be automatically
       set to the system memory size (Default: 0).
 
-    * ``optimize_ideal_threshold`` (int): Sets the qubit threshold for
-      applying circuit optimization passes on ideal circuits.
-      Passes include gate fusion and truncation of unused qubits
-      (Default: 5).
-
-    * ``optimize_noise_threshold`` (int): Sets the qubit threshold for
-      applying circuit optimization passes on ideal circuits.
-      Passes include gate fusion and truncation of unused qubits
-      (Default: 12).
-
     These backend options only apply when using the ``"statevector"``
     simulation method:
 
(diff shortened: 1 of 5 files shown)


chriseclectic (commented):
We should probably add a short `upgrade` reno saying these were removed, since if someone does try and use these options after merge it will raise an exception for unsupported options (rather than just doing nothing as previous).

jakelishman:
@chriseclectic: done in 1b7c711

jakelishman:
These repeated mac failures I've had over the past couple of days feel suspiciously like something's segfaulting in the background. At first I thought it might just be the mac machines being a bit funny on Azure, because I was having problems on Terra CI with them, but Terra's was intermittent and definitely a provisioning issue, whereas this is repeated, and always fails in the same way, with one worker silently hanging until the job dies.

mtreinish:
Yeah, I've been meaning to open an issue about it for some time now. It's almost certainly a segfault or other system error causing the test worker to exit prior to it reporting back the result of a test. We've been seeing it pretty regularly on MacOS jobs and always in the pulse simulator. But unfortunately we don't get a lot of debug details from gha when this happens and reproducing locally has proven tricky.

chriseclectic (approved):

PR #1369, opened 2021-10-06
MPS: Normalize lambda after every svd

merav-aharoni:
### Summary
The norm of lambda after svd may be slightly off 1 even without truncation. Therefore  always check its norm, and normalize if needed.

### Details and comments
Slight errors in the computation of lambda during measurement may build up when many qubits are involved, eventually leading to a segmentation fault.

### DIFF
--- a/src/simulators/matrix_product_state/svd.cpp
+++ b/src/simulators/matrix_product_state/svd.cpp
@@ -127,14 +127,15 @@ double reduce_zeros(cmatrix_t &U, rvector_t &S, cmatrix_t &V,
     for (uint_t i=new_SV_num; i<SV_num; i++) {
       discarded_value += std::norm(S[i]);
     }
-    // After approximation, we may need to re-normalize the values of S
-    double new_sum_squares = 0;
-    for (uint_t i=0; i<S.size(); i++) 
-      new_sum_squares +=std::norm(S[i]);
-    
-    double sqrt_sum = std::sqrt(new_sum_squares);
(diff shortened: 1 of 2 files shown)


yaelbh (dismissed):
What causes these errors?

merav-aharoni:
@yaelbh , I haven't been able to identify the root cause for these differences. Since, from what I see, the difference is relatively small, it could be numerical issues, but I can't be sure.

src/simulators/matrix_product_state/svd.cpp
```diff
@@ -127,14 +127,15 @@ double reduce_zeros(cmatrix_t &U, rvector_t &S, cmatrix_t &V,
...
-    
-    double sqrt_sum = std::sqrt(new_sum_squares);
-    for (uint_t i=0; i<S.size(); i++)
-      S[i] /= sqrt_sum;
+  }
+  // Check if we need to re-normalize the values of S
+  double new_sum_squares = 0;
+  for (uint_t i=0; i<S.size(); i++) 
+    new_sum_squares +=std::norm(S[i]);
+  if (!Linalg::almost_equal(new_sum_squares, 1.0, THRESHOLD)) {
```
chriseclectic, line 135: When comparing floats it's probably safer to compare to 0 because of the way relative differences are computed: 
```suggestion
  if (!Linalg::almost_equal(1.0 - new_sum_squares, 0., THRESHOLD)) {
```

PR #1371, opened 2021-10-07
Fix automatic method selection when some circuits have invalid instructions

### LINKED ISSUE #1370: Partial result is not schema compliant
### Informations

- **Qiskit Aer version**: 0.9.0
- **Python version**:
- **Operating system**:

Related to #1214 
When part of a circuit fails (running on simulator), the resulting JSON can not be loaded with `Result.from_dict()`because there is no "results" field. This is causing a [test](https://github.com/Qiskit-Partners/qiskit-ibm/issues/125) to fail in qiskit-ibm 

Create a qobj with one good and one bad instruction 

```
qc_new = transpile(ReferenceCircuits.bell())
qobj = assemble([qc_new]*2)
qobj.experiments[1].instructions[1].name = 'bad_instruction'
```

Run the qobj on the simulator, and the `raw_data` returned does not contain "results" 

Partial results should still be schema compliant and be able to loaded from  `Result.from_dict()`

mtreinish:
### Summary

When the automatic method detection loop was moved to occur once ahead
of time in #1312 the loop over all circuits there was not tolerant of
any circuits in a job with multiple circuits being invalid for all
simulation methods. This was an unexpected change from the previous
behavior where we would return a partial result and only the circuits
with no matching methods would error and the other results would be
returned. This commit fixes the behavior and restores the partial result
generation and only errors for the jobs where the circuit is not
runnable.

### Details and comments

Fixes #1370

### DIFF
--- a/src/controllers/aer_controller.hpp
+++ b/src/controllers/aer_controller.hpp
@@ -1727,17 +1727,7 @@ Controller::simulation_methods(std::vector<Circuit> &circuits,
     bool superop_enabled = false;
     bool kraus_enabled = false;
     for (const auto& circ: circuits) {
-      Controller::Method method;
-      try {
-        method = automatic_simulation_method(circ, noise_model);
-
-      }
-      // If setting a method for this circuit raised an exception fallback to
-      // a default of statevector. The execution will fail but we will get
-      // partial result generation and generate a user facing error message
-      catch (std::exception &e) {
-        method = Method::statevector;
-      }
+      auto method = automatic_simulation_method(circ, noise_model);
       sim_methods.push_back(method);
       if (!superop_enabled && (method == Method::density_matrix || method == Method::superop)) {
         noise_model.enable_superop_method(max_parallel_threads_);
@@ -1748,9 +1738,7 @@ Controller::simulation_methods(std::vector<Circuit> &circuits,
         kraus_enabled = true;
       }
     }
-    if (!sim_methods.empty()) {
-        return sim_methods;
-    }
+    return sim_methods;
   }
 
   // Use non-automatic default method for all circuits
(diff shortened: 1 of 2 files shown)


src/controllers/aer_controller.hpp
```diff
@@ -1727,7 +1727,17 @@ Controller::simulation_methods(std::vector<Circuit> &circuits,
     bool superop_enabled = false;
     bool kraus_enabled = false;
     for (const auto& circ: circuits) {
-      auto method = automatic_simulation_method(circ, noise_model);
+      Controller::Method method;
+      try {
```
chriseclectic, line 1731: Better to remove this try block and leave it as it was, and instead update the `automatic_simulation_method` at https://github.com/Qiskit/qiskit-aer/blob/872bdbaba177a4fb0eb2241b37dfcfac2233a79b/src/controllers/aer_controller.hpp#L1802 to return `return Method::statevector` at the end instead of throwing the the final exception which this added try block is always hiding.

If we get any other exceptions they should be legitimate errors which we want to throw normally.

src/controllers/aer_controller.hpp
```diff
@@ -1738,7 +1748,9 @@ Controller::simulation_methods(std::vector<Circuit> &circuits,
         kraus_enabled = true;
       }
     }
-    return sim_methods;
+    if (!sim_methods.empty()) {
```
chriseclectic, line 1751: I don't think this is necessary since sim methods will never be empty unless there are no circuits.

chriseclectic (changes requested):

src/controllers/aer_controller.hpp
mtreinish: Done in: https://github.com/Qiskit/qiskit-aer/pull/1371/commits/42cbc7e72d302b3203c868d4459fa663c64161b5

chriseclectic (approved):
LGTM

PR #1376, opened 2021-10-19
add support of control-flow in qiskit

hhorii:
### Summary

This PR supports control-flow which will be introduced in https://github.com/Qiskit/qiskit-terra/pull/7123 .

### Details and comments

This PR includes two main changes: `AerCompiler` and introduction of `jump/mark`.

`AerCompiler` works as:

- Unroll all the `for_loop`
- Inject `mark` instruction for each iteration of an unrolled loop
- Inject `mark` instruction at the end of an unrolled loop
- Convert `breakLoop` and `continueLoop` with `jump` instruction
- Convert `while_loop` to `jump` instructions at start (with conditional) and end points of the loop.
- Convert `if_else` to `jump` instruction with conditional and inject `mark` instructions at start of `else_body` and end of the `else_body` (or `true_body` if no `else_body` is specified).

Then, Aer supports new instructions `jump` and `mark`, which are handled in `apply_ops` of the base class of `State`. Extended Stabilizer extends `apply_ops`, so this PR does not cover the method.

```
from qiskit import transpile
from qiskit.circuit import Clbit, Instruction, Parameter, QuantumCircuit, QuantumRegister, ClassicalRegister, Barrier
from qiskit.circuit.controlflow import *
from qiskit.quantum_info.random import random_unitary
from qiskit.providers.aer import AerSimulator, AerCompiler

loop_body = QuantumCircuit(3, 1)
loop_parameter = Parameter("foo")
indexset = range(0, 7, 2)

loop_body.unitary(random_unitary(2**2), [0, 1])
loop_body.unitary(random_unitary(2**2), [1, 2])
loop_body.unitary(random_unitary(2**2), [2, 0])
loop_body.rx(loop_parameter, 0)
loop_body.ry(loop_parameter * 2, 1)
loop_body.rz(loop_parameter * 4, 2)
# loop_body.measure(1, 0)

# true_body = QuantumCircuit(3, 0)
# true_body.append(BreakLoopOp(3, 0), [0, 1, 2], [])
# else_body = QuantumCircuit(3, 0)
# else_body.append(ContinueLoopOp(3, 0), [0, 1, 2], [])
# body.append(IfElseOp(body.clbits[0], true_body, else_body), [0, 1, 2], [])
# loop_body.append(IfElseOp(loop_body.clbits[0], true_body), [0, 1, 2], [])

for_loop = ForLoopOp(loop_parameter, indexset, loop_body)

loop_circ = QuantumCircuit(3, 1)
loop_circ.append(for_loop, [0, 1, 2], [0])

while_body = QuantumCircuit(3, 1)
while_body.unitary(random_unitary(2**2), [0, 1])
while_body.unitary(random_unitary(2**2), [1, 2])
while_body.unitary(random_unitary(2**2), [2, 0])
while_body.measure(0, 0)
while_loop = WhileLoopOp(loop_circ.clbits[0], while_body)

loop_circ.append(while_loop, [0, 1, 2], [0])    
loop_circ.measure_all()

circ = AerCompiler().compile_circuit(loop_circ)
print(circ)

simulator = AerSimulator()
circ = transpile(circ, simulator, optimization_level=0)
simulator.run(circ, shots=1).result()


        ┌──────────┐                     ┌──────────┐┌───────┐»
   q_0: ┤0         ├─────────────────────┤1         ├┤ Rx(0) ├»
        │  Unitary │┌──────────┐┌───────┐│          │└───────┘»
   q_1: ┤1         ├┤0         ├┤ Ry(0) ├┤  Unitary ├─────────»
        └──────────┘│  Unitary │└───────┘│          │┌───────┐»
   q_2: ────────────┤1         ├─────────┤0         ├┤ Rz(0) ├»
                    └──────────┘         └──────────┘└───────┘»
   c: 1/══════════════════════════════════════════════════════»
                                                              »
meas: 3/══════════════════════════════════════════════════════»
                                                              »
«        ┌─────────────────┐┌──────────┐                     ┌──────────┐»
«   q_0: ┤0                ├┤0         ├─────────────────────┤1         ├»
«        │                 ││  Unitary │┌──────────┐┌───────┐│          │»
«   q_1: ┤1 Mark(loop_0_0) ├┤1         ├┤0         ├┤ Ry(4) ├┤  Unitary ├»
«        │                 │└──────────┘│  Unitary │└───────┘│          │»
«   q_2: ┤2                ├────────────┤1         ├─────────┤0         ├»
«        └─────────────────┘            └──────────┘         └──────────┘»
«   c: 1/════════════════════════════════════════════════════════════════»
«                                                                        »
«meas: 3/════════════════════════════════════════════════════════════════»
«                                                                        »
«        ┌───────┐┌─────────────────┐┌──────────┐                     »
«   q_0: ┤ Rx(2) ├┤0                ├┤0         ├─────────────────────»
«        └───────┘│                 ││  Unitary │┌──────────┐┌───────┐»
«   q_1: ─────────┤1 Mark(loop_0_2) ├┤1         ├┤0         ├┤ Ry(8) ├»
«        ┌───────┐│                 │└──────────┘│  Unitary │└───────┘»
«   q_2: ┤ Rz(8) ├┤2                ├────────────┤1         ├─────────»
«        └───────┘└─────────────────┘            └──────────┘         »
«   c: 1/═════════════════════════════════════════════════════════════»
«                                                                     »
«meas: 3/═════════════════════════════════════════════════════════════»
«                                                                     »
«        ┌──────────┐┌───────┐ ┌─────────────────┐┌──────────┐            »
«   q_0: ┤1         ├┤ Rx(4) ├─┤0                ├┤0         ├────────────»
«        │          │└───────┘ │                 ││  Unitary │┌──────────┐»
«   q_1: ┤  Unitary ├──────────┤1 Mark(loop_0_4) ├┤1         ├┤0         ├»
«        │          │┌────────┐│                 │└──────────┘│  Unitary │»
«   q_2: ┤0         ├┤ Rz(16) ├┤2                ├────────────┤1         ├»
«        └──────────┘└────────┘└─────────────────┘            └──────────┘»
«   c: 1/═════════════════════════════════════════════════════════════════»
«                                                                         »
«meas: 3/═════════════════════════════════════════════════════════════════»
«                                                                         »
«                  ┌──────────┐┌───────┐ ┌─────────────────┐»
«   q_0: ──────────┤1         ├┤ Rx(6) ├─┤0                ├»
«        ┌────────┐│          │└───────┘ │                 │»
«   q_1: ┤ Ry(12) ├┤  Unitary ├──────────┤1 Mark(loop_0_6) ├»
«        └────────┘│          │┌────────┐│                 │»
«   q_2: ──────────┤0         ├┤ Rz(24) ├┤2                ├»
«                  └──────────┘└────────┘└─────────────────┘»
«   c: 1/═══════════════════════════════════════════════════»
«                                                           »
«meas: 3/═══════════════════════════════════════════════════»
«                                                           »
«        ┌───────────────────┐┌──────────────────────┐┌────────────────────┐»
«   q_0: ┤0                  ├┤0                     ├┤0                   ├»
«        │                   ││                      ││                    │»
«   q_1: ┤1 Mark(loop_0_end) ├┤1 Mark(while_1_start) ├┤1 Jump(while_1_end) ├»
«        │                   ││                      ││                    │»
«   q_2: ┤2                  ├┤2                     ├┤2                   ├»
«        └───────────────────┘└──────────────────────┘└─────────╥──────────┘»
«                                                          ┌────╨────┐      »
«   c: 1/══════════════════════════════════════════════════╡ c_0 = F ╞══════»
«                                                          └─────────┘      »
«meas: 3/═══════════════════════════════════════════════════════════════════»
«                                                                           »
«        ┌──────────┐            ┌──────────┐┌─┐┌──────────────────────┐»
«   q_0: ┤0         ├────────────┤1         ├┤M├┤0                     ├»
«        │  Unitary │┌──────────┐│          │└╥┘│                      │»
«   q_1: ┤1         ├┤0         ├┤  Unitary ├─╫─┤1 Jump(while_1_start) ├»
«        └──────────┘│  Unitary ││          │ ║ │                      │»
«   q_2: ────────────┤1         ├┤0         ├─╫─┤2                     ├»
«                    └──────────┘└──────────┘ ║ └──────────────────────┘»
«   c: 1/═════════════════════════════════════╬═════════════════════════»
«                                             ║                         »
«meas: 3/═════════════════════════════════════╩═════════════════════════»
«                                             0                         »
«        ┌────────────────────┐ ░ ┌─┐      
«   q_0: ┤0                   ├─░─┤M├──────
«        │                    │ ░ └╥┘┌─┐   
«   q_1: ┤1 Mark(while_1_end) ├─░──╫─┤M├───
«        │                    │ ░  ║ └╥┘┌─┐
«   q_2: ┤2                   ├─░──╫──╫─┤M├
«        └────────────────────┘ ░  ║  ║ └╥┘
«   c: 1/══════════════════════════╬══╬══╬═
«                                  ║  ║  ║ 
«meas: 3/══════════════════════════╩══╩══╩═
«                                  0  1  2 

```


qiskit/providers/aer/backends/__init__.py
```diff
@@ -19,3 +19,5 @@
 from .statevector_simulator import StatevectorSimulator
 from .unitary_simulator import UnitarySimulator
 from .pulse_simulator import PulseSimulator
+
+from .aer_compiler import AerCompiler
```
chriseclectic, line 23: This should just be an internal part of the AerBackend so it happens automatically during the `run` call, but I think this requires qiskit terra transpiler support for the qasm3 instructions first.

qiskit/providers/aer/backends/aer_compiler.py
```diff
@@ -0,0 +1,218 @@
...
+    Jump instruction
+
+    This instruction sets a program counter to specified mark instruction.
+    """
+
+    def __init__(self, jump_to, num_qubits):
+        super().__init__("jump", num_qubits, 0, [jump_to])
+
+
+class AerCompiler:
```
chriseclectic, line 47: It would be good if this could be built into an aer `assemble` method so it only happens when we construct the qobj payload for the C++ simulator (same as with the noise replacement in #1350) so we can avoid iterating over all the circuits too many times.

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -337,23 +337,27 @@ class AerSimulator(AerBackend):
             'rzz', 'rzx', 'ccx', 'cswap', 'mcx', 'mcy', 'mcz', 'mcsx',
             'mcp', 'mcphase', 'mcu', 'mcu1', 'mcu2', 'mcu3', 'mcrx', 'mcry', 'mcrz',
             'mcr', 'mcswap', 'unitary', 'diagonal', 'multiplexer',
-            'initialize', 'delay', 'pauli', 'mcx_gray'
+            'initialize', 'delay', 'pauli', 'mcx_gray',
+            'jump', 'mark'
```
chriseclectic, line 341: If the compiler becomes part of the internal assemble we dont need to add these custom internal gates to the basis gates. These are only needed for terra transpiler, which should only be applied to circuits that don't contain these instructions.

src/framework/circuit.hpp
```diff
@@ -265,6 +265,13 @@ void Circuit::set_params(bool truncation) {
   size_t last_initialize_pos = 0;
   bool ops_to_remove = false;
 
+  for (size_t i = 0; i < size; ++ i) {
```
chriseclectic, line 268: This adds an unnecessary full iteration over each circuit which is not good for performance. In principle jump operations shouldn't effect truncation, but they may need special handling for how to remap the new locations.

src/framework/circuit.hpp
```diff
@@ -358,7 +365,9 @@ void Circuit::set_params(bool truncation) {
         case OpType::save_stabilizer:
         case OpType::save_unitary:
         case OpType::save_mps:
-        case OpType::save_superop: {
+        case OpType::save_superop:
+        case OpType::jump:
```
chriseclectic, line 369: I don't think jump and mark need to be here, since they dont produce output or change clbits themselves.

src/simulators/state.hpp
```diff
@@ -344,9 +344,57 @@ void State<state_t>::apply_ops(InputIterator first, InputIterator last,
                                ExperimentResult &result,
                                RngEngine &rng,
                                bool final_ops) {
+
+  std::unordered_map<std::string, InputIterator> marks;
+  // Simple loop over vector of input operations
+  for (auto it = first; it != last; ++it) {
```
chriseclectic, line 350: This is also adding an unnecessary full iteration over all circuits just to find the jump locations. I think a more efficient way to handle this would be to add an unordered_map to `Circuit` that stores the mark locations, and this can be populated during the `set_params` method without an extra iteration, furthermore doing it this way would allow you to still apply truncation properly with these instructions since you can update their positions in the truncated circuit. IT will also allow you to catch errors for duplicate marks during circuit loading rather than execution.

chriseclectic (changes requested):
This is a nice feature to support, but I am a bit worried that the current implementation will have some inefficiencies for low-qubit number simulations due to multiple extra iterations over all instructions in each circuit.

Most of the logic done in these iterations should be able to be streamlined to be done in the existing methods that already iterate over the instructions (whether assemble on the python side, or set_params on the C++ Circuit).

I also think that the AerCompiler should not be something exposed to a user, this should just be an internal part of the assembly in `AerBackend.run` method that is used to produce the qobj to be executed.

qiskit/providers/aer/backends/__init__.py
hhorii: I removed `AerCompiler` from `__init__.py`.

qiskit/providers/aer/backends/aer_compiler.py
hhorii: I integrated `AerCompiler` into `AerBackend`.

qiskit/providers/aer/backends/__init__.py
hhorii: I removed `AerCompiler`.

src/simulators/state.hpp
hhorii: I will add check codes in `set_params`. This loop is not necessary because the next loop registers and looks-up `mark` instructions.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
+                           range(ret.num_qubits), [])
+            elif isinstance(inst, ContinueLoopOp):
+                ret.append(AerJump(continue_label, ret.num_qubits),
+                           range(ret.num_qubits), [])
+            else:
+                ret.append(inst, qargs, cargs)
+
+        return ret
+
+    def _compile(self, circuits):
```
chriseclectic, line 520: I think this function could be moved to a separate file as `_compile_qasm3(circuits)`

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
+                return circuits
+            basis_gates = self.configuration().basis_gates + ['mark', 'jump']
+            return [transpile(self._inline_circuit(self._decompose_subcircuits(circuit),
+                                                   None, None),
+                              basis_gates=basis_gates) if self._is_dynamic(circuit)
+                    else circuit for circuit in circuits]
+        else:
+            return (transpile(self._inline_circuit(self._decompose_subcircuits(circuits),
+                                                   None, None),
+                              basis_gates=basis_gates) if self._is_dynamic(circuits)
```
chriseclectic, line 533: `basis_gates` is not defined here, since it is only defined in the `if` block of this conditional

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
+
+    def _compile(self, circuits):
+        if isinstance(circuits, list):
+            all_static = all([not self._is_dynamic(circuit) for circuit in circuits])
+            if all_static:
+                return circuits
+            basis_gates = self.configuration().basis_gates + ['mark', 'jump']
+            return [transpile(self._inline_circuit(self._decompose_subcircuits(circuit),
+                                                   None, None),
+                              basis_gates=basis_gates) if self._is_dynamic(circuit)
```
chriseclectic, line 528: You are calling `_is_dynamic` multiple times per circuit, this should be called once and the result saved as a variable if you need to refer to it again.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
 
+    def _is_dynamic(self, circuit):
+        if not isinstance(circuit, QuantumCircuit):
+            return False
+        for inst, _, _ in circuit.data:
+            if isinstance(inst, (WhileLoopOp, ForLoopOp, IfElseOp, BreakLoopOp, ContinueLoopOp)):
+                return True
+        return False
+
+    def _decompose_subcircuits(self, circuit, basis_gates=None):
```
chriseclectic, line 333: Can all these qasm3 compilation functions be moved back to a separate file, and just have a `compile_qasm3` function that is imported from there. It makes the class here very cluttered when all of these functions do not need to refer to the class instance.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
+        if circuit is None:
+            return None
+        if basis_gates is None:
+            basis_gates = self.configuration().basis_gates + ['for_loop', 'while_loop',
```
chriseclectic, line 337: I dont think any of these helper functions need `basis_gates=None`. They should have a required basis gates arg, and you should do this formatting of basis gates to add the loop functions in the call to the `_compile` function

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
+        if inlined_body:
+            ret.append(AerMark(break_label, inlined_body.num_qubits),
+                       range(inlined_body.num_qubits), [])
+        return ret
+
+    def _inline_while_loop_op(self, inst):
+        condition_tuple = inst.condition
+        body, = inst.params
+
+        self._last_flow_id += 1
```
chriseclectic, line 396: This variable shouldn't be stored in the class. If its a counter it should be passed as an argument and returned from the required functions.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
+
+        return ret
+
+    def _compile(self, circuits):
+        if isinstance(circuits, list):
+            all_static = all([not self._is_dynamic(circuit) for circuit in circuits])
+            if all_static:
+                return circuits
+            basis_gates = self.configuration().basis_gates + ['mark', 'jump']
+            return [transpile(self._inline_circuit(self._decompose_subcircuits(circuit),
```
chriseclectic, line 526: This function is very difficult to read. It would be a lot clearer if it was something like:
```python
if not isinstance(circuits, list):
    circuits = [circuits]

compiled_circuits = []
for circuit in circuits:
    if not _is_dynamic(circuit):
         compiled_circuits.append(circuit)
         continue

    decomposed_circuit = _decompose_subcircuits(circuit)
    compiled_circuits.append(_inline_circuit(decomposed_circuit, None, None))

return compiled_circuits
```

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
+
+        ret.append(AerMark(if_end_label, ret.num_qubits),
+                   range(ret.num_qubits), [])
+
+        return condition_tuple, if_true_label, if_else_label if false_body else if_end_label, ret
+
+    def _convert_c_if_args(self, cond_tuple):
+        return [1 if elem is True else 0 if elem is False else elem for elem in cond_tuple]
+
+    def _inline_circuit(self, circ, continue_label, break_label):
```
chriseclectic, line 470: Can you add doc strings for all these these inline helper functions. They are quite complicated and it is not obvious what they do

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -313,6 +322,217 @@ def _run(self, qobj, job_id=''):
...
+                   range(ret.num_qubits), [])
+        ret.append(AerMark(break_label, inlined_body.num_qubits),
+                   range(inlined_body.num_qubits), [])
+        return condition_tuple, continue_label, loop_start_label, break_label, ret
+
+    def _inline_if_else_op(self, inst, continue_label, break_label):
+        condition_tuple = inst.condition
+        true_body, false_body = inst.params
+
+        self._last_flow_id += 1
```
chriseclectic, line 427: This should be passed as a variable to the function, and returned from function if it needs to be updated so you aren't tracking some internal state of a function call for a single circuit in the simulator class

chriseclectic (changes requested):
Can you move the qasm3 compile functions back to a separate py file and just import that `_compile` function into aerbackend? Having so many helper methods added to the backend class make it very difficult to read that class and almost all these functions do not need to access the simulator class instance at all.

Doing this will help fix some issues with these functions storing temporaries in the simulator state (since there will be no state to store them, they will need to be passed as function arguments and returns which would be better). Same with basis gates, which should always be passed as concrete arguments to these internal functions.

There is also an inefficiencies with checking if the circuits have the qasm3 instructions, since the is_dynamics is called multiple times, resulting in an unnecessary iteration.

src/framework/circuit.hpp
```diff
@@ -382,6 +382,8 @@ void Circuit::set_params(bool truncation) {
   // Counter for current position in ops as we shuffle ops
   size_t op_idx = 0;
   size_t head_end = 0;
+  std::set<std::string> marks;
```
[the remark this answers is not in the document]
hhorii: Current code does not use this information. I wanted to cache addresses or offset to destination instructions here, but transpilation may change instructions.

requirements-dev.txt
```diff
@@ -13,3 +13,4 @@ sphinx-automodapi
 jupyter-sphinx;python_version<'3.8'
 reno>=3.4.0
 ddt>=1.2.0,!=1.4.0
+astunparse
```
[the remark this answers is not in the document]
hhorii: I take it. I used it for tests but I replaced them with new circuit-builder of Terra.

setup.py
```diff
@@ -83,7 +83,7 @@ def install_needed_req(import_name, package_name=None, min_version=None, max_ver
 if not _DISABLE_CONAN:
     setup_requirements.append('conan>=1.22.2')
 
-requirements = common_requirements + ['qiskit-terra>=0.17.0', 'scipy>=1.0']
+requirements = common_requirements + ['qiskit-terra>=0.19.0', 'scipy>=1.0']
```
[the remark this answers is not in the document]
hhorii: Yes.

chriseclectic (approved):
LGTM.

This is missing a release note, I will merge as is but one should be added before release.

PR #1368, opened 2021-10-06
Performance improvement in measurement in MPS

merav-aharoni:
### Summary
Small performance improvement in measurement in MPS.

### Details and comments
Wrote a method that computes the probability of measuring 0 for a single qubit. The previous method used the expectation value, and had some overhead. 
The performance gain was measured on instances of RealAmplitudes with 100 qubits. 
The reduction in time was 17% (from 3.410sec to 2.870sec)

PR #1377, opened 2021-10-21
MPS: New measurement sampling method

merav-aharoni:
### Summary
New measurement sampling algorithm for MPS.

### Details and comments
The algorithm implemented here is based on https://arxiv.org/abs/1709.01662.


yaelbh:
I'm not sure that I agree with the removal of two options for sampling.

If I understand correctly, these are the two algorithms (assume for simplicity that we measure all the qubits) that are currently implemented in the main branch:
1. An algorithm that converts the MPS into a statevector. Involves many contractions.
2. An algorithm that measures the first qubit, updates the MPS, and so on.

Can you please elaborate a bit on the price of each of these two algorithms? I'm at loss at:
- Algorithm 1: how many contractions, which tensor sizes are reached?
- Algorithm 2: what does it take measure to a qubit, and to update the MPS, in terms of time and memory?

Now for the new algorithm of this PR, let's call it Algorithm 3. Like Algorithm 1, it involves of many contractions. But these contractions seem to be more lightweight, because they involve only one matrix per tensor. Can you quantify time and memory for Algorithm 3?

You see that there are many things that I'm missing now, but it seems to me that Algorithm 3 can get to very large matrices that are built contraction after contraction, which perhaps may be sometimes avoided by the strategy of Algorithm 2.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1586,6 +1543,115 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
...
+// We build the randomly selected base value for every shot as follows:
+// For the first qubit, compute its probability for 0 and then randomly select 
+//        the measurement. 'mat' is initialized to the suitable matrix (0 or 1).
+// For qubit i, we store in 'mat'(the contraction of the matrices that were selected up to i-1).
+//        We compute the probability that qubit i is 0 by contracting with matrix 0. 
+//        We randomly select a measurement according to this probability. We then update 'mat' 
+//        by contracting it with the suitable matrix (0 or 1).
+
+reg_t MPS::sample_measure_internal(const reg_t &qubits, const rvector_t &rnds) const {
+  uint_t size = qubits.size();
```
yaelbh, line 1564: As long as the only supported case is the case of all qubits, I think `qubits` should disappear everywhere

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1586,6 +1543,115 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
...
+  double prob = 1;
+  std::string current_measure="";
+  char measure_1_qubit;
+  bool is_first_qubit = true;
+  cmatrix_t mat;
+
+  for (uint_t i=0; i<size; i++) {
+    measure_1_qubit = sample_measure_single_qubit(i, is_first_qubit, current_measure,
```
yaelbh, line 1572: Currently `i` is 0 iff `is_first_qubit` is `True`

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1586,6 +1543,115 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
...
+  double prob0 = 0;
+  if (is_first_qubit) {
+    reg_t qubits_to_update;
+    qubits_to_update.push_back(qubit);
+    // step 1 - measure qubit in Z basis
+    double exp_val = real(expectation_value_pauli_internal(qubits_to_update, "Z", qubit, qubit, 0));
+    // step 2 - compute probability for 0 or 1 result
+    prob0 = (1 + exp_val ) / 2;
+  } else {
+    std::string new_string = '0' + prev_measure;
```
yaelbh, line 1594: `new_string` is not used

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1586,6 +1543,115 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
...
+  prob *= new_prob;
+
+  // Now update mat for the next qubit
+  // mat represents the accumulated product of the matrices of the current
+  // measurement outcome
+  uint_t bit = measurement == '0'? 0 : 1;
+  if (is_first_qubit) {
+    mat = q_reg_[qubit].get_data(bit);
+    if (qubit != 0)  // multiply mat by left lambda
+      for (uint_t col=0; col<mat.GetColumns(); col++)
```
yaelbh, line 1610: I still prefer to separate between implementations and PRs. If this PR doesn't support the case where `qubit` can't be 0, then this code segment should be removed.

yaelbh:
I think I see, that Algorithm 3 really deals only with matrices and matrix multiplications, and not with tensors, so the data structure won't explode. Is it correct?

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1586,6 +1543,115 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
...
+  if (qubit != num_qubits_-1) {
+    for (uint_t row=0; row<temp_mat.GetRows(); row++) {
+      for (uint_t col=0; col<temp_mat.GetColumns(); col++) {
+	temp_mat(row, col) *= lambda_reg_[qubit][col];
+      }
+    }
+  }
+  //prob0 = the probability to measure 0
+  double prob0 = real(AER::Utils::sum( AER::Utils::elementwise_multiplication(temp_mat, AER::Utils::conjugate(temp_mat))));
+  return prob0;
```
yaelbh, line 1637: Could be optimized by not computing `mat * q_reg_[qubit].get_data(0)` again in the cases that measurement 0 is randomized

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1586,6 +1543,115 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
...
+}
+
+reg_t MPS::create_outcome_vector(const reg_t &qubits, const std::string &current_measure) const {
+  uint_t size = qubits.size();
+  reg_t outcome_vector(size), ordered_outcome(size), final_outcome(size);
+
+  // copy string of outcome to vector while reversing the direction of the qubits
+  for (uint_t i=0; i<size; i++) {
+    outcome_vector[size-1-i] = (current_measure[i] == '0') ? 0 : 1;
```
yaelbh, line 1646: What's the benefit of working with a string all the time, and not directly on the vector?

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -1017,46 +1012,17 @@ std::vector<reg_t> State::sample_measure(const reg_t &qubits,
   // of qubits,and the number of shots.
   // The parameters used below are based on experimentation.
   // The user can override this by setting the parameter "mps_sample_measure_algorithm"
-  if (MPS::get_sample_measure_alg() == Sample_measure_alg::PROB){
-    return sample_measure_using_probabilities(qubits, shots, rng);
+  if (MPS::get_sample_measure_alg() == Sample_measure_alg::PROB && 
+      qubits.size() == qreg_.num_qubits()){
+    return sample_measure_all(qubits, shots, rng);
```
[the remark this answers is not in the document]
merav-aharoni: In this case, `sample_measure_using_apply_measure` is invoked, just like before.

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -1089,6 +1055,27 @@ std::vector<reg_t> State::
...
+  rnds_list.reserve(shots);
+  for (int_t i = 0; i < shots; ++i) {
+    rvector_t rands;
+    rands.reserve(qubits.size());
+    for (int_t j = 0; j < qubits.size(); ++j)
+      rands.push_back(rng.rand(0., 1.));
+    rnds_list.push_back(rands);
+  }
+  for (int_t i=0; i<static_cast<int_t>(shots);  i++) {
+    auto single_result = qreg_.sample_measure(qubits, rnds_list[i]);
```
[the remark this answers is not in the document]
merav-aharoni: I agree. Done.

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -1089,6 +1055,27 @@ std::vector<reg_t> State::
...
+  std::vector<rvector_t> rnds_list;
+  rnds_list.reserve(shots);
+  for (int_t i = 0; i < shots; ++i) {
+    rvector_t rands;
+    rands.reserve(qubits.size());
+    for (int_t j = 0; j < qubits.size(); ++j)
+      rands.push_back(rng.rand(0., 1.));
+    rnds_list.push_back(rands);
+  }
+  for (int_t i=0; i<static_cast<int_t>(shots);  i++) {
```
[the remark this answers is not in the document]
merav-aharoni: Changed all the `int_t` to `uint_t`. The reason they were `int_t` was because they were in a parallel section. However, I cancelled the parallelization because it was giving errors.

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -1089,6 +1055,27 @@ std::vector<reg_t> State::
   return all_samples;
 }
 
+std::vector<reg_t> State::sample_measure_all(const reg_t &qubits, 
+		     uint_t shots, 
+		     RngEngine &rng) {
```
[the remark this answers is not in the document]
merav-aharoni: ok. done.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1586,6 +1543,115 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
   }
 }
 
+reg_t MPS::sample_measure(const reg_t &qubits, const rvector_t &rnds) const {
+  reg_t internal_qubits = get_internal_qubits(qubits);
+  return sample_measure_internal(internal_qubits, rnds);
```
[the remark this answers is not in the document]
merav-aharoni: I removed it for now. I will probably need it when I implement the subsets.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: Right. Removed.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: You are right - good comment! I was simply caught up in thinking of measurement as a string. Changed the string to a reg_t. This simplifies the code.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: ok, changed to `if (qubit == 0)`

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: ok, done.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: This is a good idea, but I prefer to do this after I merge in #1368.

merav-aharoni:
In response to your general questions on the algorithm, I am writing a short description of each, including a discussion on complexity.
**Algorithm 1 – compute outcomes from the probabilities vector**
1.	Compute the probabilities vector once O(2^n) matrix multiplications where each matrix multiplication is O(D^3), where n is the number of qubits and D is the bond dimension. Total O(2^n*D^3).
2.	Compute the accumulated probabilities vector O(2^n)
3.	For every shot, determine the outcome by binary search over the accumulated probabilities vector O(n * s), where s is the number of shots.

Total O(2^n*D^3 + s * n). 
This algorithm is best for small numbers of qubits (up to ~20) and large number of shots (>10000). I think it is not worth maintaining this algorithm, just for these cases, especially as the user might as well use the statevector simulator.

**Algorithm 2 – compute outcomes by successive apply_measure operations**
1.	Move all the qubits to sorted position – O(D^3 * n^2). 
This step is very heavy in practice and can probably be eliminated. That would significantly improve this algorithm.
2.	For every shot:
a.	Copy the MPS to a temporary MPS O(n * D^2)
b.	For each qubit
i.	Measure it  O(D^2) (After the improvement in PR #1368)
ii.	Contract with its neighbor O(D^3)
Total – O(s * n * D^3), assuming n < s

**Algorithm 3 – Compute probabilities per qubit per outcome**
1.	For every shot
a.	For every qubit
i.	Compute the probability for an outcome of 0/1 : O(D^2)
ii.	Contract its matrix with the matrix that is the contraction of all previous matrices: O(D^3)
               Total O(s * n * D^3)
               Note that here, for every shot, we contract n matrices, so we have n-1 matrix multiplications. In theory, algorithm 3 appears to have the same complexity as algorithm 2. However, in practice, its performance is much better, as seen in the graph above. My guess is that steps (1) and (2)a. cause a large overhead in algorithm 2.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1586,6 +1543,115 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
...
+}
+
+// The algorithm implemented here is based on https://arxiv.org/abs/1709.01662.
+// Given a particular base value, e.g., 11010, its probability is computed by contracting
+// the suitable matrices per qubit (from right to left), i.e., mat(0) for qubit 0, mat(1) 
+// for qubit 1, mat(0) for qubit 2, mat(1) for qubit 3, mat(1) for qubit 4.
+// We build the randomly selected base value for every shot as follows:
+// For the first qubit, compute its probability for 0 and then randomly select 
+//        the measurement. 'mat' is initialized to the suitable matrix (0 or 1).
+// For qubit i, we store in 'mat'(the contraction of the matrices that were selected up to i-1).
```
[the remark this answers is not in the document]
merav-aharoni: Done.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
merav-aharoni: done.

mishmash:
Regarding the scaling, for Algorithm 3, if the bits getting measured are connected to the boundary, then part i. is O(D) and part ii. is O(D^2) (per bit). Also, I had benchmarked Algorithm 2 a while back and generation of n-bit samples (measuring all qubits) seemed to be O(n * D^2) per sample (after implementation of https://github.com/Qiskit/qiskit-aer/pull/1277). The new Algorithm 3 is just much lighter, without the copying and so on.

merav-aharoni:
I believe both algorithms 2 and 3 have complexity O(n * D^3) per shot in the worst case, because matrix multiplication is D^3. I agree that algorithm 2 has much more overhead, and perhaps that is the reason for the dramatic difference in performance.

merav-aharoni:
@yaelbh , can you approve this PR?

mishmash:
Algorithm 3 should ideally be O(n * D^2) per shot if you start from one boundary, since you only have to update a length-D vector at each bit with a matrix-vector multiplication, which is O(D^2) [see right above Eq. (9) in https://journals.aps.org/prx/pdf/10.1103/PhysRevX.8.031012]. These are the relevant equations from my notes:

image

If you start in the middle of the system, and say only measure a subset of qubits not connected to the boundary, then that will be O(D^3) per bit, since then instead you have to update a D x D matrix with a matrix-matrix multiplication [O(D^3)]. If you measure qubit k first, then the first such matrix (norm) that you have to compute looks like

image

And then the matrix X gets updated in subsequent bit measurements. At least that's the algorithm I have in mind for both cases. Do you concur?

Also, can you clarify how Algorithm 2 should have O(D^3) scaling? I guess it's the SVD step?

merav-aharoni:
I agree with your description. I was thinking of the general case of measuring a subset of the qubits. In that case, we will also have the additional overhead of moving the measured qubits together.

merav-aharoni:
Regarding algorithm 2, I am not quite sure how to compute its complexity. We measure a qubit, and then contract it with its neighbor(s), and then SVD. However, because the qubit was measured, it tensor will be quite small.

yaelbh (approved):

yaelbh (approved):

chriseclectic (approved):

PR #1384, opened 2021-11-05
fix wrong memory free in avx2

hhorii:
### Summary

This PR fixes wrong memory free in avx2 for diagonal matrix multiplication if a number of shots is less than of cores.

### Details and comments

Memory free does not work correctly in avx2. Especially, the following code fails if a number of cores is less than 10, which is a configured `shots`.

```
from qiskit import Aer, transpile
from qiskit.circuit.library import QuantumVolume
from qiskit.providers.aer.noise import depolarizing_error

noise_model = NoiseModel()
noise_model.add_all_qubit_quantum_error(depolarizing_error(0.05, 1), ['u'])
qbackend = AerSimulator(noise_model=noise_model, method="statevector")

circ_list = []
circ = transpile(QuantumVolume(num_qubits=15, depth=1), basis_gates=['u','cx'])
circ.measure_all()

result = qbackend.run(circ, shots=10).result()
```

This PR will fix this issue.

### DIFF
--- a/src/simulators/statevector/qv_avx2.cpp
+++ b/src/simulators/statevector/qv_avx2.cpp
@@ -1207,10 +1207,8 @@ Avx apply_diagonal_matrix_avx<double>(double* qv_data_,
 
   avx_apply_lambda(data_size >> (batch + 1), 1, lambda, omp_threads, input_vec);
 
-#pragma omp parallel for if (omp_threads > 1) num_threads(omp_threads)
-  for (int i = 0; i < omp_threads; ++i) {
-    free(double_tls);
-  }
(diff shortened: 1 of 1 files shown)


jakelishman:
The correction to the frees looks right to me, but I think there might be a memory leak in the allocation block for `double_tls` here: https://github.com/Qiskit/qiskit-aer/blob/eb29c91e5f3e78bf44634892d724367563e1b808/src/simulators/statevector/qv_avx2.cpp#L1175-L1185
 
Looks like we're allocating twice for `double_tls` in all situations there, but the first point gets overwritten?

jakelishman (approved):
Looks correct to me!

chriseclectic (approved):

PR #1386, opened 2021-11-07
[WIP] MPS: Extension of new measurement sampling algorithm to a subset of the qubits

merav-aharoni:
### Summary
This is an extension of PR #1377 to subsets of the qubits.


merav-aharoni:
I made some performance comparisons between this algorithm and the previous one (`apply_measure`), and it appears that the new algorithm does not improve performance. So I think it is not worth making this update. Closing the PR.

PR #1391, opened 2021-11-26
Add transpiler passes for adding instruction-dependent noises

### LINKED ISSUE #1390: Transpiler pass to add gate dependent noise

* Add transpilation pass for inserting noise into a circuit for specific instructions based on a user supplied callable function.
* Add transpilation pass that is a specific case of this pass for thermal relaxation noise model based on gate durations.

The transpilation pass should look something like:

```python
class NoisePass(TransformationPass):

    """Transpiler pass to insert noise into a circuit.
    
    The noise in this pass is defined by a noise function or callable with signature 
    
    .. code:: python
        
            def fn(
                inst: Instruction,
                qubits: Optional[List[int]] = None,
                clbits: Optional[List[int]] = None,
            ) -> Union[QuantumError, Instruction, QuantumCircuit, None]:
    
    For every instance of one of the reference instructions in a circuit the
    supplied function is called on that instruction and the returned noise
    is added to the circuit. This noise can depend on properties of the
    instruction it is called on (for example parameters or duration) to
    allow inserting parameterized noise models.

    Several methods for adding the constructed errors to circuits are supported
    and can be set by using the ``method`` kwarg. The supported methods are

    * ``"append"``: add the return of the callable after the instruction.
    * ``"prepend"``: add the return of the callable before the instruction.
    * ``"replace"``: replace the instruction with the return of the callable.

    """

    def __init__(
        self,
        fn: Callable,
        instructions: Optional[Union[Instruction, Sequence[Instruction] = None,
        method: Optional[str] = 'append'
    ):
        """Initialize noise pass.

        Args:
            fn: noise function `fn(inst, qubits, clbits)`.
            instructions: Optional, single or list of instructions to apply the
                                  noise function to. If None the noise function will be
                                  applied to all instructions in the circuit.
            method: method for inserting noise. Allow methods are
                           'append', 'prepend', 'replace'.
        """
        ...

    # other methods for transpiler pass implementation...
```

This should work to iterate over all instructions in a circuit and for each instance of the specified instructions (or for every instruction is `instructions` is None) depending on the initialized method it should append, prepend, or replace the circuit instruction with the return of the callable.

If the callable returns None, then that should not be added to the circuit. The edge case of `method="replace"` and the callable returns None, the instruction should be removed from the circuit.

In addition to this general pass we should also add a specific thermal relaxation pass.

```python
class RelaxationNoisePass(NoisePass):
    """Add duration dependent thermal relaxation noise after instructions."""
    
    def __init__(
        self,
        instructions: Optional[Union[Instruction Sequence[Instructions]] = None,
        t1s: Optional[List[float]] = None,
        t2s: Optional[List[float]] = None,
        excited_state_populations: Optional[List[float]] = None,
        backend: Optional[Backend] = None,
        unit: str = 's',
    ):
        """Initialize noise pass.
        
        Args:
            instruction: Optional, the instruction to add relaxation to. If None
                 relaxation will be added to all instructions.
            t1s: Optional, list of T1 times for each qubit.
            t2s: Optional, list of T2 times for each qubit.
            excited_state_populations: Optional, list of excited state populations
                for each qubit at thermal equilibrium. If not supplied or obtained
                from the backend this will be set to 0 for each qubit.
            backend: Optional, backend object to extract T1, T2 values from.
            unit: Time unit for T1 and T2 values if not obtained from a
                  backend (s, us, ms, ns, etc).
        """
        # If `t1s` and `t2s` are supplied use them for getting T1, T2 values
        # for each qubit. Otherwise use backend.properties to construct
        # these values.
        ...

        self._t1s = np.asarray(t1s)
        self._t2s = np.asarray(t2s)
        if excited_state_populations is not None:
            self._p1s = np.asarray(excited_state_populations)
        else:
            self._p1s = np.zeros(len(t1s))
        super().__init__(self._relaxation_error, instructions=instructions, method="append")

        # Note: We might need extra kwarg such as dt for converting gate durations
        # in circuit into same time units as T1 T2 times...

        # Note: In principle you could compute excited_state_populations from backend
        # based on each qubit frequency and the effective temperature of the
        # qubit. I don't think this temperature info is in the backend though...
   
    # pylint: disable = unused-argument
    def _thermal_relaxation_error(
        gate: Instruction, =
        qubits: Sequence[int],
        clbits: Optional[Sequence[int]] = None.
        ):
        """Return thermal relaxation error on each gate qubit"""
        duration = gate.duration
        if duration == 0:
            return None
    
        t1s = self._t1s[qubits]
        t2s = self._t2s[qubits]
        p1s = self._p1s[qubits]

        # Note: Make sure to convert duration and t1, t2 to same time units!
        
        if gate.num_qubits == 1:
            t1, t2, p1 = t1s[0], t2s[0], p1s[0]
            if t1 == np.inf and t2 == np.inf:
                return None
            return thermal_relaxation_error(t1, t2, duration, p1)
        
        # General multi-qubit case
        noise = QuantumCircuit(gate.num_qubits)
        for qubit, (t1, t2, p1) in enumerate(zip(t1s, t2s, p1s)):
            if t1 == np.inf and t2 == np.inf:
                # No relaxation on this qubit
                continue
            error = thermal_relaxation_error(t1, t2, duration, p1)
            noise.append(error, [qubit])

        return noise
```

#### Basic Example 

After adding these passes one could add duration dependent relaxation noise to delay gates as follows:

```python
from qiskit.circuit import Delay

backend = AerSimulator.from_backend(ibmq_backend)
delay_pass = RelaxationNoisePass(Delay, backend=backend)

# Build circuits
circuits = ...

# Schedule to insert delays
sched_circuits = transpile(circuits, backend, scheduling_method='asap')

# Add delay noise
sched_circuits = [delay_pass(circ) for circ in sched_circuits]

# Run on simulator with rest of device noise model for other gates
result = backend.run(sched_circuits).result()
```

itoko:
### Summary
Add transpiler passes for inserting instruction-dependent noises and upgrade `NoiseModel.from_backend` so that it adds relaxation noises on delays as default.

This changes the default noisy simulators' behavior against circuits with delays such as T1 experiments. Previously, simulation with default noise model produces the same state |1> probabilities unexpectedly because it adds noises only on X gates and measurements and does not add any noises on the delays. After this PR, the default noisy simulator will expectedly produce different state |1> probabilities depending the duration of delays.

### Details and comments
Closes #1390
- Add two noise passes that insert noises to circuits: `LocalNoisePass` and `RelaxationNoisePass`
- Upgrade `NoiseModel.from_backend` so that it adds relaxation noises on delays as default (that means the resulting model requires circuits to be scheduled before running)
- Deprecate `BackendProperties` support in `NoiseModel.from_backend`


qiskit/providers/aer/noise/noise_model.py
```diff
@@ -1045,3 +1063,34 @@ def _nonlocal_quantum_errors_equal(self, other):
                     if iinner_dict1[iinner_key] != iinner_dict2[iinner_key]:
                         return False
         return True
+
+    def pass_manager(self, only_custom: bool = False) -> Optional[PassManager]:
```
chriseclectic, line 1067: We could just return a trival (empty) pass manager for the case where there are no passes. Then later you can always assume you can do `pm.run(circuits)`.

This would require changes to pass manager that if it has no passes it just returns the input circuits 
```python
class PassManager:
   ...
    def run(
        self,
        circuits: Union[QuantumCircuit, List[QuantumCircuit]],
        output_name: str = None,
        callback: Callable = None,
    ) -> Union[QuantumCircuit, List[QuantumCircuit]]:
    if not self._pass_set and output_name is None and callback is None:
        return circuits
    # current code...
```

otherwise it will have bad performance

chriseclectic: I would also suggest leaving this as in internal method for this release until we have implemented converting the full noise model to a pass manager (which needs to either handle non-local errors or wait till they have been removed after deprecation period as mentioned below)
```suggestion
    def _pass_manager(self) -> Optional[PassManager]:
```

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -182,6 +182,11 @@ def run(self,
                         run_options[key] = value
             qobj = self._assemble(circuits, **run_options)
         else:
+            noise_model = run_options.get("noise_model", self.options.get("noise_model"))
+            if noise_model:
```
chriseclectic, line 186: This should be built into either the `_assemble` or  `_assemble_noise_model` function, not handled as a special case outside it. The result of calling `qobj = backend._assemble(circuits)` should be exactly the qobj that is run.

My suggestion would to modify `_assemble_noise_model` like:
```python
        updated_noise = False
        noise_model = run_options.get(
            'noise_model', getattr(self.options, 'noise_model', None))
        
        # Add custom pass noise
        if noise_model:
            npm = noise_model._pass_manager()
            if npm is not None:
                circuits = npm.run(circuits)
         
        # Check if circuits contain quantum error instructions
        run_circuits = []
        # .. rest of code...

```

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -535,9 +531,15 @@ def from_backend(cls, backend, **options):
...
         # Use automatic noise model if none is provided
         if 'noise_model' not in options:
-            noise_model = NoiseModel.from_backend(backend)
-            if not noise_model.is_ideal():
-                options['noise_model'] = noise_model
+            # pylint: disable=import-outside-toplevel
+            # Avoid cyclic import
+            from ..noise.noise_model import NoiseModel
+
+            if 'delay_noise' in options:
```
chriseclectic, line 538: I dont think this is necessary here. If you want to do custom configuration of the noise model you should do `noise_model=NoiseModel.from_backend(backend, **custom_noise_options)`.

It is strange to only have handling for one of the noise model options here.

qiskit/providers/aer/noise/passes.py
```diff
@@ -0,0 +1,209 @@
...
+        Args:
+            dag: DAG to be changed.
+        Returns:
+            A changed DAG.
+        Raises:
+            TranspilerError: if generated operation is not valid.
+        """
+        qubit_indices = {qubit: idx for idx, qubit in enumerate(dag.qubits)}
+        for node in dag.topological_op_nodes():
+            if self._ops and node.op.name not in self._ops:
```
chriseclectic, line 96: I originally thought we should use instances of ops rather than names. Names strings for instructions can be a bit problematic because you can get situations where a circuit might end up with the same name as an instruction you added to the noise model.

The original motivation for using instances is so you could define diferent noise for cases of the same instruction by subclassing and doing something like

```
class CX1(CXGate):
    pass

class CX2(CXGate):
     pass
```
And add different errors to to the alternative versions of the same gate.

If you make this change then this code would look like something like
```
qubit_indices = {qubit: idx for idx, qubit in enumerate(dag.qubits)}
op_types = tuple(self._ops)
 for node in dag.topological_op_nodes():
    if self._ops and isinstance(node.op, op_types):
```

qiskit/providers/aer/noise/passes.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        if method not in {"append", "prepend", "replace"}:
+            raise TranspilerError(
+                f'Invalid method: {method}, it must be "append", "prepend" or "replace"'
+            )
+        if isinstance(ops, str):
+            ops = [ops]
+        super().__init__()
+        self._fn = fn
+        self._ops = set(ops) if ops else {}
```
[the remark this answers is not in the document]
chriseclectic: If you use strings instead of instances / objects this doesn't match the type hints. You would need to do `instruction.name` to add to the set. Personally I think using specific instructions would be better than using strings so this would be something like:

`self._ops = set(type(op) if isinstance(op, Instruction) else op for op in ops)`

chriseclectic (changes requested):
This is looking good but is very little light on tests, especially for the correctness of the quantum errors inserted by the relaxation noise model.

My main request to get this in to the next release is to simplify some of the changes to the `NoiseModel` class so that it only has the custom pass for delay noise, and to make `pass_manager` and internal function until we can fully convert all the noise to a pass manager after the depreciation period for non-local errors has finished.

Other than that there a a few other issues and suggestions i've pointed out in specific comments below.

qiskit/providers/aer/noise/passes.py
```diff
@@ -0,0 +1,209 @@
...
+from qiskit.dagcircuit import DAGCircuit
+from qiskit.transpiler import TransformationPass
+from qiskit.transpiler.exceptions import TranspilerError
+from .errors import QuantumError
+from .errors.standard_errors import thermal_relaxation_error
+
+InstructionLike = Union[Instruction, QuantumError]
+
+
+class LocalNoisePass(TransformationPass):
```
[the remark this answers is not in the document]
itoko: 37e8e657

qiskit/providers/aer/noise/noise_model.py
```diff
@@ -208,7 +213,8 @@ def from_backend(cls, backend,
                      gate_lengths=None,
                      gate_length_units='ns',
                      standard_gates=None,
-                     warnings=True):
+                     warnings=True,
+                     delay_noise=False):
```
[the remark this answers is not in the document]
itoko: I agree with the change of the meaning of `delay_noise` option.
However, I have some reservations on changing its default value.
It will drastically change the default behavior of `NoiseModel.from_backend`. All the existing codes using default noise model will stop working and raise errors that tells circuits must be scheduled. That's why I'm hesitating to set `delay_noise=True`. Do you think that is OK?

qiskit/providers/aer/noise/noise_model.py
```diff
@@ -1045,3 +1063,34 @@ def _nonlocal_quantum_errors_equal(self, other):
...
+        passes = []
+        if not only_custom:
+            noise_model_pass = LocalNoisePass(self._noise_model_fn, self._noise_instructions)
+            passes.append(noise_model_pass)
+        passes.extend(self._custom_noise_passes)
+        if len(passes) > 0:
+            return PassManager(passes)
+        return None
+
+    def _noise_model_fn(self, inst, qubits):
```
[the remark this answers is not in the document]
itoko: 3f6239a8

qiskit/providers/aer/noise/noise_model.py
itoko: Yeah, as you pointed out, I wanted to return empty pass manager but I checked the implementation of the current PassManager and ended up with returning None in the case. In 3f6239a8, I've kept the return type optional but made the function private. So we can change not to return None without deprecation after improving the PassManager implementation.

qiskit/providers/aer/backends/aerbackend.py
itoko: 631108d9

qiskit/providers/aer/noise/noise_model.py
chriseclectic: We can change it to an empty pass manger after release. I made a PR to Terra which will be included in the 0.19 release to make pass managers return the input circuits if there are no passes so performance shouldn't be an issue

qiskit/providers/aer/noise/noise_model.py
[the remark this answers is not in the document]
chriseclectic: No errors should be raised (if they are that is something that needs to change).

The relaxation noise noise pass should not care if the input is a scheduled circuit or not. It only looks for the specified instructions (delay in this case) and add noise if present with a non-zero duration. Scheduling is an optional step you can do to add delays. If you don't schedule your circuit, your simulation will be exactly the same as before (unless you had manually added delays to your circuit). If you do schedule your circuit then you will get the extra delay noise (this should be added to API documentation for noise model, and basic device noise tutorial)

In general the relaxation noise pass should also work on a non-scheduled circuit by doing nothing (because the instructions would have `None` for duration if not scheduled).

This reminds me that we need to add a release note explaining new additions. In particular it should have both a `feature` section (for the new passes and addition to noise model) and `upgrade` section that points out this change if you run a scheduled circuit or circuit with delays.

qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
```diff
@@ -0,0 +1,105 @@
...
+        """Run the RelaxationNoisePass pass on `dag`.
+        Args:
+            dag: DAG to be changed.
+        Returns:
+            A changed DAG.
+        Raises:
+            TranspilerError: if failed to insert noises to the dag.
+        """
+        if dag.duration is None:
+            raise TranspilerError("This pass accepts only scheduled circuits")
```
chriseclectic, line 103: This needs to be removed. If dag duration is None or 0 it should just return the input dag rather than raise an exception
```python
if dag.duration:
    return super().run(dag)
return dag
```

qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
```diff
@@ -0,0 +1,105 @@
...
+        super().__init__(self._thermal_relaxation_error, ops=ops, method="append")
+
+    def _thermal_relaxation_error(
+            self,
+            op: Instruction,
+            qubits: Sequence[int]
+    ):
+        """Return thermal relaxation error on each gate qubit"""
+        duration = op.duration
+        if duration == 0:
```
chriseclectic, line 65:
```suggestion
        if not duration:
```
So duration = None also returns None

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -360,6 +360,14 @@ def _assemble_noise_model(self, circuits, **run_options):
         noise_model = run_options.get(
             'noise_model', getattr(self.options, 'noise_model', None))
 
+        # Add custom pass noise only to QuantumCircuit objects
+        if noise_model and all(isinstance(circ, QuantumCircuit) for circ in circuits):
+            npm = noise_model._pass_manager()
+            if npm is not None:
+                circuits = npm.run(circuits)
```
chriseclectic, line 367: Since we only have the delay pass at the moment we could improve the efficiency of this by checking if circuits have duration so we can avoid the circuit <-> dag conversion if they dont by doing:
```python
# Check if any circuits have a non-zero duration.
# Since the noise pass currently only contains delay noise we can skip it for
# zero duration circuits to avoid the overhead of circuit <-> dag conversions.
has_duration = any(circ.duration for circ in circuits)
if has_duration and npm is not None:
    circuits = npm.run(circuits)
```

qiskit/providers/aer/noise/noise_model.py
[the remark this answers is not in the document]
itoko: I understand no errors should be raised for non-scheduled circuit. In 1672eb0, I've updated the relaxation noise pass so that it can add noises to gates as well as delays relying on `InstructionDurations` supplied instead of `dt`. It's different from your suggestion but what do you think of it? does it make sense? It could be seen as a problem who the relaxation noise pass should ask to know durations, `InstructionDurations` or `Instruction.duration`. (Both of them might be deprecated after circuit has durations info in its metadata, though)

qiskit/providers/aer/backends/aer_simulator.py
itoko: Done at 053130f

qiskit/providers/aer/backends/aerbackend.py
itoko: Since https://github.com/Qiskit/qiskit-terra/pull/7334 is merged in Terra, we don't need to add this hack, right?

qiskit/providers/aer/noise/passes.py
itoko: Done at 5753075

qiskit/providers/aer/noise/passes.py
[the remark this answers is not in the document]
itoko: 5753075

qiskit/providers/aer/noise/passes.py
```diff
@@ -0,0 +1,209 @@
...
+            TranspilerError: if generated operation is not valid.
+        """
+        qubit_indices = {qubit: idx for idx, qubit in enumerate(dag.qubits)}
+        for node in dag.topological_op_nodes():
+            if self._ops and node.op.name not in self._ops:
+                continue
+
+            qubits = [qubit_indices[q] for q in node.qargs]
+            new_op = self._fn(node.op, qubits)
+            if new_op is None:
```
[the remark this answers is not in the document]
itoko: Done at d478674

qiskit/providers/aer/backends/aer_simulator.py
itoko: I also removed `delay_noise` option from `NoiseModel.from_backend` (e959be86).
We now have no public way to replicate old behavior but we still have a private way with one additional line for it:
```
noise_model = NoiseModel.from_backend(backend)
noise_model._custom_noise_passes = []	 # a bit hack though
```

qiskit/providers/aer/noise/noise_model.py
```diff
@@ -341,9 +348,18 @@ def from_backend(cls, backend,
                 warnings=warnings)
         for name, qubits, error in gate_errors:
             noise_model.add_quantum_error(error, name, qubits, warnings=warnings)
+        # Add delay errors
+        if delay_noise:
+            delay_pass = RelaxationNoisePass(
```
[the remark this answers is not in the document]
itoko: Done at c692a890

qiskit/providers/aer/noise/passes.py
```diff
@@ -0,0 +1,209 @@
...
+
+            dag.substitute_node_with_dag(node, new_dag)
+
+        return dag
+
+
+class RelaxationNoisePass(LocalNoisePass):
+    """Add duration dependent thermal relaxation noise after instructions."""
+
+    def __init__(
```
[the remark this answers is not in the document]
itoko: I think `NoiseModel.from_backend` is enough for end users and I don't really like a lot of classes have `from_backend` method, which introduces dependency to `Backend` class (i.e. reduces the modularity of codes). In addition, Terra is now migrating from Backend V1 to Backend V2. So if we introduced `from_backend`, we would need to update it immediately. I don't think we should add it at least in this PR.

qiskit/providers/aer/backends/aerbackend.py
itoko: I know we need to care about performance very much.
But what if we have ordinary (non-scheduled) circuits with delays?
I think to do this correctly, we need to count delays in circuits efficiently.
That could be done (using `ContainsInstruction("delay")` pass) after we reconstruct the entire NoiseModel as a pass manager in the future.
I have no good idea so far, so I'll leave this line as it is.

itoko:
I've tested T1 noisy simulation with this commit and qiskit-experiment and confirmed that it works as expected.

Test code:
```
from qiskit_experiments.library import T1 as T1Experiment

from qiskit import IBMQ

IBMQ.load_account()
provider = IBMQ.get_provider(hub='ibm-q', group='open', project='main')
device_backend = provider.get_backend('ibmq_armonk')
print(f"(Backend: {device_backend.name()}, Qubit: 0)")
print(f"Reported T1={device_backend.properties().t1(0):.6f}[s]")

new_sim = AerSimulator.from_backend(device_backend)
t1exp = T1Experiment(0, delays=list(range(1, 200, 10)), unit="us")
t1exp.set_transpile_options(scheduling_method="alap")
exp_data = t1exp.run(new_sim, seed_simulator=123).block_for_results()
# display(exp_data.figure(0))
result = exp_data.analysis_results()[0]
print(f"Simulated T1={result.value}")
```
Console output:
```
(Backend: ibmq_armonk, Qubit: 0)
Reported T1=0.000184[s]
Simulated T1=0.00018127391979888208 ± 2.586368725374051e-05 s
```

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: Does such a circuit still register as `circuit.duration = 0`? If so then yes we couldn't include this. If we need to do another pass to check, it's not worth doing  since we might as well just run the pass manager.

chriseclectic: I just checked this and indeed `circuit.duration` is `None` if you manually add delays, so this should be left as it currently is

qiskit/providers/aer/noise/noise_model.py
```diff
@@ -299,15 +305,24 @@ def from_backend(cls, backend,
...
+                ' has been deprecated as of qiskit-aer 0.10.0 and'
+                ' will be removed no earlier than 3 months from that release date.',
+                DeprecationWarning, stacklevel=2)
             properties = backend
             basis_gates = set()
             for prop in properties.gates:
                 basis_gates.add(prop.gate)
             basis_gates = list(basis_gates)
+            num_qubits = len(properties.qubits)
+            dt = 1  # dummy value (may cause unrealistic noises)
```
chriseclectic, line 325: I think it would be better to set `dt=0` and disable delay noise and update the deprecation warning to add to the above warning that "duration dependent delay noise on scheduled circuits is disabled when initialized with backend properties".
```suggestion
            dt = 0 # disable delay noise if dt is unknown
```

qiskit/providers/aer/noise/noise_model.py
```diff
@@ -1034,3 +1066,15 @@ def _nonlocal_quantum_errors_equal(self, other):
...
+        """
+        Return the pass manager that add custom noises defined as noise passes
+        (stored in the _custom_noise_passes field). Note that the pass manager
+        does not include passes to add other noises (stored in the different field).
+        """
+        passes = []
+        passes.extend(self._custom_noise_passes)
+        if len(passes) > 0:
+            return PassManager(passes)
+        return None
```
chriseclectic, line 1080: The fix was included in terra 0.19 that an empty pass manager just returns the input circuits without dag conversions.

qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class RelaxationNoisePass(LocalNoisePass):
+    """Add duration dependent thermal relaxation noise after instructions."""
+
+    def __init__(
+            self,
+            t1s: List[float],
+            t2s: List[float],
+            dt: float,
```
chriseclectic, line 32: General question: Are delay units in scheduled circuits always in terms of a backends dt?

qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
```diff
@@ -0,0 +1,93 @@
...
+            op_types: Optional[Union[type, Sequence[type]]] = None,
+            excited_state_populations: Optional[List[float]] = None,
+    ):
+        """Initialize RelaxationNoisePass.
+
+        Args:
+            t1s: List of T1 times in seconds for each qubit.
+            t2s: List of T2 times in seconds for each qubit.
+            dt: Backend sample time (resolution) in seconds,
+                the time unit used in instruction durations of a scheduled circuit.
```
chriseclectic, line 42:
```suggestion
                the value is used as the time unit of durations in scheduled circuits.
```

qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
```diff
@@ -0,0 +1,93 @@
...
+            t1s: List[float],
+            t2s: List[float],
+            dt: float,
+            op_types: Optional[Union[type, Sequence[type]]] = None,
+            excited_state_populations: Optional[List[float]] = None,
+    ):
+        """Initialize RelaxationNoisePass.
+
+        Args:
+            t1s: List of T1 times in seconds for each qubit.
```
chriseclectic, line 39: One thing I thought that would be nice is if this pass could work for an arbitrary number of qubits all with the same T1 and T2 by doing `RelaxationNoisePass(t1s = t1, t2s=t2, dt, op_types)`, where t1,t2 are floats. (Same with excited state populations if its a float rather than list). I'll leave code comments below for how this could be implemented

qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
```diff
@@ -0,0 +1,93 @@
...
+            excited_state_populations: Optional, list of excited state populations
+                for each qubit at thermal equilibrium. If not supplied or obtained
+                from the backend this will be set to 0 for each qubit.
+        """
+        self._t1s = np.asarray(t1s)
+        self._t2s = np.asarray(t2s)
+        if excited_state_populations is not None:
+            self._p1s = np.asarray(excited_state_populations)
+        else:
+            self._p1s = np.zeros(len(t1s))
```
chriseclectic, line 54:
```suggestion
            self._p1s = np.zeros(self._t1s.shape)
```

qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
```diff
@@ -0,0 +1,93 @@
...
+        if not op.duration:
+            if op.duration is None:
+                warnings.warn("RelaxationNoisePass ignores instructions without duration,"
+                              " you may need to schedule circuit in advance.", UserWarning)
+            return None
+
+        # convert duration in seconds
+        duration = op.duration * self._dt
+
+        t1s = self._t1s[qubits]
```
chriseclectic, line 73: For arbitrary number of qubits with same T1/T2s
```python
        t1s = self._t1s[qubits] if self._t1s.shape else len(qubits) * [float(self._t1s)]
        t2s = self._t2s[qubits] if self._t2s.shape else len(qubits) * [float(self._t2s)]
        p1s = self._p1s[qubits] if self._p1s.shape else len(qubits) * [float(self._p1s)]
```

chriseclectic (approved):
This looks great now. I have a suggestion for making the RelaxationNoise pass work with arbitrary number of qubits with same T1 / T2 (which is handy for quick simulations), but this can be done in a separate PR.

chriseclectic (approved):

PR #1392, opened 2021-11-26
remove threadprivate memaloc in omp

hhorii:
### Summary

https://github.com/Qiskit/qiskit-aer/pull/1384 is not enough to fix all the cases.
This fix will resolve issues related to allocate temporal memory in avx.

### Details and comments

The fix of https://github.com/Qiskit/qiskit-aer/pull/1384 is not enough to run following codes.

```
from qiskit import transpile
from qiskit.circuit.library import QFT

circ = QFT(20)
circ.measure_all()

simulator = AerSimulator()

result = simulator.run([transpile(circ, simulator, basis_gates=['u3', 'cx'])] * 4,
                       **{'_parallel_experiments': 2,
                          '_parallel_state_update': 2,
                          'shots': 10}).result()
```

Because of the structure of codes, temporal memory is allocated and freed in the other OMP blocks in avx codes.
There is no guarantee to use the same set of threads to run these blocks, especially when nested omp is enabled.

This PR changes memory allocation to malloc redundant spaces. Though memory is consumed slightly more, memory is allocated and freed always-correctly.

### DIFF
--- a/src/simulators/statevector/qv_avx2.cpp
+++ b/src/simulators/statevector/qv_avx2.cpp
@@ -1155,12 +1155,6 @@ static m256_t<float> _load_diagonal_input(const std::complex<float>* input_vec,
   return _mm256_load(reinterpret_cast<float*>(tmp));
 }
 
-// Thread-local storage for apply_diagonal_matrix_avx
-std::complex<double>* double_tls;
-#pragma omp threadprivate(double_tls)
-std::complex<float>* float_tls;
-#pragma omp threadprivate(float_tls)
-
 template <>
 Avx apply_diagonal_matrix_avx<double>(double* qv_data_,
                                       const uint64_t data_size,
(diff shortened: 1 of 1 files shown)


src/simulators/statevector/qv_avx2.cpp
```diff
@@ -1172,16 +1166,16 @@ Avx apply_diagonal_matrix_avx<double>(double* qv_data_,
   auto qv_data = _to_complex(qv_data_);
   const auto input_vec = _to_complex(vec_);
 
-#pragma omp parallel if (omp_threads > 1) num_threads(omp_threads)
-  {
+  // conservatively allocate temporal thread-local memory
+  // some omp implementation does not correctly manage threadprivate
+  auto omp_max_threads = omp_get_max_threads();
```
jakelishman, line 1171: Can you absolutely rely on OpenMP being available at build time?  The `#include <omp.h>` at the top is protected in an `#ifdef _OPENMP`.  The previous `#pragma omp` forms silently degrade to no-ops if OpenMP isn't there, but this would be a compilation error due to the missing `omp_get_max_threads` symbol.

hhorii: Thanks. Very good point.

src/simulators/statevector/qv_avx2.cpp
```diff
@@ -1198,16 +1192,15 @@ Avx apply_diagonal_matrix_avx<double>(double* qv_data_,
     const auto until = base + (1UL << (batch + 1));
     for (auto i = base; i < until; i+=2) {
       auto tgt_qv_data = _mm256_load(reinterpret_cast<double*>(&(qv_data[i])));
-      auto input_data = _load_diagonal_input(in_vec, double_tls, i, qregs, qregs_size, q0_mask);
+      auto input_data = _load_diagonal_input(in_vec, double_tmp, i, qregs, qregs_size, q0_mask);
```
jakelishman, line 1195: In the OpenMP case here, how are we ensuring that there's no thread collision in the pointer?  I see that `double_tmp` is allocated space for `2 * omp_max_num_threads` complex doubles, but `_load_diagonal_input` seems to access this pointer at 0 offset.  Should the lambda get its thread number outside of the loop, and access `double_tmp` at an offset of `2 * omp_get_thread_num()` (or 0 if not OpenMP)?

hhorii: Thanks... This is my mistake. The reason why memory is allocated with `thradprivate` was to avoid conflict. Allocating two `double` or four `float` are necessary for each thread. `double_tmp` and `float_tmp` must be accessed via indexers.

jakelishman (approved):
This looks correct to me.  I had wondered if it were possible to allocate `double_tmp` on the stack instead of on the heap to avoid the unpleasant call to `malloc`, but I'm guessing the issue is that `omp_get_num_threads()` is only known at run-time?

hhorii:
@jakelishman Thank you so much. Results of `omp_get_max_threads()` and `omp_get_thread_num()` depend on OMP settings and thread-scheduling. I think `alloca()` is not necessary because `double_tmp` will be always on LLC.

chriseclectic (approved):

PR #1394, opened 2021-12-03
[WIP] temporally disable avx for diagonal matrix multiplication on windows

hhorii:
### Summary

This PR disables use of AVX in diagonal matrix multiplication only on Windows.

### Details and comments

CI fails only on Windows and the last commit is related AVX in diagonal matrix multiplication.
This PR temporally disable AVX only on Windows for debug this issue.


hhorii:
#1395 will fix the same issue.

PR #1400, opened 2021-12-13
Add experimental support of cuQuantum

doichanj:
### Summary
This is the experimental support for NVIDIA's cuQuantum Beta 2 (ver 0.1.0).

### Details and comments
We can use cuStateVec APIs instead of Aer's GPU implementations by setting options at runtime (see CONTRIBUTING.md for details). cuStateVec is enabled when building with `CUSTATEVEC_ROOT` with the path to cuQuantum.
By using cuStateVec, we can speed up x2 for large qubits (larger than 22 qubits) but Aer's implementation is still faster for smaller qubits.

Since cuQuantum is beta version, there are some limitations:
- cuStateVec is not thread safe, multi-chunk parallelization (cache blocking) is done by single thread (slow)
- Multi-shots parallelization is disabled (single thread, slow)
- Multi-shots batched optimization is not support for cuStateVec


chriseclectic:
I added On-hold until 0.10 release is out

CONTRIBUTING.md
```diff
@@ -643,6 +643,29 @@ Few notes on GPU builds:
...
+if the number of qubits of input circuit is equal or greater than 22 qubits by default.
+This threshold can be modified by setting `cuStateVec_threshold` option.
+By setting `cuStateVec_enable=False` to disable using cuStateVec. 
+Following example shows how you accelerate 10 or more qubits simulations using cuStateVec.
+
+```
+sim = AerSimulator(method='statevector', device='GPU')
+results = execute(circuit,sim,cuStateVec_enable=True,cuStateVec_threshold=10).result()
+```
+
```
hhorii, line 666: It is better to explain other methods (not only `statevector`) are also supported.

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -216,6 +220,19 @@ class AerSimulator(AerBackend):
       values (16 Bytes). If set to 0, the maximum will be automatically
       set to the system memory size (Default: 0).
 
+    * ``cuStateVec_enable`` (bool): This option enables accelerating by
+      cuStateVec library of cuQuantum from NVIDIA, that has highly optimized
+      kernels for GPUs. This option is enabled when the number of qubits of
+      the input circuit is equal or greater than ``cuStateVec_threshold``.
+      Currently this option only works well for large number of qubits.
+      Also this option will be disabled for noise simulation
+      (Default: True).
```
hhorii, line 229: can we add "this option will be ignored if AerSimulator is not built with cuStateVec support" ?

src/simulators/state.hpp
```diff
@@ -342,6 +342,8 @@ class State {
   complex_t global_phase_ = 1;
 
   int_t max_matrix_qubits_ = 0;
+
+  std::string sim_device_name_;  //name of device
```
hhorii, line 346: no use?

src/simulators/statevector/chunk/chunk_container.hpp
```diff
@@ -534,6 +154,12 @@ class ChunkContainer : public std::enable_shared_from_this<ChunkContainer<data_t
     return matrix_bits_;
   }
 
+  int_t num_pow2_qubits(void)
+  {
+    return num_pow2_qubits_;
+  }
+  void update_pow2_qubits(void);
```
hhorii, line 161: Is `update_pow2_qubits()` called only by `DeviceChunkContainer:Allocate()`?

src/simulators/statevector/chunk/chunk_manager.hpp
```diff
@@ -43,12 +43,15 @@ class ChunkManager
   int num_qubits_;             //number of global qubits
 
   uint_t num_chunks_;          //number of chunks on this process
+  uint_t chunk_index_;         //global chunk index for the first chunk
```
hhorii, line 46: I'm confused why previous builds work well without `chunk_index_`. I guess that this is necessary in MPI execution for each process to know what is the first chunk in the process.

hhorii (commented):
I think introducing a new namespace is necessary for `AER::QV` to identify chunk-based or not. For example, `GateFuncBase` sounds too general and it should be in `AER::QV::CHUNK` or something.

hhorii:
@doichanj a release note is necessary.

src/controllers/aer_controller.hpp
```diff
@@ -636,9 +663,19 @@ void Controller::set_parallelization_circuit(const Circuit &circ,
                                              const Method method)  
 {
   enable_batch_multi_shots_ = false;
-  if(batched_shots_gpu_ && sim_device_ == Device::GPU && circ.shots > 1 && max_batched_states_ >= num_gpus_ && 
-              batched_shots_gpu_max_qubits_ >= circ.num_qubits ){
-    enable_batch_multi_shots_ = true;
+  if(batched_shots_gpu_ && sim_device_ == Device::GPU && 
+     circ.shots > 1 && max_batched_states_ >= num_gpus_ && 
+     batched_shots_gpu_max_qubits_ >= circ.num_qubits ){
+    //cuStateVec is not supported currently, because cuStateVec does not handle conditional functions
```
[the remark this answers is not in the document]
doichanj: I already disable cuStateVec in StateChunk class if multi-shots parallelization is enabled. 
So I delete this from controller.
https://github.com/doichanj/qiskit-aer/blob/da0f42d8cac08de8e3277024d3da9eba02d8a7c8/src/simulators/state_chunk.hpp#L644-L646

src/controllers/aer_controller.hpp
```diff
@@ -1460,7 +1497,7 @@ void Controller::run_circuit_without_sampled_noise(Circuit &circ,
         shot_state.set_parallelization(parallel_state_update_);
         shot_state.set_global_phase(circ.global_phase_angle);
 
-        state.set_max_matrix_qubits(max_bits);
+        shot_state.set_max_matrix_qubits(max_bits);
```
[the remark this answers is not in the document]
doichanj: yes, I found a bug here.

src/simulators/density_matrix/densitymatrix_state.hpp
```diff
@@ -1344,7 +1344,7 @@ void State<densmat_t>::apply_gate_u3(const int_t iChunk, uint_t qubit, double th
 template <class densmat_t>
 void State<densmat_t>::apply_diagonal_unitary_matrix(const int_t iChunk, const reg_t &qubits, const cvector_t & diag)
 {
-  if(BaseState::thrust_optimization_){
+  if(BaseState::thrust_optimization_ || !BaseState::multi_chunk_distribution_){
```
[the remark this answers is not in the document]
doichanj: This is added for cuStateVec to call `apply_diagonal_unitary_matrix,` because `BaseState::thrust_optimization_ == false` when cuStateVec is enabled.

src/simulators/state.hpp
doichanj: I forgot to delete this. deleted

src/simulators/statevector/chunk/chunk_container.hpp
doichanj: Yes. I moved this to DeviceChunkContainer

src/simulators/statevector/chunk/chunk_manager.hpp
doichanj: This is handled in `QubitVectorThrust` previously to pass this value to kernel functions. But in this PR I moved all the kernel functions in `ChunkContainer` that requires this parameter.

CONTRIBUTING.md
doichanj: added

qiskit/providers/aer/backends/aer_simulator.py
doichanj: added

hhorii (approved):

jakelishman:
This probably should also wait for Aer 0.11 - it's a big new feature, and patch releases are usually for bugfixes.

hhorii (changes requested):
Now I found that OpenMP does not work well in any devices. Let me investigate this phenomena is from only my configuration or common.

src/controllers/aer_controller.hpp
```diff
@@ -636,9 +666,16 @@ void Controller::set_parallelization_circuit(const Circuit &circ,
...
+  if(batched_shots_gpu_ && sim_device_ == Device::GPU && 
+     circ.shots > 1 && max_batched_states_ >= num_gpus_ && 
+     batched_shots_gpu_max_qubits_ >= circ.num_qubits ){
+      enable_batch_multi_shots_ = true;
+  }
+
+  if(cuStateVec_enable_){
+    enable_batch_multi_shots_ = false;    //cuStateVec does not support batch execution of multi-shots
+    parallel_shots_ = 1;    //cuStateVec is currently not thread safe
+    return;
```
hhorii, line 678: if `cuStateVec_enable=True` is configured in `AerSimulator.run()`, `parallel_state_update_` is not set. This will produce performance regression if application accidientaly sets `cuStateVec_enable` with `device='CPU'`.

leofang: Question: when `enable_batch_multi_shots_=true` would you create `nShots` copies of the statevector for parallelization? If so & IIUC, I think a proper "workaround" is to create multiple cuStateVec handles (or just retain and reuse a pool of handles at init time to reduce overhead) and use them in parallel.

IMHO though it's beyond a "workaround": even after we fix the thread safety issue, generally speaking it is still challenging for library handles to be shared by multiple host threads. For example, despite cuBLAS supports this usage pattern [they explicitly recommend to not do so](https://docs.nvidia.com/cuda/cublas/index.html#thread-safety2). Thus the handle pool approach is commonly seen in ML/DL frameworks.

doichanj: `enable_batch_multi_shots_=true` is not applicable for cuStateVec currently, because multiple state vectors are calculated in a single CUDA kernel and each state vector refers classical registers to handle branch operations, this is not implemented in cuStateVec. 
Multiple cuStateVec handle is required when `enable_batch_multi_shots_=false` and shot level parallelization is required. In this case, state vectors are independently calculated using OpenMP threads. (Currently cuStateVec is not thread safe and we disable OpenMP parallelization)

leofang: Thanks for explanation @doichanj. I understand better now. So once we fix thread safety we can unblock you for the shot-level parallelization.

test/terra/backends/simulator_test_case.py
```diff
@@ -66,12 +74,39 @@ def _method_device(methods):
...
     data_args = []
     for method in methods:
         if method in gpu_methods:
             for device in available_devices:
                 data_args.append((method, device))
+            #add test cases for cuStateVec if available using special device = 'GPU_cuStateVec'
+            #'GPU_cuStateVec' is used only inside tests not available in Aer
+            #and this is converted to "device='GPU'" and option "cuStateVec_enalbe = True" is added
+            if cuStateVec:
+                data_args.append((method, 'GPU_cuStateVec'))
```
hhorii, line 90: @chriseclectic could you review this change? This is a hack to minimize changes of tests for tests of `cuStateVec` option. `cuStateVec` is an option only for `device=GPU`. Current annotator `supported_methods()` requires tests to take two argument `method` and `device`. Adding new option `cuStateVec_enable` to all the tests is not productive, I believe.

hhorii (approved):

hhorii:
I confirmed that no regressions will be happened with this PR.

PR #1401, opened 2021-12-13
Add Python iterable magic methods to compatibility classes

### LINKED ISSUE #1399: #1388 breaks some existing codes

As @garrison commented https://github.com/Qiskit/qiskit-aer/pull/1388#issuecomment-991795063, #1388 breaks some existing codes, for example,

```python
state = result.get_statevector(qc)
size = len(state)
# need to change as follows
# size = len(state.data)
```

It broke CI of Qiskit applications and we fixed the unit tests.
It may affect users' codes as well. Is it an intentional behavior?

References
https://github.com/Qiskit/qiskit-optimization/pull/298
https://github.com/Qiskit/qiskit-finance/pull/124

jakelishman:
### Summary

This is not entirely complete for all possible magic methods, but covers
the use cases in downstream projects that we know about.

### Details and comments

Fix #1399.

### DIFF
--- a/qiskit/providers/aer/backends/compatibility.py
+++ b/qiskit/providers/aer/backends/compatibility.py
@@ -64,6 +64,11 @@ class Statevector(qi.Statevector):
             np.allclose(self.data, other.data, rtol=self.rtol, atol=self.atol)
         )
 
+    def __bool__(self):
+        # Explicit override to the default behaviour for Python objects to
+        # prevent the new `__len__` from messing with it.
+        return True
+
     # Magic methods for the iterable/collection interface that need forwarding,
     # but bypass `__getattr__`.  `__getitem__` is defined by `Statevector`, so
     # that can't be forwarded (and consequently neither can `__setitem__`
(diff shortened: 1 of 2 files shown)


qiskit/providers/aer/backends/compatibility.py
```diff
@@ -77,19 +107,37 @@ def __eq__(self, other):
...
+
+    def __contains__(self, value):
+        _deprecation_warning(self, "density matrices")
+        return self._data.__contains__(value)
+
+    def __reversed__(self):
+        _deprecation_warning(self, "density matrices")
+        return self._data.__reversed__()
+
+    def __getitem__(self, key):
```
chriseclectic, line 126: These should probably be added to Statevector too for vector access `vec[i]`

chriseclectic (commented):
I think the get/set item can be added to Statevector too

qiskit/providers/aer/backends/compatibility.py
chriseclectic: Never mind, i see this is already in statevector (but has some issues there with only working with str or int keys, not slices/ranges/lists)

chriseclectic (approved):

PR #1404, opened 2021-12-13
Fixes to from_backend for V2

chriseclectic:
### Summary

The `*Simulator.from_backend` and `NoiseModel.from_backend` assume the input backend is a V1 backend.
This changes these functions to raise an exception if a V2 backend is passed until we can work out how to make them support V2 backends.

### Details and comments

This also reworks some of the functionality added in #1391 to be more robust when looking for required backend attributes (like dt, t1, t2)

### DIFF
--- a/qiskit/providers/aer/backends/aer_simulator.py
+++ b/qiskit/providers/aer/backends/aer_simulator.py
@@ -15,7 +15,6 @@ Qiskit Aer qasm simulator backend.
 
 import copy
 import logging
-import warnings
 from qiskit.providers.options import Options
 from qiskit.providers.models import QasmBackendConfiguration
 from qiskit.providers.backend import BackendV2
@@ -569,7 +568,7 @@ class AerSimulator(AerBackend):
         """Initialize simulator from backend."""
         if isinstance(backend, BackendV2):
             raise AerError(
-                "AerSimulator.from_backend is not yet supported for V2 Backends."
+                "AerSimulator.from_backend does not currently support V2 Backends."
             )
         # Get configuration and properties from backend
         configuration = copy.copy(backend.configuration())
--- a/qiskit/providers/aer/backends/aer_simulator.py
+++ b/qiskit/providers/aer/backends/aer_simulator.py
@@ -15,8 +15,10 @@ Qiskit Aer qasm simulator backend.
 
 import copy
 import logging
+import warnings
 from qiskit.providers.options import Options
 from qiskit.providers.models import QasmBackendConfiguration
+from qiskit.providers.backend import BackendV2
 
 from ..version import __version__
 from .aerbackend import AerBackend, AerError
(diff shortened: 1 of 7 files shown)


qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -565,6 +567,10 @@ def name(self):
     @classmethod
     def from_backend(cls, backend, **options):
         """Initialize simulator from backend."""
+        if isinstance(backend, BackendV2):
+            raise AerError(
+                "AerSimulator.from_backend is not yet supported for V2 Backends."
```
jakelishman, line 572:
```suggestion
                "AerSimulator.from_backend does not yet support V2 Backends."
```
(I think the grammar is the wrong way round)

qiskit/providers/aer/noise/noise_model.py
```diff
@@ -302,11 +302,15 @@ def from_backend(cls, backend,
...
+                "V2 Backends are not yet supported by `NoiseModel.from_backend`")
+        if isinstance(backend, (BaseBackend, BackendV1)):
             properties = backend.properties()
-            basis_gates = backend.configuration().basis_gates
-            num_qubits = backend.configuration().num_qubits
-            dt = backend.configuration().dt
+            configuration = backend.configuration()
+            basis_gates = configuration.basis_gates
+            num_qubits = configuration.num_qubits
+            dt = getattr(configuration, "dt", 0)
```
jakelishman, line 313: I'd be tempted to restore this to the previous behaviour, especially this close to release.  Now that we're filtering for `BackendV1`, those properties should be guaranteed to exist in those locations.

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -15,8 +15,10 @@
 
 import copy
 import logging
+import warnings
```
jakelishman, line 18: ?

jakelishman (approved):

HuangJunye:
I am working on this Terra issue of adding mocked V2 backends: https://github.com/Qiskit/qiskit-terra/issues/7391. @mtreinish pointed to me that Aer noise model does not support V2 backend yet. What's the plan for supporting V2 backends? @chriseclectic If you can give me some directions, I can work on it.

HuangJunye:
I had a closer look at this. It seems to me that supporting V2 backend is quite straightforward. If understand the code correctly, the noise model is generated from backend properties, basis gates, num_qubits and dt. So we just need to translate the code to extract these information from V2 backends.

V1:
```python
properties = backend.properties()
configuration = backend.configuration()
basis_gates = configuration.basis_gates
num_qubits = configuration.num_qubits
dt = getattr(configuration, "dt", 0)
```
V2:
```python
properties = backend.properties()
basis_gates = backend.operation_names
num_qubits = backend.num_qubits
dt = backend.dt
```

I think V2 backends support non-homogenous gate sets so `backend.operation_names` will raise an warning if these operations are not global. https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/providers/backend.py#L352-L363

Issue #1398, opened 2021-12-13
time_taken is overwritten


doichanj:
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?
`time_taken` tags are written twice in the metadata.

https://github.com/Qiskit/qiskit-aer/blob/fb4cfa205f6b6860f69980a91a8dbe85e7d0dd22/src/controllers/aer_controller.hpp#L866-L868

and

https://github.com/Qiskit/qiskit-aer/blob/fb4cfa205f6b6860f69980a91a8dbe85e7d0dd22/src/controllers/aer_controller.hpp#L997-L999

so prior one overwrites `time_taken` 

### Suggested solutions

I think one of them should be renamed.

chriseclectic:
Fixed in 1408

PR #1408, opened 2021-12-15
Fix performance regression from QuantumError.to_dict

### LINKED ISSUE #1407: Performance regression in noise model loading for Qiskit Aer 0.10.0
### Informations

- **Qiskit Aer version**: 0.10.0
- **Python version**: 3.7
- **Operating system**: MacOS

The performance of simulations with large noise models greatly decreased in Aer 0.10.0 This is worst for FakeBackend simulators for large number of qubit devices since they have a lot of error terms in their noise models. 

```python
from qiskit.test.mock import FakeParis

# 27-qubit fake backend simulator
backend = AerSimulator.from_backend(FakeParis())

# 1-qubit circuit
circ = QuantumCircuit(1)
circ.x(0)
circ.measure_all()

# Run sim
%timeit result = backend.run(circ).result()
```

Times on 0.9.1
```
%timeit result = backend.run(circ).result()
11.8 ms ± 3.51 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Times on 0.10.0
```
%timeit -r 3 result = backend.run(circ).result()
16.5 s ± 84.6 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)
```

This is about a 1000x slowdown to `run`

### Suggested solutions

After some debugging I believe this is due to #1141 which changed `QuantumError` instructions to use `QuantumCircuit` internally instead of qobj.

The assembly of these circuits to qobj is now being called through the C++ pybind code and has a huge overhead. Doing this before calling the C++ code via calling `noise_model.to_dict` as part of run also results in big overhead. 

Oddly this issue also only seems to effect some version of python, particularly 3.7 which may be due to how assemble is implemented wrt python multiprocessing.

So far I've managed to largely fix the problem by manually assemble quantum errors and bypassing the terra assemble function using
```python
instructions = [[op[0].assemble().to_dict() for op in circ.data for circ in self._circs]
```
instead of current
```python
qobj = assemble(self.circuits)
instructions = [exp.to_dict()['instructions'] for exp in qobj.experiments]
```

With this change I get

```
%timeit result = backend.run(circ).result()
331 ms ± 4.21 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

chriseclectic:
### Summary

Fixes #1407 #1398 

Fixes issue with QuantumError.to_dict causing a huge performance regression in noisy simulations from its calling of `qiskit.assemble` inside Pybind11 code.

### Details and comments

This also fixed an inefficiency in the C++ controller `execute` function where the qobj config was deserialized/parsed twice (which doubled the performance regression in #1407)

### DIFF
--- a/src/controllers/aer_controller.hpp
+++ b/src/controllers/aer_controller.hpp
@@ -332,8 +332,6 @@ protected:
     }
     return max_memory_mb_;
   }
-  // Truncate and remap input qubits
-  bool enable_truncation_ = true;
 
   // The maximum number of threads to use for various levels of parallelization
   int max_parallel_threads_;
@@ -390,8 +388,6 @@ protected:
 //-------------------------------------------------------------------------
 
 void Controller::set_config(const json_t &config) {
-  // Load validation threshold
-  JSON::get_value(enable_truncation_, "enable_truncation", config);
 
   // Load validation threshold
   JSON::get_value(validation_threshold_, "validation_threshold", config);
@@ -847,7 +843,7 @@ Result Controller::execute(const inputdata_t &input_qobj) {
     auto timer_start = myclock_t::now();
 
     // Initialize QOBJ
-    Qobj qobj(input_qobj, enable_truncation_);
+    Qobj qobj(input_qobj);
     auto qobj_time_taken =
         std::chrono::duration<double>(myclock_t::now() - timer_start).count();
 
--- a/src/controllers/aer_controller.hpp
+++ b/src/controllers/aer_controller.hpp
@@ -846,37 +846,31 @@ Result Controller::execute(const inputdata_t &input_qobj) {
     // Start QOBJ timer
     auto timer_start = myclock_t::now();
 
-    Noise::NoiseModel noise_model;
-    json_t config;
+    // Initialize QOBJ
+    Qobj qobj(input_qobj, enable_truncation_);
+    auto qobj_time_taken =
+        std::chrono::duration<double>(myclock_t::now() - timer_start).count();
 
-    // Check for config
-    if (Parser<inputdata_t>::get_value(config, "config", input_qobj)) {
-      // Set config
-      set_config(config);
-      // Load noise model
-      Parser<json_t>::get_value(noise_model, "noise_model", config);
-    }
+    // Set config
+    set_config(qobj.config);
 
-    // Initialize qobj
-    Qobj qobj;
-    if (noise_model.has_nonlocal_quantum_errors()) {
-       // Non-local noise does not work with optimized initialization
-       qobj = Qobj(input_qobj, false);
-    } else {
-      qobj = Qobj(input_qobj, enable_truncation_);
-    }
+    // Run qobj circuits
+    auto result = execute(qobj.circuits, qobj.noise_model, qobj.config);
 
-    auto result = execute(qobj.circuits, noise_model, config);
+    // Add QOBJ loading time
+    result.metadata.add(qobj_time_taken, "time_taken_load_qobj");
+
     // Get QOBJ id and pass through header to result
     result.qobj_id = qobj.id;
     if (!qobj.header.empty()) {
       result.header = qobj.header;
     }
+
     // Stop the timer and add total timing data including qobj parsing
-    auto timer_stop = myclock_t::now();
     auto time_taken =
-        std::chrono::duration<double>(timer_stop - timer_start).count();
+        std::chrono::duration<double>(myclock_t::now() - timer_start).count();
     result.metadata.add(time_taken, "time_taken");
+    
     return result;
   } catch (std::exception &e) {
     // qobj was invalid, return valid output containing error message
@@ -1007,7 +1001,7 @@ Result Controller::execute(std::vector<Circuit> &circuits,
     auto timer_stop = myclock_t::now();
     auto time_taken =
         std::chrono::duration<double>(timer_stop - timer_start).count();
-    result.metadata.add(time_taken, "time_taken");
+    result.metadata.add(time_taken, "time_taken_execute");
   }
   // If execution failed return valid output reporting error
   catch (std::exception &e) {
--- a/test/terra/backends/aer_simulator/test_noise.py
+++ b/test/terra/backends/aer_simulator/test_noise.py
@@ -17,9 +17,11 @@ from ddt import ddt
 from qiskit.providers.aer import noise
 
 import qiskit.quantum_info as qi
+from qiskit import transpile
 from qiskit.circuit import QuantumCircuit, Reset
 from qiskit.circuit.library import QFT
 from qiskit.circuit.library.standard_gates import IGate, HGate
+from qiskit.quantum_info.states.densitymatrix import DensityMatrix
 from test.terra.backends.simulator_test_case import (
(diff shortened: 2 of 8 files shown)


test/terra/noise/test_quantum_error.py
```diff
@@ -283,8 +283,9 @@ def depol_error(self, param, **kwargs):
         )
 
     @staticmethod
-    def aslist(qargs):
-        return [q.index for q in qargs]
+    def qubits_list(circ, qargs):
+        qubit_map = {qubit: i for i, qubit in enumerate(circ.qubits)}
+        return [qubit_map[q] for q in qargs]
```
jakelishman, line 288: `return [circ.find_bit(q).index for q in qargs]` is a one-liner that does the same thing for Terra 0.19+.

test/terra/reference/ref_kraus_noise.py
```diff
@@ -60,15 +60,6 @@ def kraus_gate_error_noise_models():
...
 
-def kraus_gate_error_noise_models_full():
-    """Kraus gate error noise models on many gate types"""
-
-    # Amplitude damping error on "u1", "u2", "u3", "cx"
-    error = amplitude_damping_error(0.2)
-    noise_model = NoiseModel()
-    noise_model.add_all_qubit_quantum_error(error, ['h'])
-    noise_model.add_all_qubit_quantum_error(error.tensor(error), ['cp', 'swap'])
-    return noise_model
```
jakelishman, line 63: Was this function completely unused?

test/terra/backends/aer_simulator/test_noise.py
```diff
@@ -156,15 +158,35 @@ def test_kraus_gate_noise(self, method, device):
...
+        # compute target counts
+        noise_state = DensityMatrix(noise_circuit)
+        ref_target = {i: shots * p for i, p in noise_state.probabilities_dict().items()}
+
+        # Run sim
+        ideal_circuit.measure_all()
+        result = backend.run(ideal_circuit, shots=shots).result()
         self.assertSuccess(result)
-        self.compare_counts(result, [circuit], [ref_target], delta=0.1 * shots)
+        self.compare_counts(result, [ideal_circuit], [ref_target], hex_counts=False, delta=0.1 * shots)
```
jakelishman, line 189: I'm in favour of getting rid of the hard-coded counts, but is there another reason for changing the test in this PR?

src/controllers/aer_controller.hpp
```diff
@@ -846,37 +842,31 @@ Result Controller::execute(const inputdata_t &input_qobj) {
...
       result.header = qobj.header;
     }
+
     // Stop the timer and add total timing data including qobj parsing
-    auto timer_stop = myclock_t::now();
     auto time_taken =
-        std::chrono::duration<double>(timer_stop - timer_start).count();
+        std::chrono::duration<double>(myclock_t::now() - timer_start).count();
     result.metadata.add(time_taken, "time_taken");
+
```
jakelishman, line 869: This is also an `execute` function - should this metadata be changed to `time_taken_execute` too?

src/controllers/aer_controller.hpp
```diff
@@ -846,37 +842,31 @@ Result Controller::execute(const inputdata_t &input_qobj) {
...
-       // Non-local noise does not work with optimized initialization
-       qobj = Qobj(input_qobj, false);
-    } else {
-      qobj = Qobj(input_qobj, enable_truncation_);
-    }
+    // Run qobj circuits
+    auto result = execute(qobj.circuits, qobj.noise_model, qobj.config);
+
+    // Add QOBJ loading time
+    result.metadata.add(qobj_time_taken, "time_taken_load_qobj");
```
jakelishman, line 857: I mean, it's not _wrong_, but it's kind of weird that you add the time-taken metadata for the load _after_ you've done the execute, and it's not grouped with the other metadata adds (the header stuff is between them).

src/controllers/aer_controller.hpp
```diff
@@ -1007,7 +997,7 @@ Result Controller::execute(std::vector<Circuit> &circuits,
     auto timer_stop = myclock_t::now();
     auto time_taken =
         std::chrono::duration<double>(timer_stop - timer_start).count();
-    result.metadata.add(time_taken, "time_taken");
+    result.metadata.add(time_taken, "time_taken_execute");
   }
```
jakelishman, line 1001: This feels like it could be a change to the public API?  If so, it maybe needs a release note.

jakelishman (commented):
I don't think I have the benchmarks you and Hiroshi were using, but this seems fine.  I'm not certain I understand _how_ the change to the `assemble` call helps fix any regression though - perhaps a comment in that file might help.

test/terra/noise/test_quantum_error.py
chriseclectic: Cool, didn't know this was added

jakelishman: yeah, turns out we had that `qubit_map` line in like 1,000 places, so it was starting to become really necessary.  We still need to add something similar to DAGCircuit, because like 80% of the transpiler passes have something similar as well.

test/terra/reference/ref_kraus_noise.py
chriseclectic: It was used in a single test so I just moved it to make it easier to read

jakelishman: Oh yeah, I totally misread the test you modified - I only noticed the other helper, not this one.

test/terra/backends/aer_simulator/test_noise.py
chriseclectic: It started failing but on some jobs, and after debugging it seemed that the ref counts were wrong but I couldn't find any reason in Aer for why they would have changed.

src/controllers/aer_controller.hpp
chriseclectic: The name change was just to distinguish the outer one that includes loading vs the inner one that doesn't

src/controllers/aer_controller.hpp
chriseclectic: You can't add it to the result, until you've created the result (which is done via the second execute call)

src/controllers/aer_controller.hpp
chriseclectic: This was just overriden by the outer time taken call before so was never seen by users. Now it will be a new entry in `result.metadata` so you can actually see it

src/controllers/aer_controller.hpp
jakelishman: Ah, I get it - I got confused a little, because it definitely wouldn't have been my choice to use function overloading to represent the "public" and "private" versions of a function, but that's fine.

src/controllers/aer_controller.hpp
jakelishman: oh yeah, dunno how I missed that sorry

src/controllers/aer_controller.hpp
jakelishman: ok, sounds good.

jakelishman (approved):

PR #1409, opened 2021-12-16
Reduce unnecessary circuit iteration

chriseclectic:
### Summary

This PR complements #1408 to reduce some other possible, but less serious, regressions from Aer 0.10.

### Details and comments

This adds checks to avoid transpiling or iterating over circuits for noise model transpilation, quanutm error replacement, qasm3 compiling, and qobj splitting if the required instructions are not in the circuits. This is done by collecting sets of all instructions in each circuit in advance, and using that set for checks before running the other transpilation/compilation steps.

### DIFF
--- a/qiskit/providers/aer/backends/aer_compiler.py
+++ b/qiskit/providers/aer/backends/aer_compiler.py
@@ -20,6 +20,7 @@ from qiskit.circuit.controlflow import (
     BreakLoopOp,
     ContinueLoopOp)
 from qiskit.compiler import transpile
+from .backend_utils import circuit_optypes
 from ..library.control_flow_instructions import AerMark, AerJump
 
 
@@ -40,27 +41,29 @@ class AerCompiler:
                             (default: None).
 
         Returns:
-            QuantumCircuit or list: QuantumCircuit (or list
-                of QuantumCircuit objects) without control-flow instructions
+            tuple: Typle of list QuantumCircuit without control-flow
+                   instructions and list of optypes for each circuit.
         """
-        if isinstance(circuits, QuantumCircuit):
+        if not isinstance(circuits, list):
             circuits = [circuits]
+        if optypes is None:
+            optypes = len(circuits) * [None]
         if isinstance(circuits, list):
             basis_gates = basis_gates + ['mark', 'jump']
             compiled_circuits = []
             for idx, circuit in enumerate(circuits):
-                optype = optypes[idx] if optypes else None
-                if isinstance(circuit, QuantumCircuit) and self._is_dynamic(
-                        circuit, optype):
+                if self._is_dynamic(circuit, optypes[idx]):
                     compiled_circ = transpile(
                         self._inline_circuit(circuit, None, None),
                         basis_gates=basis_gates
                     )
                     compiled_circuits.append(compiled_circ)
+                    # Recompute optype for compiled circuit
+                    optypes[idx] = circuit_optypes(compiled_circ)
                 else:
                     compiled_circuits.append(circuit)
-            return compiled_circuits
-        return circuits
+            return compiled_circuits, optypes
+        return circuits, optypes
 
     @staticmethod
     def _is_dynamic(circuit, optype=None):
--- a/qiskit/providers/aer/backends/aer_compiler.py
+++ b/qiskit/providers/aer/backends/aer_compiler.py
@@ -76,8 +76,8 @@ class AerCompiler:
         )
 
         # Check via optypes
-        if isinstance(optype, set) and optype.intersection(controlflow_types):
-            return True
+        if isinstance(optype, set):
+            return bool(optype.intersection(controlflow_types))
 
         # Check via iteration
         for inst, _, _ in circuit.data:
--- a/qiskit/providers/aer/backends/aer_compiler.py
+++ b/qiskit/providers/aer/backends/aer_compiler.py
@@ -29,37 +29,58 @@ class AerCompiler:
     def __init__(self):
         self._last_flow_id = -1
 
-    def compile(self, circuits, basis_gates=None):
+    def compile(self, circuits, basis_gates=None, optypes=None):
         """compile a circuit that have control-flow instructions.
 
         Args:
-            circuits (QuantumCircuit or list): The QuantumCircuit (or list
-                of QuantumCircuit objects) to be compiled
-            basis_gates (list): basis gates to decompose sub-circuits(default: None).
+            circuits (QuantumCircuit or list): The QuantumCircuits to be compiled
+            basis_gates (list): basis gates to decompose sub-circuits
+                                (default: None).
+            optypes (list): list of instruction type sets for each circuit
+                            (default: None).
 
         Returns:
             QuantumCircuit or list: QuantumCircuit (or list
                 of QuantumCircuit objects) without control-flow instructions
         """
-        if basis_gates:
-            basis_gates = basis_gates + ['mark', 'jump']
+        if isinstance(circuits, QuantumCircuit):
+            circuits = [circuits]
         if isinstance(circuits, list):
-            return [transpile(self._inline_circuit(circ, None, None),
-                              basis_gates=basis_gates) if self._is_dynamic(circ)
-                    else circ for circ in circuits]
-        else:
-            return (transpile(self._inline_circuit(circuits, None, None),
-                              basis_gates=basis_gates) if self._is_dynamic(circuits)
-                    else circuits)
+            basis_gates = basis_gates + ['mark', 'jump']
+            compiled_circuits = []
+            for idx, circuit in enumerate(circuits):
+                optype = optypes[idx] if optypes else None
+                if isinstance(circuit, QuantumCircuit) and self._is_dynamic(
+                        circuit, optype):
+                    compiled_circ = transpile(
+                        self._inline_circuit(circuit, None, None),
+                        basis_gates=basis_gates
+                    )
+                    compiled_circuits.append(compiled_circ)
+                else:
+                    compiled_circuits.append(circuit)
+            return compiled_circuits
+        return circuits
 
-    def _is_dynamic(self, circuit):
-        """check whether a circuit contains control-flow instructions
-        """
+    @staticmethod
+    def _is_dynamic(circuit, optype=None):
+        """check whether a circuit contains control-flow instructions"""
         if not isinstance(circuit, QuantumCircuit):
             return False
+
+        controlflow_types = (
+            WhileLoopOp, ForLoopOp, IfElseOp, BreakLoopOp, ContinueLoopOp
+        )
+
+        # Check via optypes
+        if isinstance(optype, set) and optype.intersection(controlflow_types):
+            return True
+
+        # Check via iteration
         for inst, _, _ in circuit.data:
-            if isinstance(inst, (WhileLoopOp, ForLoopOp, IfElseOp, BreakLoopOp, ContinueLoopOp)):
+            if isinstance(inst, controlflow_types):
                 return True
+
         return False
 
     def _inline_circuit(self, circ, continue_label, break_label):
@@ -187,8 +208,8 @@ class AerCompiler:
         parent.append(AerMark(if_end_label, true_body.num_qubits), qargs, [])
 
 
-def compile_circuit(circuits, basis_gates=None):
+def compile_circuit(circuits, basis_gates=None, optypes=None):
     """
     compile a circuit that have control-flow instructions
     """
-    return AerCompiler().compile(circuits, basis_gates)
+    return AerCompiler().compile(circuits, basis_gates, optypes)
--- a/qiskit/providers/aer/backends/aer_compiler.py
+++ b/qiskit/providers/aer/backends/aer_compiler.py
@@ -13,6 +13,7 @@
 Compier to convert Qiskit control-flow to Aer backend.
 """
 from qiskit.circuit import QuantumCircuit
+from qiskit.pulse import Schedule, ScheduleBlock
 from qiskit.circuit.controlflow import (
     WhileLoopOp,
     ForLoopOp,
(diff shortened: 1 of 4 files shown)


hhorii (dismissed):

qiskit/providers/aer/backends/aer_compiler.py
```diff
@@ -29,37 +29,58 @@ class AerCompiler:
...
-
-    def _is_dynamic(self, circuit):
-        """check whether a circuit contains control-flow instructions
-        """
+            basis_gates = basis_gates + ['mark', 'jump']
+            compiled_circuits = []
+            for idx, circuit in enumerate(circuits):
+                optype = optypes[idx] if optypes else None
+                if isinstance(circuit, QuantumCircuit) and self._is_dynamic(
+                        circuit, optype):
```
jakelishman, line 54: the `isinstance` check isn't necessary, since `_is_dynamic` does it anyway

qiskit/providers/aer/backends/aer_compiler.py
```diff
@@ -29,37 +29,58 @@ class AerCompiler:
...
+        if isinstance(optype, set) and optype.intersection(controlflow_types):
+            return True
+
+        # Check via iteration
         for inst, _, _ in circuit.data:
-            if isinstance(inst, (WhileLoopOp, ForLoopOp, IfElseOp, BreakLoopOp, ContinueLoopOp)):
+            if isinstance(inst, controlflow_types):
                 return True
+
         return False
```
jakelishman, line 84: I think you want an early `False` return here if `optype` is a set, but there's no intersection.  Otherwise you're looping through the data even though you know it should fail.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -330,11 +331,21 @@ def _format_results(output):
...
         return Result.from_dict(output)
 
+    @staticmethod
+    def _circuit_optypes(circuit):
+        """Retur set of all operation names in a circuit"""
+        if not isinstance(circuit, QuantumCircuit):
+            return set()
+        optypes = set()
+        for instr, _, _ in circuit._data:
+            optypes.add(type(instr))
```
jakelishman, line 341: I think you need to recurse into control-flow ops here.  It'll work for the use case you have in mind, because the only way you could be "hiding" a control-flow op is if it's inside another (so it'd be detected), but it's basically a latent bug waiting to happen otherwise.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -367,10 +393,17 @@ def _assemble(self, circuits, parameter_binds=None, **run_options):
     def _assemble_noise_model(self, circuits, **run_options):
         """Move quantum error instructions from circuits to noise model"""
         if isinstance(circuits, (QasmQobj, PulseQobj)):
-            return circuits, run_options
+            optypes = [set() for _ in range(len(circuits.experiments))]
```
jakelishman, line 396: Unimportant, but since you just need to match the length, it could just be `[set() for _ in circuits.experiments]`.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -379,24 +412,35 @@ def _assemble_noise_model(self, circuits, **run_options):
...
-                    circuits = [circuits]
+                # Get indicies of circuits that need noise transpiling
+                transpile_idxs = [
+                    idx for idx, opset in enumerate(optypes) if Delay in opset
+                ]
+
+                # Transpile only the required circuits
+                transpiled_circuits = npm.run([run_circuits[i] for i in transpile_idxs])
+                if isinstance(transpiled_circuits, QuantumCircuit):
+                    transpiled_circuits = [transpiled_circuits]
```
jakelishman, line 429: You always pass a list - can you still get a scalar back?

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -379,24 +412,35 @@ def _assemble_noise_model(self, circuits, **run_options):
         noise_model = run_options.get(
             'noise_model', getattr(self.options, 'noise_model', None))
 
-        # Add custom pass noise only to QuantumCircuit objects
-        if noise_model and all(isinstance(circ, QuantumCircuit) for circ in circuits):
+        # Add custom pass noise only to QuantumCircuit objects that contain delay
+        # instructions since this is the only instruction handled by the noise pass
+        # at present
```
jakelishman, line 417: It would perhaps be better to add a function to `noise_model` that takes (for example) the circuit and the opset, and returns whether it would affect it - it feels like `AerBackend` shouldn't need to know this for all models, because different noise models might do different things.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -379,24 +412,35 @@ def _assemble_noise_model(self, circuits, **run_options):
...
-                run_circuits.append(circ)
-            else:
+        for idx, circ in enumerate(run_circuits):
+            if QuantumChannelInstruction in optypes[idx] and not isinstance(
+                    circ, (Schedule, ScheduleBlock)):
                 updated_circ = False
                 new_data = []
                 for inst, qargs, cargs in circ.data:
-                    if inst.name == "quantum_channel":
+                    if isinstance(inst, QuantumChannelInstruction):
```
jakelishman, line 443: I don't know if `isinstance` is what you mean here.  It's usually correct in Python, but your `optypes` set is of _exact_ classes, so you'll only enter this loop if there exists an instruction whose type is _exactly_ `QuantumChannelInstruction`.  It's a general problem with storing instructions in the set.

Perhaps my issue here is wiuth `optypes`: instead of just adding `type(instruction)` to the set, you might need to add all types in `type(instruction).__bases__` that are subclasses of `Instruction`/whatever you care about.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -379,24 +412,35 @@ def _assemble_noise_model(self, circuits, **run_options):
...
-        for circ in circuits:
-            if isinstance(circ, (Schedule, ScheduleBlock)):
-                run_circuits.append(circ)
-            else:
+        for idx, circ in enumerate(run_circuits):
+            if QuantumChannelInstruction in optypes[idx] and not isinstance(
+                    circ, (Schedule, ScheduleBlock)):
                 updated_circ = False
                 new_data = []
                 for inst, qargs, cargs in circ.data:
```
jakelishman, line 442: This doesn't recurse into control flow.

qiskit/providers/aer/jobs/utils.py
```diff
@@ -120,14 +120,20 @@ def _split_qobj(qobj, max_size, qobj_id, seed):
...
+        # Optypes store class names as strings
+        return any(
+            "Save" in name or "Snapshot" in name
+            for optype in optypes for name in optype
+        )
+    # Otherwise iterate over instructions
+    return any(
+        "save" in inst.name or "snapshot" in inst.name
+        for exp in experiments for inst in exp.instructions
+    )
```
jakelishman, line 136: This seems very fragile.  It'd be better to test against the exact names and classes - for `optype` it'd even likely be more efficient, because the complexity would scale with the number of "custom instruction types" rather than the number of instruction types in the circuit.

jakelishman (commented):
My main issue here is that you use `optypes: Set[Type[Instruction]]` as a type-test for instructions, but then you frequently use `isinstance` checks.  I'm concerned that things like `QuantumError` may well be subclassed in the future, in which case `optypes` will silently begin to fail.  If you still want it to work, then it needs to add everything in `type(instruction).__bases__` that's a subclass of things you care about.

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: This is a good point. I suppose a more robust thing should be to do the control flow unrolling first, before the noise assembling, and you rebuild the opset after doing each unrolling.

qiskit/providers/aer/backends/aer_compiler.py
chriseclectic: In this case i think returning True is the early case since its basically an `if any` check for a circuit containing any of hte control flow instructions.

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: Probably worth doing something like this eventually

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: I think passmanger  always returns a length one list as the first element, so if you are only transpiling 1 circuit it coudl return as the single circuit

qiskit/providers/aer/jobs/utils.py
chriseclectic: Yeah, I would also like to remove this completely in the future since you should be able to support these instructions now that the type and subtype of save instructions is in metadata.

qiskit/providers/aer/backends/aer_compiler.py
jakelishman: No, I agree with that point, the trick is that `if isinstance(optype, set) and not optype.intersection(controlflow_types)` then you know that the the iterative loop will return `False`.  If `optype` is not `None`, then you should never need to iterate through the instructions.

qiskit/providers/aer/backends/aerbackend.py
jakelishman: if that's the case, then fair enough - I'm only used to `transpile` and things like that that will return a scalar if they're given a scalar.

qiskit/providers/aer/backends/aer_compiler.py
chriseclectic: Ah i see

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: This was a bug introduced assemble when control flow was added. I restructured assemble so that it nows does the control flow compiling first before any of the noise assembly to avoid this issue.

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: Is instance should be correct here now that we've fixed the MRO stuff in optypes

qiskit/providers/aer/backends/aer_compiler.py
```diff
@@ -40,27 +41,29 @@ def compile(self, circuits, basis_gates=None, optypes=None):
...
-            QuantumCircuit or list: QuantumCircuit (or list
-                of QuantumCircuit objects) without control-flow instructions
+            tuple: Typle of list QuantumCircuit without control-flow
+                   instructions and list of optypes for each circuit.
         """
-        if isinstance(circuits, QuantumCircuit):
+        if not isinstance(circuits, list):
             circuits = [circuits]
+        if optypes is None:
+            optypes = len(circuits) * [None]
```
jakelishman, line 50: I prefer this way - much easier to work out what's going on inside the loop.

qiskit/providers/aer/backends/aer_compiler.py
```diff
@@ -40,27 +41,29 @@ def compile(self, circuits, basis_gates=None, optypes=None):
...
 
         Returns:
-            QuantumCircuit or list: QuantumCircuit (or list
-                of QuantumCircuit objects) without control-flow instructions
+            tuple: Typle of list QuantumCircuit without control-flow
+                   instructions and list of optypes for each circuit.
         """
-        if isinstance(circuits, QuantumCircuit):
+        if not isinstance(circuits, list):
             circuits = [circuits]
```
jakelishman, line 48: The previous form was probably better - this way means that if you pass `circuits` as a tuple or any other iterable, it won't work, whereas previously it would (type-hint notwithstanding).

qiskit/providers/aer/backends/aer_compiler.py
```diff
@@ -40,27 +41,29 @@ def compile(self, circuits, basis_gates=None, optypes=None):
...
-                if isinstance(circuit, QuantumCircuit) and self._is_dynamic(
-                        circuit, optype):
+                if self._is_dynamic(circuit, optypes[idx]):
                     compiled_circ = transpile(
                         self._inline_circuit(circuit, None, None),
                         basis_gates=basis_gates
                     )
                     compiled_circuits.append(compiled_circ)
+                    # Recompute optype for compiled circuit
+                    optypes[idx] = circuit_optypes(compiled_circ)
```
jakelishman, line 62: This mutates the input list, if the user supplies it - you just make a shallow copy (`optypes = list(optypes)`) before the loop if you want to do this.

qiskit/providers/aer/backends/aer_compiler.py
```diff
@@ -40,27 +41,29 @@ def compile(self, circuits, basis_gates=None, optypes=None):
...
                 else:
                     compiled_circuits.append(circuit)
-            return compiled_circuits
-        return circuits
+            return compiled_circuits, optypes
+        return circuits, optypes
```
jakelishman, line 66: This is 100% a public API change, and I'd be super nervous about just changing the return type like this.  You can get around it simply by adding a `return_optypes` keyword argument to the function, and return `compiled_circuits` or `circuits` (the previous behaviour) only, unless it's specified as `True`, and update calling library code to use that.

(Putting this in now is how we change it safely - in the next Aer release you can make it raise a deprecation warning if the value isn't given, etc etc.)

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -331,53 +331,46 @@ def _format_results(output):
...
-                self,
-                parameter_binds=assemble_binds,
-                parameterizations=parameterizations)
         else:
-            qobj = assemble(
-                compile_circuit(
+            # Generate optypes for circuit
+            # Generate opsets of instructions
+            if not isinstance(circuits, list):
+                circuits = [circuits]
```
jakelishman, line 342: Same point as before about trying to allow arbitrary iterables - in particular, it's usually surprising if you can't use `tuple` as a collection for these purposes (_technically_ it's more efficient than `list` for a few things).

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -331,53 +331,46 @@ def _format_results(output):
...
+                    parameter_binds=assemble_binds,
+                    parameterizations=parameterizations)
+            else:
+                qobj = assemble(circuits, backend=self)
+
+            # Add optypes to qobj
+            # We convert to strings to avoid pybinding of types
+            qobj.config.optypes = [
+                [i.__name__ for i in optype] if optype else []
+                for optype in optypes]
```
jakelishman, line 373: Gross, but there's not really an alternative in this case, I think.  Would it be better to pass this as a `set` still - I don't know how that's handled in C++, though.

qiskit/providers/aer/backends/backend_utils.py
```diff
@@ -207,3 +207,14 @@ def func(data):
         return {key: func(val) for key, val in data.items()}
 
     return func(data)
+
+
+def circuit_optypes(circuit):
+    """Return set of all operation names in a circuit"""
```
jakelishman, line 213: They're types, not names - that's rather more specific. It might be good to mention in the docstring that it includes all subclasses as well, so we don't forget / other developers know that the function is safe with regards to subclasses.

qiskit/providers/aer/jobs/utils.py
```diff
@@ -126,12 +126,13 @@ def _check_custom_instruction(experiments, optypes=None):
     if optypes is not None:
         # Optypes store class names as strings
         return any(
-            "Save" in name or "Snapshot" in name
-            for optype in optypes for name in optype
+            {"SaveData", "Snapshot"}.intersection(optype)
+            for optype in optypes
         )
```
jakelishman, line 131: Do these come from `Qobj` in this case?  If so, then we really _need_ to make sure we're storing `optype` as a `set`, not a `list` to avoid the nasty iteration penalties here.

qiskit/providers/aer/backends/aer_compiler.py
chriseclectic: this function isn't and everything in this module isn't included in any init files or import paths, so its not really public

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: There isnt really a JSON equivalent for set, so its easier to just convert to a list.

qiskit/providers/aer/backends/aer_compiler.py
chriseclectic: Changed this to return just circuits in cases where optypes is None

qiskit/providers/aer/backends/aerbackend.py
chriseclectic: Well, it seemed to work changing to set, so i guess thats ok.

jakelishman (approved):

PR #1410, opened 2021-12-16
Speed up assembling of noise model for circuits without delays

### LINKED ISSUE #1407: Performance regression in noise model loading for Qiskit Aer 0.10.0

itoko:
### Summary
The commit fixes the remaining minor performance issue introduced at #1391 by skipping the run of noise pass manager for circuits without delays.

Fixes #1407 (performance regression) completely in addition to the major fix #1408.

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -21,7 +21,7 @@ import uuid
 import warnings
 from abc import ABC, abstractmethod
 
-from qiskit.circuit import QuantumCircuit, ParameterExpression
+from qiskit.circuit import QuantumCircuit, ParameterExpression, Delay
 from qiskit.compiler import assemble
 from qiskit.providers import BackendV1 as Backend
 from qiskit.providers.models import BackendStatus
@@ -379,8 +379,16 @@ class AerBackend(Backend, ABC):
         noise_model = run_options.get(
             'noise_model', getattr(self.options, 'noise_model', None))
 
-        # Add custom pass noise only to QuantumCircuit objects
-        if noise_model and all(isinstance(circ, QuantumCircuit) for circ in circuits):
+        # Add custom pass noise only to QuantumCircuit objects with no delays
+        def _contains_delay(circuit):
+            for op, _, _ in circuit.data:
+                if isinstance(op, Delay):
+                    return True
+            return False
+
+        if (noise_model and
+                all(isinstance(circ, QuantumCircuit) for circ in circuits) and
+                any(_contains_delay(circ) for circ in circuits)):
             npm = noise_model._pass_manager()
             if npm is not None:
                 circuits = npm.run(circuits)


itoko:
I've confirmed that this PR significantly reduces the assemble time of noisy simulation of circuits without delays.

Without this PR:
TRANSPILE TIME TAKEN: 66.19
ASSEMBLE TIME TAKEN: 5.05
RUN QOBJ TIME TAKEN: 6.21

With this PR:
TRANSPILE TIME TAKEN: 63.55
ASSEMBLE TIME TAKEN: 1.17
RUN QOBJ TIME TAKEN: 6.41

Using the following code
```
import time
import numpy as np
from qiskit_experiments.library import StandardRB
from qiskit.test.mock import FakeParis

backend = AerSimulator.from_backend(FakeParis())

start = time.time()
lengths = np.arange(1, 1000, 100)
num_samples = 10
seed = 1010
qubits = [0]

# Run an RB experiment on qubit 0
exp = StandardRB(qubits, lengths, num_samples=num_samples, seed=seed)

# Step by step
from qiskit import transpile

# Generate and transpile circuits
start = time.time()
circuits = transpile(exp.circuits(), backend)
stop = time.time()
print("TRANSPILE TIME TAKEN: {:.2f}".format(stop - start))

# Simulator Assemble
start = time.time()
qobj = backend._assemble(circuits)
stop = time.time()
print("ASSEMBLE TIME TAKEN: {:.2f}".format(stop - start))

# Simulator Run
start = time.time()
result = backend._run(qobj)
stop = time.time()
print("RUN QOBJ TIME TAKEN: {:.2f}".format(stop - start))
```

jakelishman (commented):
This looks fine to me, but #1409 also has the same optimisation, but with less data iteration due to a new data structure it introduces.  I think #1409 likely supersedes this.

chriseclectic (commented):
As @jakelishman said this is already covered by #1409 which includes this fix along with a fixes for other unnecessary iteration from qasm3 compiling, qerror_loc insertion, and qobj splitting.

itoko:
Close as it's already covered by #1409

PR #1412, opened 2021-12-16
Backports for Aer 0.10.1

chriseclectic:
### Summary

Aer 0.10.1 patch to fix performance regressions.

On-hold until @jakelishman tests on Terra CI


jakelishman (commented):
Terra's CI works with this (see Qiskit/qiskit-terra#7424), and locally it appears to have fixed the major performance regressions for me on Python 3.7.  There's a potential issue with `NoiseModel` accessing T2 times which don't exist on one fake backend (`OpenPulse2Q`), but to me that's kind of both an Aer bug and a Terra bug.  I've made a PR to add those properties in Terra (Qiskit/qiskit-terra#7425), but we may want to make the calls more resilient in Aer too.

We also need a release note prelude for 0.10.1.

jakelishman (approved):
Still works with Terra CI for me locally.

PR #1419, opened 2022-01-04
Fix simulation of for loops with no parameter

### LINKED ISSUE #1418: QASM 3 fails if a `for` loop doesn't have a parameter
### Informations

- **Qiskit Aer version**: 0.10.1
- **Python version**: any
- **Operating system**: any

The QASM 3 simulator can't cope with Terra `for` loops created that don't use the loop parameter.  If such a circuit is built with the builder interface the loop parameter is set to `None`, indicating that nothing should be bound into the loop body.

```python
import qiskit

qc = qiskit.QuantumCircuit(2)
with qc.for_loop(range(4)) as i:
    # Note that this doesn't use `i`, so the for loop
    # parameter is set to `None`.
    qc.h(0)
    qc.cx(0, 1)

AerSimulator(method="statevector").run(qc)
```

This gives a `CircuitError`:
```python
~/code/qiskit/aer/qiskit/providers/aer/backends/aer_compiler.py in _inline_for_loop_op(self, inst, parent, qargs, cargs)
    160                                                 continue_label,
    161                                                 break_label)
--> 162             inlined_body = inlined_body.bind_parameters({loop_parameter: index})
    163             parent.append(inlined_body, qargs, cargs)
    164             parent.append(AerMark(continue_label, inlined_body.num_qubits), qargs, [])

~/code/qiskit/terra/qiskit/circuit/quantumcircuit.py in bind_parameters(self, values)
   2545                     "Found ParameterExpression in values; use assign_parameters() instead."
   2546                 )
-> 2547             return self.assign_parameters(values)
   2548         else:
   2549             if any(isinstance(value, ParameterExpression) for value in values):

~/code/qiskit/terra/qiskit/circuit/quantumcircuit.py in assign_parameters(self, parameters, inplace)
   2500             ]
   2501             if len(params_not_in_circuit) > 0:
-> 2502                 raise CircuitError(
   2503                     "Cannot bind parameters ({}) not present in the circuit.".format(
   2504                         ", ".join(map(str, params_not_in_circuit))

CircuitError: 'Cannot bind parameters (None) not present in the circuit.'
```

The for loop should run, and just be equivalent to repeating the body `len(indices)` times.

jakelishman:
### Details and comments

Fix #1418.


chriseclectic (approved):

jakelishman:
Those windows test failures are weird.  Looking at the error tracebacks, they're not using the version of the code from this PR, but from `main` - they say line 162 of `aer_compiler.py` is `inlined_body = inlined_body.bind_parameters(...)`, but that was the state _before_ this PR.  Not sure how that happened.

chriseclectic (approved):

PR #1420, opened 2022-01-04
Fix bug in QuantumError.to_dict

### LINKED ISSUE #1415: Different pauli_error result from previous releases for two-qubit gate.
### Informations

- **Qiskit Aer version**: 0.10.1
- **Python version**: 3.9.9
- **Operating system**: MacOS

The tensor of two `pauli_error` gates after a two qubit gate is different from previous versions. Applying the gate
```python
import qiskit.providers.aer.noise as noise
error_1 = noise.pauli_error([('X', 1/3), ('Y', 1/3), ('Z', 1/3), ('I', 0)])
error_2 = error_1.tensor(error_1)
```
to a circuit with only a `cx` gate. (essentially identity circuit) results in the dictionary
```python
{'00': 557, '01' 467}
```

```python3
from qiskit import QuantumCircuit, execute, Aer
import qiskit.providers.aer.noise as noise
error_1 = noise.pauli_error([('X', 1/3), ('Y', 1/3), ('Z', 1/3), ('I', 0)])
err = error_1.tensor(error_1)

noise_model = noise.NoiseModel()
noise_model.add_all_qubit_quantum_error(err, ['cx'])

circ = QuantumCircuit(2, 2)
circ.cx(0, 1)
circ.measure([0, 1], [0, 1])

result = execute(circ, Aer.get_backend('qasm_simulator'), noise_model=noise_model).result()
counts = result.get_counts()
print(counts)
```

In previous versions, it results in something close to what I would expect given my understanding of the `pauli_error` which would be.
```python
counts = {'01': 2 / 9 * shots, '11': 4 / 9 * shots, '10': 2 / 9 * shots, '00': 1 / 9 * shots}
```

### Suggested solutions
Revert to the old behavior or explain in the documentation what a tensored pauli_error is supposed to do in practice.

chriseclectic:
### Summary

Fixes #1415 

Fixes a bug in `QuantumError.to_dict` where N-qubit circuit instructions where the assembled instruction always applied to qubits `[0, ..., N-1]` rather than the instruction qubits. This bug was also affected device and fake backend


qiskit/providers/aer/noise/errors/quantum_error.py
```diff
@@ -452,12 +452,21 @@ def error_term(self, position):
 
     def to_dict(self):
         """Return the current error as a dictionary."""
+        # Assemble noise circuits
+        instructions = []
+        for circ in self._circs:
+            circ_inst = []
+            for inst, qregs, _ in circ.data:
+                qobj_inst = inst.assemble()
+                qobj_inst.qubits = [circ.find_bit(q).index for q in qregs]
```
jakelishman, line 461: The name `qregs` is misleading here - they're not registers, they're qubits.  I think perhaps you meant `qargs`?

test/terra/backends/aer_simulator/test_noise.py
```diff
@@ -223,18 +223,26 @@ def test_kraus_circuit_noise(self, method, device):
         """Test simulation with Kraus quantum errors in circuit."""
         backend = self.backend(method=method, device=device)
         shots = 1000
-        error1 = noise.amplitude_damping_error(0.05)
-        error2 = error1.tensor(error1)
-
+        error0 = noise.amplitude_damping_error(0.05)
+        error1 = noise.amplitude_damping_error(0.15)
+        error01 = error1.tensor(error0)
```
jakelishman, line 228: I think it might be good to test an error with the qargs the other way round in the same circuit as well?

releasenotes/notes/fix_qerror_to_dict-13a7683ac4adddd4.yaml
```diff
@@ -0,0 +1,10 @@
+---
+fixes:
+  - |
+    Fixes a bug in :meth:`.QuantumError.to_dict` where N-qubit circuit
+    instructions where the assembled instruction always applied to
+    qubits ``[0, ..., N-1]`` rather than the instruction qubits. This
+    bug was also affected device and fake backend noise models.
```
jakelishman, line 7: "was also affected" should presumably be either "was also affecting" or "also affected"?

test/terra/backends/aer_simulator/test_noise.py
chriseclectic: I think its fine one way, as long as the error rates are asymetric. But I will swap the gates to be applied to reverse order for (ie [1,0] instead of [0, 1])

jakelishman: The point was more to ensure that we're _certainly_ passing in one error that acts on `[0, 1]` and one on `[1, 0]` to make sure that the remapping works.

jakelishman (approved):

PR #1422, opened 2022-01-07
Fix diagonal fusion in parallel transpilations

hhorii:
### Summary

fusion to generate a diagonal matrix has a bug when transpilation is in parallel

### Details and comments

If a number of gates exceeds `fusion_parallelization_threshold`, fusion transpiler divides the sequence of gates into sub-sequences and multiple threads run transpilation for each sub-sequence. A thread can fuse gates only in a sub-sequence. However, current diagonal-fusion searches gates across sub-sequences and can produce wrong results. This PR prevent from a thread not to fuse gates across sub-sequences.

### DIFF
--- a/src/transpile/fusion.hpp
+++ b/src/transpile/fusion.hpp
@@ -479,7 +479,7 @@ public:
 private:
   bool is_diagonal_op(const op_t& op) const;
 
-  int get_next_diagonal_end(const oplist_t& ops, const int from, std::set<uint_t>& fusing_qubits) const;
+  int get_next_diagonal_end(const oplist_t& ops, const int from, const int end, std::set<uint_t>& fusing_qubits) const;
 
   const std::shared_ptr<FusionMethod> method_;
   uint_t min_qubit = 3;
@@ -511,8 +511,13 @@ bool DiagonalFusion::is_diagonal_op(const op_t& op) const {
   return false;
 }
 
+// Returns an index in `ops` or `-1`.
+// If gates from `from` to the returned index are fused to a gate, the fused gate is a diagonal gate.
+// The returned index is equal or more than `from` and is lower than `end`.
+// If -1 is returned, no pattern to generate a diagonal gate is identified from `from`.
 int DiagonalFusion::get_next_diagonal_end(const oplist_t& ops,
                                           const int from,
+                                          const int end,
                                           std::set<uint_t>& fusing_qubits) const {
 
   if (is_diagonal_op(ops[from])) {
@@ -537,11 +542,11 @@ int DiagonalFusion::get_next_diagonal_end(const oplist_t& ops,
   //        ■ [from,pos]
 
   // find first cx list
-  for (; pos < ops.size(); ++pos)
+  for (; pos < end; ++pos)
     if (ops[from].type != Operations::OpType::gate || ops[pos].name != "cx")
       break;
 
-  if (pos == from || pos == ops.size())
+  if (pos == from || pos == end)
     return -1;
 
   auto cx_end = pos - 1;
@@ -558,7 +563,7 @@ int DiagonalFusion::get_next_diagonal_end(const oplist_t& ops,
 
   bool found = false;
   // find diagonals
-  for (; pos < ops.size(); ++pos)
+  for (; pos < end; ++pos)
     if (is_diagonal_op(ops[pos]))
       found = true;
     else
@@ -567,7 +572,7 @@ int DiagonalFusion::get_next_diagonal_end(const oplist_t& ops,
   if (!found)
     return -1;
 
-  if (pos == ops.size())
+  if (pos == end)
     return -1;
 
   auto u1_end = pos;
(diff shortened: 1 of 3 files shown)


jakelishman:
I think something's gone slightly wrong in the history here - looks like there's a cherry-picked commit rather than a merge with the base.

hhorii:
@jakelishman Thanks. I will rebuild a history based on the current main with cherry-picking 7cbcb1e.

src/releasenotes/notes/fix_parallel_diag_fusion-a7f914b3a9f491f7.yaml
```diff
@@ -0,0 +1,10 @@
+---
+fixes:
+  - |
+    If a number of gates exceeds ``fusion_parallelization_threshold``,
+    fusion transpiler divides the sequence of gates into sub-sequences
+    and multiple threads run transpilation for each sub-sequence.
+    A thread can fuse gates only in a sub-sequence. However, current
+    diagonal-fusion searches gates across sub-sequences and can produce
+    wrong results. This commit prevents from a thread not to fuse gates
+    across sub-sequences.
```
jakelishman, line 10: The release note should be an explanation to the user, rather than to other developers - I think this is the text from the PR/commit body, but for users, we want to say something (very roughly) along the lines of

> Multi-threaded transpilations with a ``fusion_parallelization_threshold`` will now work correctly if the circuit can be fused with the diagonal strategy.  Previously, different threads would occasionally fuse the same element into multiple blocks, causing incorrect results.

(I don't 100% understand what `DiagonalFusion` is doing so feel free to correct it, but we don't need to be _super_ detailed about what caused a bug if the details weren't visible to the user, other than just giving incorrect results.)

src/transpile/fusion.hpp
```diff
@@ -628,22 +629,22 @@ bool DiagonalFusion::aggregate_operations(oplist_t& ops,
...
+    // find instructions to generate the next diagonal gates
     auto next_diagonal_start = next_diagonal_end + 1;
 
     while (true) {
-      auto nde = get_next_diagonal_end(ops, next_diagonal_start, checking_qubits_set);
-      if (nde < 0)
-        break;
-      if (checking_qubits_set.size() > max_fused_qubits)
+      auto nde = get_next_diagonal_end(ops, next_diagonal_start, fusion_end, checking_qubits_set);
+      if (nde < 0 || nde >= fusion_end || checking_qubits_set.size() > max_fused_qubits)
```
jakelishman, line 647: I think in theory `nde` can't ever be greater than `fusion_end`, right?  In `get_next_diagonal_end` it returns `-1` if it reaches the end.  It might be good to write that into the documentation/contract of `get_next_diagonal_end`, and simplify the extraneous condition here.

jakelishman (commented):
In principle this looks like a good, straightforward fix to me, and the test is good.  I only had two comments - one about the release note, the other about an extraneous condition. The condition feels like it's trying to be slightly overly defensive against something that we should be able to guarantee earlier - it's not really a terrible thing to have it, but if it triggers, I think it would indicate a bug earlier in the logic.

test/terra/backends/aer_simulator/test_fusion.py
```diff
@@ -511,3 +512,45 @@ def test_fusion_diagonal(self):
...
+        for q in range(num_of_qubits):
+            circuit.h(q)
+        
+        for qubit in range(num_of_qubits):
+            for q in range(size):
+                ctrl = (qubit + q) % num_of_qubits
+                tgt = (ctrl + 1) % num_of_qubits
+                circuit.cx(ctrl, tgt)
+        
+            circuit.p(np.random.random(), tgt)
```
jakelishman, line 534: It'd be best to make this a fixed value, rather than making the test stochastic - if it were to ever break, it'd be impossible to reproduce.

test/terra/backends/aer_simulator/test_fusion.py
```diff
@@ -511,3 +512,45 @@ def test_fusion_diagonal(self):
...
+                circuit.cx(ctrl, tgt)
+                
+        circuit.save_statevector()
+        
+        thread = 8
+        result = backend.run(circuit,
+                               fusion_threshold=1,
+                               **{'fusion_enable.diagonal': True,
+                                  'fusion_enable.cost_based': False,
+                                  'fusion_enable.n_qubits': False},
```
jakelishman, line 548: I had no idea you could pass invalid identifier names like this!  That's cool, although I'm not certain it's the best user interface...  Still, that's beyond the scope of this PR.

test/terra/backends/aer_simulator/test_fusion.py
```diff
@@ -511,3 +512,45 @@ def test_fusion_diagonal(self):
...
+                               fusion_parallelization_threshold=3,
+                               max_parallel_threads=thread).result()
+        actual = result.data(0)['statevector']
```
jakelishman, line 551: Largely irrelevant, but isn't this just `result.get_statevector(0)`?

jakelishman (commented):
It looks much cleaner with the "-1 means error" behaviour more encoded in the contract, thanks.  I had a couple of minor things about the test, but nothing much.

test/terra/backends/aer_simulator/test_fusion.py
hhorii: 'fusion_enable.*' options are just for debug use. If users require to control fusion strategies, we should rename option names :-)

test/terra/backends/aer_simulator/test_fusion.py
hhorii: `result.get_statevector(0)` is the same with `result.data(0)['statevector']` and looks simpler than it.

jakelishman (approved):
Looks good to me now.

mtreinish (approved):

PR #1423, opened 2022-01-10
Fix all compiler warnings and enable -Werror

jakelishman:
### Summary

This fixes all warnings that are currently enabled in our CMake configuration, and then turns warnings into errors (if not on Windows).

The most dangerous behaviour fixed here was likely in some loop variables that were started uninitialised, then used directly as indexes into arrays.  If we were unlucky, this could have been clobbering random data, and if we were a little luckier, we'd have had some segfaults.

### Details and comments

There are several more categories of warnings that we may want to also suppress, mostly contained within `-Wextra` on `clang` and `gcc`.  A lot of these are `sign-compare` and `unused-parameter` though, which will require rather more fiddling through every part of the codebase, but will make us a bit more resilient overall.

There are a few separate commits in this PR, which explain what they were doing individually.


CMakeLists.txt
```diff
@@ -127,6 +128,7 @@ if(NOT MSVC)
 	enable_cxx_compiler_flag_if_supported("-Wredundant-decls")
 	enable_cxx_compiler_flag_if_supported("-Wshadow")
 	enable_cxx_compiler_flag_if_supported("-Woverloaded-virtual")
+	enable_cxx_compiler_flag_if_supported("-Werror")
```
garrison, line 131: As a developer, I am very supportive of building with `-Werror` locally and on CI, and I applaud any effort to make this possible.  But I'd caution against enabling `-Werror` by default for everyone who builds Aer.  While the code might build without warnings on specific versions of specific compiler(s), it is still quite possible for a different compiler (or even a newer release of the same compiler, e.g. gcc) to trigger a new warning.  Is it really better for users to error out rather than to continue with the build when this happens?  In my experience, supporting users through the resulting (and ongoing) fallout can be a major headache.

jakelishman:
Closing because this is very out-of-date, since I never had time to return to this.

PR #1428, opened 2022-01-17
Run truncation after resolving parameters

### LINKED ISSUE #1427: `instruction param position out of range` when truncation is processed
### Informations

- **Qiskit Aer version**: 0.34.1
- **Python version**: any
- **Operating system**: 0.10.2

The current parameterized qobj is not processed correctly when truncation is enabled.

Once truncation is processed, indices of parameters must be changed to skip truncated instructions.
However, the current implementation uses specified indices and then throw exception `Invalid parameterized qobj: instruction param position out of range`.

```
import numpy as np
from qiskit import transpile, QuantumCircuit
from qiskit.circuit.library import RealAmplitudes

simulator = AerSimulator()

circuit = QuantumCircuit(10, 10)
circuit = circuit.compose(RealAmplitudes(10), range(10))
circuit = transpile(circuit, simulator)
for q in range(5):
    circuit.measure(q, q)

num_of_params = 10
param_map = {}
for param in circuit.parameters:
    param_values = [ np.random.random() for _ in range(num_of_params) ]
    param_map[param] = param_values

simulator.run(circuit, parameter_binds=[param_map]).result().success
```

No runtime exception when the above script runs.

### Suggested solutions

Resolve parameters after truncation or resolve parameters with consideration of truncation results.

hhorii:
### Summary

Resolve #1427.

### Details and comments

Perform truncation after resolving parameters.
This solution is not optimal but acceptable. Truncation for all sets of parameters results the same. Ideally, one truncation for all the sets of parameters.
However, truncation can change indices of instructions in parameter map and mapping instruction indices of before- and after-truncation needs codes.


mtreinish (commented):
This should have a test to validate this works after the fix.

test/terra/backends/aer_simulator/test_truncate.py
```diff
@@ -135,3 +137,32 @@ def test_delay_measure(self):
...
+        circuit = QuantumCircuit(10, 10)
+        circuit = circuit.compose(RealAmplitudes(5), range(5))
+        circuit = transpile(circuit, backend)
+        for q in range(5):
+            circuit.measure(q, q)
+        
+        num_of_param_sets = 3
+        param_map = {}
+        for param in circuit.parameters:
+            param_values = [ np.random.random() for _ in range(num_of_param_sets) ]
```
jakelishman, line 153: Similar to the other PR, it'd be good to fix the seed for this.

test/terra/backends/aer_simulator/test_truncate.py
```diff
@@ -135,3 +137,32 @@ def test_delay_measure(self):
...
+        result = backend.run(circuit, parameter_binds=[param_map]).result()
+        self.assertSuccess(result)
+
+        result_without_parameters = backend.run(resolved_circuits).result()
+        self.assertSuccess(result_without_parameters)
+
+        for actual_result, expected_result in zip(result.results, result_without_parameters.results):
+            metadata = actual_result.metadata
+            self.assertEqual(metadata["active_input_qubits"], [q for q in range(5)])
+            self.assertDictAlmostEqual(actual_result.data.counts, expected_result.data.counts, )
```
jakelishman, line 168: Would it be possible/better to run a statevector simulator, to reduce the possible variation?  I'm slightly concerned that comparing two counts distributions is naturally going to be a bit flaky due to the randomness in the simulator (or is the simulator seeded before each call?), whereas comparing the statevectors is only subject to floating-point tolerance issues.

test/terra/backends/aer_simulator/test_truncate.py
```diff
@@ -135,3 +137,32 @@ def test_delay_measure(self):
         metadata = result.results[0].metadata
         self.assertIn('measure_sampling', metadata)
         self.assertTrue(metadata['measure_sampling'])
+
+    def test_parameterized_circuits(self):
+        """Test truncation for parameterized circuits"""
+        backend = self.backend()
+        circuit = QuantumCircuit(10, 10)
+        circuit = circuit.compose(RealAmplitudes(5), range(5))
+        circuit = transpile(circuit, backend)
```
jakelishman, line 146: If there's non-trivial transpilation being done here, it wouldn't hurt to seed the transpiler as well, but in practice it's _probably_ ok - I think all the random elements are moot for Aer because it's got all-to-all connectivity.

test/terra/backends/aer_simulator/test_truncate.py
hhorii: I have moved this test to https://github.com/Qiskit/qiskit-aer/blob/47f2765b8c117e8f29792c45f50f64b80b32136f/test/terra/backends/test_parameterized_qobj.py, and then used statevectors for assertions.

test/terra/backends/aer_simulator/test_truncate.py
hhorii: I changed this [test circuit](https://github.com/Qiskit/qiskit-aer/blob/47f2765b8c117e8f29792c45f50f64b80b32136f/test/terra/backends/test_parameterized_qobj.py#L303-L312) for lightweight execution.

jakelishman (approved):

Issue #1435, opened 2022-01-24
Simulation fails when circuit contains metadata


kevinsung:
### Informations

- **Qiskit Aer version**: 0.10.2
- **Python version**: 3.8
- **Operating system**: Arch Linux

### What is the current behavior?
Simulation fails when circuit contains metadata.

### Steps to reproduce the problem
Input:
```python
from qiskit import Aer, QuantumCircuit

backend = Aer.get_backend("statevector_simulator")
circuit = QuantumCircuit(1, metadata = {1: object})
backend.run(circuit)
```
Output:
```
Simulation failed and returned the following error message:
ERROR: Failed to load qobj: to_json not implemented for this type of object: <class 'type'>
```
Input:
```python
from qiskit import Aer, QuantumCircuit

backend = Aer.get_backend("statevector_simulator")
circuit = QuantumCircuit(1, metadata = {object: 1})
backend.run(circuit)
```
Output:
```
Simulation failed and returned the following error message:
ERROR: Failed to load qobj: Unable to cast Python instance to C++ type (compile in debug mode for details)
```

### What is the expected behavior?
Simulation should succeed. The metadata should be handled according to https://qiskit.org/documentation/stubs/qiskit.circuit.QuantumCircuit.html#qiskit.circuit.QuantumCircuit, which says
> metadata (dict) – Arbitrary key value metadata to associate with the circuit. This gets stored as free-form data in a dict in the metadata attribute. It will not be directly used in the circuit.

### Suggested solutions
None (yet).

chriseclectic:
@hhorii For this one it might be better to not attempt to serialize the header information at all, and instead store it on the python side and add it back to the results after execution.

PR #1436, opened 2022-01-25
Fix handling of circuit metadata

hhorii:
### Summary

#1435 

### Details and comments

`to_json` is called for each circuit header.
A header may include metadata and metadata can be a python object.
This PR changes serialization of circuit headers to use `py::handle` without serializing to a json.

Another solution can be in `_assemble()` to remove all unnecessary metadata from input.


jakelishman:
I'm not very knowledgeable about the C++/Python handoff with pybind11, so a slight question: what we need out of the header seems very simple - can we just leave the metadata in Python space, and reattach it to the new output circuit (if there is one)?  It feels like we could save time by just not serialising it, if we don't use it.

hhorii:
> what we need out of the header seems very simple - can we just leave the metadata in Python space, and reattach it to the new output circuit (if there is one)?

I believe that `py::handle` can be passed through C++ space. I will try it.
The current implementation of [`ExperimentResult`](https://github.com/Qiskit/qiskit-aer/blob/f778f4b53b4223c86f5e41b50794d176208c76c6/src/framework/results/experiment_result.hpp#L46) uses `json_t`. To pass `py::handle` instead of `json_t`, as other codes, we need to make [`ExperimentResult`] class a template class to handle both of `json_t` and `py::handle`, and modify [`to_python(ExperimentResult)`](https://github.com/Qiskit/qiskit-aer/blob/f778f4b53b4223c86f5e41b50794d176208c76c6/src/framework/results/pybind_result.hpp#L72-L74) to handle the both types.

kevinsung:
Please add a test that ensures the repro code from https://github.com/Qiskit/qiskit-aer/issues/1435 works.

hhorii:
> Please add a test that ensures the repro code from #1435 works.

I added [test cases](https://github.com/Qiskit/qiskit-aer/pull/1436/files#diff-d7e527587e86f4149f9adaace70916785fe803bed9aca34a54a1f1e2c44c4758R25) to check metadata is copied correctly.

hhorii:
TODO:
- [ ] Use `pop` instead of `del`
- [ ] Store an index of original circuit for parameterized circuit. 
- [ ] Change `_assemble` instead of `_run`
- [ ] Do not change codes to copy headers in controller

hhorii:
Aer does not use Metadata to simulate circuits. Therefore, I believe, we can reduce metadata from qobj before simulation.

Previously, `metadata` of `QuantumCircuit` is copied to `QobjExperimentHeader` in a qobj. This`QobjExperimentHeader` is returned as `header` of `ExperimentResult`. Whit this change, `header` of `ExperimentResult` does not keep `metadata` of `QuantumCircuit` because it is removed before generating qobj.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -352,6 +352,12 @@ def _assemble(self, circuits, parameter_binds=None, **run_options):
             circuits, optypes, run_options = self._assemble_noise_model(
                 circuits, optypes, **run_options)
 
+            # Remove metadata from circuits
+            metadata_list = [circuit.metadata for circuit in circuits]
+            for circuit in circuits:
+                if circuit.metadata:
+                    circuit.metadata.clear()
```
kevinsung, line 359: The way this is written, `metadata_list` only contains references to the metadata dictionaries, so this line actually deletes the metadata. Instead of clearing the dictionaries, set them to `None`.

test/terra/backends/aer_simulator/test_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+        self.assertEqual(len(result.results), 6)
+        self.assertIsNone(result.results[0].header.metadata)
+        self.assertIsNone(result.results[1].header.metadata)
+        self.assertIsNone(result.results[2].header.metadata)
+        self.assertIsNone(result.results[3].header.metadata)
+        self.assertIsNone(result.results[4].header.metadata)
+        self.assertIsNone(result.results[5].header.metadata)
+        self.assertEqual(circuit0.metadata, metadata0)
+        self.assertEqual(circuit1.metadata, metadata1)
+        self.assertEqual(circuit2.metadata, metadata2)
```
kevinsung, line 103: In order to catch the bug I pointed out where the metadata is inadvertently modified, these assertions should explicitly construct the expected dictionaries.

kevinsung (changes requested):

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -352,6 +352,12 @@ def _assemble(self, circuits, parameter_binds=None, **run_options):
             circuits, optypes, run_options = self._assemble_noise_model(
                 circuits, optypes, **run_options)
 
+            # Remove metadata from circuits
+            metadata_list = [circuit.metadata for circuit in circuits]
+            for circuit in circuits:
+                if circuit.metadata:
```
kevinsung, line 358: This `if` statement is unnecessary.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -366,6 +372,11 @@ def _assemble(self, circuits, parameter_binds=None, **run_options):
             else:
                 qobj = assemble(circuits, backend=self)
 
+            # Recover metadata of circuits
+            for circuit, metadata in zip(circuits, metadata_list):
+                if metadata:
```
kevinsung, line 377: This `if` statement and the one I commented on above can both be removed.

kevinsung (changes requested):

kevinsung (approved):
LGTM. @chriseclectic or someone else with write access please take a look.

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -366,6 +371,10 @@ def _assemble(self, circuits, parameter_binds=None, **run_options):
             else:
                 qobj = assemble(circuits, backend=self)
 
+            # Recover metadata of circuits
+            for circuit, metadata in zip(circuits, metadata_list):
+                circuit.metadata = metadata
```
mtreinish, line 376: So I think the issue with doing this is that the metadata doesn't get set in the output result. The idea behind metadata is that it provide more rich data around tracking the purpose of a circuit in a larger application. Prior to the introduction of `circuit.metadata` the only available information was the `circuit.name` which let to people trying to encode this all in a string. 

While metadata shouldn't influence the execution of the circuit on a backend it is important that users can correlate the results to the metadata easily (similar to how you can find a circuit name associated with a result object). Is there a way we can preserve setting the metadata in the result like it was before, but skip the json serialization in `assemble()`?

hhorii: I think there are two ways:

Approach-1: take metadata from `QuantumCircuit` before `assemble()` and store the metadata to a result before `AerJob.result()` in a caller thread
Approach-2: take metadata from `QObj` before `_execute()` and store the metadata to a result after `_execute()` in pooled threads

In approach-1, the taken metadata needs to be kept somewhere. I believe Approach-2 looks better and implemented it in fd65ac7.

releasenotes/notes/remove_circuit_metadata_from_qobj-324e7ea9b369ee67.yaml
```diff
@@ -0,0 +1,9 @@
+---
+fixes:
+  - |
+    Reduce metadata of circuits in Qobj before simulation and set the metadata
+    to ```header``` of ```ExperimentResult``` in python. 
+    Previously, metadata is serialized with JSON format and copied to
+    the result even though metadata is not used in simulation. If metadata
+    can not be serizalized, simulation was failed. This fix works around
+    this issue.
```
mtreinish, line 9:
```suggestion
    Fixed a potential issue with running simulations on circuits that have the
    :attr:`.QuantumCircuit.metadata` attribute set. The :attr:`~.QuantumCircuit.metadata`
    attribute can be any python dictionary and previously qiskit-aer would attempt to
    JSON serialize the contents of the attribute to process it with the rest of the rest
    of the circuit input, even if the contents were not JSON serializable. This no longer
    occurs as the :attr:`.QuantumCircuit.metadata`  attribute is not used to run the
    simulation so now the contents are no serialized and instead are directly attached
    to the :class:`qiskit.result.Result` object without attempting to JSON serialize
    the contents.
    Fixed `#1435 <https://github.com/Qiskit/qiskit-aer/issues/1435>`__
```

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -288,9 +288,30 @@ def _run(self, qobj, job_id='', format_result=True):
...
+        # Take metadata from headers of experiments to work around JSON serialization error
+        metadata_list = []
+        metadata_index = 0
+        for expr in qobj.experiments:
+            if hasattr(expr.header, "metadata"):
+                metadata_copy = expr.header.metadata.copy()
+                metadata_list.append(metadata_copy)
+                expr.header.metadata.clear()
+                if "id" in metadata_copy:
+                    expr.header.metadata["id"] = metadata_copy["id"]
```
mtreinish, line 300: What is this `id` for/from? and why are we preserving it in the qobj?

mtreinish (commented):
Overall LGTM, just one question inline and a suggestion on the release note. Thanks for updating this

qiskit/providers/aer/backends/aerbackend.py
hhorii: This is my mistake. I should reduce these codes.

Qiskit-Aer internally use this metadata for DASK execution. Even though metadata is cleared, this `metadata['id']` was necessary for merging results and to be kept. However, the latest commit of this PR keeps whole metadata, then we do not need to specialize it.

I reduced the codes in 461077b. Thanks.

hhorii:
I realized that `id` in metadata is necessary for clustering. QObj is divided to sub-QObjs (first level) and each sub-QObj can be divided to subsub-QObj (second level). If `id` is removed in the first level, there is no way to merge results in second level.

mtreinish (approved):

PR #1437, opened 2022-01-25
Fix time_taken value in cluster backend

hitomitak:
### Summary
`time_taken` value in Result is not updated in the cluster backend environment. 

### Details and comments
Current code stores the execution time of each experiment into `time_taken` in `Result` when the executor is DASK. This fix updates this value to the total time.


qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -335,9 +335,9 @@ def _combine_results(self,
 
         if self._start_time:
             _time_taken = self._end_time - self._start_time
-            combined_result.time_taken = _time_taken.total_seconds()
+            combined_result._metadata["time_taken"] = _time_taken.total_seconds()
         else:
-            combined_result.time_taken = 0
+            combined_result._metadata["time_taken"] = 0
```
jakelishman, line 340: I'm not 100% sure how this relates to the body of the PR - it doesn't seem to change anything about the value being set, only sets it in the metadata rather than attaching it as an instance variable?

The `Result` interface isn't really ideal at all, but since it doesn't give us a public way to modify metadata, it would probably be better if this method built up `Result` as a dictionary, and then called `Result.from_dict` to build it properly the first time, rather than mutating it after creation.  That would let us do away with the private-member access, we should avoid.

jakelishman (commented):
Is there any test we can add for this?  We probably just need to test that the time is being set to _anything_ in the result.

qiskit/providers/aer/jobs/aerjobset.py
```diff
@@ -335,9 +335,9 @@ def _combine_results(self,
 
         if self._start_time:
             _time_taken = self._end_time - self._start_time
-            combined_result.time_taken = _time_taken.total_seconds()
```
hhorii, line 338: I guess that `time_taken` should have  `_time_taken.total_seconds()`, too, as `AerJob` has.

qiskit/providers/aer/jobs/aerjobset.py
hitomitak: Thank you for the review. I added the test if `time_taken` value is set in `Result`

The result data users receive is generated from _metadata value in `Result`. So if `combined_result.time_taken` is updated, the result data does not change. 

Chris said that the process of `from_dict` and `to_dict` functions are heavy so it is avoid to use them. Do you think I should use them ?

qiskit/providers/aer/jobs/aerjobset.py
hitomitak: I added the line to set the time to `combined_result.time_taken `.

qiskit/providers/aer/jobs/aerjobset.py
mtreinish: I would avoid using the `to_dict` and `from_dict` as standard practice. They're mostly for backwards compatibility with how things used to be constructed (where the results object was more of an object wrapping the JSON api response body from the ibm devices) and when interacting they just add overhead both in memory and run time to construct intermediate objects when it's normally more efficient to just operate on the object directly. If we're missing a way for backend's to update result metadata in a result that's a terra bug which should be fixed in terra. So I think doing it this way is fine for now as long as we have a tracking issue in terra.

That being said I agree with @jakelishman and I'm not clear on what this changes functionally. Looking at the result class setting `time_taken` as an instance variable is basically identical to setting it in `_metadata` except that if it's a field in the `_metadata` dict it's included in the output of `to_dict` and it's in the `repr` for the object. Is this just about making sure it's included in `to_dict`'s output or is there something else going on?

hitomitak: Thank you for the review. I updated ` _combine_results` function to generate `Result` from JSON data.

jakelishman (approved):
This PR is nice in that makes `_combine_results` only use public interfaces.  I agree with Matthew and Chris that using `from_dict` and `to_dict` isn't right _in general_, but that's something that we should solve on Terra, not Aer .  There _is_ a public-interface-only way of solving it (i.e. like this PR), so rather than working around the issue in downstream libraries, we should use the public interface, and fix the issue at the source (Terra).

jakelishman:
I was in the process of merging this, but then realised I'd misread your answer to Matthew's question - is this just about making sure the output of `to_dict` later is correct?

hitomitak:
Yes. This PR updates the value of `to_dict` output.

PR #1438, opened 2022-01-25
fix performance regression produced by #1392

hhorii:
### Summary

Fix performance regression in diagonal gates, which were produced by #1392

### Details and comments

#1392 produced performance regression because each thread allocates large temporal memory conservatively.
This PR fixes this regression.


doichanj (dismissed):
I think performance regression caused by #1392 was memory allocation of whole temporary buffer for all the threads that causes miss align for half of threads (32B for each thread)
This fix allocates aligned temporary for each thread and this is reasonable.

jakelishman:
How does this handle the initial problem that #1384 was trying to solve?  This seems to revert #1392 entirely, but that was originally trying to fix a bug, right?

hhorii:
I have tested the following code that is failed with qiskit-aer 0.9.0. This test was added in `test_fusion.py`.

```
from qiskit import transpile
from qiskit.circuit.library import QFT

circ = QFT(5)
circ.measure_all()

simulator = AerSimulator()

result = simulator.run([transpile(circ, simulator, basis_gates=['u3', 'cx'])] * 4,
                       **{'_parallel_experiments': 2,
                          '_parallel_state_update': 2,
                          'fusion_threshold': 1,
                          'statevector_parallel_threshold': 1,
                          'shots': 10}).result()
```

In Qiskit 0.9.0, memory allocation and free to a thread-local pointer were in different `#pragma omp` blocks due to use framework of `omp` of Aer. This implementation was wrong because there is no guarantee to use the same sets of threads in two blocks. This PR change to use single `#pragma omp` to allocate and free memory (and not use the framework).

chriseclectic (approved):

PR #1441, opened 2022-01-28
Support Visual Studio 17 2022

hhorii:
### Summary

Cmake for conan should support Visual Studio 2022.

### Details and comments

Current conan.cmake does not support Visual Studio 2022.
This PR makes support it.


hhorii:
The current conan center does not have binaries of fmt and spdlog and cmake fails to buid them. It may be too early to use VS 2022. #1442 should be a better approach than this.

PR #1446, opened 2022-02-04
Use unitary equivalence instead of circuit equality in noise tests

hitomitak:
### Summary
I deleted I gate from the test to reflect this [commit](https://github.com/Qiskit/qiskit-terra/commit/9e09abe1c178526d609b198f500512ebab895ddd).


jakelishman (commented):
This seems right for the immediate test, but the change in Terra was because of a misalignment of expectations around `IGate` - for historical IBM Q reasons, `IGate` represents a delay rather than an identity, so it was wrong for `opflow` to emit them.

I'm slightly concerned in Aer that the whole `QuantumError` noise model seems to want `IGate`s as input, and `depolarising_error` emits `IGate`.  I suppose in the context of creating noise models, it's possibly ok?

Here's the docstring about it:
https://github.com/Qiskit/qiskit-terra/blob/226f4646140bd26e0ea70fdfd018a8fd7f156fa8/qiskit/circuit/library/standard_gates/i.py#L20-L24

mtreinish:
We can confirm with @itoko on this but I think that makes sense in this context, at least while the `IGate` is defined as a single qubit gate delay (which I think is just confusing). It might make sense if we could have things work natively in delay. Eventually we're going to have to do it once we're ready for Qiskit/qiskit-terra#7146 (I'm still not clear how it really breaks that assumption because it translating between `I` and `U(0,0,0)` will keep the single gate delay properties) and this will come up again.

hhorii:
I believe that, in this case, we can just compare unitary matrices of target and expected circuits, which can include `IGate` or not.

test/terra/noise/test_standard_errors.py
```diff
@@ -520,7 +520,7 @@ def test_pauli_error_2q_gate_from_string_1qonly(self):
         for i in range(actual.size):
             circ, prob = actual.error_term(i)
             expected_circ, expected_prob = expected.error_term(i)
-            self.assertEqual(circ, expected_circ)
+            self.assertEqual(qi.Operator(circ), qi.Operator(expected_circ))
```
mtreinish, line 523: We can do:

```suggestion
            self.assertTrue(qi.Operator(circ).equiv(qi.Operator(expected_circ)))
```

if we want to ignore phase. But I don't think that matters here

mtreinish (dismissed):
LGTM

jakelishman (approved):

PR #1064, opened 2020-12-02
Break main GHA into multiple workflows

chriseclectic:
Update cibuildwheel version

### Summary

Breaks main GHA workflow into separate workflows for:
* Build (standalone and wheel)
* Tests Linux
* Tests MacOS
* Tests Windows
* Doc and Tutorials

### Details and comments

There is still a problem with the wheel build from cibuildwheel that needs to be fixed (https://github.com/pypa/manylinux/issues/836)

PR #1448, opened 2022-02-04
Simplify Aer CI configuration

mtreinish:
### Summary

Currently the aer ci configuration is too onerous and prevents us from
making forward progress on development because we're constantly having
to adapt to upstream changes or system errors in CI. This has been
blocking us from potentially getting the 0.10.3 bugfix release prepared.
To improve the situation and make ci maintainable this commit makes a
few key changes to the CI system. The first is to stop installing
qiskit-terra from source. While the goal when this was done was to catch
failures from terra sooner in practice this just means we're playing
catch up constantly. Aer has a much slower patch rate compare to
qiskit-terra and in practice all doing this means is we have to fix ci
anytime someone pushes a new PR because things have bit rotted. These
breakages aren't aer's fault (at least normally) and are caused by
breaking api changes or bugs introduced during the development of terra,
which only guarantees api stability on releases. So we should treat
terra just like any other upstream dependency and install it from it's
release. Hopefully when we have proper integration testing setup in
terra the spurious failures will decrease as well. The second key change
made here is to decrease the duplication between jobs. In aer
historically we've tested every permutation of install method on
platform and python version. This was in response to early versions of
aer which were packaged incorrectly and quite poorly and had issues
installing everywhere without issue. However, this has led to a
proliferation of jobs that are testing the same thing in slightly
different ways. Additionally, because of all these jobs and limitations
in github actions around retying failed jobs things were split into 3
workflows to enable retrying subsets of jobs by platform. This however
further duplicated the jobs so that we had fast failures on each
workflow. This commit undoes this, it reduces the 3 test workflows back
into a single shared one. In that workflow we only have 1 lint job, one
set of sdist jobs, 1 json input test, and then 1 platform specific set
of unittest runs. This should hopefully decrease the wait time for CI
and also make things easier to manage moving forward.

### Details and comments


.github/workflows/tests.yml
```diff
@@ -177,3 +172,99 @@ jobs:
...
+        shell: bash
+      - name: Run Tests
+        env:
+          QISKIT_SUPPRESS_PACKAGING_WARNINGS: Y
+          LANG: 'C.UTF-8'
+          PYTHONIOENCODING: 'utf-8:backslashreplace'
+        run: |
+          set -e
+          chcp.com 65001
+          pip check
```
jakelishman, line 270: What's this about for Windows?

.github/workflows/tests.yml
```diff
@@ -177,3 +172,99 @@ jobs:
...
+      - name: Run Tests macOS 3.8
+        env:
+          QISKIT_SUPPRESS_PACKAGING_WARNINGS: Y
+          QISKIT_IN_PARALLEL: TRUE
+        run: |
+          set -e
+          pip check
+          stestr run --slowest
+        shell: bash
+        if: runner.os == 'macOS' && matrix.python-version == '3.8'
```
jakelishman, line 225: The Python 3.8 special parallel handling was introduced in #1064, before Python 3.9 was released, and I'm assuming because of issues with `spawn`?  Given that the previous test runs without the environment variable in Python 3.9, is the special handling still necessary?

.github/workflows/tests.yml
mtreinish: The code page piece is about circuit drawing. The default code page in windows is not utf8 and the text circuit drawer will emit characters the windows character set doesn't understand. This would cause an encoding error when a test errored or tried to print a circuit diagram. Changing the code page to utf8 prior to running the tests solves this.

The pip check line is mostly legacy from the run up to pip having a depsolver. We added it per the general recommendation of the pypa to test how the depsolver would work in our environment prior to the depsolver going live (in pip 21.0 iirc). We can probably remove it but it also doesn't hurt anything it's a quick check

.github/workflows/tests.yml
mtreinish: Yeah, we can remove this, it's actually encoded in terra by default now: https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/tools/parallel.py#L62-L79

mtreinish: Done in: https://github.com/Qiskit/qiskit-aer/pull/1448/commits/89dab544f87dff70a8677d2757c4445306b09642

.github/workflows/tests.yml
jakelishman: it was the code page thing I was really wondering about, yeah.  Good to know, thanks!

jakelishman (approved):

hhorii:
@mtreinish can we keep a workflow to test the latest terra as optional? I know other integration tests should cover tests with the latest qiskit-*. But, maybe, it is good to know what tests will be failed with the latest-terra in this repository.

PR #1430, opened 2022-01-18
Drop support for Python 3.6

mtreinish:
### Summary

Now that Qiskit Aer 0.10.0 has been released that Qiskit no longer
supports python 3.6 (in the lead up to adding support for python 3.10).
This commit officially drops the support for python 3.6 and makes python
3.7 required to install qiskit-aer. It also removes all the ci jobs
still using 3.6 both for testing and building wheels.

PR #1449, opened 2022-02-04
Only install numpy and scipy from wheels in build jobs

mtreinish:
### Summary

With recent changes to the upstream numpy and scipy packaging our CI
build jobs are spending most of their time build numpy and/or scipy from
source to test the built wheels function correctly. This is unecessary
as there are compatible wheels available on pypi. To correct this issue
this commit updates the cibuildwheel config used in the wheel build jobs
(both for ci and release artifacts) to only use binary wheels for numpy
and scipy before testing.

### Details and comments

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,13 +134,6 @@ jobs:
       - name: Install deps
         run: python -m pip install -U cibuildwheel==2.1.2
       - name: Build Wheels
-        env:
-          CIBW_BEFORE_ALL_LINUX: "yum install -y openblas-devel"
-          CIBW_SKIP: "pp*"
-          CIBW_BEFORE_TEST: "pip install --only-binary=numpy,scipy numpy scipy"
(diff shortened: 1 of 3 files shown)


mtreinish:
We really should unify the cibuildwheel config in the pyproject.toml. But I do that in #1430 so I pushed this separately to the existing config so we can backport it as part of 0.10.3.

jakelishman (approved):
It'd be good to split out a bunch of the CIBW config into `pyprojet.toml` so we can control it centrally, but I'm guessing that's perhaps what you were already thinking with this series of PRs.

jakelishman:
We're still using CIBW 2.1 on Aer, so we probably need to manually set the `manylinux` wheel VMs to 2014 for Python 3.10 to get Numpy suppport.

chriseclectic (dismissed):

chriseclectic (approved):

PR #1450, opened 2022-02-06
fix asv version to 0.4.2

hhorii:
### Summary

Build of the latest asv (0.5.0) is failed in tox

### Details and comments

Fix the version of asv to work around build fails in asv 0.5.0.


hhorii:
This may be no longer necessary.

PR #1454, opened 2022-02-07
Fix unit handling in RelaxationNoisePass

### LINKED ISSUE #1453: RelaxationNoisePass assumes dt delay units
### Informations

- **Qiskit Aer version**: 10.2
- **Python version**:
- **Operating system**:

Running the delay noise pass (or fake backend) on a non-scheduled circuit with a delay gate will apply delay noise assuming the delay duration is in `dt` units, even if it is defined in other units.

The follow example illustrates this descrepency

```python
from qiskit.test.mock import FakeMontreal
import matplotlib.pyplot as plt

sim = AerSimulator.from_backend(FakeMontreal())


circs = []
for unit in ['dt', 'ns', 'us']:
    circuit = QuantumCircuit(1,1)
    circuit.x(0)
    circuit.delay(50000, [0], unit=unit)
    circuit.measure([0],[0])
    circs.append(circuit)

shots = 1000
seed = 99
result_sched = sim.run(transpile(circs, sim, scheduling_method='alap'),
                       seed_simulator=seed, shots=shots).result()
result_not_sched = sim.run(circs, seed_simulator=seed, shots=shots).result()

p1s_sched = [counts['1'] / shots for counts in result_sched.get_counts()]
p1s_not_sched = [counts['1'] / shots for counts in result_not_sched.get_counts()]
plt.plot(p1s_sched, 'o-', label='scheduled')
plt.plot(p1s_not_sched, 's-', label='not scheduled')
plt.ylim(0, 1)
plt.legend()
```
Screen Shot 2022-02-07 at 11 38 41 AM

The blue and orange curves should look the same in the above plot.

### Suggested solutions

The RelaxationNoisePass should check the units of the the gate when accessing its duration. This also means the dt parameter of the pass init function could be made optional and would then raise an exception if it was not provided and a dt unit duration gate is attempted to run.

chriseclectic:
### Summary

Fixes #1453

RelaxationNoisePass assumes all instructions have dt unit durations. This fix changes it to check the specified unit for each op and convert non-dt units to seconds correctly.

### DIFF
--- a/qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
+++ b/qiskit/providers/aer/noise/passes/relaxation_noise_pass.py
@@ -18,8 +18,10 @@ from typing import Optional, Union, Sequence, List
 import numpy as np
 
 from qiskit.circuit import Instruction, QuantumCircuit
+from qiskit.utils.units import apply_prefix
 from .local_noise_pass import LocalNoisePass
 from ..errors.standard_errors import thermal_relaxation_error
+from ..noiseerror import NoiseError
 
 
 class RelaxationNoisePass(LocalNoisePass):
@@ -29,7 +31,7 @@ class RelaxationNoisePass(LocalNoisePass):
             self,
             t1s: List[float],
             t2s: List[float],
-            dt: float,
+            dt: Optional[float] = None,
             op_types: Optional[Union[type, Sequence[type]]] = None,
             excited_state_populations: Optional[List[float]] = None,
     ):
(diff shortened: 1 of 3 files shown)


itoko (approved):
LGTM, thank you for the quick fix!

evmckinney9:
@chriseclectic The documentation specifies 

```
t1s: List of T1 times in seconds for each qubit.
t2s: List of T2 times in seconds for each qubit.
```
however, I have found that keeping them in nanoseconds (as well as specifying `dt=1e-9`) gives me a far more accurate simulated fidelity. 

Is this a case of the documentation needing a correction, something is off with unit conversions, or possibility my simulations being totally incorrect?

PR #1455, opened 2022-02-07
Fix LocalNoisePass for circuit returns from noise func

### LINKED ISSUE #1447: RelaxationNoisePass creates noisy circuits that cannot be run directly on simulator for circuits with multi-qubit gates
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

`RelaxationNoisePass` produces noisy circuits that cannot be run directly on Aer simulator when circuits with multi-qubit gates are supplied. Because the output circuit contains composite instructions labeled like "circuit-123". So before running the circuit on simulator, users must transpile the circuit to decompose the composite instructions.

```
from qiskit.providers.aer.noise.passes import RelaxationNoisePass

from qiskit import transpile
from qiskit.circuit import Delay, QuantumCircuit
from qiskit.circuit.library.standard_gates import SXGate, CXGate

circ = QuantumCircuit(2)
circ.h(0)
circ.cx(0, 1)
circ.measure_all()
circ = transpile(circ,
                 basis_gates=["rz", "sx", "cx"],
                 scheduling_method="alap",
                 instruction_durations=[
                     ("rz", None, 0),
                     ("sx", None, 160),
                     ("cx", None, 800),
                     ("measure", None, 20000)
                 ])
t1s = t2s = [1e-4, 2e-4]
noise_pass = RelaxationNoisePass(t1s=t1s, t2s=t2s, dt=1e-9, op_types=[Delay, SXGate, CXGate])
noisy_circ = noise_pass(circ)
job = AerSimulator().run(noisy_circ)  # error
```

We can avoid the above error by transpiling the circuit (decomposing "circuit instructions") before running on the simulator.
```
noisy_circ = transpile(noisy_circ, basis_gates=["rz", "sx", "cx", "quantum_channel", "delay"])
```
However, it feels a little awkward.

Circuits created by`RelaxationNoisePass` should be able to be run directly on Aer simulator.

### Suggested solutions
- RelaxationNoisePass produces circuits that do not contains composite instructions.
(or AerSimulator becomes to be able to run circuits with composite instructions.)

chriseclectic:
### Summary

Fixes #1447 

IF the local noise pass function returns a QuantumCircuit instead of an instruction or operator this composes the circuit into the DAG so that it is unrolled, rather than added as an opaque circuit instruction.

### DIFF
--- a/qiskit/providers/aer/noise/passes/local_noise_pass.py
+++ b/qiskit/providers/aer/noise/passes/local_noise_pass.py
@@ -14,13 +14,14 @@ Local noise addition pass.
 """
 from typing import Optional, Union, Sequence, Callable, Iterable
 
-from qiskit.circuit import Instruction
+from qiskit.circuit import Instruction, QuantumCircuit
 from qiskit.dagcircuit import DAGCircuit
+from qiskit.converters import circuit_to_dag
 from qiskit.transpiler import TransformationPass
 from qiskit.transpiler.exceptions import TranspilerError
 from ..errors import QuantumError, ReadoutError
 
-InstructionLike = Union[Instruction, QuantumError]
+InstructionLike = Union[Instruction, QuantumError, QuantumCircuit]
 
 
 class LocalNoisePass(TransformationPass):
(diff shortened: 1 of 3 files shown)


qiskit/providers/aer/noise/passes/local_noise_pass.py
```diff
@@ -100,31 +101,52 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            # to conver to an Instruction
+            if not isinstance(new_op, (QuantumCircuit, Instruction)):
                 try:
                     new_op = new_op.to_instruction()
                 except AttributeError as att_err:
                     raise TranspilerError(
                         "Function must return an object implementing 'to_instruction' method."
                     ) from att_err
+
+            # Valide the instruciton matches the number of qubits and clbits of the node
```
mtreinish, line 133:
```suggestion
            # Validate the instruction matches the number of qubits and clbits of the node
```

mtreinish (dismissed):
LGTM, seems straightforward enough one small comment typo inline. It'd be good to get @itoko to review this too as he's more familiar with this pass though.

Also something I couldn't think of a good place to ask inline, but is there a scenario where `new_dag` contains a single op node? If so it'd probably be much more efficient to just do `substitute_node(node, new_op, inplace=True)` instead of `substitute_node_with_dag()` as it's just a reference update and doesn't need to traverse the new dag or deal with any of those complexities. (but that should be a follow-up PR optimization if it can be done, and not something I'd want to backport either)

chriseclectic:
@mtreinish I think it is usually at least 2 nodes (the original + the error) unless the pass is being run with `replace=True`.

If replace is True then in principle it could be a single node substitution (depending on function return being a single N-qubit instruction/op vs a quantum circuit). If it has a major effect on performance it's something that could be handled as a special case, and we could exploit it with pre-made passes by replacing gates with a quantum channel that is `ideal.compose(error)` rather than appending just the error after the ideal gate.

test/terra/noise/passes/test_local_noise_pass.py
```diff
@@ -94,3 +94,19 @@ def out_readout_error(op, qubits):
...
+            circ = QuantumCircuit(2)
+            for q in qubits:
+                circ.x(q)
+            return circ
+
+        noise_pass = LocalNoisePass(func=composite_error, op_types=CXGate)
+        noise_qc = noise_pass(qc)
+        self.assertEqual(len(noise_qc.data), 3)
+        self.assertEqual(noise_qc.data[1][0].name, "x")
+        self.assertEqual(noise_qc.data[2][0].name, "x")
```
itoko, line 112:
```suggestion

        expected = QuantumCircuit(2)
        expected.cx(0, 1)
        expected.x(0)
        expected.x(1)

        self.assertEqual(expected, noise_qc)
```

itoko (approved):
LGTM, I cannot think of any better way to fix. I just left one minor suggestion.

mtreinish:
> If replace is True then in principle it could be a single node substitution (depending on function return being a single N-qubit instruction/op vs a quantum circuit). If it has a major effect on performance it's something that could be handled as a special case, and we could exploit it with pre-made passes by replacing gates with a quantum channel that is `ideal.compose(error)` rather than appending just the error after the ideal gate.

For a single node replacement `substitute_node` is about 10-15x faster with `inplace=True` (and 5-7x with `in_place=False`) than `substitute_node_with_dag()`. That's also ignoring construction time to create the subcircuit equivalent dag to replace the node with. I made a little benchmark script to test this:

```python
import time

from qiskit.converters import circuit_to_dag
from qiskit.circuit.library import U2Gate

qc = QuantumCircuit(2)
qc.h(0)
qc.cx(0, 1)
qc.measure_all()
# Build dags for benchmarking
dag = circuit_to_dag(qc)
dag_two = circuit_to_dag(qc)
# equiv replacement for sub node with dag
equiv_h = QuantumCircuit(1)
equiv_h.u2(0, 3.14159, 0)
equiv_dag = circuit_to_dag(equiv_h)
# sub in place
dag_node_to_replace = dag.op_nodes()[0]
equiv_node = U2Gate(0, 3.14159)
start = time.time()
dag.substitute_node(dag_node_to_replace, equiv_node, inplace=False)
stop = time.time()
print(f"substitute_node: {stop - start}")
# sub with dag
dag_node_to_replace = dag_two.op_nodes()[0]
start = time.time()
dag_two.substitute_node_with_dag(dag_node_to_replace, equiv_dag)
stop = time.time()
print(f"substitute_node_with_dag: {stop - start}")
```
But like I said earlier it's not really something we should worry about for this PR or the bugfix release. Just a potential optimization in the future.

Issue #1452, opened 2022-02-07
MPI chunk/shot distribution is not working


doichanj:
### Informations

- **Qiskit Aer version0.11.0:
- **Python version**:
- **Operating system**:

### What is the current behavior?
When simulating with noise model using MPI + GPU, multiple shots will be distributed to multiple processes over MPI, but shots are not distributed.
Also parallelizing large qubits circuit is not working correctly.

### Steps to reproduce the problem
Building Aer with MPI support and run simulation on 2 or more processes.

### What is the expected behavior?
Simulation time should be decreased by increasing number of processes of MPI.

### Suggested solutions
I will recover MPI distribution codes later.

PR #1456, opened 2022-02-08
Fix for MPI distribution issue

doichanj:
### Summary
This PR is a fix for issue #1452 

### Details and comments
I added missed function call to StateChunk::set_distribution that sets MPI configurations.
This fix only affects to MPI build (built by user), so there is no side effect for binary distribution.


src/simulators/state_chunk.hpp
```diff
@@ -508,6 +508,12 @@ class StateChunk : public State<state_t> {
 
   uint_t mapped_index(const uint_t idx);
 
+  //apply OpenMP parallelization to loop with lambda function if OpenMP can be applied
+  template<typename Lambda>
+  void apply_omp_parallel(bool enabled, int_t i_begin, int_t i_end, Lambda& func);
```
hhorii, line 513: Can we move `apply_omp_parallel*` to framewrok/utils.hpp ?

hhorii (approved):
LGTM

PR #1460, opened 2022-02-09
stop installing qiskit-terra from source in tox

hhorii:
### Summary

stop installing qiskit-terra from source in tox

### Details and comments

This changes installation of qiskit-terra in tox.
Recent commit of #1448  changed installation of qiskit-terra to use pypi in most CI workflow, but not tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,6 @@ setenv =
   LC_ALL=en_US.utf-8
 deps =
   -r requirements-dev.txt
-  qiskit-terra
 commands =
     python setup.py bdist_wheel -- -- -j4
     pip install --find-links={toxinidir}/dist qiskit_aer
@@ -31,7 +30,6 @@ commands =
 deps =
   pycodestyle
   pylint
-  qiskit-terra
 commands =
   pycodestyle --ignore=E402,W504 --max-line-length=100 qiskit/providers/aer
   pylint -j 2 -rn qiskit/providers/aer
@@ -39,7 +37,6 @@ commands =
 [testenv:docs]
 deps =
   -r requirements-dev.txt
-  qiskit-terra
   qiskit-ibmq-provider
 commands =
   python setup.py bdist_wheel -- -- -j4
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ setenv =
   LC_ALL=en_US.utf-8
 deps =
   -r requirements-dev.txt
-  git+https://github.com/Qiskit/qiskit-terra.git
+  qiskit-terra
 commands =
     python setup.py bdist_wheel -- -- -j4
     pip install --find-links={toxinidir}/dist qiskit_aer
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -31,15 +31,15 @@ commands =
...
+  qiskit-terra
 commands =
   pycodestyle --ignore=E402,W504 --max-line-length=100 qiskit/providers/aer
   pylint -j 2 -rn qiskit/providers/aer
 
 [testenv:docs]
 deps =
   -r requirements-dev.txt
-  git+https://github.com/Qiskit/qiskit-terra.git
+  qiskit-terra
```
mtreinish, line 42:
```suggestion
```

mtreinish (commented):
You don't actually need to list `qiskit-terra` here, it's included in the requirements list for the project and will always be installed when pip installs aer. The additional line here was to pre-install terra from source outside of pip's normal behavior

mtreinish:
> I'm not sure if that change was intended @mtreinish can you comment? The original reason for installing qiskit from main was to catch when qiskit changes broke Aer before release, which used to happen quite frequently.

It was very much intentional (see the commit message on: https://github.com/Qiskit/qiskit-aer/commit/1d19b6724d16e0a05364f8b534229aeb108e4caa ). I was always opposed to doing that because it puts the burden on the wrong project. Catching API breaks in terra should be terra's responsibility not aer's, especially because for that strategy to work we need someone to actively be on top of failures.

garrison:
> it puts the burden on the wrong project. Catching API breaks in terra should be terra's responsibility not aer's

Why not add a workflow to Terra's CI that ensures that a given pull request does not break Aer's test suite?  This would allow breakage to be detected and dealt with even sooner than is possible currently, rather than later (which is what #1448 does).  Even if this Aer test within the Terra CI occurs on a single python version, it will detect the vast majority of changes that lead to Aer breaking.

mtreinish (approved):

mtreinish:
I think we can have a separate issue (probably on terra) to discuss this as this is off-topic for this mostly mechanical fix. I agree having some automated opt-in testing of the wider downstream ecosystem is very valuable. I wasn't familiar with how julia's testing was setup, but I've seen similar systems before. Rust has a similar opt-in mechanism they use around release time to test compilation of everything on crates.io. The issue with this for qiskit right now is we don't have the automation systems in place to do that kind of testing and it would have to be built (there are also mechanical issues around how we find the wider downstream, install them, and test them which isn't as simple as in a standalone language like julia or rust that have their own package repos). The idea behind having more coupled integration tests running pre-merge on every PR is more an acknowledgement that the qiskit-* projects under the qiskit github org are more tightly coupled and we want to ensure that as a cohesive unit that is always working. To that end having an integration test suite run on every commit is what I was planning for as a first step because that will get us a lot and also should make dealing with this api breakages between qiskit projects easier to manage.

PR #1467, opened 2022-02-22
Change to truncate T2 up to 2 * T1 in NoiseModel.from_backend

### LINKED ISSUE #1464: Warning T2 > 2 * T1 in simulation with device noise model rather than failing
Currently, noisy simulation with default noise model created from a device backend fails when the backend reports T2 value greater than 2 * T1 for a qubit.

For example, 
```
noise_model = NoiseModel.from_backend(device_backend)
AerSimulator(noise_model=noise_model).run(scheduled_circ)
```
occasionally fails with an error like
```
FAILURE: Can not get job id, Resubmit the qobj to get job id. Terra job error: 
'Invalid T_2 relaxation time parameter: T_2 greater than 2 * T_1.'
```

Instead of failing to run, the T2 value should be truncated to 2 * T_1 within the noise model and the warning like below should be issued.
```
Device model returned an invalid T_2 relaxation time parameter greater than the theoretical maximum value 2 * T_1 (??? > 2 * ???). Truncating to maximum value.
```

Note: Warnings are issued for the other types of errors in device parameters, e.g.
```
Device model returned a depolarizing error parameter greater than maximum allowed value (1.067061 > 1.066667). Truncating to maximum value.
```

itoko:
Fixes #1464 by changing the behavior of `NoiseModel.from_backend` when T2 value greater than 2 * T_1 is supplied such that it truncates the T2 value up to 2 * T_1 with issuing a user warning. Previously, thermal relaxation errors were created based on non-truncated T2 values, which was causing errors when running noise simulations as reported in #1464.


hhorii:
This needs a test case at least.

test/terra/noise/test_noise_model.py
```diff
@@ -229,6 +229,55 @@ def test_noise_model_from_mumbai(self):
...
+                """defaults == configuration"""
+                return self._configuration
+
+            def properties(self):
+                """properties == configuration"""
+                return self._configuration
+
+        backend = InvalidT2Fake1Q()
+        with self.assertWarns(UserWarning):
+            NoiseModel.from_backend(backend)
```
hhorii, line 279: checking corrected T2 is necessary here, isn't it?

hhorii (approved):
LGTM

PR #1469, opened 2022-02-27
Do not build with `-ffast-math`

### LINKED ISSUE #1461: UserWarning: The value of the smallest subnormal for <class 'numpy.float64'> type is zero.
### Informations

- **Qiskit Aer version**: 0b0913ca6f62f72c7ddb25c7e91ed5e58846280f
- **Python version**: 3.9.6
- **Operating system**: RHEL 8.5

Importing AerSimulator gives a Numpy warning.

```python
```
```
/home/kevinsung/.local/share/virtualenvs/qiskit-terra-rZn7sWYf/lib64/python3.9/site-packages/numpy/core/getlimits.py:499: UserWarning: The value of the smallest subnormal for <class 'numpy.float64'> type is zero.
  setattr(self, word, getattr(machar, word).flat[0])
/home/kevinsung/.local/share/virtualenvs/qiskit-terra-rZn7sWYf/lib64/python3.9/site-packages/numpy/core/getlimits.py:89: UserWarning: The value of the smallest subnormal for <class 'numpy.float64'> type is zero.
  return self._float_to_str(self.smallest_subnormal)
/home/kevinsung/.local/share/virtualenvs/qiskit-terra-rZn7sWYf/lib64/python3.9/site-packages/numpy/core/getlimits.py:499: UserWarning: The value of the smallest subnormal for <class 'numpy.float32'> type is zero.
  setattr(self, word, getattr(machar, word).flat[0])
/home/kevinsung/.local/share/virtualenvs/qiskit-terra-rZn7sWYf/lib64/python3.9/site-packages/numpy/core/getlimits.py:89: UserWarning: The value of the smallest subnormal for <class 'numpy.float32'> type is zero.
  return self._float_to_str(self.smallest_subnormal)
```

No warning.

### Suggested solutions
None yet.

jakelishman:
### Summary

This flag enables several unsafe floating-point transformations that are
not generally appropriate for a distributed mathematical library.  Most
importantly, however, `-ffast-math` on `gcc` also forcibly enables the
floating-point "flush to zero" mode for subnormal numbers when the
dynamic library is loaded, which affects not just Qiskit Aer, but every
other floating-point operation in the running process.  This is not our
decision to make for users, and it shouldn't be distributed.

### Details and comments

Fix #1461.

This will hopefully also fix testing failures in the randomised tests on Terra (see https://github.com/Qiskit/qiskit-terra/issues/2645#issuecomment-1053059832 and the small comment chain immediately below it).

I can't reproduce the errors locally easily because I can only reliably build Aer with `clang` on my Mac, which doesn't mess with the subnormal handling.  If @mtreinish or somebody could verify this fixes the issue, that'd be great.


jakelishman:
I would imagine you can check the subnormal handling in a roundabout manner by checking to see if `sys.float_info.min / 2 == 0`.  If it's set correctly (i.e. subnormals are valid and non-zero), then it should be `False`, whereas if something's messing with it, it'll be `True`.

hhorii:
I tested several benchmarks to compare 8c400ca and e88a5a0 and no regression exists.

RealAmplitudes (full entanglement):
```
[  0.00%] · For qiskit-aer commit 8c400ca3:
[  0.00%] ·· Benchmarking conda-py3.8
[ 50.00%] ··· statevector_cpu.Benchmark.track_real_amplitudes_full                                                                                        ok
[ 50.00%] ··· ======== ======================
               qubits                        
              -------- ----------------------
                 5      0.011002302169799805 
                 15     0.06725907325744629  
                 25      5.768640518188477   
              ======== ======================

[ 50.00%] · For qiskit-aer commit e88a5a0e:
[ 50.00%] ·· Building for conda-py3.8..
[ 50.00%] ·· Benchmarking conda-py3.8
[100.00%] ··· statevector_cpu.Benchmark.track_real_amplitudes_full                                                                                        ok
[100.00%] ··· ======== ======================
               qubits                        
              -------- ----------------------
                 5      0.011059761047363281 
                 15     0.08462214469909668  
                 25      5.381572961807251   
              ======== ======================

```
QuantumVolume:
```
[  0.00%] · For qiskit-aer commit 8c400ca3:
[  0.00%] ·· Benchmarking conda-py3.8
[ 50.00%] ··· statevector_cpu.Benchmark.track_qv                                                                                                          ok
[ 50.00%] ··· ======== =====================
               qubits                       
              -------- ---------------------
                 5        0.0101318359375   
                 15     0.07631731033325195 
                 25      1.8449420928955078 
              ======== =====================

[ 50.00%] · For qiskit-aer commit e88a5a0e:
[ 50.00%] ·· Building for conda-py3.8..
[ 50.00%] ·· Benchmarking conda-py3.8
[100.00%] ··· statevector_cpu.Benchmark.track_qv                                                                                                          ok
[100.00%] ··· ======== ======================
               qubits                        
              -------- ----------------------
                 5      0.010067224502563477 
                 15     0.07092523574829102  
                 25      1.781517505645752   
              ======== ======================
```

jakelishman:
Yeah, it's pretty unlikely that `-ffast-math` is doing anything much helpful for us; we're far more likely constrained by memory throughput and cache efficiency.

hhorii (approved):
Because `fast-math` no longer matter performance, this option is not necessary, I believe. (Currently, SIMD works for large qubit simulation, then I guess no performance impact on this option.)

jakelishman:
Both @mtreinish and I checked compilation with `gcc` on Linux (which in theory would be the only problematic compiler - `clang` doesn't mess with subnormal handling by default), and this change should be sufficient to remove the errors and the bad behaviour.

For what it's worth, the check I proposed with `sys.float_info` does work (the version of Aer in the snippet is `main`, including the buggy behaviour):
```python
>>> import sys
>>> sys.float_info.min / 2 == 0
False
>>> import qiskit.providers.aer
/home/jake/code/qiskit/venv/py310/lib/python3.10/site-packages/numpy/core/getlimits.py:499: UserWarning: The value of the smallest subnormal for <class 'numpy.float64'> type is zero.
  setattr(self, word, getattr(machar, word).flat[0])
/home/jake/code/qiskit/venv/py310/lib/python3.10/site-packages/numpy/core/getlimits.py:89: UserWarning: The value of the smallest subnormal for <class 'numpy.float64'> type is zero.
  return self._float_to_str(self.smallest_subnormal)
/home/jake/code/qiskit/venv/py310/lib/python3.10/site-packages/numpy/core/getlimits.py:499: UserWarning: The value of the smallest subnormal for <class 'numpy.float32'> type is zero.
  setattr(self, word, getattr(machar, word).flat[0])
/home/jake/code/qiskit/venv/py310/lib/python3.10/site-packages/numpy/core/getlimits.py:89: UserWarning: The value of the smallest subnormal for <class 'numpy.float32'> type is zero.
  return self._float_to_str(self.smallest_subnormal)
>>> sys.float_info.min / 2 == 0
True
```

mtreinish (approved):

1ucian0:
https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html noticed this.

moyix:
> Both @mtreinish and I checked compilation with `gcc` on Linux (which in theory would be the only problematic compiler - `clang` doesn't mess with subnormal handling by default), and this change should be sufficient to remove the errors and the bad behaviour.

Happy this was already noticed and fixed! One quick note here, `clang` will also link in `crtfastmath.o` if it's present on the system from an existing installation of `gcc`. Ref: https://github.com/llvm/llvm-project/issues/57589#issuecomment-1238629834

jakelishman:
Ah, thanks for that comment about the extra lib - it's definitely good to know for next time.  I'm guessing I didn't see that because my work laptop is a Mac, and `gcc` can't even find its own sysroot on that without an annoying amount of prodding!

PR #1471, opened 2022-03-01
Scrub old Cython-related code and comments

jakelishman:
### Summary

Aer has not used Cython at all since late 2020 (see #1014), and there
were still a few minor comment references to it in the code.  This also
removes two CMake files that were used during the Cython build process,
but are no longer needed, except for a trivial variable that was only
used in `FindPybind11.cmake`, and is now calculated there.

### Details and comments

This is just bookkeeping - mostly just deleting old CMake files we're not using.  At the time of making this commit, it looks like all the rest of the files in `/cmake` are used in build processes in some form.  It was on my mind because Qiskit/qiskit#1439 scrubs some public references to Cython, and I was interested enough to have a check through the old history to see how Aer used to use it.


hhorii (approved):
LGTM

PR #1472, opened 2022-03-02
Split out gpu Python 3.10 wheel build

mtreinish:
### Summary

The aer gpu package wheel job for python 3.10 builds in the manylinux2014
container image, while for the older Python versions it uses the
manylinux2010 image. The manylinux2014 image is based on Centos 7 while
the manylinux2010 image is based on Centos 6. During the qiskit-aer
0.10.3 release the gpu wheel job failed because when running the py3.10
build we installed the cuda package for centos 6 in the manylinux2014
centos 7 based image. This caused the job to fail when nvcc was called
because it was not installed properly. To fix this issue the Python 3.10
build is split out into a separate job. This separate job will install
the same cuda version but for centos 7 which is appropriate for the
manylinux2014 image.

### Details and comments

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -134,8 +134,40 @@ jobs:
           python -m pip install cibuildwheel==2.2.2
       - name: Build wheels
         env:
-          CIBW_BEFORE_ALL: "yum install -y yum-utils wget && wget https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && rpm -i cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && yum clean all && yum -y install cuda-10-1 openblas-devel"
-          CIBW_SKIP: "*-manylinux_i686 pp* cp36* *musllinux*"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -134,8 +134,40 @@ jobs:
           python -m pip install cibuildwheel==2.2.2
       - name: Build wheels
         env:
-          CIBW_BEFORE_ALL: "yum install -y yum-utils wget && wget https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && rpm -i cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && yum clean all && yum -y install cuda-10-1 openblas-devel"
-          CIBW_SKIP: "*-manylinux_i686 pp* cp36* *musllinux*"
+          CIBW_BEFORE_ALL: "yum install -y yum-utils wget && wget -q https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && rpm -i cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && yum clean all && yum -y install cuda-10-1 openblas-devel"
```
jakelishman, line 137: Surprised you need to `yum install wget` - is `curl` not available and preconfigured in the `manylinux` images?

(This doesn't need changing or for you to look up if you don't immediately know, I'm just curious if you happen to know off the top of your head.)

jakelishman (approved):
This looks fine to me.  There's a Windows-specific build step in these jobs that run on `ubuntu-latest` which is funny, but not an issue at all.

PR #1474, opened 2022-03-04
Add NoiseModel.from_backend_properties

yoshida-ryuhei:
### Summary
- Support date information when we construct the noise model from the information of Backend.

### Details and comments
- The motivation is to use [Measurement Error Mitigation](https://qiskit.org/documentation/tutorials/noise/3_measurement_error_mitigation.html) correctly.
- Now, if I construct the NoiseModel from the backend, the information about the current error is available.
- However, to use the Measurement Error Mitigation to the result on a real machine, such as ibmq_jakarta, we need to construct the NoiseModel based on information about errors when executing a job.
- 
### TODO
- [x] Update the document (e.g. refer `from_backend` and describe the difference)
- [x] Remove deprecating options standard_gates and warnings from arguments
- [x] Add a release note (https://github.com/Qiskit/qiskit-aer/blob/main/CONTRIBUTING.md#release-notes)


qiskit/providers/aer/noise/noise_model.py
```diff
@@ -307,7 +314,11 @@ def from_backend(cls, backend,
             raise NoiseError(
                 "NoiseModel.from_backend does not currently support V2 Backends.")
         if isinstance(backend, (BaseBackend, BackendV1)):
-            properties = backend.properties()
+            # TODO This if state below is not so good.
+            if datetime:
+                properties = backend.properties(datetime=datetime)
```
hhorii, line 319: [`BackendV1.properties`](https://github.com/Qiskit/qiskit-terra/blob/77219b5c7b7146b1545c5e5190739b36f4064b2f/qiskit/providers/backend.py#L147) does not take any arguments.

(Though deprecation warning is shown), use of `BackendProperties` instead of `BackendV1` is a work around. Also, defining custom backend that returns specific `BackendProperties` can resolve this issue.

yoshida-ryuhei: Thank you for reviewing it.
I forgot that [BackendV1.properties](https://github.com/Qiskit/qiskit-terra/blob/77219b5c7b7146b1545c5e5190739b36f4064b2f/qiskit/providers/backend.py#L147) does not have that argument.
I understand that by using `BackendProperties` what I want can be done, but I don't think it is a good idea to use a 
deprecated function.

I have added a code to check the class of backend is  `IBMQBackend`, whose class method `properties` has the argument `datetime`.
Could you please review it?

hhorii: `IBMQBackend` is in [qiskit-ibmq-provider](https://github.com/Qiskit/qiskit-ibmq-provider). qiskit-aer can not has dependency on it.
I believe that this issue should be resolved in `qiskit-ibmq-provider` (or user codes) and not in Aer. [`IBMQBackend.properties()`](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/ibmqbackend.py#L410-L413) may allow cache not-the-latest properties ([current code](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/ibmqbackend.py#L456-L457) does not).

yoshida-ryuhei: Thank you for your comment.

> IBMQBackend is in [qiskit-ibmq-provider](https://github.com/Qiskit/qiskit-ibmq-provider). qiskit-aer can not have dependency on it.
I understand your policy.

> I believe that this issue should be resolved in qiskit-ibmq-provider (or user codes) and not in Aer.
I disagree with that, I think the code to use local simulator should be in this repository, otherwise many similar codes should be included in a lot of repositories.
For example in this noise simulator, each library to control real quantum computers needs similar code to use a local simulator, or users must write each time, which is not good to use and maintain, I think.`

I understand the reason, thanks.

itoko:
Thank you for working on this and sorry for our slow response. @yoshida-ryuhei 
I think it's nice to have a separate method for generating noise model from backend properties. Because I've heard from several users that they want an alternative since we deprecated the support of `BackendProperties` from a `backend` option in `NoiseModel.from_backend`.
So I think the direction is good but needs improvement on implementation at least in the following points.
- Update the document (e.g. refer `from_backend` and describe the difference)
- Remove deprecating options `standard_gates` and `warnings`
- Add a release note

yoshida-ryuhei:
Thank you for your reply.
I would very much like to complete it.

I understand the first two improvements and the deadline clearly. 

I have a question about the last improvements, about the release note.
Should I write the release note just as the YAML files in the [releasenotes/notes](https://github.com/Qiskit/qiskit-aer/tree/main/releasenotes/notes)?
Thanks.

itoko:
Thanks. Regarding the release note, that's right. You can create a new YAML file for this PR using `reno` command. Please refer 
https://github.com/Qiskit/qiskit-aer/blob/main/CONTRIBUTING.md#release-notes for the details.

yoshida-ryuhei:
Now, I have revised this patch based on the 3 comments.
@itoko Could you check this again?

releasenotes/notes/add-NoiseModel-from_backendproperties-1a3d6d976133a661.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Add a new function :func:`from_backend_properties` to :class:`NoiseModel`.
+    This function returns a noise model derived from a device backend properties.
```
itoko, line 5: Again :meth:`~.NoiseModel.from_backend`

It might be good to mention the background why this is added. E.g.
The function allows to construct a noise model from a ``BackendProperties`` object, which can seen as a spin-off of a removed functionality of ``NoiseModel.from_backend``.

qiskit_aer/noise/noise_model.py
```diff
@@ -441,6 +441,110 @@ def from_backend(cls, backend,
...
+            gate_errors = basic_device_gate_errors(
+                backend_properties,
+                gate_error=gate_error,
+                thermal_relaxation=thermal_relaxation,
+                gate_lengths=gate_lengths,
+                gate_length_units=gate_length_units,
+                temperature=temperature,
+            )
+        for name, qubits, error in gate_errors:
+            noise_model.add_quantum_error(error, name, qubits, warnings=True)
```
itoko, line 519:
```suggestion
            noise_model.add_quantum_error(error, name, qubits)
```

qiskit_aer/noise/noise_model.py
```diff
@@ -441,6 +441,110 @@ def from_backend(cls, backend,
...
+            basis_gates.add(prop.gate)
+        basis_gates = list(basis_gates)
+        num_qubits = len(backend_properties.qubits)
+        dt = 0  # disable delay noise if dt is unknown
+        noise_model = NoiseModel(basis_gates=basis_gates)
+
+        # Add single-qubit readout errors
+        if readout_error:
+            for qubits, error in basic_device_readout_errors(backend_properties):
+                noise_model.add_readout_error(error, qubits, warnings=True)
```
itoko, line 501:
```suggestion
                noise_model.add_readout_error(error, qubits)
```

itoko (commented):
Thank you for the updates. Now, overall looks good to me. I leave minor suggestions inline.

yoshida-ryuhei:
Thank you very much for reviewing the codes.

Now I have revised the codes based on your comments.

Also, I corrected the wrong document about the expected error.

qiskit_aer/noise/noise_model.py
```diff
@@ -441,6 +441,107 @@ def from_backend(cls, backend,
...
+            temperature (double): qubit temperature in milli-Kelvin (mK) for
+                                  thermal relaxation errors (Default: 0).
+            gate_lengths (list): Custom gate times for thermal relaxation errors.
+                                  Used to extend or override the gate times in
+                                  the backend properties (Default: None))
+            gate_length_units (str): Time units for gate length values in
+                                     gate_lengths. Can be 'ns', 'ms', 'us',
+                                     or 's' (Default: 'ns').
+            dt (float): The system time resolution of input signals (Default: 0).
+                        This is required to be implemented if the backend supports Pulse scheduling.
```
itoko, line 481:
```suggestion
            dt (float): Backend sample time (resolution) in seconds (Default: None).
                        Required to convert time unit of durations to seconds
                        if including thermal relaxation errors on delay instructions.
```

itoko (commented):
Thanks, it's about to be merged. Followings are my last nit-picks. (Please double check the spec around `dt`. I might be wrong.)

yoshida-ryuhei:
Thank you @itoko.
All your comments are good, so I revised them as you said.
Additionally, I change the type of the document about the arguments,  `dt` and `gate_lengths` in commit b4698667082d2e65eca5ae63829aabf6f19e4832

itoko:
@yoshida-ryuhei Ah, thank you for bring up the issue on types in docstring. I missed that. Since we have type hints within the code, you can remove all type specs in docstrings. For example, if you write `dt: Optional[float] = None` in the code, the document like `dt (Optional[float] ...)` will be automatically generated. (see e.g. [source code](https://qiskit.org/documentation/_modules/qiskit_aer/noise/passes/relaxation_noise_pass.html#RelaxationNoisePass) and [generated doc](https://qiskit.org/documentation/stubs/qiskit_aer.noise.RelaxationNoisePass.html) on RelaxationNoisePass)

yoshida-ryuhei:
@itoko Thank you for the information.
Generating type hints automatically is a very good feature. I will use that next time.

I fixed all the codes I was concerned about. I would appreciate it if you could let me know if there is anything I should add.

itoko (approved):
LGTM, congratulations on your first contribution @yoshida-ryuhei !

jakelishman (approved):
Proxying Itoko's approval.

Issue #1479, opened 2022-03-11
Multi-chunk parallelization is not applied to density_matrix with noises


doichanj:
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?
Multi-chunk parallelization (cache blocking) is not applied when using density_matrix method with noise models.

### Steps to reproduce the problem
```
from qiskit import *
from qiskit.circuit.library import *
from qiskit.providers.aer import *
from qiskit.providers.aer.noise import *
import sys

noise_model = NoiseModel()

dev = sys.argv[2]
qubits = int(sys.argv[3])
shots = int(sys.argv[4])
error_rate = float(sys.argv[5])

if sys.argv[1] == "Kraus":
    #insert Kraus error
    error1 = amplitude_damping_error(error_rate, 1)
    noise_model.add_all_qubit_quantum_error(error1, ['u1', 'u2', 'u3','x','h','p'])
    print(">>> using Kraus noise")
else:
    #insert pauli errors
    error1 = depolarizing_error(error_rate, 1)
    error2 = depolarizing_error(error_rate, 2)
    noise_model.add_all_qubit_quantum_error(error1, ['u1', 'u2', 'u3','x','h','p'])
    noise_model.add_all_qubit_quantum_error(error2, ['cx'])
    print(">>> using Pauli noise")

sim = AerSimulator(method='density_matrix', device=dev, noise_model = noise_model)

circuit = transpile(QFT(qubits), backend=sim,optimization_level=0)
circuit.measure_all()

result = execute(circuit,sim,shots=shots,noise_model=noise_model,seed_simulator=12345,fusion_threshold=22,blocking_qubits=11).result()
if result.to_dict()['metadata']['mpi_rank'] == 0:
    print(result)
    print(result.to_dict()['backend_name'])
    print(result.to_dict()['results'][0]['time_taken'])
    print(sorted(result.to_dict()['results'][0]['data']['counts'].items(),key=lambda x:x[0]))

```

### What is the expected behavior?
cache_blocking metadata will be recorded in the result when is is applied. 

### Suggested solutions
cache blocking transpiler does not recognize some operator in density_matrix, add them to enable cache blocking.

hhorii:
fixed with #1480.

PR #1480, opened 2022-03-11
Fix: cache blocking transpiler for density_matrix with noises

doichanj:
### Summary
This is fix for issue #1479 

A test case to test Kraus noise on QFT with cache blocking (in` test_noise.py`)
Also this PR fixes issue #1483 that was found when adding the test case. 

### Details and comments
superop was missing in `CacheBlocking::can_reorder` function so recognize it by calling `CacheBlocking::is_blockable_operation`

I could not make separate PR for issue #1483 because `test_noise.test_kraus_gate_noise_on_QFT_cache_blocking` fails if we separate PR.

For issue #1483,
` QubitVectorThrust<data_t>::norm()` originally returns sum of norm for all chunks. 
This causes incorrect measurement result when qubit is out of chunk qubits.
I changed ` QubitVectorThrust<data_t>::norm()` returns norm for one chunk if ` QubitVectorThrust<data_t>::enable_batch_` is not set. ` sample_measure` only uses batched sum of norm at this time.

### DIFF
--- a/releasenotes/notes/density-multi-chunk-fix-e9effc67d0365418.yaml
+++ b/releasenotes/notes/density-multi-chunk-fix-e9effc67d0365418.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    Fix cache blocking transpiler to recognize superop to be cache blocked.
+    This is fix for
+    `issue 1479 <https://github.com/Qiskit/qiskit-aer/issues/1479>`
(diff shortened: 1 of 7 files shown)


hhorii:
Can we add tests for this?

hhorii (approved):
This looks good to me. No side effects on the other methods.

Issue #1473, opened 2022-03-03
Mult-shots batched optimization is slow for Pauli noise with multiple GPUs


doichanj:
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?
Batched multi-shots optimization for GPU is slower than without batched optimization when using Pauli noises.

Especially it is slow when using multiple GPUs

### Steps to reproduce the problem

```
from qiskit import *
from qiskit.circuit.library import *
from qiskit.providers.aer import *
from qiskit.providers.aer.noise import *
import sys

noise_model = NoiseModel()

dev = sys.argv[2]
shots = int(sys.argv[3])
niter = int(sys.argv[4])
error_rate = float(sys.argv[5])
max_batch = int(sys.argv[6])
max_qubits = int(sys.argv[7])

depth=10

if sys.argv[1] == "Kraus":
    #insert Kraus error
    error1 = amplitude_damping_error(error_rate, 1)
    noise_model.add_all_qubit_quantum_error(error1, ['u1', 'u2', 'u3','x','h','p'])
    print(">>> using Kraus noise")
else:
    #insert pauli errors
    error1 = depolarizing_error(error_rate, 1)
    error2 = depolarizing_error(error_rate, 2)
    noise_model.add_all_qubit_quantum_error(error1, ['u1', 'u2', 'u3','x','h','p'])
    noise_model.add_all_qubit_quantum_error(error2, ['cx'])
    print(">>> using Pauli noise")

sim = AerSimulator(method='statevector', device=dev, noise_model = noise_model)

print("=== device = {0}, shots = {1}, error_rate = {2}, max_batch = {3} ===".format(dev,shots,error_rate,max_batch))

for qubits in range (9, max_qubits):
    circuit = transpile(QFT(qubits), backend=sim,optimization_level=0)
    circuit.measure_all()

    time = 0.0
    for iter in range (0,niter):
        result = execute(circuit,sim,shots=shots,noise_model=noise_model,seed_simulator=12345,batched_shots_gpu=True,batched_shots_gpu_max_qubits=max_batch,fusion_threshold=30).result()
        time = time + float(result.to_dict()['results'][0]['time_taken'])

    if result.to_dict()['metadata']['mpi_rank'] == 0:
        print("{0},{1}".format(qubits,time/float(niter)))
```

This is example benchmarking **without** batched optimization.
```
[...]

PR #1482, opened 2022-03-15
Fix for GPU batched optimization with Pauli noise

doichanj:
### Summary
This PR is fix for issue #1473 

### Details and comments
Added option to allocate multiple threads per GPU and nested parallelism is applied to parallelize runtime noise sampling for Pauli noise in batched optimization.


src/noise/quantum_error.hpp
```diff
@@ -387,6 +390,20 @@ void QuantumError::load_from_json(const json_t &js) {
   set_circuits(circuits, probs);
 }
 
+
+bool QuantumError::pauli_only(void) const
```
hhorii, line 394: According to the local discussion with @itoko -san, it is bbetter to rename `pauli_only()` with `is_mixed_pauli()`.

src/simulators/state_chunk.hpp
```diff
@@ -419,7 +420,8 @@ class StateChunk : public State<state_t> {
                                const Noise::NoiseModel &noise,
                                ExperimentResult &result,
                                uint_t rng_seed,
-                               bool final_ops);
+                               bool final_ops,
+                               bool pauli_only);
```
hhorii, line 424: can we rename `pauli_only` with `batched_pauli_ops`?

qiskit/providers/aer/backends/aer_simulator.py
```diff
@@ -252,6 +252,11 @@ class AerSimulator(AerBackend):
       number of active circuit qubits is greater than this value batching of
       simulation shots will not be used. (Default: 16).
 
+    * ``num_threads_per_group`` (int): This option sets the number of
+      threads per group. For GPU simulation, this value sets number of
+      threads per GPU. This parameter is used to optimize Pauli noise
+      simulation with multiple-GPUs (Default: 1).
+
```
hhorii, line 259: `group` sounds too general. Can we rename this with `num_threads_per_device`?

src/simulators/state_chunk.hpp
```diff
@@ -890,6 +899,8 @@ void StateChunk<state_t>::apply_ops_multi_shots(InputIterator first, InputIterat
   int_t i;
   int_t i_begin,n_shots;
 
+  bool pauli_only = noise.pauli_only();
```
hhorii, line 902: I believe that `pauli_only` should be renamed with `batched_pauli_ops`.
Also, it is better to check `NoiseModel.opset()` here instead of adding a new method `pauli_only()`.

hhorii (changes requested):
I think `NoiseModel` should not have a new method `pauli_only()`. `StateChunk<state_t>::apply_ops_multi_shots()` can efficiently check whether all the quantum errors consist of pauli gates only by iterating `NoiseMode.opset()`.

src/noise/quantum_error.hpp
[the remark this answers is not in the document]
doichanj: removed

src/noise/noise_model.hpp
```diff
@@ -1047,6 +1050,19 @@ inline void from_json(const json_t &js, NoiseModel &model) {
   model = NoiseModel(js);
 }
 
+bool NoiseModel::pauli_only(void) const
```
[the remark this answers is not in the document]
doichanj: I recover the prior code to check if sampled gates only contains Pauli gates or not.

src/simulators/state_chunk.hpp
doichanj: removed

qiskit/providers/aer/backends/aer_simulator.py
doichanj: changed to num_threads_per_device

src/simulators/state_chunk.hpp
doichanj: removed

hhorii (approved):
LGTM. In my understanding, this change does not need tests because this improves performance only by parallelizing noise sampling.

PR #1489, opened 2022-03-23
Fix for the latest cuStateVec APIs

doichanj:
### Summary
This is the fix to support the latest cuQuantum (ver. 0.40).

### Details and comments
Compilation failed with cuStateVec support because some cuStateVec APIs' name and parameters are changed from the previous version. This PR follows the latest cuStateVec.

Also cuQuantum 0.40 newly supports memory pool mechanism to allocate work space inside cuStateVec APIs.
Qiskit Aer now use memory pool instead of allocating static work buffer if the version of CUDA installed on the system supports it.

### DIFF
--- a/src/simulators/statevector/chunk/cuStateVec_chunk_container.hpp
+++ b/src/simulators/statevector/chunk/cuStateVec_chunk_container.hpp
@@ -20,8 +20,6 @@
 
 #include "custatevec.h"
 
-#define SUPPORTS_MEMORY_POOL ( __CUDACC_VER_MAJOR__ > 11 || (__CUDACC_VER_MAJOR__ == 11 && __CUDACC_VER_MINOR__ >= 2))
-
 namespace AER {
 namespace QV {
 namespace Chunk {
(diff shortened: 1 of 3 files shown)


src/simulators/statevector/chunk/cuStateVec_chunk_container.hpp
```diff
@@ -20,6 +20,8 @@
 
 #include "custatevec.h"
 
+#define SUPPORTS_MEMORY_POOL ( __CUDACC_VER_MAJOR__ > 11 || (__CUDACC_VER_MAJOR__ == 11 && __CUDACC_VER_MINOR__ >= 2))
```
hhorii, line 23: Can we always use memory pool if it provides performance in any simulation? The latest cuQuantum changed its API (such as `custatevecSamplerCreate`) and the current `cuStateVec_chunk_container.hpp` does not support previous API (`custatevecSampler_create`). I guess, if the current code is based on the latest API,  memory pool is always available.

doichanj: I removed codes for older version of CUDA and now 11.2 or newer version is required to use cuStateVec support (added in CONTRIBUTING.md)

I think sampling does not support memory pool (https://docs.nvidia.com/cuda/cuquantum/custatevec/api/functions.html#sampling)
so work buffer should be allocated if needed.

hhorii (approved):

PR #1496, opened 2022-04-01
[WIP] Reduce BLAS dependency

hhorii:
### Summary

Work around memory allocation errors in OpenBLAS when multi-threads run

### Details and comments

OpenBLAS sometimes fail test cases of Aer when multiple threads run.

```
$ python -m unittest test.terra.backends.aer_simulator.test_noise.TestNoise 
....BLAS : Program is Terminated. Because you tried to allocate too many memory regions.
BLAS : Program is Terminated. Because you tried to allocate too many memory regions.
BLAS : Program is Terminated. Because you tried to allocate too many memory regions.
BLAS : Program is Terminated. Because you tried to allocate too many memory regions.
BLAS : Program is Terminated. Because you tried to allocate too many memory regions.
BLAS : Program is Terminated. Because you tried to allocate too many memory regions.
BLAS : Program is Terminated. Because you tried to allocate too many memory regions.
Segmentation fault
```

This error can be worked around by setting `OPENBLAS_NUM_THREADS=1 `:
```
$ OPENBLAS_NUM_THREADS=1 python -m unittest test.terra.backends.aer_simulator.test_noise.TestNoise 
.......................................................
----------------------------------------------------------------------
Ran 55 tests in 4.844s

OK
```

In CI, stestr runs multiple threads or multiple processes and then encounters unsuitability:
```
{0} test.terra.backends.aer_simulator.test_noise.TestNoise.test_kraus_gate_noise_on_QFT_cache_blocking_2___statevector____Thrust__ [] ... inprogress
```

This PR sets `OPENBLAS_NUM_THREADS=1 ` when `AerBackend` is loaded.


jakelishman:
Is this something we are capable of detecting within the library?  It feels like some bug in our defaults that this can happen.

hhorii:
[The official FAQ](https://github.com/xianyi/OpenBLAS/wiki/Faq#how-can-i-use-openblas-in-multi-threaded-applications) said:

> If your application is already multi-threaded, it will conflict with OpenBLAS multi-threading.

Maybe `openblas_set_num_threads(1)` is necessary if Aer uses OpenBLAS.

Another solution may be not to use BLAS. Currently, only matrix.hpp uses it to support matrix operators. Aer uses them only for small size of matrices. I guess that few performance impact from reduction of BLAS.

BTW, the current CI errors are from recent terra release. CI does not download binary from pypi, and then fails to compile qiskit-terra because of lack of rust compiler.

mtreinish:
We need to update our wheel jobs to use manylinux2014 so the terra binaries are compatible with the build env

jakelishman:
> Maybe `openblas_set_num_threads(1)` is necessary if Aer uses OpenBLAS

Yeah, I was thinking something along these lines - we know if we're spawning threads/processes already, and we should know whether we're linked against OpenBLAS or something else, so we should make sure we're not conflicting with their assumptions.

We should set up our defaults to "just work" for the most part - if we're having trouble with a simple `python -munittest TestNoise`, it looks like a behaviour problem on our side.

pyproject.toml
```diff
@@ -33,4 +33,4 @@ environment = { CMAKE_GENERATOR = "Visual Studio 16 2019"}
 # For older Pythons we maintain our manylinux2010 support.
 [[tool.cibuildwheel.overrides]]
 select = "cp3{6,7,8,9}-manylinux*"
-manylinux-x86_64-image = "manylinux2010"
+# manylinux-x86_64-image = "manylinux2010"
```
mtreinish, line 36: I fixed this for real in #1498 (and a corresponding fix for the stable branch in #1499)
```suggestion
```

mtreinish:
I'm not sure this fixes the underlying issue. For me locally when I run with main I'm able to still get a segfault when I run with `OPENBLAS_NUM_THREADS=1`:

```
OPENBLAS_NUM_THREADS=1 .tox/py310/bin/python -m unittest test.terra.backends.aer_simulator.test_noise.TestNoise 
.............OpenBLAS : Program will terminate because you tried to allocate too many memory regions.
...OpenBLAS : Program will terminate because you tried to allocate too many memory regions.
...<frozen importlib._bootstrap>:914: ImportWarning: _SixMetaPathImporter.find_spec() not found; falling back to find_module()
....................................
----------------------------------------------------------------------
Ran 55 tests in 2.141s

OK
zsh: segmentation fault (core dumped)  OPENBLAS_NUM_THREADS=1 .tox/py310/bin/python -m unittest 
```

I also tried with this PR applied and I also get a segfault in openblas although a without the allocation error message. It might be something with my local openblas install though. But I think we should avoid setting things via env variables via python like this if there is a bug in how we're using openblas I think we should fix that in the source.

hhorii:
OK. I believe that we need more time to resolve OpenBLAS issue and this PR should not block 0.10.4 release.

hhorii:
matrix multiplication is frequently used in `matrix_product_state` method. Here are performance comparisons of QV with 0.10.4 and 168f7b7 (Intel(R) Xeon(R) Gold 6140 CPU, 18cores x 2HT, Ubuntu 18.0.4, gcc 8.4.0). Probably, for low-qubit circuits, library calls to OpenBLAS library were bottlenecks.

image
image

Another comparisons on Macbook Pro(Intel Core i7, 6core x 2HT, macOS BigSur).
image
image

Aer does not use matrix multiplication heavily, in general.

hhorii:
In `statevector` method, matrix multiplication is used in fusion, mainly. Here are performance comparisons (QV) of `statevector` with 0.10.4 and https://github.com/Qiskit/qiskit-aer/commit/168f7b783719a44e49a850448eb79e9b112f51cc (Intel(R) Xeon(R) Gold 6140 CPU, 18cores x 2HT, Ubuntu 18.0.4, gcc 8.4.0). 
image
image

hhorii:
https://github.com/Qiskit/qiskit-aer/blob/a21260c5030c4bf01256e911e715e4bf9d176254/src/simulators/matrix_product_state/matrix_product_state.hpp#L1055-L1064

The above OMP loop produces SEGV in `test.terra.backends.aer_simulator.test_noise.TestNoise.test_readout_noise_5___matrix_product_state____CPU__`. By taking OMP pragma, no error is thrown.
If a bug exists in it, we need to fix it instead of work-around by reducing use of BLAS.

hhorii:
#1555 will resolve this BLAS issue

PR #1497, opened 2022-04-04
Fix device=Thrust was very slow for small number of qubits

doichanj:
### Summary
This is a fix for device=Thrust, Thrust running on CPU was very slow for small number of qubits.

### Details and comments
This was because Thrust CPU always applied OpenMP for threading kernel programs without referring to the number of qubits.
This fix applies OpenMP referring to statevector_parallel_threshold parameter as same as device=CPU.


hhorii (approved):

hhorii:
on hold until CI is fixed with #1498 or #1500.

hhorii:
This PR was included in #1498.

PR #1498, opened 2022-04-04
Use manylinux2014 for wheel builds

mtreinish:
### Summary

Starting with the recently released qiskit-terra 0.20.0 qiskit only
builds packages on manylinux2014 now. This commit updates the wheel jobs
for aer to only use manylinux2014 for builds now as we don't want to be
building terra from source (especially now that it requires rust).

### Details and comments


jakelishman (approved):
Did we decide we were ok to do this in a patch release?  I think maybe we did it once before on Aer, so probably not the end of the world.

mtreinish:
I was thinking that for stable we would stick with manylinux2010, as using manylinux2014 seems like a potentially disruptive change for a patch release. I pushed up #1499 to do just install an old terra release for 0.10.x wheel jobs

mtreinish:
Sigh, it looks like the wheel jobs are failing because epel dropped support for i686 in centos 7 so switching to manylinux2014 for i686 wheels fails because there is no openblas-devel package.

mtreinish:
It looks like this is blocked on https://github.com/Qiskit/qiskit-aer/pull/1497 I'm going to bundle that into this PR so we can unblock CI

hhorii (approved):
LGTM

PR #1499, opened 2022-04-04
Only install terra from binary for wheel jobs

mtreinish:
### Summary

In the recently released terra 0.20.0 release the minimum python
packaging spec supported is manylinux2014 now. On the main branch we
bumped the manylinux base image we use for wheel jobs to manylinux2014
in #1498. However, on the stable 0.10.x branch we don't want to do that
since we probably should not drop support for older environments on a
stable release. This commit updates the wheel job config to instead
install terra from a compatible binary wheel instead of using the latest
release. This should hopefully avoid the CI failure but still enable us
to run without building terra from source or dropping support for
manylinux2010.

### Details and comments


jakelishman (approved):

hhorii:
#1446 is necessary to pass tests. I pushed it to this branch.

PR #1500, opened 2022-04-05
Only install terra from binary for wheel jobs

hhorii:
### Summary

Only install terra from binary for wheel jobs

### Details and comments

In the recently released terra 0.20.0 release the minimum python
packaging spec supported is manylinux2014 now.  This commit updates
the wheel job config to instead install terra from a compatible binary wheel
instead of using the latest release. This will avoid compilation fails of terra in terra.

This commit will be necessary until main starts using only manylinux2014 in #1498.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ test-command = "python {project}/tools/verify_wheels.py"
 # tendency to crash if they're installed from source by `pip install`, and since
 # Numpy 1.22 there are no i686 wheels, so we force pip to use older ones without
 # restricting any dependencies that Numpy and Scipy might have.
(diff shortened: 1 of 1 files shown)


mtreinish:
I think instead of this we should just disable i686 builds on main. We don't want to be testing compatibility with terra 0.19.x forever. Even if I get things fixed in #1498 so that we're building openblas on i686 that takes ~15 min to build in CI and we don't want that done on every PR.

hhorii:
> I think instead of this we should just disable i686 builds on main. We don't want to be testing compatibility with terra 0.19.x forever. Even if I get things fixed in #1498 so that we're building openblas on i686 that takes ~15 min to build in CI and we don't want that done on every PR.

I agree.

PR #1504, opened 2022-04-08
Add cross-build wheel jobs for ppc64le and s390x

mtreinish:
### Summary

This commit adds a new CI job at release time to build precompiled
binaries for s390x and ppc64le linux platforms.  These platforms do not
have precompiled wheels for any upstream dependencies (except for
retworkx and tweedledum, which both publish binaries for both platforms).
Since the only publically available CI service that has native s390x and
ppc64le environments available is travis which doesn't provide
sufficient quota to open source projects anymore to make it usable we
have to rely on either cross compilation or emulation. Cross compiling
with python wheels while possible is quite tricky to setup and configure
in practice so emulation is used here. This is a much simpler path to
configure, especially because cibuildwheel has support for emulating
non-x86 architectures via QEMU to build wheels. While QEMU emulation of
other architectures is exceedingly slow, we have a 6 hour job time limit
with github actions which should hopefully be sufficient to compile the
binary and build wheels for all our supported python versions.

However, because this is so slow we don't have a sufficient time budget
for compiling anything but qiskit-aer. To test the wheels we'd have to
build most (not qiskit-terra or retworkx) dependencies from source and
many are slow to build under emulation. Scipy alone could take our
entire time budget. So to avoid this issue this commit skips the tests
and solely builds the wheels. This also means installing scipy and other
dependencies is still an exercise for the s390x and ppc64le users. But,
if this works it will mean that at least for Qiskit there will not be a
need to compile anything from source.

### Details and comments


mtreinish:
@hhorii if you want to give https://github.com/Qiskit/qiskit-aer/suites/5993576300/artifacts/208823384 a try on a ppc64le that would be good to confirm the build is working. I'll give the s390x wheels a try too. Assuming they work I'll revert https://github.com/Qiskit/qiskit-aer/pull/1504/commits/03addfc120d533173bf33465eb649a4e8454a106 and we can move forward with publishing wheels for these platforms in the next release and adding ppc64le and s390x as tier 3 supported (per https://qiskit.org/documentation/getting_started.html#tier-3)

mtreinish:
I tested the s390x wheels this morning and it seem to work fine. I did find/confirm a bug when running on the platform with passing unitaries in the circuit, I assume because s390x is big endian while all our other supported platforms are little endian, see: #1506 (this was originally reported in qiskit-experiments but I'm pretty sure it's an aer bug). But we can fix this independently of adding support for building on the platform to the release process.

dongcc:
@mtreinish is it possible for me to build the s390x wheels locally to downloaded from somewhere to have a try? thanks

mtreinish:
You can download the wheels built by this job from https://github.com/Qiskit/qiskit-aer/suites/5993576300/artifacts/208823384 that link will redirect to download a zip file containing all the precompiled binaries for all the python versions built in CI. Then you can use pip to locally install the appropriate file for your environment.

dongcc:
> You can download the wheels built by this job from https://github.com/Qiskit/qiskit-aer/suites/5993576300/artifacts/208823384 that link will redirect to download a zip file containing all the precompiled binaries for all the python versions built in CI. Then you can use pip to locally install the appropriate file for your environment.

thanks @mtreinish , is that by using of the s390x wheel file will fix the issue https://github.com/Qiskit/qiskit-aer/issues/1506?
I tried the wheel `qiskit_aer-0.11.0-cp38-cp38-manylinux_2_17_s390x.manylinux2014_s390x.whl` and it seemed that it still fails w/ the following code:
```
(qiskit361) [root@lpar38 q]# cat test.py
from qiskit import Aer
from qiskit.quantum_info import random_unitary

backend = Aer.get_backend("aer_simulator")
qc = QuantumCircuit(2)
qc.unitary(random_unitary(4), [0, 1])
qc.measure_all()
backend.run(qc)
(qiskit361) [root@lpar38 q]# python test.py
ERROR: Failed to load qobj: "unitary" matrix is not unitary.
```
Is that more code needs to be changed from the qiskit aer? thanks in advance,

mtreinish:
This was not intended to fix #1506 and does not change anything related to that issue. All this does is add the CI automation to enable publishing precompiled binaries for s390x (and ppc64le) at release time. As I mentioned earlier https://github.com/Qiskit/qiskit-aer/pull/1504#issuecomment-1095129834 that using the wheels built in CI still face this issue and this is still an issue until an independent fix is developed.

releasenotes/notes/non-x86_ibm_cpu-493e51313ba222a6.yaml
```diff
@@ -0,0 +1,21 @@
...
+    This release will start publishing pre-compiled binaries for ppc64le and
+    s390x Linux platforms on all Python versions. However, unlike other
+    supported platforms not all of Qiskit's upstream dependencies support these
+    platforms yet. So a C/C++ compiler may be required to build and install
+    these dependencies and a simple ``pip install qiskit-aer`` with just a
+    working Python environment will not be sufficient to install Qiskit Aer.
+    Additionally, these same constraints prevent us from testing the
+    pre-compiled wheels before publishing them, so the same guarantees around
+    platform support that exist for the other platforms don't apply to these
+    platforms.
```
jakelishman, line 14: Shall we point to the ["tiered support" table](https://qiskit.org/documentation/getting_started.html#platform-support) for this too?  We approximately support s390x and ppc64le for Terra too, though I don't think we've got them in the table yet.

jakelishman (approved):
This looks sensible to me, assuming the wheels build and work (within known issues) - I don't have an immediate way to test them.

Approving now, in case we dismiss the comment on the release note.

Issue #1503, opened 2022-04-08
Statevector backend produces incorrect results when inverting isometry


AlbertJP:
*BUG TEMPLATE* 

### Informations

- **Qiskit AER version**: 0.10.3
- **Operating system**: Ubuntu 20.04.3 in WSL (Windows 10 21H1)

### What is the current behavior?

A circuit containing an inverted isometry produces incorrect results on the Aer statevector backend. The BasicAer "qasm" simulator does produce the expected result.

### Steps to reproduce the problem

This script shows the same circuit on two different simulators:
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

from qiskit import QuantumCircuit, QuantumRegister, ClassicalRegister, transpile, BasicAer
from qiskit.providers.aer import AerSimulator
import numpy as np

qc=QuantumCircuit(1,1)
vector=[0.2,0.1]
vector_circuit = QuantumCircuit(1)
vector_circuit.isometry(vector / np.linalg.norm(vector), [0], None)
vector_circuit = vector_circuit.inverse() #inverse gives adjoint!
qc.append(vector_circuit, [0])
qc.measure(0,0)

shots=2048

print('Aer statevector')
backend=AerSimulator(method='statevector')
t_circ = transpile(qc,backend, optimization_level = 2)
print(t_circ)
job = backend.run(t_circ, shots=shots)
result = job.result().get_counts()
print(result)

print('BasicAer qasm')
backend=BasicAer.get_backend('qasm_simulator')
t_circ_qasm = transpile(qc,backend, optimization_level = 2)
print(t_circ_qasm)
job = backend.run(t_circ_qasm, shots=shots)
result = job.result().get_counts()
print(result)
```

The output on my machine is:
```
Aer statevector
     ┌─────────────┐┌─┐
  q: ┤ Multiplexer ├┤M├
     └─────────────┘└╥┘
c: 1/════════════════╩═
                     0
{'0': 2048}
BasicAer qasm
     ┌──────────────────┐┌─┐
  q: ┤ U3(0.9273,-π,-π) ├┤M├
     └──────────────────┘└╥┘
c: 1/═════════════════════╩═
                          0
{'0': 1636, '1': 412}
```
It can be seen, that the Aer statevector simulator always returns 0 when measuring the state after the "multiplexer" gate.

### What is the expected behavior?

Both simulators produce the same result.

hhorii:
#1508 fixed this issue.

[...]

PR #1508, opened 2022-04-12
Apply multiplexer to statevector even if no control qubit exists

hhorii:
### Summary

Apply multiplexer to statevector even if no control qubit exists.

### Details and comments

Currently, multiplexer needs control qubits to apply matrices to target qubits.
However, as reported in #1503, control-qubits may not exist.
This fix allows multiplexer to apply updates without control qubits.

### DIFF
--- a/src/framework/operations.hpp
+++ b/src/framework/operations.hpp
@@ -491,6 +491,9 @@ inline Op make_multiplexer(const reg_t &qubits,
   if (1ULL << num_controls != num_mats) {
     throw std::invalid_argument("invalid number of multiplexer matrices.");
   }
+  if (num_controls == 0) { // mats.size() must be 1
+    return make_unitary(qubits, mats[0]);
+  }
   // Check number of targets and controls matches qubits
   if (num_controls + num_targets != qubits.size()) {
     throw std::invalid_argument("multiplexer qubits don't match parameters.");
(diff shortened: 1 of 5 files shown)


src/framework/operations.hpp
```diff
@@ -491,6 +491,9 @@ inline Op make_multiplexer(const reg_t &qubits,
   if (1ULL << num_controls != num_mats) {
     throw std::invalid_argument("invalid number of multiplexer matrices.");
   }
+  if (num_controls == 0 && mats.size() == 1) {
+    return make_unitary(qubits, mats[0]);
+  }
```
jakelishman, line 496: I'm not sure I understand what's special about there only being a single matrix here?  It feels like either the existing loop should be updated to work with `num_controls=0`, or this return value should have some special logic to multiplex several matrices without controls.  (Updating the regular loop seems like the more likely choice.)

But it may just be that I don't fully understand what the code is doing.

hhorii: if `num_controls` is `0`, a size of matrices must be `1` because `num_controls` is defined with `std::log2(num_mats)`.

I simplified the guard condition to enter this special case.

jakelishman (approved):
This looks sensible enough - before, I didn't realise that `multiplexer` was a gate that Terra provides with a specific meaning, so I thought we were missing something.

PR #1513, opened 2022-04-26
Reduce dependency to ClassicalRegister from AER::QV::QubitVector

hhorii:
### Summary

Reduce dependency to `ClassicalRegister` to construct `AER::QV::QubitVector` objects.

### Details and comments

`qubitvector.hpp` is not self-sufficient. This PR adds some include-statements for some header files to the file.
Additionally, `ClassicalRegister` becomes not referred. This odd dependency is added for batch-execution on GPU noise optimization and it is not necessary in `AER::QV::QubitVector`. This PR reduces this dependency by templating the class.

With this change, the following code will be compiled successfully.

```
#include <iostream>
#include "simulators/statevector/qubitvector.hpp"

int main() {
    AER::QV::QubitVector<double> qv(21);
    qv.initialize();
    std::cout << qv[0] << std::endl;
    return 1;
}
```

### DIFF
--- a/src/simulators/state_chunk.hpp
+++ b/src/simulators/state_chunk.hpp
@@ -944,7 +944,7 @@ void StateChunk<state_t>::apply_ops_multi_shots(InputIterator first, InputIterat
 
     //collect measured bits and copy memory
     for(i=0;i<n_shots;i++){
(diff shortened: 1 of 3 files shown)


doichanj (approved):
I think the change is reasonable

jakelishman (approved):
Looks fine to me.

PR #1515, opened 2022-04-28
Optimized MPI simulator

doichanj:
### Summary
This PR is the optimization for MPI simulator.

### Details and comments
A cache blocking transpiler puts some swap operators to reorder qubits,
and some swaps require data exchange between processes for MPI simulator.
Previously each swap operator is applied independently, and this PR merged some swaps to decrease data exchange.

By using this PR, scalability will be improved.
image
This graph shows weak scaling of Quantum Volume simulation, 30, 31, 32 and 33 qubits for 1, 2, 4 and 8 nodes (30 qubits per node) on IBM Power System AC922 (6x Tesla V100 GPUs). QW20 is reported in the paper (https://arxiv.org/pdf/2102.02957.pdf) , Base is the latest Qiskit Aer without this PR and Optimized is with this PR.

### DIFF
--- a/src/simulators/state_chunk.hpp
+++ b/src/simulators/state_chunk.hpp
@@ -84,6 +84,7 @@ public:
     distributed_procs_ = 1;
     distributed_rank_ = 0;
     distributed_group_ = 0;
+    distributed_proc_bits_ = 0;
 
     chunk_omp_parallel_ = false;
     thrust_optimization_ = false;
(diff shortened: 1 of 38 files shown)


src/simulators/state_chunk.hpp
```diff
@@ -386,6 +388,10 @@ class StateChunk : public State<state_t> {
 
   reg_t qubit_map_;             //qubit map to restore swapped qubits
 
+  bool multi_chunk_swap_enable_ = true;     //enable multi-chunk swaps
+  uint_t chunk_swap_buffer_qubits_ = 15;    //maximum buffer size in qubits for chunk swap
+  uint_t max_multi_swap_;                 //maximum swaps can be applied at a time, calculated by chunk_swap_buffer_bits_
```
hhorii, line 393: These values should be included in metadata of a result if MPI works.

hhorii:
I added a minor comment about metadata, but basically all changes look good to me.

hhorii (approved):
LGTM

PR #1518, opened 2022-05-03
Add mergify configuration

jakelishman:
### Summary

This adds a configuration similar to what already exists on Terra that
causes the mergify bot to handle merging automatically on our behalf, if
a PR passes all its status checks and has the `automerge` label applied.
Procedure on Aer has usually been to group stable backports rather than
doing them one at a time, so this configuration does not add a backport
rule (unlike Terra's version).

GitHub Actions (what Aer uses for CI) tests appear as separate status
check entries, so we have to manually specify all of these to get the
desired behaviour.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,22 +1,6 @@
-queue_reules:
+queue_rules:
   - name: automerge
-    conditions:
-      - check-success=lint
-      - check-success=sdist
-      - check-success=tests_linux
-      - check-success=tests_macos
-      - check-success=tests_windows
-      - check-success=tests-json-input
-      - check-success=standalone
-      - check-success=mpi_standalone
-      - check-success=wheel-arm64
-      - check-success=wheel
-      - check-success=docs
-      - check-success=tutorials
-      - or:
-        - check-success=license/cla
-        - check-neutral=license/cla
-        - check-skipped=license/cla
+    conditions: []  # No additional rules, because branch-protection does them.
 
 pull_request_rules:
   - name: automatic merge on CI success and review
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,47 @@
...
+      - check-success=tests-json-input
+      - check-success=standalone
+      - check-success=mpi_standalone
+      - check-success=wheel-arm64
+      - check-success=wheel
+      - check-success=docs
+      - check-success=tutorials
+      - or:
+        - check-success=license/cla
+        - check-neutral=license/cla
```
mtreinish, line 42: Is this one ok? Doesn't this mean cla assistant is lost at sea and hasn't asserted the commit authors have signed the cla or am I misinterpreting the "neutral" here?

mtreinish (commented):
Do we want to add the auto backport support here too?

jakelishman:
I'm just checking on a few things - after I made the PR, mergify let me check the actual configuration and the rules, so I'll make a couple of changes.  I think all the explicit checks will actually be unnecessary because the GitHub branch-protection rules should apply all of them automatically on our behalf.

.mergify.yml
jakelishman: I think the "lost" status is "Expected" rather than "Neutral".  I think "neutral" is the case that everyone is a second-time contributor, so the check never even ran.  On Terra, this set of three rules is what's applied by mergify in order to fulfill the full branch-protection rules from GitHub.

mtreinish (approved):

PR #1523, opened 2022-05-26
Support for cuQuantum 22.5.0.41

doichanj:
### Summary
This PR is update for the latest cuQuantum release.

### Details and comments
Thread safety issues in some cuStateVec APIs are fixed in the latest release.
So OpenMP parallel threads on the host is now turned on when enabling cuStateVec for multi-shots and multi-chunk parallelization.

### DIFF
--- a/src/controllers/aer_controller.hpp
+++ b/src/controllers/aer_controller.hpp
@@ -531,7 +531,7 @@ void Controller::set_config(const json_t &config) {
       }
       sim_device_ = Device::GPU;
 
-#ifdef AER_CUSTATEVEC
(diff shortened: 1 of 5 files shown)


src/controllers/aer_controller.hpp
```diff
@@ -531,7 +531,7 @@ void Controller::set_config(const json_t &config) {
       }
       sim_device_ = Device::GPU;
 
-#ifdef AER_CUSTATEVEC
+#ifdef AER_CUSTATEVEC_
```
hhorii, line 534: Why `_` was added here?

hhorii (approved):

PR #1524, opened 2022-05-31
Initial input of AerState

hhorii:
### Summary

A new API for shared library and interactive simulator

### Details and comments

This PR adds a new pybind for users to allow interactive access to quantum state.

```
state = AerState()
state.configure('method', 'statevector')
state.configure('device', 'CPU')

state.allocate_qubits(10)

state.apply_unitary([0, 1], random_unitary(2**2))
state.apply_unitary([1, 2], random_unitary(2**2))
state.apply_unitary([2, 3], random_unitary(2**2))
state.apply_unitary([3, 4], random_unitary(2**2))
state.apply_unitary([4, 5], random_unitary(2**2))
state.apply_unitary([5, 6], random_unitary(2**2))
state.apply_unitary([6, 7], random_unitary(2**2))
state.apply_unitary([7, 8], random_unitary(2**2))
state.apply_unitary([8, 9], random_unitary(2**2))
state.apply_unitary([9, 0], random_unitary(2**2))

print(state.sample_measure(10))
# {669: 1, 941: 1, 61: 1, 876: 1, 172: 1, 535: 1, 428: 1, 615: 1, 44: 1, 1014: 1}

state.clear_qubits()
```

All of the methods and devices will be available and `ndarray` of statevector and densitymatrix will be returned with zero-copy. Because the simulator can skip qobj construction and parsing, performance for low-qubits becomes better.

image
(Note that currently gate fusion is not supported and performance with 16 or more qubits is worse than `AerSimulator`).

In addition, standalone build will produce a shared library `libaer.so`, which wraps `AerState` for C and C++ programs.

- [x] Buffer gates and fuse buffered gates
- [x] Support all the methods
- [x] Support GPUs
- [x] Add tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -392,50 +392,52 @@ else() # Standalone build
 	# This will add the linter as part of the compiling build target
 	#add_linter(qasm_simulator)
 
-	set(AER_RUNTIME_SOURCES "${PROJECT_SOURCE_DIR}/contrib/runtime/aer_runtime.cpp")
-	if(CUDA_FOUND AND AER_THRUST_BACKEND STREQUAL "CUDA")
-		set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES LANGUAGE CUDA)
-		set_source_files_properties(${AER_RUNTIME_SOURCES} PROPERTIES LANGUAGE CUDA)
-		set_source_files_properties(${AER_RUNTIME_SOURCES} PROPERTIES COMPILE_FLAGS "${CUDA_NVCC_FLAGS}")
-		if(DEFINED SIMD_FLAGS_LIST)
-			nvcc_add_compiler_options_list("${SIMD_FLAGS_LIST}" SIMD_FLAGS)
-			set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES COMPILE_FLAGS "${CUDA_NVCC_FLAGS} ${SIMD_FLAGS}")
-		endif()
-		add_executable(aer ${AER_RUNTIME_SOURCES} ${SIMD_SOURCE_FILE})
-		target_link_libraries(aer ${AER_LIBRARIES})
-		string(STRIP ${AER_COMPILER_FLAGS} AER_COMPILER_FLAGS_STRIPPED)
-		nvcc_add_compiler_options(${AER_COMPILER_FLAGS_STRIPPED} AER_COMPILER_FLAGS_OUT)
+	if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64" OR CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "amd64")
+		set(AER_RUNTIME_SOURCES "${PROJECT_SOURCE_DIR}/contrib/runtime/aer_runtime.cpp")
+		if(CUDA_FOUND AND AER_THRUST_BACKEND STREQUAL "CUDA")
+			set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES LANGUAGE CUDA)
+			set_source_files_properties(${AER_RUNTIME_SOURCES} PROPERTIES LANGUAGE CUDA)
+			set_source_files_properties(${AER_RUNTIME_SOURCES} PROPERTIES COMPILE_FLAGS "${CUDA_NVCC_FLAGS}")
+			if(DEFINED SIMD_FLAGS_LIST)
+				nvcc_add_compiler_options_list("${SIMD_FLAGS_LIST}" SIMD_FLAGS)
+				set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES COMPILE_FLAGS "${CUDA_NVCC_FLAGS} ${SIMD_FLAGS}")
+			endif()
+			add_executable(aer ${AER_RUNTIME_SOURCES} ${SIMD_SOURCE_FILE})
+			target_link_libraries(aer ${AER_LIBRARIES})
+			string(STRIP ${AER_COMPILER_FLAGS} AER_COMPILER_FLAGS_STRIPPED)
+			nvcc_add_compiler_options(${AER_COMPILER_FLAGS_STRIPPED} AER_COMPILER_FLAGS_OUT)
 
-		set_target_properties(aer PROPERTIES
-			LINKER_LANGUAGE CXX
-			CXX_STANDARD 14
-			COMPILE_FLAGS ${AER_COMPILER_FLAGS_OUT}
-			LINK_FLAGS ${AER_LINKER_FLAGS}
-			RUNTIME_OUTPUT_DIRECTORY_DEBUG Debug
-			RUNTIME_OUTPUT_DIRECTORY_RELEASE Release)
-	else()
+			set_target_properties(aer PROPERTIES
+				LINKER_LANGUAGE CXX
+				CXX_STANDARD 14
+				COMPILE_FLAGS ${AER_COMPILER_FLAGS_OUT}
+				LINK_FLAGS ${AER_LINKER_FLAGS}
+				RUNTIME_OUTPUT_DIRECTORY_DEBUG Debug
+				RUNTIME_OUTPUT_DIRECTORY_RELEASE Release)
+		else()
 			string(REPLACE ";" " " SIMD_FLAGS "${SIMD_FLAGS_LIST}")
-		set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES COMPILE_FLAGS "${SIMD_FLAGS}")
-		add_library(aer SHARED ${AER_RUNTIME_SOURCES} ${SIMD_SOURCE_FILE})
-		target_link_libraries(aer PRIVATE ${AER_LIBRARIES})
-		set_target_properties(aer PROPERTIES
-			LINKER_LANGUAGE CXX
-			CXX_STANDARD 14
-			COMPILE_FLAGS ${AER_COMPILER_FLAGS}
-			LINK_FLAGS ${AER_LINKER_FLAGS}
-			RUNTIME_OUTPUT_DIRECTORY_DEBUG Debug
-			RUNTIME_OUTPUT_DIRECTORY_RELEASE Release)
-	endif()
-	target_include_directories(aer
-		PRIVATE ${AER_SIMULATOR_CPP_SRC_DIR}
-		PRIVATE ${AER_SIMULATOR_CPP_EXTERNAL_LIBS})
-	target_compile_definitions(aer
-		PRIVATE ${AER_COMPILER_DEFINITIONS})
-	if(WIN32 AND NOT AER_BLAS_LIB_PATH)
-		add_custom_command(TARGET aer POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different
-				${BACKEND_REDIST_DEPS}
-				$<TARGET_FILE_DIR:aer>)
-		install(FILES ${BACKEND_REDIST_DEPS} DESTINATION bin)
+			set_source_files_properties(${SIMD_SOURCE_FILE} PROPERTIES COMPILE_FLAGS "${SIMD_FLAGS}")
+			add_library(aer SHARED ${AER_RUNTIME_SOURCES} ${SIMD_SOURCE_FILE})
+			target_link_libraries(aer PRIVATE ${AER_LIBRARIES})
+			set_target_properties(aer PROPERTIES
+				LINKER_LANGUAGE CXX
+				CXX_STANDARD 14
+				COMPILE_FLAGS ${AER_COMPILER_FLAGS}
+				LINK_FLAGS ${AER_LINKER_FLAGS}
+				RUNTIME_OUTPUT_DIRECTORY_DEBUG Debug
+				RUNTIME_OUTPUT_DIRECTORY_RELEASE Release)
+		endif()
+		target_include_directories(aer
+			PRIVATE ${AER_SIMULATOR_CPP_SRC_DIR}
+			PRIVATE ${AER_SIMULATOR_CPP_EXTERNAL_LIBS})
+		target_compile_definitions(aer
+			PRIVATE ${AER_COMPILER_DEFINITIONS})
+		if(WIN32 AND NOT AER_BLAS_LIB_PATH)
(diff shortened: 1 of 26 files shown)


hhorii:
I would like to ask
- @ikkoham to review `AerStatevector`,
- @doichanj to review state_controller.hpp and refactoring of state.hpp, state_chunk.hpp, and experiment_result.hpp
- @mtreinish or @jakelishman to review a new sub-package `qiskit.providers.aer.quantum_info.states` and other packaging issues.

src/controllers/state_controller.hpp
```diff
@@ -0,0 +1,1101 @@
...
+    parallel_state_update_ = omp_get_max_threads();
+  }
+#endif
+  state_->set_config(configs_);
+  state_->set_distribution(1);
+  state_->set_max_matrix_qubits(max_gate_qubits_);
+  state_->set_parallelization(parallel_state_update_);
+
+  state_->initialize_qreg(num_of_qubits_);
+  state_->initialize_creg(num_of_qubits_, num_of_qubits_);
```
doichanj, line 592: I think number of classical registers and memories sometimes maybe larger than number of qubits. These should be modified by option.

ikkoham (commented):
When data is a `Statevector` or `ndarray` (`list[complex]`), is it possible to send information to the `aer_state` side using `set_statevector`?
It seems that basic operations (addition, scalar multiplication, tensor product, etc.) use the methods of the parent class, so the methods of `AerStatevector` cannot be used in this case? (i.e. after basis operations, it can be simply Statevector.) Is it difficult to support basic operations in `AerStatevector`?

hhorii:
@ikkoham thank you for your comments. Initialization of `AerStatevector` with `ndarray` is now supported with b285e6d.

CMakeLists.txt
```diff
@@ -26,14 +26,14 @@ if(NOT DEFINED AER_THRUST_BACKEND AND DEFINED ENV{AER_THRUST_BACKEND})
...
+	set(CUDA_FOUND TRUE)
+	#include(FindCUDA) # for cuda_select_nvcc_arch_flags, CUDA_FOUND
+	include(FindCUDA/select_compute_arch)
+	enable_language(CUDA)
 else()
-  # idiosyncrasy of CMake that it still creates a reference to this
-  set(CMAKE_CUDA_COMPILE_WHOLE_COMPILATION "")
+	# idiosyncrasy of CMake that it still creates a reference to this
+	set(CMAKE_CUDA_COMPILE_WHOLE_COMPILATION "")
 endif()
```
mtreinish, line 37: Here and most of the changes below seem like unrelated whitespace changes. It'd probably be better to remove this from this already large PR to make it easier for people to bisect and review

CMakeLists.txt
```diff
@@ -391,6 +391,55 @@ else() # Standalone build
...
+
+		set_target_properties(aer PROPERTIES
+			LINKER_LANGUAGE CXX
+			CXX_STANDARD 14
+			COMPILE_FLAGS ${AER_COMPILER_FLAGS_OUT}
+			LINK_FLAGS ${AER_LINKER_FLAGS}
+			RUNTIME_OUTPUT_DIRECTORY_DEBUG Debug
+			RUNTIME_OUTPUT_DIRECTORY_RELEASE Release)
+	else()
+			string(REPLACE ";" " " SIMD_FLAGS "${SIMD_FLAGS_LIST}")
```
mtreinish, line 417: Why the extra indent here?

qiskit/providers/aer/quantum_info/states/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+=============================================
+State (:mod:`qiskit.providers.aer.quantum_info.state`)
+=============================================
+"""
+
+from .aer_statevector import AerStatevector
```
mtreinish, line 19: You don't want to re-export `AerState` here either?

qiskit/providers/aer/quantum_info/states/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+=============================================
+State (:mod:`qiskit.providers.aer.quantum_info.state`)
```
mtreinish, line 15: I think you're missing a file for this path to work. There should be a `qiskit/providers/aer/quantum_info/__init__.py` too to make this subpackage importable. We might also want to re-export `AerStatevector` and `AerState` off that root quantum info path too.

qiskit/providers/aer/quantum_info/states/aer_state.py
```diff
@@ -0,0 +1,297 @@
...
+# that they have been altered from the originals.
+"""
+State class that handles internal C++ state safely
+"""
+import numpy as np
+from qiskit.providers.aer.backends.controller_wrappers import AerStateWrapper
+from ...backends.aerbackend import AerError
+
+
+class AerState:
```
mtreinish, line 20: This really could use some docs to explain how to use this with some examples.

qiskit/providers/aer/quantum_info/states/aer_state.py
```diff
@@ -0,0 +1,297 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2018, 2019, 2020, 2021, 2022
```
mtreinish, line 3: It doesn't really matter but I'd just put:

```suggestion
# (C) Copyright IBM 2022
```

You could also just do a date range if you though the full date was needed

test/terra/states/test_aer_statevector.py
```diff
@@ -0,0 +1,112 @@
...
+
+from test.terra import common
+from qiskit.providers.aer.aererror import AerError
+
+class TestAerStatevector(common.QiskitAerTestCase):
+    """AerState tests"""
+
+    def test_generate_aer_statevector(self):
+        """Test generation of Aer's StateVector"""
+        from qiskit.providers.aer.quantum_info.states import AerStatevector
```
mtreinish, line 35: Why is this a runtime import? Also in every test method

test/terra/states/test_aer_statevector.py
```diff
@@ -0,0 +1,112 @@
...
+        circ = QuantumCircuit(5)
+        state = AerStatevector(circ)
```
mtreinish, line 37: Shouldn't this have an assertion that it's an empty statevector?

test/runtime/runtime_sample.c
```diff
@@ -0,0 +1,53 @@
...
+      printf("  %08x [%lf, %lf]\n", i, creal(sv[i]), cimag(sv[i]));
+  }
+
+  aer_state_finalize(state);
+
+  free(qubits);
+  free(sv);
+
+  return 1;
+}
```
mtreinish, line 53: We should build and install this in Does this get built and run in CI?

test/terra/states/test_aer_state.py
```diff
@@ -0,0 +1,465 @@
...
+        state = AerState()
+        state.allocate_qubits(4)
+        state.initialize()
+        sv = state.move_to_ndarray()
+
+        try:
+            state.allocate_qubits(4)
+            self.fail("do not reach here")
+        except:
+            pass
```
mtreinish, line 63: This should be something like:

```suggestion
        with self.assertRaises(Exception):
            state.allocate_qubits(4)
```

and you should make the exception specific to what `AerState` actually raises as that is part of the API.

test/terra/states/test_aer_statevector.py
```diff
@@ -0,0 +1,112 @@
...
+from qiskit.circuit import QuantumCircuit
+from qiskit.quantum_info.random import random_unitary
+from qiskit.quantum_info.states import Statevector
+from qiskit.circuit.library import QuantumVolume
+from qiskit.providers.aer import AerSimulator
+
+from test.terra import common
+from qiskit.providers.aer.aererror import AerError
+
+class TestAerStatevector(common.QiskitAerTestCase):
```
mtreinish, line 30: I think it might be a good idea if we copied more from https://github.com/Qiskit/qiskit-terra/blob/main/test/python/quantum_info/states/test_statevector.py and update it to use aer statevector so we can test full compatibility (or as much as we can get)

qiskit/providers/aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,204 @@
...
+            data, aer_state = AerStatevector._from_ndarray(data, **configs)
+            self._aer_state = aer_state
+        else:
+            self._aer_state = None
+        super().__init__(data)
+        self._result = None
+        self._configs = configs
+
+    def aer(self):
+        return self._aer_state is not None
```
mtreinish, line 58: What does this do? This is a public method it should have some docstring explaining the purpose of it. But I'm not clear on the purpose.

qiskit/providers/aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,204 @@
...
+        """Initialize a statevector object.
+
+        Args:
+            data (np.array or list or Statevector or QuantumCircuit or qiskit.circuit.Instruction):
+                Data from which the statevector can be constructed. This can be either a complex
+                vector, another statevector or a ``QuantumCircuit`` or ``Instruction``
+                (``Operator`` is not supportted in the current implementation).  If the data is
+                a circuit or instruction, the statevector is constructed by assuming that all
+                qubits are initialized to the zero state.
+            configs (kwargs): configurations of ```AerState`. All the keys are handeld as string
```
mtreinish, line 37: I think you should expand on this a bit more, isn't this all the configuration for `AerState` which mirrors the run time options for `AerSimulator`. You should explicitly outline this here and link to the source of all the valid options.

qiskit/providers/aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,204 @@
...
+    def device(self):
+        return self._metadata()['device']
+
+    @property
+    def fusion_config(self):
+        return self._metadata()['fusion']
+
+    @property
+    def parallel_state_update(self):
+        return self._metadata()['parallel_state_update']
```
mtreinish, line 90: These also need docstrings so users know what they are.

Also why are these options hard coded but not other ones?

qiskit/providers/aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,204 @@
...
+"""
+import copy
+import numpy as np
+
+from qiskit.quantum_info.states import Statevector
+from qiskit.circuit import QuantumCircuit, Instruction
+from qiskit.providers.aer.quantum_info.states.aer_state import AerState
+
+
+class AerStatevector(Statevector):
```
mtreinish, line 24: Did you want to have a method to do the expectation value calculation more efficiently in the C++ code. Terra's version is partially in rust and multithreaded (and the rest is in numpy) but I feel like aer state can compute that faster directly.
Similarly what about `probabilities()`, I feel like we can do the sampling much more efficiently on the aer state.

Also some of the mutation methods like `reset()` I feel like we need overloaded versions of here otherwise won't `_aer_state` get out of sync with data?

qiskit/providers/aer/quantum_info/states/__init__.py
mtreinish, line 15: This module needs to be add to the doc tree in `docs/apidocs` with a new rst file with an `automodule` directive. Also you need to update the api ref toctree here: https://github.com/Qiskit/qiskit-aer/blob/main/docs/apidocs/aer.rst

See https://qiskit.org/documentation/maintainers_guide.html#documentation-structure for an explanation of the structure.

mtreinish (changes requested):
I just did a quick pass to start (and skipped most of the c++ code for this review). This is a very big PR and part of me would much prefer this to be split into 3 PRs, one for `AerState`, one for `AerStatevector` on top of `AerState`, and a third for the runtime C api. As they're all self contained logical changes that build on each other. This would make something like this significantly easier to review. But given the deadline for release and this being a headline feature I"ll just work through it.

I left a bunch of comments inline, but the big thing missing for me is the lack of release notes and documentation which are critical for a large new feature like this. We need to document how people are going to use this and what the new feature provides users.

qiskit/providers/aer/quantum_info/states/aer_state.py
hhorii: I do not expect that end-users use `AerState` directly. In such case, do we need docs?

test/terra/states/test_aer_statevector.py
hhorii: I wanted that tests are not failed even if pybind of AerStateWrapper fails. I will move this import statements to the head in new PRs.

test/runtime/runtime_sample.c
hhorii: I wanted to keep libaer.so (or dylib or dll) experimental and do not want to include it in CI.

qiskit/providers/aer/quantum_info/states/aer_statevector.py
hhorii: Let me keep it internal. `AerStatevector` has two modes to manage statevector: use of `AerState` and use of super class (`ndarray` of `q.qi.Statevector`). `aer()` returns `True` if `AerStatevector` is in the former state, otherwise, returns `False.`

qiskit/providers/aer/quantum_info/states/aer_state.py
mtreinish: Ok, if it's an internal class we don't need a full docstring then since it won't be published on the docs page. But having some minimal docstrings will help us as most IDEs will be able to read docstrings and present them to developers. So having some minimal explanation of the purpose of the class and some explanation for the methods would be good. You can also mention it's for internal aer use only in the docstring to make it clear

test/runtime/runtime_sample.c
mtreinish: I still think even if it's experimental we'll want to test it in CI. We want to make sure a proposed change doesn't break functionality. We can still say it's experimental and the api may change and there are probably lots of bugs. But at least verifying what we have works as expected is good to do.

hhorii: Sure. I will try to include this into CI in #1582.

hhorii:
This PR was divided into #1582, #1586, and #1590.

PR #1527, opened 2022-05-31
Pin macOS version in wheel jobs

mtreinish:
### Summary

The recent version of macOS 11 stopped including the necessary headers
to 10.x compatible builds. This is causing failures in CI and would also
bump our published binaries. To avoid this issue, this commit pins the
macOS version we use in the wheel jobs to the 10.15 image which will
maintain compatibility moving forward.

### Details and comments

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ["macOS-latest", "ubuntu-latest", "windows-2019"]
+        os: ["macos-10.15", "ubuntu-latest", "windows-2019"]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python 3.7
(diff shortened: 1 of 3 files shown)


jakelishman (approved):
This is fine for probably the next Aer release, but [the macOS 10.15 image is deprecated as of today](https://github.com/actions/virtual-environments/issues/5583), and due to be completely removed by the end of August.  We may need to consider dropping support for macOS 10.15 at that point, if we can't emulate it, or vendor in the headers from somewhere else.

jakelishman:
Actually, for the failing build right now, I think we need to tell cibuildwheel to target a particular macOS version as well?  I can't entirely remember.

mtreinish:
cibuildwheel should be implicitly setting: `MACOSX_DEPLOYMENT_TARGET=10.9` at least I think it does (it has in the past). I can add it to explicitly set the env variable in the build though it probably doesn't hurt to do that anyway

mtreinish:
Superseded by #1536

PR #1536, opened 2022-06-09
Make CI to build OpenMP from source through conan for mac

hhorii:
### Summary

llvm-openmp in conan is built for macosx_11

### Details and comments

The reason why recent CI encounters errors in build of mac is that generated wheel file is for macosx_11 though OS target is 10.9.
Binaries of llvm-openmp in conan were for macosx_11 and then skbuild generates a wheel for 11.

With this PR, CI builds llvm-openmp from source in conan. The built library files are for 10.9 and then skbuild generates a wheel for 10.9.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,6 +20,8 @@ jobs:
         run: |
           python -m pip install cibuildwheel==2.2.2
       - name: Build wheels
+        env:
+          AER_CMAKE_OPENMP_BUILD: build
         run: python -m cibuildwheel --output-dir wheelhouse
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -20,6 +20,8 @@ jobs:
         run: |
           python -m pip install cibuildwheel==2.2.2
       - name: Build wheels
+        env:
+          AER_CMAKE_OPENMP_BUILD: 1
```
mtreinish, line 24: Do you think it would be better to set this via `CIBW_ENVIRONMENT`: https://cibuildwheel.readthedocs.io/en/stable/options/#environment so it gets set in the build environment directly? I think this is working because macOS builds don't use any virtualization and run natively on the host but I feel like leveraging the CIBW env variable to set this is probably more future proof.

So probably something like:

```suggestion
        env:
         CIBW_ENVIRONMENT: AER_CMAKE_OPENMP_BUILD=1
```

mtreinish (approved):
LGTM as it's working. One inline suggestion but since this is working it's not worth blocking over.

PR #1538, opened 2022-06-14
fix issues from mismatch of binding specs in assemble and backend of aer

### LINKED ISSUE #1530: Bug or spec regarding parameter_binds
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

Passing different parameters to multiple circuits will result in an error.

```python
from qiskit.circuit import Parameter, QuantumCircuit

a = Parameter("a")
b = Parameter("b")
qc1 = QuantumCircuit(1)
qc1.rx(a, 0)
qc2 = QuantumCircuit(1)
qc2.rx(b, 0)

AerSimulator().run([qc1, qc2], parameter_binds=[{a: [1]}, {b: [2]}])
```
```
QiskitError: 'Mismatch between run_config.parameter_binds and all circuit parameters. Parameter binds: [dict_keys([Parameter(a), Parameter(b)])] Circuit parameters: [ParameterView([Parameter(a)]), ParameterView([Parameter(b)])]'
```

Binds correctly.

### Suggested solutions

Frankly, I don't know if this is a bug or a specification. However, from reading the documentation, it seems to me that the validation is too strict (bug). If it's a spec, I'd like to know how to bind it correctly (and should be documented.).

hhorii:
### Summary

Fix #1530

### Details and comments

To use Aer's parameterized qobj, all `Parameter` must be resolved.
Originally, parameters were resolved with `assemble` by setting all pairs of parameters specified for circuits.
However, `assemble` will raise an error if unrelated parameters are specified.

This PR changes parameter resolving outside of `assemble` by calling `bind_parameters` directly.

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -384,13 +384,18 @@ class AerBackend(Backend, ABC):
             if parameter_binds:
                 # Handle parameter binding
                 parameterizations = self._convert_binds(circuits, parameter_binds)
-                circuits = [circuit.bind_parameters({param: 1 for param in circuit.parameters})
-                            for circuit in circuits]
-
-                qobj = assemble(
-                    circuits,
-                    backend=self,
-                    parameterizations=parameterizations)
+                qobj = None
+                for circuit in circuits:
+                    assemble_bind = {param: 1 for param in circuit.parameters}
+                    qobj_tmp = assemble(
+                        [circuit],
+                        backend=self,
+                        parameter_binds=[assemble_bind],
+                        parameterizations=parameterizations)
+                    if qobj:
+                        qobj.experiments.append(qobj_tmp.experiments[0])
+                    else:
+                        qobj = qobj_tmp
             else:
                 qobj = assemble(circuits, backend=self)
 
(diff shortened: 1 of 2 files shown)


qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -384,13 +384,12 @@ def _assemble(self, circuits, parameter_binds=None, **run_options):
             if parameter_binds:
                 # Handle parameter binding
                 parameterizations = self._convert_binds(circuits, parameter_binds)
-                assemble_binds = []
-                assemble_binds.append({param: 1 for bind in parameter_binds for param in bind})
+                circuits = [circuit.bind_parameters({param: 1 for param in circuit.parameters})
+                            for circuit in circuits]
```
mtreinish, line 388: I'm worried about doing this, `bind_parameters()` adds a deepcopy for each circuit to bind the dummy value. The whole parameter binding path is setup to avoid doing the bind on the circuit object so we don't need to copy it. That being said I think there is a deepcopy already in `assemble()`: https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/compiler/assembler.py#L553-L611 so there might not be any extra overhead here. But we should test it to confirm either way.

That being said I think we're seeing more evidence that aer should rewrite `assemble()` and not leverage terra's anymore. The serialization aer needs is more specific than what terra does now (this also gives us more flexibility once we have #1524 )

mtreinish: The potential issue with that is doing it all at once is run in multiple processes. But I don't think that's a big loss so this is probably a fine approach.

mtreinish (approved):

PR #1540, opened 2022-06-15
Binary distribution for cuQuantum support

doichanj:
### Summary
Add binary distribution for cuQuantum support

### Details and comments
Added binary build workflow in `.github/workflow/deploy.yml` for cuQuantum support.

User will be able to install by
`pip install qiskit-aer-cuQuantum `


.github/workflows/deploy.yml
```diff
@@ -183,3 +183,35 @@ jobs:
...
+          python-version: '3.7'
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
+        if: runner.os == 'Windows'
+      - name: Install cibuildwheel
+        run: |
+          python -m pip install cibuildwheel==2.2.2
+      - name: Build wheels
+        env:
+          CIBW_BEFORE_ALL: "yum install -y yum-utils wget && wget -q https://developer.download.nvidia.com/compute/cuda/11.7.0/local_installers/cuda-repo-rhel7-11-7-local-11.7.0_515.43.04-1.x86_64.rpm && rpm -i cuda-repo-rhel7-11-7-local-11.7.0_515.43.04-1.x86_64.rpm && yum clean all && yum -y install nvidia-driver-latest-dkms cuda && yum -y install cuda-drivers && yum -y install openblas-devel && yum-config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel7/x86_64/cuda-rhel7.repo && yum clean all && yum -y install cuquantum cuquantum-devel cuquantum-doc"
```
mtreinish, line 203: The nvidia drivers are required to build here? There aren't any GPUs present on the CI node so I typically don't bother installing the drivers.

mtreinish (commented):
Overall this looks like a reasonable first step. I had one inline comment about the packages you're installing.

Other than that before we merge this it would be good to add a test commit on the PR branch which enables this job to be triggered on a PR. The two ways to do this are to either delete all the twine lines from this file and change the trigger condition at the top of the file or to copy the new job to `wheels.yml` without the twine steps. Either way the goal is to have this job run in CI to test it before we merge. We can then revert that commit after we confirm its working.

.github/workflows/deploy.yml
doichanj: I removed CUDA driver installation and I confirm we do not need CUDA drivers for building binary.

doichanj:
https://github.com/doichanj/qiskit-aer/runs/7084200066?check_suite_focus=true

doichanj:
I think large file size issue is a common problem for CUDA 11, but cuQuantum required CUDA 11 or above.

https://discuss.python.org/t/what-to-do-about-gpus-and-the-built-distributions-that-support-them/7125

mtreinish:
Since we're hitting issues with pypi file size, maybe we can use: https://github.com/softprops/action-gh-release to add the wheels as a release artifacts on github for the time being.

.github/workflows/deploy.yml
```diff
@@ -183,3 +183,35 @@ jobs:
...
+      - uses: actions/upload-artifact@v2
+        with:
+          path: ./wheelhouse/*.whl
+      - name: Publish Wheels
+        env:
+          TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
+          TWINE_USERNAME: qiskit
+        run : |
+          pip install -U twine
+          twine upload wheelhouse/*
```
mtreinish, line 217:
```suggestion
      - name: Publish release artifacts to github release page
        uses: softprops/action-gh-release@v1
        with:
           files: ./wheelhouse/*.whl
```

leofang:
@doichanj we're preparing detailed instruction for cuQuantum downstream to avoid hitting the file size limit, please stay tuned 🙂

Generally, if it's only ~300 MB it's possible to request size limit bump by submitting a ticket to https://github.com/pypa/pypi-support/issues, but we do not consider it a nice experience for users and developers. We'd rather advocate for the following changes:
- use dynamic instead of static linking
  - cuQuantum already have wheels such as `custatevec-cu11` and `cutensornet-cu11` living on PyPI
  - NVIDIA CUDA wheels are also up (`nvidia-*-cu11` for CUDA 11 and `nvidia-*-cu12` for CUDA 12, ex: `nvidia-cublas-cu11`)
- use `auditwheel repair --exclude` (https://github.com/pypa/auditwheel/pull/368) to whitelist certain shared libraries
  - it's possible to not do repair at all, if you already build the wheel in a manylinux container and meet certain conditions

However there are some caveats in terms of the build system, we'll try to explain them in the upcoming instruction. This is how our `cuquantum-python-cuXX` wheels can be kept under very comfortable size.

leofang:
> @doichanj we're preparing detailed instruction for cuQuantum downstream to avoid hitting the file size limit, please stay tuned 🙂

FYI: https://github.com/NVIDIA/cuQuantum/tree/main/extra/demo_build_with_wheels, let us know if it helps! 🙂

leofang:
Hi @hhorii, would it be possible to keep this PR opened? Even with our cuquantum Appliance container, we can't cover all use cases, and according to our understanding (@tlubowe knows far better than I do 🙂) the majority of users still want a simple `pip-install` based solution, without container. If you don't have bandwidth for this PR, maybe @tlubowe can find someone on our side to help?

PR #1541, opened 2022-06-16
Add qiskit-neko integration test job

mtreinish:
### Summary

This commit adds a new CI job to run the Qiskit Neko test suite. Qiskit
Neko is an integration suite that's designed to validate backwards
compatibility for end user workflows and between different components in
Qiskit. This test job will hopefully eventually replace the role that
the tutorials CI job is playing for backwards compatibility testing.
Right now the test suite is still fairly small but it covers some
examples of terra, aer, experiments, nature, and machine learning. The
test suite will continue to grow and improve our coverage for testing.

This test job will install qiskit-neko and aer from main, while
the other qiskit components (nature, aer, machine learning, and
experiments) will be installed from release. This will enable us to
check that a proposed PR does not regress compatibility on an API or an
integration point with another qiskit project and ensure that
aer doesn't break CI for another project using neko.

### Details and comments

### DIFF
--- a/.github/workflows/neko.yml
+++ b/.github/workflows/neko.yml
@@ -2,7 +2,7 @@ name: Qiskit Neko Integration Tests
 on:
   push:
   pull_request:
-    branches: ['main']
+    branches: ['main', 'stable/*']
 concurrency:
   group: ${{ github.repository }}-${{ github.ref }}-${{ github.head_ref }}-${{ github.workflow }}
   cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


.github/workflows/neko.yml
```diff
@@ -0,0 +1,16 @@
+name: Qiskit Neko Integration Tests
+on:
+  push:
+  pull_request:
+    branches: ['main']
```
jakelishman, line 5: Same question from Terra - do we want to run this on PRs to stable branches as well?

mtreinish: Yeah, this should include stable too

```suggestion
  push:
  pull_request:
    branches: ['main', 'stable/*']
```

jakelishman (approved):

PR #1550, opened 2022-07-05
Use qiskit.providers.fake_provider instead of qiskit.test.mock and correct rccx and rcccx usages in tests

hhorii:
### Summary

terra 0.22 started deprecation of qiskit.test.mock
existing wrong usages of rccx and rcccx are now in error

### Details and comments

This PR changes some tests to use `qiskit.providers.fake_provider` instead of `qiskit.test.mock`.
Existing tests specify float numbers as labels of rccx and rcccx gates. 0.22 raises errors in this situation.

### DIFF
--- a/test/terra/backends/aer_simulator/test_standard_gates.py
+++ b/test/terra/backends/aer_simulator/test_standard_gates.py
@@ -84,8 +84,8 @@ MC_GATES = [
     (C3XGate, 0, False),
     (C4XGate, 0, False),
     # Parameterized Gates
-    (RC3XGate, 1, False),
-    (RCCXGate, 1, False),
+    (RC3XGate, 0, False),
+    (RCCXGate, 0, False),
     (MCXGate, 0, True),
     (MCPhaseGate, 1, True),
     (MCXGrayCode, 0, True),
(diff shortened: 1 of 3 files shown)


HuangJunye (commented):
LGTM. Thanks for updating. Just want to double check, are these the only usage of qiskit.test.mock module?

test/terra/backends/aer_simulator/test_standard_gates.py
```diff
@@ -84,8 +84,8 @@
     (C3XGate, 0, False),
     (C4XGate, 0, False),
     # Parameterized Gates
-    (RC3XGate, 1, False),
-    (RCCXGate, 1, False),
+    (RC3XGate, 0, False),
+    (RCCXGate, 0, False),
```
jakelishman, line 88: What's this change about?

jakelishman (commented):
Ah, nice - I didn't think to check for `from qiskit.test import mock` when I tried this!

test/terra/backends/aer_simulator/test_standard_gates.py
hhorii: The second argument means a number of parameters. This was a bug in previous code because [RC3XGate](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/circuit/library/standard_gates/x.py#L685) and [RCCXGate](https://github.com/Qiskit/qiskit-terra/blob/f134aeb7a9aa6f7381de5a9c9de0fb662fea3c5f/qiskit/circuit/library/standard_gates/x.py#L404) are 3-controlled Toffoli gate and normal Toffoli gate in Terra. In the previous code, a float number was specified as `label` but this becomes a fail since [recent change](https://github.com/Qiskit/qiskit-terra/commit/ef997f9ba2c3da0b1018394cd1a2649bdeb6a183) in Terra.

PR #1555, opened 2022-07-14
Disable BLAS if many threads run

hhorii:
### Summary

Work around BLAS fails when many threads run.

### Details and comments

OpenBLAS has issues when many threads attempt to allocate memory in BLAS. 

https://github.com/xianyi/OpenBLAS/issues/1882

This PR disables use of BLAS when a number of threads are more than 50.
(My 76-core machine fails when OMP_NUM_THREADS is more than 56).


jakelishman (commented):
If this fixes the issue, then it seems fine to include.  It might be good to have a comment about the `std::min` explaining why it's there.

Issue #1557, opened 2022-07-19
Result error with mpi


anavasca:
### Informations

- **Qiskit Aer version**:  0.11.0
- **Python version**: 3.8.10
- **Operating system**: Ubuntu 20.04
- **mpirun (Open MPI)**: 4.0.3

### What is the current behavior?

from qiskit import QuantumCircuit, transpile, Aer, IBMQ
from qiskit.tools.jupyter import *
from qiskit.visualization import *
from qiskit.providers.aer import QasmSimulator
import qiskit
from qiskit import QuantumCircuit, execute, ClassicalRegister, QuantumRegister, transpile
from qiskit.providers.aer import AerSimulator
import sys
import numpy as np
from qiskit import QuantumCircuit
from qiskit import Aer, transpile
from qiskit.tools.visualization import plot_histogram, plot_state_city
import qiskit.quantum_info as qi
from qiskit.opflow import I, X, Y, Z
from qiskit.quantum_info.operators import Operator, Pauli
from qiskit.opflow.primitive_ops import PauliSumOp
from qiskit.opflow.evolutions import TrotterizationBase
from qiskit.circuit.library import PauliEvolutionGate
from qiskit import QuantumCircuit,Aer,transpile
from qiskit.providers.aer import AerSimulator
from qiskit.circuit.library import RXGate
import qiskit
from qiskit import QuantumCircuit, execute, ClassicalRegister, QuantumRegister, transpile
from qiskit.providers.aer import AerSimulator
import sys
import numpy as np
from qiskit import QuantumCircuit
from qiskit import Aer, transpile
from qiskit.tools.visualization import plot_histogram, plot_state_city
import qiskit.quantum_info as qi

n=5

gamma = 0.2
beta = 0.2

operator = - 3/4 * (Z ^ I ^ I ^ Z ^ I) + 1/4 * (Z ^ I ^ I ^ I ^ I) - 2 * (I ^ Z ^ Z ^ I ^ I) - 7/4 * (I ^ Z ^ I ^ I ^ Z) + 5/4 * (I ^ Z ^ I ^ I ^ I) - 5/4 * (I ^ I ^ Z ^ I ^ Z) + 7/4 * (I ^ I ^ Z ^ I ^ I) - 1/4 * (I ^ I ^ I ^ Z ^ I) + 2 * (I ^ I ^ I ^ I ^ Z)
evo = PauliEvolutionGate(operator, time=-gamma)

qc=QuantumCircuit(n)

[...]

PR #1560, opened 2022-07-25
avaraging seed of cirucits among all MPI processes

doichanj:
### Summary
This is fix for issue #1557 

### Details and comments
Each MPI process had different seed numbers when `seed_simulator` is not set.
So I added MPI_Allreduce to average seed numbers for each circuit so that all the process has the same seed.


hhorii (approved):
This change looks good to me.

PR #1562, opened 2022-07-28
Improve the performance (Primitives)

ikkoham:
### Summary

If the parameters are empty lists, we don't need bind_circuits in assemble.
Related to https://github.com/Qiskit/qiskit-terra/pull/8407.


hhorii (approved):
LGTM

PR #1563, opened 2022-07-28
Fixed crash (SegFault) in apply_diagonal_matrix_avx

1tnguyen:
### Summary

Fixing intermittent Segfault crashes in `apply_diagonal_matrix_avx` when building without OpenMP.

### Details and comments

The temporary double array of size 4 needed for `_mm256_load_pd` should be allocated with `posix_memalign` on Linux platforms. It is accidentally wrapped within `_OPENMP` switch; hence causing the default `malloc` to be used in all cases when that compiler flag is not defined. This may cause Segfault when `_mm256_load_pd` is called on Linux builds. It occurs intermittently, especially in DEBUG builds (no optimization).

The need for `posix_memalign` should not depend on OpenMP. I think this bug was first introduced when we varied the size of the temporary array depending on the number of threads. I suppose when that logic was implemented, the else code path (`_OPENMP` is not defined) should have also had two paths for Windows/non-Windows but it was not.

Since the array size is now fixed, move the entire allocation logic out of the  `#if defined(_OPENMP)` block.

cc @hhorii

### DIFF
--- a/src/simulators/statevector/qv_avx2.cpp
+++ b/src/simulators/statevector/qv_avx2.cpp
@@ -1169,13 +1169,11 @@ Avx apply_diagonal_matrix_avx<double>(double* qv_data_,
 #if defined(_OPENMP)
 #pragma omp parallel if (omp_threads > 1) num_threads(omp_threads)
   {
+#endif
 #if !defined(_WIN64) && !defined(_WIN32)
   void* data = nullptr;
   posix_memalign(&data, 64, sizeof(std::complex<double>) * 2);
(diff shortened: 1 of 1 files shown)


1tnguyen:
@hhorii Do you know why some CI pipelines are failing? Are they related to this proposed change?

hhorii:
CI fail is independent. https://github.com/Qiskit/qiskit-aer/issues/1564
I'm now investigating its root cause.

This fix makes sense to me. I would like to approve it after pass the tests.

hhorii (approved):

Issue #1565, opened 2022-08-03
GAS error with mpi


anavasca:
### Informations

- **Qiskit Aer version**: 11.0
- **Python version**: 3.8
- **Operating system**: Ubuntu 22.04

### What is the current behavior?

I execute the following program with error:

import numpy as np
from qiskit import QuantumCircuit, transpile, Aer, IBMQ
from qiskit.tools.jupyter import *
from qiskit.visualization import *
from qiskit.providers.aer import QasmSimulator
from qiskit.algorithms import NumPyMinimumEigensolver
from qiskit_optimization.algorithms import GroverOptimizer, MinimumEigenOptimizer
from qiskit_optimization.problems import QuadraticProgram
from qiskit_optimization.translators import from_docplex_mp
from qiskit import BasicAer
from docplex.mp.model import Model
from qiskit_optimization import QuadraticProgram
from qiskit_optimization.translators import from_docplex_mp
from qiskit_optimization.converters import QuadraticProgramToQubo
from qiskit.providers.aer import StatevectorSimulator
from qiskit.providers.aer import AerSimulator

backend = AerSimulator(method='statevector',blocking_enable=True, blocking_qubits=2)

model = Model()

#Establecemos las variables y su tipo

x0 = model.binary_var(name="x0")
x1 = model.binary_var(name="x1")
x2 = model.binary_var(name="x2")
x3 = model.binary_var(name="x3")
x4 = model.binary_var(name="x4")

model.minimize(-(x0 + 5x1 + 3x2 + 2x3 + 2x4 - 3x0x3 - 8x1x2 - 7x1x4 - 5x2x4))

qp = from_docplex_mp(model)

cubo=QuadraticProgramToQubo().convert(qp)

grover_optimizer = GroverOptimizer(10, num_iterations=10, quantum_instance=backend)
results = grover_optimizer.solve(cubo)
print(results.prettyprint())

### Steps to reproduce the problem

when I launch the programme with the command:

!time mpirun -np 1 -host 172.31.41.211:1 python /home/ubuntu/qiskit_pruebas/scripts/Qiskit_GAS_Parallelization.py

I get the correct result:

objective function value: -7.0
variable values: x0=0.0, x1=1.0, x2=0.0, x3=1.0, x4=0.0
status: SUCCESS

however when I launch the command :

[...]

PR #1569, opened 2022-08-08
Fix for save_statevector and fix for global indexing for Thrust

doichanj:
### Summary
This is fix for issue #1565 

### Details and comments
Non-gate operations such as save_statevector are not recognized correctly in cache blocking transpiler that causes error if there are full-qubits operations.
The  save_statevector operation only returns statevector to the 1st MPI process previously. Now all the process receive statevector.

Also this PR included fix for global chunk indexing for Thrust implementations.


hhorii (approved):
LGTM. `MPI_Allgatherv` makes initialization very simple in MPI run.

PR #1572, opened 2022-08-08
set_options in get_backend

### LINKED ISSUE #1568: Aer.get_backend("aer_simulator", statevector_parallel_threshold=1) ignores the keyword arguments
### Informations

- **Qiskit Aer version**: 0.10.4 but also applies to latest main
- **Python version**: 3.10.5
- **Operating system**: Linux

The `statevector_parallel_threshold=1` is ignored.

Initialize with `Aer.get_backend("aer_simulator", statevector_parallel_threshold=1)` and run a simulation. I didn't observe any matrix ops parallelization.

### Suggested solutions

I think this is the cause. In https://github.com/Qiskit/qiskit-aer/blob/8ca389a53314cf0f5d921277006ebf1795a51697/qiskit/providers/aer/aerprovider.py#L64-L77, the `kwargs` is not passed to the `backend_cls(**opts)`. This should be a quick fix.

hhorii:
### Summary

Fix #1568 by calling `set_options()` in `Aer.get_backend()`

### Details and comments

Keyword options are ignored in original `Aer.get_backend()`. This PR adds a call of `set_options()` to set keyword arguments as backend options.

### DIFF
--- a/qiskit/providers/aer/aerprovider.py
+++ b/qiskit/providers/aer/aerprovider.py
@@ -59,7 +59,9 @@ class AerProvider(Provider):
             AerProvider._BACKENDS = backends
 
     def get_backend(self, name=None, **kwargs):
-        return super().get_backend(name=name, **kwargs)
+        backend = super().get_backend(name=name, **kwargs)
+        backend.set_options(**kwargs)
+        return backend
 
     def backends(self, name=None, filters=None, **kwargs):
         # pylint: disable=arguments-differ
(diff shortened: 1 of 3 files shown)


jakelishman (commented):
I'm not 100% sure this is something we should include: the base `Provider.get_backend` documents `kwargs` as being used for backend filtering, not configuration.  It could be argued that doing `Aer.get_backend(...).set_options(...)` is more in the spirit of the interface.

I don't know what other backends do in this respect, though.

hhorii:
@jakelishman Thank you for your comment. Right. `Aer.get_backend(...).set_options(...)` will be the best for users. In Aer, historically, we prepared several backends (qasm_simulator, statevector_simulator, ...) and it is natural that `method` is a condition of filtering when `AerSimulator` was introduced.

Issue #1567, opened 2022-08-04
Differences between AerJob and IBMQJob for failing jobs


yaelbh:
### Informations

- **Qiskit AER version**: 0.10.4
- **Operating system**: Linux

### What is the current behavior?

Suppose that a job fails. For example, it contains a circuit with gates that don't belong to the backend's set of basis gates.
The job failure looks pretty different if we run it on a real device or on a simulator:
1. `job.result()` will raise an exception for real devices, but not for simulators.
2. The job's status will be `ERROR` for real devices, and `DONE` for simulators (for simulators, the information that an error has occured is hiding in a `status` field inside the result).
3. `job.error_message()` will work only for devices, for simulators it will raise an `AttributeError` exception.

I'd like to point out that this is a real and even annoying problem. I'm writing code for qiskit-experiments. Our experiments typically run on devices, but our tests and tutorials must run on the simulator (when applicable, i.e. measurement level 2 etc.), either directly or via a fake backend. Now:
1. We don't want to fill our code with ifs or try-excepts to distinguish between job types.
2. Even if we did, we would still have an issue with the coverage of our tests. Because the tests wouldn't cover the if sections that apply only to devices.

Specifically I have a concrete bug in qiskit-experiments right now (https://github.com/Qiskit/qiskit-experiments/issues/866) that I can test and debug only on a device. When I fix it, I'll want to add a test to our test suite, that will be able to catch the bug in the future if it returns. To overcome what I've just described, here is what I intend to do (I still don't know if it will work): for the sake of the test, I'll write a subclass of `AerSimulator` and a subclass of `AerJob`; I'll tweak their behavior to mimic devices.

### Steps to reproduce the problem

Code:
```python
from qiskit import QuantumCircuit, Aer, IBMQ

provider = IBMQ.load_account()
backends = [Aer.get_backend("aer_simulator_stabilizer"), provider.backend.ibmq_lima]

circ = QuantumCircuit(1, 1)
circ.t(0)
circ.measure(0, 0)

for backend in backends:
    print("\nRunning with", backend.name())
    job = backend.run(circ)

    try:
        print("Job result:", job.result())
    except Exception as e:
        print("Exception when trying to fetch job result:", e)
        print(f"     (exception type is {type(e).__name__})")

    print("Job status:", job.status())
    try:
        print("Job's error message:", job.error_message())
    except Exception as e:
        print("Exception when trying to fetch job's error message:", e)
[...]

Issue #1551, opened 2022-07-05
`parameter_binds` behavior is different from that of Terra's documentation


t-imamichi:
### Informations

- **Qiskit Aer version**: 0.10.4
- **Python version**: 3.10.5
- **Operating system**: macOS 12.4

### What is the current behavior?

`parameter_binds` of `IBMQBackend.run` or `execute` are explained as follows.
> parameter_binds (Optional[List[Dict[[Parameter](https://qiskit.org/documentation/stubs/qiskit.circuit.Parameter.html#qiskit.circuit.Parameter), float]]]) – List of Parameter bindings over which the set of experiments will be executed. Each list element (bind) should be of the form {Parameter1: value1, Parameter2: value2, …}. All binds will be executed across all experiments; e.g., if parameter_binds is a length-n list, and there are m experiments, a total of m x n experiments will be run (one for each experiment/bind pair). 

[Terra's execute](https://qiskit.org/documentation/apidoc/execute.html)
[Terra's assemble](https://qiskit.org/documentation/stubs/qiskit.compiler.assemble.html)
[IBMQBackend.run](https://qiskit.org/documentation/stubs/qiskit.providers.ibmq.IBMQBackend.run.html)

The value part is expected to be a value for Terra and IBMQ provider.
 
But, Aer seems to expect a list of values.

Which behavior is correct?

### Steps to reproduce the problem

```python
from qiskit import QuantumCircuit, Aer, IBMQ
from qiskit.circuit import Parameter

qc = QuantumCircuit(1, 1)
param = Parameter('x')
qc.p(param, 0)
qc.measure(0, 0)

prov = IBMQ.load_account()

def test(parameter_binds):
    backend = Aer.get_backend('aer_simulator')
    try:
        print('\naer')
        print(backend.run(qc, parameter_binds=parameter_binds).result().get_counts())
    except Exception as ex:
        print(repr(ex))

    backend = prov.get_backend('ibmq_qasm_simulator')
    try:
        print('\nibmq_qasm_simulator')
        print(backend.run(qc, parameter_binds=parameter_binds).result().get_counts())
    except Exception as ex:
        print(repr(ex))

parameter_binds = [{param: 0}]
print('\nCase 1: parameter_binds =', parameter_binds)
test(parameter_binds)

parameter_binds = [{param: [0]}]
print('\nCase 2: parameter_binds =', parameter_binds)
test(parameter_binds)
```

[...]

PR #1575, opened 2022-08-09
Add `ibmq_semantics` to support semantics of IBMQ provider

hhorii:
### Summary

Fill semantics differences when `ibmq_semantics=True`

### Details and comments

As reported in #1567 and #1551, ibmq provider and Aer have different semantics. A new parameter `ibmq_semantics` converts their semantics:

**1. Job status**

**IBMQ**: `JobStatus.ERROR` if experiments or Job scheduler fail
**Aer**: `JobStatus.ERROR` if Job scheduler (thread pool executor) fails

**2. Parameter binds**

**IBMQ**: list of parameter sets (`dict` with a parameter as a key and its value as a value). If multiple circuits are specified, all the parameter sets are applied to all the circuits (if `n` circuits and `m` parameter sets are specified, `n*m` experiments run). 
**Aer**: list of parameter vector sets (`dict` with a parameter as a key and a list of values as value). The number of circuits and parameter vector sets must be the same. All the length of parameter vector in a `dict` must be the same.

If `ibmq_semantics=True` is set, Aer uses `IBMQ` semantics, otherwise uses `Aer` semantics.

### DIFF
--- a/qiskit/providers/aer/backends/aerbackend.py
+++ b/qiskit/providers/aer/backends/aerbackend.py
@@ -161,6 +161,12 @@ class AerBackend(Backend, ABC):
         Raises:
             ValueError: if run is not implemented
         """
+
+        ibmq_semantics = run_options.get('ibmq_semantics',
+                                         self.options.get('ibmq_semantics', False))
+        if 'ibmq_semantics' in run_options:
+            ibmq_semantics = run_options['ibmq_semantics']
+
         if isinstance(circuits, (QasmQobj, PulseQobj)):
             warnings.warn(
                 'Using a qobj for run() is deprecated as of qiskit-aer 0.9.0'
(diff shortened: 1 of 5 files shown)


yaelbh:
Thanks a lot @hhorii for addressing #1567. This is a good solution. Could you also add an `erorr_message` method to `AerJob`?

qiskit/providers/aer/backends/aerbackend.py
```diff
@@ -343,6 +364,9 @@ def _run(self, qobj, job_id='', format_result=True):
             if "status" in output:
                 msg += f" and returned the following error message:\n{output['status']}"
             logger.warning(msg)
+            if raise_error:
+                raise AerError(msg)
```
yaelbh, line 368: For devices, I don't think it's `run` that triggers an error, but `result`.

hhorii:
@mtreinish, @chriseclectic, and I discussed and decided that Aer should not maintain two semantics.

yaelbh:
Did you come up with some kind of resolution for #1567 ?

PR #1586, opened 2022-08-27
Add AerState that allows python access simulator state directly

hhorii:
### Summary

Add `AerState` that allows python access simulator state directly.

### Details and comments

This PR is the second PR from #1524 (which is being divided to three PRs based on [this request](https://github.com/Qiskit/qiskit-aer/pull/1524#pullrequestreview-1082244663)).
`AerState` is an internal class and end-users should not use it because memory management is sensitive.
This PR depends #1582, which is the first PR from #1524.


qiskit_aer/quantum_info/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+=============================================
+State (:mod:`qiskit_aer.quantum_info`)
```
mtreinish, line 15:
```suggestion
Aer Quantum Information (:mod:`qiskit_aer.quantum_info`)
```

releasenotes/notes/add_aer_state-f8bcee6d844f6dad.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Add new class AerState to support access simulator state from python directly.
+    This class is just for internal usages and end users can access it through
+    other classes such as AerStatevector.
```
mtreinish, line 6: If it's private we don't need a release note since release notes are just for documenting changes for users in the release.

src/controllers/state_controller.hpp
```diff
@@ -446,7 +449,7 @@ void AerState::configure(const std::string& _key, const std::string& _value) {
     configs_[_key] = std::stod(value);
   } else {
     std::stringstream msg;
-    msg << "not supportted configuration" << key << "=" << value << std::endl;
+    msg << "not supportted configuration: " << key << "=" << value << std::endl;
```
mtreinish, line 452:
```suggestion
    msg << "not supported configuration: " << key << "=" << value << std::endl;
```

qiskit_aer/quantum_info/states/aer_state.py
```diff
@@ -0,0 +1,306 @@
...
+    """Internal class to access state of Aer."""
+
+    def __init__(self):
+        """State that handles cpp quantum state safely"""
+        self._shift_to_created_state()
+        self._state = AerStateWrapper()
+        self._method = 'statevector'
+        self._last_qubit = -1
+        self._initialized = False
+        self._closed = False
```
mtreinish, line 30: Do you think it would be useful to have options to set the method and stuff at construction time instead of needing to call `configure()`

mtreinish (commented):
This LGTM (the split makes it so much easier to review the new interface thanks for doing that). I left some small comments inline but I think besides that this is good to merge, especially as it's explicitly an internal class so we don't have to worry too much about compatibility if the interface doesn't work in some way moving forward.

test/terra/states/test_aer_state.py
```diff
@@ -0,0 +1,502 @@
...
+
+        state = AerState()
+        state.allocate_qubits(5)
+        state.initialize(init_state.data)
+
+        expected = init_state.probabilities()
+        actual = state.probabilities()
+
+        for idx in range(0, 2**5):
+            self.assertAlmostEqual(actual[idx], expected[idx])
```
jakelishman, line 448: This test has a leading underscore - it might not be getting run?

jakelishman (commented):
I didn't read the C++ parts super carefully, because I really don't know the context of any of that code, but nothing jumped out at me.

I think the Python wrapper here might be a bit rough.  If it's meant to be accessed by users directly, I think it might want a little bit of attention on making sure it's not possible with standard Python methods for the object to get into an invalid state (e.g. pickle, `copy.{deep,}copy`, calling public methods, etc).  If it's meant to only be used privately by `AerStatevector`, it might be worth putting it in a private class.

(If this is a private class, and _that's_ the reason that no part of the Python imports it into an upper namespace, then that's probably ok too.  Ideally we'd be more explicit about it being private, but we've not done that very clearly elsewhere.)

qiskit_aer/quantum_info/states/aer_state.py
```diff
@@ -0,0 +1,362 @@
...
+        """Safely release all releated memory."""
+        self._assert_allocated_or_mapped_or_moved()
+        if self._state == _STATE.ALLOCATED:
+            self.move_to_ndarray()
+
+        self._assert_mapped_or_moved()
+        if self._state == _STATE.MAPPED:
+            # intentional memory leak
+            self._native_state.move_to_buffer()
+            AerState._not_in_use(self._init_data)
```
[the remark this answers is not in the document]
hhorii: Right. This native memory was mapped with other ndarray and will be freed when the ndarray is collected. I will update this comment.

qiskit_aer/quantum_info/states/aer_state.py
```diff
@@ -0,0 +1,362 @@
...
+
+    _data_in_use = {}
+
+    @staticmethod
+    def _in_use(data):
+        AerState._data_in_use[id(data)] = data
+
+    @staticmethod
+    def _not_in_use(data):
+        del AerState._data_in_use[id(data)]
```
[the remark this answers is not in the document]
hhorii: Hmmm, I do not expect that `AerState` is pickled.  If `AerStatevector` is pickled, it's `AerState` should not be serialized..

hhorii:
@jakelishman `AerState` will be used not only by `AerStatevector`, but also by `AerDensitymatrix` and other state classes. That is reason why I did not set it as a private class of `AerStatevector`.

jakelishman (approved):
By "private class", I just meant private to Aer - being used by `AerDensityMatrix` and `AerStatevector` and other stuff would still be "private" by that.  It's not super important, though - I don't think the Python parts of Aer have ever been super strict about explicitly marking private components, so that's no reason to hold this up now.

PR #1589, opened 2022-08-31
Add docs publishing job

mtreinish:
### Summary

This commit adds a new CI job for publishing documentation on release.
Now that qiskit-aer is a self contained package with it's own namespace
the documentation build no longer needs to be combined into the main
qiskit docs anymore. As such this commit will enable aer ci to publish
built documentation for the project to:

https://qiskit.org/documentation/aer

on every release.

### Details and comments

Fixes Qiskit/qiskit#1508


jakelishman:
If we're publishing it separately this time round, we probably need to run through the whole documentation adding intersphinx to Terra, and making sure the monkey-patched `QuantumCircuit` methods are still getting documented.

jakelishman:
I'm mostly concerned with the things like `QuantumCircuit.save_statevector` that aren't going to be the target of any `autosummary` directive any more after the move (previously Terra's `qiskit.circuit` module would have pulled them in as they'd be attached to `QuantumCircuit` during the build process), and I'm worried about any places in the Aer documentation that might link to Terra objects (e.g. `` :class:`.QuantumCircuit` `` will fail in a separate build).

I think the stubs that are generated by the Aer build itself should work, as you say.

jakelishman:
Happy to hold off on making changes to the documentation build process til later - we don't necessarily need to hold Aer 0.11 on that.

hhorii:
We need to change a branch protection rule from "docs (3.7)" to "docs (3.8)".

hhorii (approved):
This looks good to me. After this PR is merged, we need some follow up PRs to add missing APIs (such as `QuantumCircuit.save_*`) and more contents.

PR #1590, opened 2022-09-01
Add AerStatevector

hhorii:
### Summary

Add a new class `AerStatevector` that supports `q.i.s.Statevector` interface.

### Details and comments

By using `AerState` (#1586), `AerStatevector` generates `ndarray` of a statevector.


qiskit_aer/quantum_info/states/aer_state.py
```diff
@@ -108,8 +111,13 @@ def configure(self, key, value):
             raise AerError('AerState is configured with a str key')
         if not isinstance(value, str):
             value = str(value)
+        self._configs[key] = value
         self._native_state.configure(key, value)
 
+    def configuration(self):
+        """return configuration"""
+        return self._configs
```
jakelishman, line 119: This leaks an internal mutable object, where mutating lets things get out of sync (as `self._native_state` won't get updated.  You maybe want to have this be `return self._configs.copy()` to avoid that.

qiskit_aer/quantum_info/states/aer_state.py
```diff
@@ -147,6 +155,11 @@ def _initialize_with_ndarray(self, data, copy):
 
         self._last_qubit = num_of_qubits - 1
 
+    def __del__(self):
+        """"destructor. Call close() if not called."""
+        if self._state in (_STATE.ALLOCATED, _STATE.MOVED):
+            self.close()
```
jakelishman, line 161: `__del__` in Python isn't guaranteed to be called at all, even when the Python interpreter exits.  It _usually_ gets called, but it's not for sure.  A reliable way to ensure the finaliser _always_ gets called is [`weakref.finalize`](https://docs.python.org/3/library/weakref.html#weakref.finalize) - you'd call that immediately after allocating the memory in C++ space, I imagine.

docs/apidocs/aer_quantum_info.rst
```diff
@@ -0,0 +1,6 @@
+.. _aer-quantum_info:
+
+.. automodule:: qiskit_aer.quantum_info.states
```
jakelishman, line 3: The link name and the referenced module (and the file name) don't match up here.  I'm not sure how you want the documentation to be structured precisely, but it looks like there's a slight mistake either way.

qiskit_aer/quantum_info/states/__init__.py
```diff
@@ -15,3 +15,4 @@
 State (:mod:`qiskit_aer.quantum_info.states`)
 =============================================
 """
+from .aer_statevector import AerStatevector
```
jakelishman, line 18: This doesn't add `AerStatevector` to the documentation, but it probably should.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+from qiskit.quantum_info.states import Statevector
+from qiskit.quantum_info.operators.operator import Operator
+
+from qiskit_aer import AerSimulator
+from .aer_state import AerState
+from ...backends.aerbackend import AerError
+
+
+class AerStatevector(Statevector):
+    """AerStatevector class"""
```
jakelishman, line 29: It would be good to expand on this, and briefly explain what the purpose of the class is, since this is the main part of documentation for the class.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def __init__(self, data, **configs):
+        """Initialize a statevector object.
+
+        Args:
```
jakelishman, line 34:
```suggestion
        """
        Args:
```

(the top line isn't necessary, because the `__init__` docstring just gets appended to the class one directly, so it ends up with no context)

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+            data (np.array or list or Statevector or QuantumCircuit or qiskit.circuit.Instruction):
+                Data from which the statevector can be constructed. This can be either a complex
+                vector, another statevector or a ``QuantumCircuit`` or ``Instruction``
+                (``Operator`` is not supportted in the current implementation).  If the data is
+                a circuit or instruction, the statevector is constructed by assuming that all
+                qubits are initialized to the zero state.
+            configs (kwargs): configurations of ``AerState``. All the keys are handeld as string
```
jakelishman, line 41: This either needs explaining here, or `AerState` needs to be added to the documentation build, and its documentation will need expanding (for example, to example the configuration).  Otherwise there's no way to find out what the available configuration options are other than reading the code.

What if somebody passes a `configs` dictionary that contains `"method": "density_matrix"`?

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+            AerError: if input data is not valid.
+
+        Additional Information:
+            The ``dims`` kwarg is used to ``Statevector`` constructor.
+
+        """
+        if 'dims' in configs:
+            dims = configs.pop('dims')
+        else:
+            dims = None
```
jakelishman, line 53: This might be better as a regular keyword argument in Python, rather than popping it out of the dictionary?

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+
+        if isinstance(data, (QuantumCircuit, Instruction)):
+            data, aer_state = AerStatevector._from_instruction(data, None, **configs)
+        elif isinstance(data, list):
+            data, aer_state = AerStatevector._from_ndarray(np.array(data, dtype=complex), **configs)
+        elif isinstance(data, np.ndarray):
+            data, aer_state = AerStatevector._from_ndarray(data, **configs)
```
jakelishman, line 60: It would be cleaner (and more performant) just to pass the `configs` dictionary directly, rather than unpacking it and re-packing it (with `**`).

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+        if self._result is None:
+            self._result = self._aer_state.last_result()
+        return self._result
+
+    @property
+    def metadata(self):
+        """Return result metadata of an operation that executed lastly."""
+        if self._last_result() is None:
+            raise AerError('AerState was not used and metdata does not exist.')
+        return self._last_result()['metadata']
```
jakelishman, line 84: Unless this property is meant to be mutable, it would be clearer to make it a function, not a property, since it's often recalculated, and access can raise an exception.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def __deepcopy__(self, _memo=None):
+        data, aer_state = AerStatevector._from_instruction(
+            QuantumCircuit(self._aer_state.num_qubits),
+            self._data, **self._configs)
+        return AerStatevector(data, _aer_state=aer_state, **self._configs)
```
jakelishman, line 90: This looks odd, since it does some of the `__init__` manually, then calls `__init__` again.  It might be cleaner just to return `AerStatevector(self._data.copy(), **self._config)` instead, which will just let the `__init__` method reinitialise `AerState` instead.  It'll also let you get rid of the unnecessary middle argument to `AerStatevector._from_instruction`.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def __eq__(self, other):
+        eq_quantumqtate = isinstance(other, Statevector) and self.dims() == other.dims()
+        eq_statevector = np.allclose(self._data, other._data, rtol=self.rtol, atol=self.atol)
+        return eq_quantumqtate and eq_statevector
```
jakelishman, line 95: Generating both booleans and _then_ doing `x and y` bypasses the shortcircuiting, which means if `other` isn't a `Statevector`, then the second line will raise an `AttributeError` on `other._data`.  I'm not sure you need to override this method at all, though?  It just does the same thing as the base class's.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+        Raises:
+            AerError: if the operator dimension does not match the
+                         specified Statevector subsystem dimensions.
+        """
+        if qargs is None:
+            qargs = getattr(other, "qargs", None)
+
+        if isinstance(other, (QuantumCircuit, Instruction)):
+            data, aer_state = AerStatevector._from_instruction(other, self._data, **self._configs)
+            return AerStatevector(data, _aer_state=aer_state, **self._configs)
```
jakelishman, line 117: This passes `_aer_state` as an argument, but that's not interpreted by the actual `__init__` method (see the top review comment for a potentially simpler method here).

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+            qubits = np.array(qargs)
+        self._aer_state.close()
+        self._aer_state = AerState(**self._aer_state.configuration())
+        self._aer_state.initialize(self._data, copy=False)
+        samples = self._aer_state.sample_memory(qubits, shots)
+        self._data = self._aer_state.move_to_ndarray()
+        return samples
+
+    @classmethod
+    def _from_ndarray(cls, init_data, **configs):
```
jakelishman, line 163: This (and the others of its kind) is logically a `staticmethod`, not a `classmethod` (since they don't use `cls`).

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+
+        num_qubits = int(np.log2(len(init_data)))
+
+        aer_state.allocate_qubits(num_qubits)
+        aer_state.initialize(data=init_data)
+
+        return aer_state.move_to_ndarray(), aer_state
+
+    @classmethod
+    def _from_instruction(cls, inst, init_data, **configs):
```
jakelishman, line 182: The `init_data` here isn't super neat.  It might be a shade cleaner to get rid of that, and then in cases where you would use it, do
```python
data, state = AerStatevector._from_ndarray(init_data, config)
AerStatevector._evolve_circuit(state, inst, qubits)
```
(or whatever it is).  Then you've got a bit less duplication of the setup logic.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+            aer_state.apply_global_phase(inst.global_phase)
+
+        AerStatevector._aer_evolve_circuit(aer_state, circuit, range(num_qubits))
+
+        return aer_state.move_to_ndarray(), aer_state
+
+    @classmethod
+    def _aer_evolve_circuit(cls, aer_state, circuit, qubits):
+        """Apply circuit into aer_state"""
+        for inst, qargs, _ in circuit.data:
```
jakelishman, line 211: If Aer 0.11 depends on Terra 0.21+, then this iterator should be
```python
for instruction in circuit.data:
    # Then `inst` and `qargs` are accessed by:
    inst = instruction.operation
    qargs = instruction.qubits
```

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+                                                   [qubits[circuit.find_bit(qarg).index]
+                                                    for qarg in qargs])
+
+    @classmethod
+    def _aer_evolve_instruction(cls, aer_state, inst, qubits):
+        """Apply instruction into aer_state"""
+        params = inst.params
+        if inst.name == 'u':
+            aer_state.apply_mcu(qubits[0:len(qubits) - 1], qubits[len(qubits) - 1],
+                                params[0], params[1], params[2])
```
jakelishman, line 222: Does Aer support `u3` as well?

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+
+        return aer_state.move_to_ndarray(), aer_state
+
+    @classmethod
+    def _aer_evolve_circuit(cls, aer_state, circuit, qubits):
+        """Apply circuit into aer_state"""
+        for inst, qargs, _ in circuit.data:
+            AerStatevector._aer_evolve_instruction(aer_state, inst,
+                                                   [qubits[circuit.find_bit(qarg).index]
+                                                    for qarg in qargs])
```
jakelishman, line 214: What if an instruction has classical bits in it?

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -0,0 +1,273 @@
...
+                            allowed values).
+
+        Returns:
+            AerStatevector: The N-qubit basis state density matrix.
+
+        Raises:
+            QiskitError: if the label contains invalid characters, or the
+                         length of the label is larger than an explicitly
+                         specified num_qubits.
+        """
```
jakelishman, line 272: (I think) you can omit the docstring if you just want to inherit it from the parent class.

src/simulators/state_chunk.hpp
```diff
@@ -143,7 +143,7 @@ class StateChunk : public State<state_t> {
   // Return the expectation value of a N-qubit Pauli operator
   // If the simulator does not support Pauli expectation value this should
   // raise an exception.
-  double expval_pauli(const reg_t &qubits,const std::string& pauli) override final {return 0.0;}
+  double expval_pauli(const reg_t &qubits,const std::string& pauli) override final {return expval_pauli(0, qubits, pauli);}
```
jakelishman, line 146: This seems like an unrelated bugfix?  If it _is_ a bugfix, it'd be good to add a test and a release note about it.

test/terra/states/test_aer_statevector.py
```diff
@@ -0,0 +1,1154 @@
...
+
+@ddt
+class TestAerStatevector(common.QiskitAerTestCase):
+    """AerState tests"""
+
+    def test_generate_aer_statevector(self):
+        """Test generation of Aer's StateVector"""
+        
+        circ = QuantumCircuit(5)
+        state = AerStatevector(circ)
```
jakelishman, line 52: This test isn't really doing anything the other tests don't already do.

test/terra/states/test_aer_statevector.py
```diff
@@ -0,0 +1,1154 @@
...
+        circ1 = QuantumVolume(5, seed=1111)
+        circ2 = circ1.compose(circ1)
+        circ3 = circ2.compose(circ1)
+
+        state1 = AerStatevector(circ1)
+        state2 = AerStatevector(circ2)
+        state3 = AerStatevector(circ3)
+
+        for pa1, pa2 in zip(state1.evolve(circ1), state2):
+            self.assertAlmostEqual(pa1, pa2)
```
jakelishman, line 89: I think this is largely equivalent to `self.assertEqual(state1.evolve(circ1), state2))`, since the equality comparison in `Statevector` is defined to be fuzzy (though that certainly wouldn't have been my choice...).

test/terra/states/test_aer_statevector.py
```diff
@@ -0,0 +1,1154 @@
...
+        import copy
+        circ1 = QuantumVolume(5, seed=1111)
+
+        state1 = AerStatevector(circ1)
+        state2 = copy.deepcopy(state1)
+
+        for pa1, pa2 in zip(state1, state2):
+            self.assertAlmostEqual(pa1, pa2)
+        
+        self.assertNotEqual(id(state1._data), id(state2._data))
```
jakelishman, line 106:
```suggestion
        self.assertIsNot(state1._data, state2._data)
```

test/terra/states/test_aer_statevector.py
```diff
@@ -0,0 +1,1154 @@
...
+from test.terra import common
+from qiskit_aer.aererror import AerError
+from qiskit_aer.quantum_info.states import AerStatevector
+
+
+logger = logging.getLogger(__name__)
+
+@ddt
+class TestAerStatevector(common.QiskitAerTestCase):
+    """AerState tests"""
```
jakelishman, line 46: It would be good to provide some more stress tests of the memory allocation, moving and de-allocation routines, since there's quite a lot of complexity in there.

jakelishman (commented):
I left quite a few comments (sorry) - since this is the actual user-facing interface, I had a bit more to say about structure and things like that.

I'm excited to see this in Aer, since it really starts making it easier to work with massive qubit systems from Python space.  Do you happen to have any short benchmarks to hand about the performance of this compared to Terra's implementation?

A couple of top-level points:

1. There's a few places where you call `<ExplicitSuperClass>.<method>` directly, or override stuff to do the same thing - just a reminder that _all_ methods in Python are virtually dispatch, so `self.<method>` or `super().<method>` are usually what you want in those cases.
2. There's a lot of faff around memory moving, and passing an underlying `AerState` instance around secretly to `AerStatevector.__init__`, which doesn't actually seem to accept that in its arguments.  If you're constructing the Numpy arrays using the C API at some point, you might consider using [`PyArray_SetBaseObject`](https://numpy.org/doc/stable/reference/c-api/array.html#c.PyArray_SetBaseObject) to point to the owning `AerState`, then it'll always be accessible as `data.base`, and you won't need the "moved" status in `AerState` - `AerState` will _always_ be responsible for freeing its underlying Numpy array (all `ndarray`s that view this one set their `base` to the same object, so the refcount of the `AerState` can't ever become 0 while there's active `ndarray` views).

qiskit_aer/quantum_info/states/aer_state.py
hhorii: To call `self.close()` in a callback, `weakref.finalize()` needs to take `self` as an argument for the callback. However, taking `self` means `self` will never be collected. I guess that another way is necessary.

However, `AerState` is a private class. `AerStatevector` will not keep `AerState` as `_STATE.ALLOCATED`. I believe that `__del__` is not necessary.

qiskit_aer/quantum_info/states/aer_statevector.py
hhorii: good point. configurations are the same with `AerSimulator`. `statevector` and `matrix_product_state` are only available now.

qiskit_aer/quantum_info/states/aer_statevector.py
hhorii: I accidentally deleted `_aer_state` argument in `__init__`. It is private and used by only `AerStatevector`.

qiskit_aer/quantum_info/states/aer_statevector.py
hhorii: Right. I copied most of methods from `Statevector`, first. Interestingly, it uses `@classmethod` :-).

qiskit_aer/quantum_info/states/aer_statevector.py
hhorii: `_evolve_circuit` will not release memory of `AerState` (because it does not call `move_to_ndarray()`). On the other hand, `_from_*` returns `AerState` that releases memory. Also, currently, `_from_ndarray()` copies arrays array and `_from_instruction()` does not.  I would like to keep set `init_data` though it looks wired.

src/simulators/state_chunk.hpp
hhorii: This is not a bug in existing usages.

`expval_pauli` should not be enabled in `AerStatevector` in this PR. I would like to revert codes related to `expval_pauli`.

hhorii:
@jakelishman 
> There's a lot of faff around memory moving, and passing an underlying AerState instance around secretly to AerStatevector.__init__, which doesn't actually seem to accept that in its arguments. If you're constructing the Numpy arrays using the C API at some point, you might consider using [PyArray_SetBaseObject](https://numpy.org/doc/stable/reference/c-api/array.html#c.PyArray_SetBaseObject) to point to the owning AerState, then it'll always be accessible as data.base, and you won't need the "moved" status in AerState - AerState will always be responsible for freeing its underlying Numpy array (all ndarrays that view this one set their base to the same object, so the refcount of the AerState can't ever become 0 while there's active ndarray views).

Returned ndarray can be huge and it should be freed as early as possible. I think its lifecycle may be shorter than owning `AerState`. Also, Aer has functions to free memory allocated in C++ as ndarray. I guess, though the current complex usages of `move()` will remain, memory should be freed when its ndarray is collected.

jakelishman (commented):
A couple of minor comments, but nothing that hugely needs to hold anything up.

jakelishman (approved):
The last change squashed the last couple of major issues I had seen, thanks.  I'm not ideally placed to evaluate whether the interface satisfies exactly what you need, and I don't want to hold this up any further.  Let's move to merge and get Aer 0.11 out soon.

PR #1599, opened 2022-09-14
Add control-flow instruction names to basis gates

jakelishman:
### Summary

This adds the Terra control-flow instructions to the custom instructions lists for the simulator methods that support them.  This is likely to be a temporary measure; as classical-processing descriptions are expanded in Terra, this method of describing them feels like it will become unworkable, but for now, this is sufficient to allow `qiskit.transpile(qc, AerSimulator(...))` to work as expected when there is control flow.

### Details and comments

I mostly just added these gates to the simulator methods that could run a test circuit:
```python
import qiskit

qc = qiskit.QuantumCircuit(2, 2)
qc.x(0)
qc.measure(0, 0)
with qc.if_test((qc.clbits[0], True)):
    qc.x(1)
with qc.for_loop((0, 1)):
    qc.x(1)
```

`unitary` and `superop` seem to clearly not work with classical control-flow, but I'm not sure if `extended_stabilizer` should.

I didn't add a release note because I don't know if we want to publicise this yet.

### DIFF
--- a/qiskit_aer/backends/aer_simulator.py
+++ b/qiskit_aer/backends/aer_simulator.py
@@ -436,12 +436,14 @@ class AerSimulator(AerBackend):
             'save_amplitudes', 'save_amplitudes_sq',
             'save_density_matrix', 'save_state', 'save_statevector',
             'save_statevector_dict', 'set_statevector',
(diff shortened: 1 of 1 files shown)


hhorii (approved):
LGTM

PR #1600, opened 2022-09-14
fix an issue of while_loop at the tail of a circuit

hhorii:
### Summary

This fixes a bug of simulation of while loops.

### Details and comments

while_loop is translated to mark and jump instructions. However, some of instructions are truncated before simulation runs. This PR is to always leave mark and jump instructions.


jakelishman (approved):
This looks right to me.  I also checked manually, and this fixes all the circuits I'd been trying to run, thanks!

PR #1598, opened 2022-09-14
Primitives support run method in Terra 0.22

ikkoham:
### Summary

This PR depends on the following PRs.
https://github.com/Qiskit/qiskit-terra/pull/8684
https://github.com/Qiskit/qiskit-terra/pull/8604
https://github.com/Qiskit/qiskit-terra/pull/8837

Once it is determined what these PRs will be, I believe merging this PR is possible even before 0.22.

PR #1601, opened 2022-09-14
Prepare 0.11.0 release

mtreinish:
### Summary

This commit prepares for the 0.11.0 release. It moves the release notes to a self contained directory for the release and updates the wording and details of the release notes. Once this commit merges it should be the tagged commit for the release.


mtreinish:
The only potential blocker on this is if we want to wait for https://github.com/Qiskit/qiskit-aer/pull/1598 or not. I'm thinking maybe we should skip that for 0.11 and wait until the primitives api is stabilized in terra. Then after terra 0.22 we can push out aer 0.11.1 with a fix to update the primitives for the breaking changes in terra.

jakelishman (approved):
Ok, let's go for Aer 0.11 (finally!).

PR #1603, opened 2022-09-15
Use mainlinux2010 skip CIBW tests to release packages for GPU

hhorii:
### Summary

Build packages for qiskit-aer-gpu was failed

### Details and comments

mainlinux2014 needs recent GCC versions for qiskit-aer-gpu. Downgrading to mainlinux2010 introduces build of qiskit-terra from source codes, which producing another errors in tests of cibuildwheel.

This PR changes to
- Use mainlinux2010
- Skip tests in cibuildwheel (`CIBW_TEST_SKIP: "cp*"`)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -173,8 +173,9 @@ jobs:
           python -m pip install cibuildwheel==2.2.2
       - name: Build wheels
         env:
-          CIBW_BEFORE_ALL: "yum install -y yum-utils wget && wget -q https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && rpm -i cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && yum clean all && yum -y install cuda-10-1 openblas-devel"
+          CIBW_BEFORE_ALL: "yum install -y yum-utils wget && wget -q https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && rpm -i cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && yum clean all && yum -y install cuda-10-1 openblas-devel && scl enable devtoolset-9"
           CIBW_MANYLINUX_X86_64_IMAGE: quay.io/pypa/manylinux2010_x86_64:latest
+          CIBW_SKIP: "*-manylinux_i686 cp310* pp* cp36* *musllinux*"
           CIBW_ENVIRONMENT: QISKIT_AER_PACKAGE_NAME=qiskit-aer-gpu AER_THRUST_BACKEND=CUDA CUDACXX=/usr/local/cuda/bin/nvcc
         run: |
           python -m cibuildwheel --output-dir wheelhouse
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -173,7 +173,7 @@ jobs:
           python -m pip install cibuildwheel==2.2.2
       - name: Build wheels
         env:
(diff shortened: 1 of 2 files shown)


mtreinish (commented):
This shouldn't be necessary because if I remember correctly the devtoolset should already be installed in the manylinux container environment we use for building the wheels. I think the failure is because it's using the wrong base image, I think it should be using manylinux2010 but the failed build tried to use manylinux2014. The issue there is it's installing cuda for centos6 (for manylinux2010), but I haven't had a chance to debug it yet this morning.

.github/workflows/build.yml
```diff
@@ -160,3 +160,29 @@ jobs:
...
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.7'
+      - name: Install cibuildwheel
+        run: |
+          python -m pip install -U cibuildwheel==2.2.2
+      - name: Install terra
+        run: |
+          pip install -U qiskit-terra
```
mtreinish, line 176: FWIW, this won't actually install terra in the right place. Since you're installing it before running cibuildwheel this installs terra in the host environment while cibuildwheel operates in a container which won't see this install. You'd have to do this as part of the the cibuildwheel options to get it to install.

.github/workflows/build.yml
```diff
@@ -160,3 +160,29 @@ jobs:
...
+      - name: Build wheels
+        env:
+          CIBW_BEFORE_ALL: "yum install -y yum-utils wget && wget -q https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && rpm -i cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && yum clean all && yum -y install cuda-10-1 openblas-devel"
+          CIBW_MANYLINUX_X86_64_IMAGE: quay.io/pypa/manylinux2010_x86_64:latest
+          CIBW_SKIP: "*-manylinux_i686 cp310* pp* cp36* *musllinux*"
+          CIBW_TEST_SKIP: "cp37*"
```
mtreinish, line 182: Let's just skip all the tests and move on, the testing doesn't add much here anyway

```suggestion
          CIBW_TEST_SKIP: "cp*"
```

.github/workflows/deploy.yml
```diff
@@ -137,7 +137,9 @@ jobs:
       - name: Build wheels
         env:
           CIBW_BEFORE_ALL: "yum install -y yum-utils wget && wget -q https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && rpm -i cuda-repo-rhel6-10-1-local-10.1.243-418.87.00-1.0-1.x86_64.rpm && yum clean all && yum -y install cuda-10-1 openblas-devel"
+          CIBW_MANYLINUX_X86_64_IMAGE: quay.io/pypa/manylinux2010_x86_64:latest
           CIBW_SKIP: "*-manylinux_i686 cp310* pp* cp36* *musllinux*"
+          CIBW_TEST_SKIP: "cp37*"
```
mtreinish, line 142:
```suggestion
          CIBW_TEST_SKIP: "cp*"
```

mtreinish (commented):
The issue here is that qiskit-terra only publishes wheels for manylinux 2014 now (for compat with numpy/scipy) and we're building GPU wheels with a manylinux 2010 environment which is too old for those precompiled binaries on pypi. To workaround this we'd need to build terra from source which would require a rust compiler. This shouldn't be hard to do, we use rustup for it in terra's build jobs already so you could just copy that.

But that being said we really can just skip the tests here, the testing on the GPU wheels isn't really all that valuable because we only test the package is installable and that the CPU simulation works (since we have no GPUs available in CI).  We already validate the builds work for the normal `qiskit-aer` wheels so it's very unlikely adding CUDA to the build will break anything so we can just skip tests in CI

hhorii:
@mtreinish Thank you for your review. I think we should not include `wheel-gpu` test in test.yml because it takes mostly one hour to build gpu packages. This PR only includes change of deploy.yml to switch base linux image and add CIBW_TEST_SKIP.

mtreinish (approved):
LGTM, now after this merges I'll push up a branch with this to manually trigger a GPU wheel build to get a 0.11.0 version published

PR #1604, opened 2022-09-16
Move away from `qiskit.providers.aer` in 0.11 release notes

garrison:
### Summary

The qiskit-aer release notes for 0.11 say that the release moves away from `qiskit.providers.aer`.  However, some of the other release notes for this release actually use this soon-to-be-deprecated import location, including one that is displayed before the one about the new namespace.  This fixes that awkwardness by using `qiskit_aer` uniformly throughout the 0.11 release notes.


mtreinish (approved):
Thanks for this fix it was a good catch (I missed this in #1601). Do you mind also pushing a PR to fix this here: https://github.com/Qiskit/qiskit/blob/master/docs/release_notes.rst#L252 and here https://github.com/Qiskit/qiskit/blob/master/docs/release_notes.rst#L125 too so we get this updated in the hosted docs too

PR #1607, opened 2022-09-30
Add reno branch configuration

jakelishman:
### Summary

There is currently no reno configuration in Aer, so it defaults to basing release notes off the `master` branch, which doesn't exist.  This updates it to correctly look at `main`.

### DIFF
--- a/releasenotes/config.yaml
+++ b/releasenotes/config.yaml
@@ -1,0 +1,3 @@
+---
+encoding: utf8
+default_branch: main


mtreinish (approved):
Oops, I thought that I already had done this a while ago. When I was cleaning up the branch I used to push gpu wheels I also deleted the long stale `master` branch. I should have checked first that we created the config file first before doing that.

hhorii (approved):

PR #1618, opened 2022-10-11
Sampler caches transpiled circuits

### LINKED ISSUE #1594: Add transpilation cache to the primitives
Currently, if I have something like `job = sampler.run([qc] * 50, parameters)`, the only circuit I have here is transpiled 50 times despite it is the same circuit. If I re-use a sampler instance, the circuits I pass are also transpiled once again.

I suggest to introduce a cache of transpiled circuits at the primitive level. So, if the same circuit is re-used, then it is not transpiled, but a transpiled version of the circuit is passed down to a backed.

Here is a snippet:
```python
import numpy as np
from qiskit.circuit.library import ZZFeatureMap
from qiskit_aer.primitives import Sampler as AerSampler


parameters = np.random.random((50, 2))
parameters = 2 * parameters - 1

qc = ZZFeatureMap(2)
qc.measure_all()

sampler = AerSampler()
job = sampler.run([qc] * 50, parameters)
print(job.result())
```

And in `sampler.py` on the lines https://github.com/Qiskit/qiskit-aer/blob/main/qiskit_aer/primitives/sampler.py#L117-L119 the circuit is transpiled many times.

ikkoham:
### Summary

Resolves #1594

### Details and comments

Before:
image

After:
image

### DIFF
--- a/qiskit_aer/primitives/sampler.py
+++ b/qiskit_aer/primitives/sampler.py
@@ -97,27 +97,33 @@ class Sampler(BaseSampler):
         if seed is not None:
             run_options.setdefault("seed_simulator", seed)
 
-        # Key for cache
-        key = tuple(circuits)
+        # Transpile circuits if not in cache
+        no_cache_circuits = []
+        for circuit_index in set(circuits):
+            if circuit_index not in self._cache:
+                no_cache_circuits.append(circuit_index)
+        if no_cache_circuits:
+            transpiled_circuits = (self._circuits[i] for i in no_cache_circuits)
+            if "shots" in run_options and run_options["shots"] is None:
+                transpiled_circuits = (
+                    self._preprocess_circuit(circ) for circ in transpiled_circuits
+                )
+            if not self._skip_transpilation:
+                transpiled_circuits = transpile(
+                    list(transpiled_circuits),
+                    self._backend,
+                    **self._transpile_options,
+                )
+        for i, circuit in zip(no_cache_circuits, transpiled_circuits):
+            self._cache[i] = circuit
 
         # Prepare circuits and parameter_binds
         experiments = []
         parameter_binds = []
-        if key in self._cache:  # Use cached circuits
-            experiments = self._cache[key]
-            for i, value in zip(circuits, parameter_values):
-                self._validate_parameter_length(value, i)
-                parameter_binds.append({k: [v] for k, v in zip(self._parameters[i], value)})
-        else:
-            for i, value in zip(circuits, parameter_values):
-                self._validate_parameter_length(value, i)
-                parameter_binds.append({k: [v] for k, v in zip(self._parameters[i], value)})
-                circuit = self._circuits[i]
-                if "shots" in run_options and run_options["shots"] is None:
-                    circuit = self._preprocess_circuit(circuit)
-                experiments.append(circuit)
-            if not self._skip_transpilation:
-                experiments = transpile(experiments, self._backend, **self._transpile_options)
+        for i, value in zip(circuits, parameter_values):
+            self._validate_parameter_length(value, i)
+            parameter_binds.append({k: [v] for k, v in zip(self._parameters[i], value)})
+            experiments.append(self._cache[i])
 
         result = self._backend.run(
             experiments, parameter_binds=parameter_binds, **run_options
--- a/qiskit_aer/primitives/sampler.py
+++ b/qiskit_aer/primitives/sampler.py
@@ -86,7 +86,7 @@ class Sampler(BaseSampler):
         self._transpile_options = {} if transpile_options is None else transpile_options
         self._skip_transpilation = skip_transpilation
 
-        self._cache = {}
+        self._transpiled_circuits = {}
 
     def _call(
         self,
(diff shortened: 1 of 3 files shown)


qiskit_aer/primitives/sampler.py
```diff
@@ -99,26 +101,34 @@ def _call(
...
+        no_cache_circuits = []
+        for circuit_index in set(circuits):
+            if circuit_index not in self._cache:
+                no_cache_circuits.append(circuit_index)
+        if no_cache_circuits:
+            transpiled_circuits = (self._circuits[i] for i in no_cache_circuits)
+            if "shots" in run_options and run_options["shots"] is None:
+                transpiled_circuits = (
+                    self._preprocess_circuit(circ) for circ in transpiled_circuits
+                )
```
jakelishman, line 114: I'm concerned that the runtime `run_options` here can affect the transpilation, which potentially invalidates the cache.

qiskit_aer/primitives/sampler.py
```diff
@@ -86,6 +86,8 @@ def __init__(
         self._transpile_options = {} if transpile_options is None else transpile_options
         self._skip_transpilation = skip_transpilation
 
+        self._cache = {}
```
jakelishman, line 89: I think calling this a "cache" might hide the fact that this should be the _only_ place that the transpiled circuits should be pulled from.  Really, I think what you're trying to do is just lazy evaluation of the call to `transpile`: if you don't need the laziness, it could even be an option to do the transpilation in the initialisation (but that's really an interface question of what should be blocking / slow), and have an attribute `self._transpiled_circuits` that's _always_ populated.

That said, even with the laziness, I think it's still a bit clearer to call this something like `_transpiled_circuits`.

qiskit_aer/primitives/sampler.py
```diff
@@ -99,26 +101,34 @@ def _call(
...
+                    self._preprocess_circuit(circ) for circ in transpiled_circuits
+                )
+            if not self._skip_transpilation:
+                transpiled_circuits = transpile(
+                    list(transpiled_circuits),
+                    self._backend,
+                    **self._transpile_options,
+                )
+            for i, circuit in zip(no_cache_circuits, transpiled_circuits):
+                self._cache[i] = circuit
```
jakelishman, line 122: This might want factoring out, to make the actual logic of `_call_` simpler to read - this looks to me to just be the implementation of a lazy loader, which are usually a little more maintainable if they're kept a bit more "out of sight".

There's possibly two ways of doing this, depending on whether parallel transpilation is particularly important to you.  If it isn't, then I'd consider something like:
```python
def __init__(self):
    self._transpiled_circuits = self._circuits.copy() if self._skip_transpile else [None]*len(self._circuits)

def _get_circuit(self, index):
    if self._transpiled_circuits[index] is None:
        circuit = self._preprocess(self._circuits[index])
        self._transpiled_circuits[index] = transpile(circuit, self._backend, **self._transpile_options)
    return self._transpiled_circuits[index]

def _call(self, circuits):
    experiments = [self._get_circuit(i) for i in circuits]
    ...
```
If it is, then I might do something like
```python
def _transpile(self, indices):
    if self._skip_transpile:
        return
    to_handle = [i for i in indices if self._transpiled_circuits[i] is None]
    circuits = [self._preprocess(self._circuit[i]) for i in to_handle]
    transpiled = transpile(circuits, self._backend, **self._transpile_options)
    for i, circuit in zip(to_handle, transpiled):
        self._transpiled_circuits[i] = circuit

def _call(self, circuit_indices):
    self._transpile(set(circuit_indices))
    experiments = [self._transpiled_circuits[i] for i in circuit_indices]
    ...
```

The aim just being to separate out the lazy loading from the actual logic of `_call` - if lazy loading weren't important, you'd probably have `self._transpiled_circuits` populated directly in `__init__`, so it seems to me that it's better to avoid this logic being directly in `_call` when possible.

Of course, if the `run_options` to `_call` can invalidate the transpilation, then this has problems, but then I think that's also a potentially larger problem with the interface / implementation.

jakelishman (commented):
I'm actually quite surprised that this didn't exist in `Sampler` already, which is in part why I was against "caching" previously - I thought we were talking about something different, so sorry about that.

I see that `Estimator` already has something like this - I find `Estimator`'s implementation a bit cleaner than the proposal here, so I left a couple of review comments about how this might be tidied up a little.

qiskit_aer/primitives/sampler.py
ikkoham: exactly right

ikkoham:
With so many implementations and so many requests, I forgot to add this feature to `Sampler` in Aer. This is my mistake. Sorry for that.

I followed your advices and refactored the code. I think it is now much easier to read than the first version. Thank you for your comments.

qiskit_aer/primitives/sampler.py
```diff
@@ -99,7 +101,9 @@ def _call(
         if seed is not None:
             run_options.setdefault("seed_simulator", seed)
 
-        # Prepare circuits and parameter_binds
+        is_shots_none = "shots" in run_options and run_options["shots"] is None
```
Cryoris, line 104: So if `shots` is not in the run options it is not considered to be `None`? 🤔

Cryoris (commented):
Just one question and a tiny comment below: if shots is set to None does that mean an exact simulation? 🙂

qiskit_aer/primitives/sampler.py
ikkoham: In the current implementation, the default shots value of backend will be used. (1024)
The documentation mentioned the priority of seed, but I didn't write anything about the priority of shots.

jakelishman (dismissed):
This seems like a faithful implementation of a lazy loader for the transpilation to me.

ikkoham:
I added Reno. and sorry batch experiments is out of scope of this PR. reverted

releasenotes/notes/sampler-cache-78f916cedb0c5421.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
+  - |
+    Improved the performance of :clas:`.Sampler` introducing cache of transpiled circuits.
```
Cryoris, line 4:
```suggestion
    Improved the performance of :class:`.Sampler` introducing cache of transpiled circuits.
```

ikkoham: Thank you so much!

hhorii (approved):

PR #1620, opened 2022-10-12
[perf] Avoid parameter binding

ikkoham:
### Summary

This PR avoid parameter binding not to copy QuantumCircuit deeply in assemble.

### DIFF
--- a/qiskit_aer/backends/aerbackend.py
+++ b/qiskit_aer/backends/aerbackend.py
@@ -39,6 +39,8 @@ from .backend_utils import format_save_type, circuit_optypes
 # Logger
 logger = logging.getLogger(__name__)
 
+one = ParameterExpression({}, 1)
+
 
 class AerBackend(Backend, ABC):
     """Qiskit Aer Backend class."""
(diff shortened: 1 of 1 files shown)


qiskit_aer/backends/aerbackend.py
```diff
@@ -386,12 +388,24 @@ def _assemble(self, circuits, parameter_binds=None, **run_options):
...
                 qobj = None
                 for circuit in circuits:
-                    assemble_bind = {param: 1 for param in circuit.parameters}
+                    # Replace parameters to one
+                    original_parameters = circuit._parameters
+                    circuit._parameters = []
+                    original_params = []
+                    for instruction in circuit._data:
+                        params = instruction.operation.params
+                        original_params.append(params.copy())
```
t-imamichi, line 397: If there is no `ParameterExpression`, you don't need to make a copy. We may get some performance improvement to avoid copies.

ikkoham:
This assemble is a bottleneck and very slow and I would like to improve it. Of course, if you have other ideas, that's fine too. Anyway, regression of the most important benchmark, VQE performance, is not to be missed.
Of course I understand that private access is not good, but I don't have any other ideas. Conversely, it may be a request to improve the public API to be able to do this. It is wrong to need hacks for performance.

I'm reverting the parameters so that there are no side effects.

qiskit_aer/backends/aerbackend.py
ikkoham: Thanks. Good suggestion :+1:

ikkoham:
Cashing qobj may be alternative.

mtreinish (commented):
The issue here is that aer is using terra's assemble which is designed to construct a terra `Qobj` object which is compliant with the qobj specification. THis means we can't pass `Paramete` objects to the output as there is no representation of a parameterized circuit with parameter binds in the specification. So to workaround that limitation we bind a single dummy value to get around this.

For aer however this is unnecessary as has been discussed numerous times in the past using qobj as an exchange format between python and c++ is unecessary overhead. But even if we keep the qobj format using terra's assemble is not desireable because we don't need to create a qobj object for export and external use we can just build the internal state directlly. In this case when going from circuit -> c++ object we can skip all of this and just build the c++ object directly and that would both eliminate the deepcopy overhead by calling `assemble()` with `parameter_binds` set and also generally speed up circuit execution with aer.

I do share @jakelishman's concern about messing with the internal state of a circuit like this it seems like a big layer violation to workaround a separate issue.

ikkoham:
Thanks. Surely this problem should not be solved by workarounds like my proposal, we need fundamentally improvement. I was told that this is an old problem that has been discussed and recognized for a long time, so I look forward to future improvements.

PR #1621, opened 2022-10-14
Use AerStatevector in unittests

derwind:
### Summary

This PR fixes typos.

### Details and comments

`random_statevector` returns `qiskit.quantum_info.Statevector`. So, an `AerStatevector` instance must be created as the next process.
Also, the documentation has been modified since the current specification is that an `AerStatevector` cannot be created directly from a `Statevector`.

I would like to add that I have already discussed this modification with @hhorii.

### DIFF
--- a/qiskit_aer/quantum_info/states/aer_statevector.py
+++ b/qiskit_aer/quantum_info/states/aer_statevector.py
@@ -35,7 +35,8 @@ class AerStatevector(Statevector):
     def __init__(self, data, dims=None, **configs):
         """
         Args:
-            data (np.array or list or AerStatevector or QuantumCircuit or qiskit.circuit.Instruction):
+            data (np.array or list or AerStatevector or QuantumCircuit or
+                  qiskit.circuit.Instruction):
                 Data from which the statevector can be constructed. This can be either a complex
                 vector, another statevector or a ``QuantumCircuit`` or ``Instruction``
                 (``Operator`` is not supportted in the current implementation).  If the data is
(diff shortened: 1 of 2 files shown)


hhorii (approved):
LGTM. Thank you for your contribution.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -35,7 +35,8 @@ class AerStatevector(Statevector):
     def __init__(self, data, dims=None, **configs):
         """
         Args:
-            data (np.array or list or Statevector or QuantumCircuit or qiskit.circuit.Instruction):
+            data (np.array or list or AerStatevector or QuantumCircuit or
```
mtreinish, line 38: Is `AerStatevector` here correct? We can't pass a `Statevector` object directly to the constructor and build an AerStatevector from it?

hhorii: Currently, this constructor does not take `Statevector` as an initial value. However, it may be better to take it.

```
elif isinstance(data, Statevector)
    data, aer_state = AerStatevector._from_ndarray(np.array(data.data, dtype=complex),
                                                               configs)
```

derwind: If `Statevector` will be also allowed, we should do

- Add the presented implementation above
- Update documentations
- Add related unittest

Since the aim of this PR is to fix a minor bug in some unittests, it may be better to make a separate PR for implementation improvements.

mtreinish: Yeah, we can make this larger change in a new follow up PR after this merges, that makes sense to me

PR #1622, opened 2022-10-17
Remove lsp files from version control tracking

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

### Summary

Using ccls/clangd LSPs generate some temporary files that don't need to be checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,13 @@ rhs*_op.pyx
 build/*
 *libomp.dylib
 
+# Ignore LSP config / files
+compile_commands.json
+.cache/clangd
+.ccls
+.ccls-root
+.ccls-cache/
+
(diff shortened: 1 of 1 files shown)


jakelishman (commented):
For sure these don't need to be checked in, but given they're part of your specific workflow and not part of any Aer build processes, perhaps it would be cleaner if you suppress them locally in `.git/info/exlude`?  Just for the purposes of keeping our `.gitignore` more maintainable - we can't suppress all the details of everyone's individual local setups.  (For example, I use `clangd` too, but I just suppress its output at my global git level, like my vim swap files.)

If they're _super_ common, or Aer generates them with Aer-specific options, then I totally agree we should add them to our ignores.

PR #1626, opened 2022-10-19
Release 0.11.1

hhorii:
### Summary

This PR is for 0.11.1 release


ikkoham:
I am happy if https://github.com/Qiskit/qiskit-aer/pull/1618 is included in 0.11.1. (if possible...)

Cryoris:
☝🏻that would actually be quite important as people now using the new algorithms with primitives are complaining about the slow performance -- without caching we cannot really sensibly use the primitives for iterative algorithms like VQE 🤔

mtreinish:
We discussed this during the aer meeting this week and decided to not include caching because it was a confusing concept to all on the meeting. The hit rate will be effectively zero for any caching based on input to the compiler unless you're sending identical circuits which should never happen in practice. If you're trying to just do a parameter value update on a single circuit you should compile the circuit once upfront unbound and just send new values on each iteration. Especially on aer where you can send multiple parameter values at once with very little overhead that will have much more impact that trying to do caching in the primitive itself.

woodsp-ibm:
>The hit rate will be effectively zero for any caching based on input to the compiler unless you're sending identical circuits which should never happen in practice. 

But this is exactly what algorithms are doing. The expectation has been that the primitives will do transpilation to whatever backend etc. And passing skip_transpilation or whatever flag that is, is not standardized as part of base primitives which complicates matters,

mtreinish:
Sigh, then I think this is probably a more fundamental issue with the design of primitives then, it doesn't sound like they were designed with this kind of application in mind (which is odd considering they were supposed to be for applications). Expecting to repeatedly compile an identical circuit internally for applications like algorithms where it's just a parameter update on the same circuit on subsequent calls is a poor design, especially on backends like aer that have efficient parameter assignment at runtime. Caching seems like an overly complicated solution with limited general value for something that IMO should be baked into the interface. But if it's what we're stuck with then we can try to get it in the release

woodsp-ibm:
> Expecting to repeatedly compile an identical circuit internally for applications like algorithms where it's just a parameter update on the same circuit on subsequent calls is a poor design,

My expectation was that for performance it would transpile it once; if it does not then performance is going to be bad. I expected skip_transilation for advanced usage and normal use is to pass the circuit. And since there are identifiers for the circuit any remote transmission is only going to take place once even though circuits are passed each time to the primitive by reference caching will be different based on primitives I would expect.

mtreinish (commented):
I quickly scanned this and it lgtm. I think the only thing missing here is a release note for the terra 0.22.0 primitive compatibility fixes and a prelude release note providing an overview of the release. Then this should be good to go I think

ikkoham:
The basic assumption is that the same circuit with different parameters is run frequently. The cache must be held in some layer. In the case of opflow, `CircuitSampler` has the cache.

Let's consider what the possibilities are.
1. `transpile` has cache. (but `transpile` is just a function...)
2. `primitives` has cache. (my proposal)
3. `algorithms` has cache. (algorithms depends on implementation of primitives, it's not good. algorithms should depend on abstract)

One thing, `skip_transpilation` is essentially a workaround. The original plan was to send `PassManager` for transpiling, but it was discovered that some PassManager is not serializable, so we made a workaround to transpire circuits locally if we wanted to do complex transpiling. This interface may change once the serialization issue is resolved. (not sure... we didn't discuss this problem recently.)

On second thought, it would be a cleaner design to create a class that is in responsible for `transpile with cache` in Terra and have every primitives use it.

mtreinish:
Ok, that makes sense to me, and does seem sensible.  I was confusing a proposal I saw a while ago for doing arbitrary caching for circuit compilation and keying with #1618 which are different things, which was my bad. I do agree doing it in the base class would make this a lot easier too, but we can save that for terra 0.23.0.

jakelishman:
I'm sorry, I think I (and perhaps Matthew) had got into a bit of mess conflating various things we don't in other parts of the Qiskit stack with what Ikko is actually proposing in #1618.  I can say personally that some of this was because I'm very surprised that the `Sampler` isn't _already_ doing this storing of the transpiled circuits.

Just to clarify: since the base primitive interface normalises a "whatever the user input" into an integer index, which is all that Aer's `Sampler` needs to accept, then it makes perfect sense for me that when Aer receives an index to run, it simply pulls the transpiled circuit associated with that index (and transpiles it, if it's the first access).  To me, that's not caching, it's just lazy evaluation of the initial (and only) `transpile` call - the difference in terminology in my mental model is one of the main things that threw me.

On another note (that may be moot now): I'm against anything like "transpile with cache" landing in main Terra.  Since `QuantumCircuit` is mutable and user input / modifications are unconstrained in general usage, it'll open many avenues for bugs due to mutability, and will have a time cost (to calculate the key) that affects all users.

hhorii:
> I quickly scanned this and it lgtm. I think the only thing missing here is a release note for the terra 0.22.0 primitive compatibility fixes and a prelude release note providing an overview of the release. Then this should be good to go I think

I have created a PR to add a release note for Primitive support: https://github.com/Qiskit/qiskit-aer/pull/1628
Do we need a prelude release note for a minor release?

PR #1627, opened 2022-10-20
Update pip install command to reference qiskit-aer

jakelishman:
### Summary

Since we're moving to remove Qiskit Aer from the metapackage, we should update our documentation to reflect that as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Qiskit is made up of elements that each work together to enable quantum computin
 We encourage installing Qiskit via the pip tool (a python package manager). The following command installs the core Qiskit components, including Aer.
 
 ```bash
-pip install qiskit
+pip install qiskit qiskit-aer
 ```
 
 Pip will handle all dependencies automatically for us and you will always install the latest (and well-tested) version.


mriedem (approved):
lgtm

README.md
```diff
@@ -11,7 +11,7 @@ Qiskit is made up of elements that each work together to enable quantum computin
 We encourage installing Qiskit via the pip tool (a python package manager). The following command installs the core Qiskit components, including Aer.
 
 ```bash
-pip install qiskit
+pip install qiskit qiskit-aer
```
mtreinish, line 14: Do we need to have `qiskit` here still? qiskit-aer will depend on terra from it's requirements list so we should get everything we need just from `pip install qiskit-aer`

jakelishman: It felt a bit more right to me to leave it in, since for any real-world usage of Aer, you need to non-transitively use Terra (aka `qiskit` in the new form) to build up your circuits.  So while it's not technically needed, and `pip` doesn't distinguish between direct and transitive dependencies, I thought it was more appropriate for us to be explicit about the direct dependencies for the actual user, perhaps.

Happy to go with whichever strategy we prefer overall.

mtreinish: It was just a bit odd to me, but in this case it's fine I guess. We can always change it later too so it's not a big deal either way

mtreinish (approved):

PR #1630, opened 2022-10-25
add code-formatting with black for python and with clang-format for c++

hhorii:
### Summary

Adds code-formatting with black and pre-commit

### Details and comments

Aer has not used code-formatting. This PR introduces black for all python files and clang-format for all cpp files.
`tox -elint` validates lint and `tox -eblack`, and `tox -epre-commit` format python  and c++ codes respectively.


hhorii:
I'm not sure we should apply this to 0.11 stable. There are few changes in cpp since 0.11 and this PR includes many changes in cpp.

mtreinish:
We should not apply this to the stable branch, it's a risky change with no end user benefit the exact kind of thing we should not backport.

mtreinish (commented):
In general I'm definitely in favor of making a change like this to use auto-formatting for code. We should also do the same for using black on the python code. I do have some questions, mainly around the use of precommit in particular and the creation of a separate CI job to check it.

My concern with pre-commit is it's yet another tool to manage running tests and linting. I'm just wondering what in particular it's providing here. Like if we just called `clang-format` in our tox config would that accomplish the same thing, or is pre-commit here doing some extra setup like installing dependencies that tox won't provide us? Like what happens if you run`pre-commit` without clang installed? In general I think it would be good to try and bake this style correction into our already existing workflows for this (assuming it's feasible).

For the CI job, I think we should bake running the clang-format job into our existing lint job. We're already running that job to validate python code formatting so we can just add C++ there too (we do something similar in other projects with `cargo fmt` for rust). 

The other thought I had is around stable configuration, does clang formatting change between releases? Would it make sense to define a static config file for clang format to use for this. 

Also, we should update CONTRIBUTING.md to mention how to run clang format as it will be a custom lint step for aer code.

As for the git blame we can add a `.git-blame-ignore-revs` file to the repo and document it, like what we have in terra from when we adopted black: https://github.com/Qiskit/qiskit-terra/blob/main/.git-blame-ignore-revs and https://github.com/Qiskit/qiskit-terra/blob/main/CONTRIBUTING.md#dealing-with-the-git-blame-ignore-list

hhorii:
Currently, Aer uses `pycodestyle` and `pylint` for format-checking of python codes. They can not cover cpp codes in my understanding. `pre-commit` has [hooks for various languages](https://pre-commit.com/hooks.html) and we will be able to use it for the all. Also I confirmed that `pre-commit` does not need LLVM installation.
I reused `tests.yml` for `pre-commit` instead of standalone workflow.

jakelishman:
fwiw: the `clang-format` recipe for pre-commit uses [the PyPI package of the same name](https://pypi.org/project/clang-format), which basically just stuffs the `clang-format` executable into a Python wheel for each system.  That's a pretty questionable mismatch of different package managers to me, but as long as `pre-commit` isolates its internal Python environments (which it does seem to do) to avoid polluting my `PATH` with its `clang-format`, that's not the end of the world for me.

If we're bikeshedding the use of `pre-commit` at all, then I would be ok with putting _fast_ formatting fixers in such a file (total runtime < 1s), but against putting linters in it.  I don't think commiting should be a heavy operation, and linting (even when notionally limited to changed files - linters usually need to explore more of the package statically anyway) is not sufficiently fast.  The sort of bugs it catches are often fine to handle in CI anyway - I don't think it buys anything to force devs to run a not-exactly-the-same-as-CI lint locally on every commit.

I also don't entirely see the point in allowing pre-commit hook that might cause a commit to be rejected (like lint).  We're going to have to run all the proper formatting/lint suites in CI _anyway_ because we can't enforce that somebody installs the pre-commit hooks (or doesn't use `git commit --no-verify`), so mostly we're just making certain things harder, like storing debug progress in series of commits to be rebased, for the users who might do that sort of thing, and the users who won't are the ones most likely to not have `pre-commit` configured anwyay.

(That said, even if we did put the config in the repo, I probably wouldn't actually install the hooks myself - I usually just configure my editor to auto-format on write / manual command.)

hhorii:
> I also don't entirely see the point in allowing pre-commit hook that might cause a commit to be rejected (like lint). 

@mtreinish If e8ee5f9 adds `pre-commit` with existing `Tests/lint` workflow and commit will be failed if format is wrong. Original 8ee44e1 uses a different workflow for `pre-commit` and follows @jakelishman wants.

jakelishman:
I mean, I'm not _really_ in favour of another tool either. In this case, we could just skip the middle man and add the `clang-format` PyPI package to `tox` and just continue to use that.

I don't feel super strongly about it, and I wouldn't hold anything up though, especially since I don't do much development on Aer.

hhorii:
We will introduce new code-formatters as the last commit for 0.12.0 release.

hhorii:
Compilation for standalone in windows-2019 fails after applying clang-format. Now I'm narrowing scope of formatting to identify root causes with CI runs.

tox.ini
```diff
@@ -22,20 +23,35 @@ commands =
...
+[testenv:black]
+envdir = .tox/lint
+commands = black qiskit_aer test tools setup.py
+
+[pycodestyle]
+max-line-length = 105
+# default ignores + E741 because of opflow global variable I
+# + E203 because of a difference of opinion with black
+# codebase does currently comply with: E133, E242, E704, W505
+ignore = E121, E123, E126, E133, E226, E241, E242, E704, W503, W504, W505, E741, E203
```
mtreinish, line 53: Is this used by anything?

tools/clang-format.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+cd `dirname $0`/..
+IGNORE_FILES="src/misc/warnings.hpp"
+
+ERROR=0
+for FILE in `find src -iname "*.hpp" -or -iname "*.cpp"`; do
```
mtreinish, line 7: What about the c++ files in `qiskit_aer` and `test`?

CONTRIBUTING.md
```diff
@@ -185,6 +190,22 @@ like for the current state of the repo, you can run: `tox -edocs` which will
...
 particular will be located at `docs/_build/html/release_notes.html`
 
+## Style and lint
+
+Qiskit Aer uses 3 tools for verify code formatting and lint checking. The
+first tool is [black](https://github.com/psf/black) which is a Python code formatting
+tool that will automatically update the code formatting to a consistent style.
+The second tool is [pylint](https://www.pylint.org/) which is a code linter
+which does a deeper analysis of the Python code to find both style issues and
+potential bugs and other common issues in Python.
```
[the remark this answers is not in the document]
hhorii: It was written in an old commit, but it was wrongly overwritten while resolving conflicts...

tox.ini
hhorii: no longer :-)

tools/clang-format.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+cd `dirname $0`/..
+IGNORE_FILES="src/misc/warnings.hpp"
```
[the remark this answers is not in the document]
hhorii: clang-format wrongly formats pragma.

```
   DISABLE_WARNING(-Wall)                                                       \
   DISABLE_WARNING(-Wextra)                                                     \
   DISABLE_WARNING(-Wshadow)                                                    \
-  DISABLE_WARNING(-Wfloat-equal)                                               \
+  DISABLE_WARNING(-Wfloat - equal)                                             \
   DISABLE_WARNING(-Wundef)                                                     \
   DISABLE_WARNING(-Wpedantic)                                                  \
-  DISABLE_WARNING(-Wredundant-decls)                                           \
+  DISABLE_WARNING(-Wredundant - decls)                                         \
   DISABLE_WARNING(-Wsign-compare)
```

I tried to use `WhitespaceSensitiveMacros` but not yet find a good configuration that disable formatting of white spaces in macro.

tools/clang-format.sh
hhorii: good comment. let me add them into the list. maybe `contrib` should be included.

CONTRIBUTING.md
```diff
@@ -185,6 +190,24 @@ like for the current state of the repo, you can run: `tox -edocs` which will
...
 particular will be located at `docs/_build/html/release_notes.html`
 
+## Style and lint
+
+Qiskit Aer uses 3 tools for verify code formatting and lint checking. The
+first tool is [black](https://github.com/psf/black) which is a Python code formatting
+tool that will automatically update the code formatting to a consistent style.
+The second tool is [pylint](https://www.pylint.org/) which is a code linter
+which does a deeper analysis of the Python code to find both style issues and
+potential bugs and other common issues in Python. The thir tool is
```
mtreinish, line 200:
```suggestion
potential bugs and other common issues in Python. The third tool is
```

.github/workflows/tests.yml
```diff
@@ -34,10 +34,15 @@ jobs:
...
+      - name: Run clang-format
+        run: |
+          set -e
+          sh tools/clang-format.sh --Werror -n
+        shell: bash          
       - name: Run Lint
         run: |
           set -e
-          pycodestyle --ignore=E402,W503,W504 --max-line-length=100 qiskit_aer
+          black --check qiskit_aer test tools
```
mtreinish, line 45: This is missing the `setup.py` you have in the tox config and it will fail the `tox -elint` job locally because it needs to be reformatted.

```suggestion
          black --check qiskit_aer test tools setup.py
```

mtreinish (commented):
I caught one small oversight in the CI config. Also the black config isn't present in the pyproject.toml so it has the default line set of 88 characters instead of 100 (which I assume you were going for).

Besides that I'm having trouble reproducing the same output on a local clone (I'm getting subtle differences between my test repo and your branch). I'm not sure what's going on for the most part the diff is identical except for:

```
 controllers/aer_controller.hpp                                  |    2 +-
 framework/config.hpp                                            |    1 +
 framework/pybind_casts.hpp                                      |    1 +
 open_pulse/eval_hamiltonian.hpp                                 |   10 ++++++----
 simulators/matrix_product_state/matrix_product_state.hpp        |   10 +++++-----
 simulators/matrix_product_state/matrix_product_state_tensor.hpp |   21 +++++++++++----------
 simulators/statevector/chunk/cuStateVec_chunk_container.hpp     |    6 +++---
 simulators/statevector/statevector_state.hpp                    |   10 +++++-----
 simulators/tensor_network/tensor_net.hpp                        |    2 +-
 simulators/tensor_network/tensor_net_state.hpp                  |   10 +++++-----
 10 files changed, 39 insertions(+), 34 deletions(-)
```

I'm wondering if your local branch wasn't clean when you committed things. Like there are a few places the diff is more than formatting like extra `#include` statements which is more than I'd expect clang format to do.

mtreinish:
This is the full diff the `a` side is my fresh clone that I copied the config files into and ran `tox -eclang-format` and `tox -eblack` on. The `b` side is your PR branch at the current HEAD (54399d148b4cac24bd78939156aeb6bcbcde75a8). Some of this diff looks like things I'd expect clang format to fix like comment spacing and comment types. But others I'm confused by. I might have missed a config in my copy over or maybe there is a clang format difference on linux?

```diff
diff --git a/src/controllers/aer_controller.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/controllers/aer_controller.hpp
index bf144fc2..f0ee3acd 100644
--- a/src/controllers/aer_controller.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/controllers/aer_controller.hpp
@@ -989,7 +989,7 @@ Result Controller::execute(std::vector<Circuit> &circuits,
     const int NUM_RESULTS = result.results.size();
     // following looks very similar but we have to separate them to avoid omp
     // nested loops that causes performance degradation (DO NOT use if statement
-    //in #pragma omp)
+    // in #pragma omp)
     if (parallel_experiments_ == 1) {
       for (int j = 0; j < NUM_RESULTS; ++j) {
         set_parallelization_circuit(circuits[j], noise_model, methods[j]);
diff --git a/src/framework/config.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/framework/config.hpp
index 6ebe7865..2890747a 100644
--- a/src/framework/config.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/framework/config.hpp
@@ -15,6 +15,7 @@
 #ifndef _aer_framework_config_hpp_
 #define _aer_framework_config_hpp_
 
+#include "json.hpp"
 #include "types.hpp"
 #include <optional>
 #include <string>
diff --git a/src/framework/pybind_casts.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/framework/pybind_casts.hpp
index 1db098fb..7e18ee30 100644
--- a/src/framework/pybind_casts.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/framework/pybind_casts.hpp
@@ -16,6 +16,7 @@
 #define _aer_framework_pybind_casts_hpp_
 
 #include "../simulators/stabilizer/clifford.hpp"
+#include <pybind11/numpy.h>
 
 namespace py = pybind11;
 
diff --git a/src/open_pulse/eval_hamiltonian.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/open_pulse/eval_hamiltonian.hpp
index aea06967..dfa2197e 100644
--- a/src/open_pulse/eval_hamiltonian.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/open_pulse/eval_hamiltonian.hpp
@@ -15,6 +15,8 @@
 #ifndef _EVAL_HAMILTONIAN_HPP
 #define _EVAL_HAMILTONIAN_HPP
 
+#include "framework/types.hpp"
+#include "iterators.hpp"
 #include "misc/warnings.hpp"
 #include <complex>
 #include <string>
@@ -42,10 +44,10 @@ struct hash<ParserValues> {
 } // namespace std
 
 // TODO: Document
-complex_t evaluate_hamiltonian_expression(
+AER::complex_t evaluate_hamiltonian_expression(
     const std::string &expr_string, const std::vector<double> &vars,
     const std::vector<std::string> &vars_names,
-    const std::unordered_map<std::string, complex_t> &chan_values) {
+    const std::unordered_map<std::string, AER::complex_t> &chan_values) {
 
   static std::unordered_map<std::string, std::unique_ptr<ParserValues>>
       parser_expr;
@@ -75,7 +77,7 @@ complex_t evaluate_hamiltonian_expression(
   // std::cout << "Getting parser " << std::hex << parser << "\n";
 
   auto maybe_update_value = [parser](const std::string &var_name,
-                                     const complex_t &var_value) {
+                                     const AER::complex_t &var_value) {
     if (parser->var_values.find(var_name) == parser->var_values.end()) {
       parser->var_values.emplace(var_name,
                                  std::make_unique<mup::Value>(var_value));
@@ -93,7 +95,7 @@ complex_t evaluate_hamiltonian_expression(
 
   for (const auto &idx_var : enumerate(vars)) {
     size_t index = idx_var.first;
-    auto var_value = static_cast<complex_t>(idx_var.second);
+    auto var_value = static_cast<AER::complex_t>(idx_var.second);
     maybe_update_value(vars_names[index], var_value);
   }
 
diff --git a/src/simulators/matrix_product_state/matrix_product_state.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/matrix_product_state/matrix_product_state.hpp
index 354e6021..4105fd3b 100644
--- a/src/simulators/matrix_product_state/matrix_product_state.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/matrix_product_state/matrix_product_state.hpp
@@ -274,14 +274,14 @@ const stringmap_t<Gates>
                      {"rx", Gates::rx},     // Pauli-X rotation gate
                      {"ry", Gates::ry},     // Pauli-Y rotation gate
                      {"rz", Gates::rz},     // Pauli-Z rotation gate
-                     // Waltz Gates
+                     /* Waltz Gates */
                      {"p", Gates::u1},  // zero-X90 pulse waltz gate
                      {"u1", Gates::u1}, // zero-X90 pulse waltz gate
                      {"u2", Gates::u2}, // single-X90 pulse waltz gate
                      {"u3", Gates::u3}, // two X90 pulse waltz gate
                      {"u", Gates::u3},  // two X90 pulse waltz gate
                      {"U", Gates::u3},  // two X90 pulse waltz gate
-                     // Two-qubit gates
+                     /* Two-qubit gates */
                      {"CX", Gates::cx},   // Controlled-X gate (CNOT)
                      {"cx", Gates::cx},   // Controlled-X gate (CNOT)
                      {"cy", Gates::cy},   // Controlled-Y gate
@@ -294,10 +294,10 @@ const stringmap_t<Gates>
                      {"ryy", Gates::ryy},   // Pauli-YY rotation gate
                      {"rzz", Gates::rzz},   // Pauli-ZZ rotation gate
                      {"rzx", Gates::rzx},   // Pauli-ZX rotation gate
-                     // Three-qubit gates
+                     /* Three-qubit gates */
                      {"ccx", Gates::ccx}, // Controlled-CX gate (Toffoli)
                      {"cswap", Gates::cswap},
-                     // Pauli
+                     /* Pauli */
                      {"pauli", Gates::pauli}});
 
 //=========================================================================
@@ -813,4 +813,4 @@ std::pair<uint_t, double> State::sample_measure_with_prob(const reg_t &qubits,
 //-------------------------------------------------------------------------
 } // end namespace AER
 //-------------------------------------------------------------------------
-#endif
+#endif
\ No newline at end of file
diff --git a/src/simulators/matrix_product_state/matrix_product_state_tensor.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
index f09dbe53..01559258 100644
--- a/src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/matrix_product_state/matrix_product_state_tensor.hpp
@@ -556,9 +556,11 @@ void MPS_Tensor::contract_2_dimensions(const MPS_Tensor &left_gamma,
                                          (omp_threads > 1))                    \
     num_threads(omp_threads)
 #endif
-  for (int_t l_row = 0; l_row < left_rows; l_row++)
-    for (int_t r_col = 0; r_col < right_columns; r_col++)
+  for (int_t l_row = 0; l_row < left_rows; l_row++) {
+    for (int_t r_col = 0; r_col < right_columns; r_col++) {
       result(l_row, r_col) = 0;
+    }
+  }
 
 #ifdef _WIN32
 #pragma omp parallel for if ((omp_limit > MATRIX_OMP_THRESHOLD) &&             \
@@ -568,24 +570,23 @@ void MPS_Tensor::contract_2_dimensions(const MPS_Tensor &left_gamma,
                                          (omp_threads > 1))                    \
     num_threads(omp_threads)
 #endif
-  for (int_t l_row = 0; l_row < left_rows; l_row++)
+  for (int_t l_row = 0; l_row < left_rows; l_row++) {
     for (int_t r_col = 0; r_col < right_columns; r_col++) {
-
       for (int_t size = 0; size < left_size; size++)
         for (int_t index = 0; index < left_columns; index++) {
           result(l_row, r_col) += left_gamma.data_[size](l_row, index) *
                                   right_gamma.data_[size](index, r_col);
         }
     }
+  }
 }
 //---------------------------------------------------------------
 // function name: Decompose
-// Description: Decompose a tensor into two Gamma tensors and the Lambda between
-// 				them. Usually used after applying a 2-qubit
-// gate. Parameters: MPS_Tensor &temp - the tensor to decompose.
-//			   MPS_Tensor &left_gamma, &right_gamma , rvector_t &lambda
-//-
-// 			   tensors for the result.
+// Description: Decompose a tensor into two Gamma tensors and the
+//        Lambda between them. Usually used after applying a 2-qubit gate.
+// Parameters: MPS_Tensor &temp - the tensor to decompose.
+//			       MPS_Tensor &left_gamma, &right_gamma
+//             rvector_t &lambda - tensors for the result.
 // Returns: none.
 //---------------------------------------------------------------
 double MPS_Tensor::Decompose(MPS_Tensor &temp, MPS_Tensor &left_gamma,
diff --git a/src/simulators/statevector/chunk/cuStateVec_chunk_container.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/statevector/chunk/cuStateVec_chunk_container.hpp
index 88909884..3fd95b94 100644
--- a/src/simulators/statevector/chunk/cuStateVec_chunk_container.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/statevector/chunk/cuStateVec_chunk_container.hpp
@@ -29,11 +29,11 @@ namespace Chunk {
 template <typename data_t>
 class cuStateVecChunkContainer : public DeviceChunkContainer<data_t> {
 protected:
-  custatevecHandle_t
-      custatevec_handle_; // cuStatevec handle for this chunk container
+  custatevecHandle_t custatevec_handle_; // cuStatevec handle
+                                         // for this chunk container
   uint_t custatevec_chunk_total_qubits_; // total qubits of statevector passed
                                          // to ApplyMatrix
-  uint_t custatevec_chunk_count_; // number of counts for all chunks
+  uint_t custatevec_chunk_count_;        // number of counts for all chunks
 
   custatevecDeviceMemHandler_t custatevec_mem_handler_;
   cudaMemPool_t memory_pool_;
diff --git a/src/simulators/statevector/statevector_state.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/statevector/statevector_state.hpp
index 8b9ff80a..405c1423 100644
--- a/src/simulators/statevector/statevector_state.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/statevector/statevector_state.hpp
@@ -347,18 +347,18 @@ const stringmap_t<Gates> State<statevec_t>::gateset_(
      {"p", Gates::mcp},       // Parameterized phase gate
      {"sx", Gates::mcsx},     // Sqrt(X) gate
      {"sxdg", Gates::mcsxdg}, // Inverse Sqrt(X) gate
-     // 1-qubit rotation Gates
+     /* 1-qubit rotation Gates */
      {"r", Gates::mcr},   // R rotation gate
      {"rx", Gates::mcrx}, // Pauli-X rotation gate
      {"ry", Gates::mcry}, // Pauli-Y rotation gate
      {"rz", Gates::mcrz}, // Pauli-Z rotation gate
-     // Waltz Gates
+     /* Waltz Gates */
      {"u1", Gates::mcp},  // zero-X90 pulse waltz gate
      {"u2", Gates::mcu2}, // single-X90 pulse waltz gate
      {"u3", Gates::mcu3}, // two X90 pulse waltz gate
      {"u", Gates::mcu3},  // two X90 pulse waltz gate
      {"U", Gates::mcu3},  // two X90 pulse waltz gate
-     // 2-qubit gates
+     /* 2-qubit gates */
      {"CX", Gates::mcx},       // Controlled-X gate (CNOT)
      {"cx", Gates::mcx},       // Controlled-X gate (CNOT)
      {"cy", Gates::mcy},       // Controlled-Y gate
@@ -377,10 +377,10 @@ const stringmap_t<Gates> State<statevec_t>::gateset_(
      {"csx", Gates::mcsx},     // Controlled-Sqrt(X) gate
      {"csxdg", Gates::mcsxdg}, // Controlled-Sqrt(X)dg gate
      {"ecr", Gates::ecr},      // ECR Gate
-     // 3-qubit gates
+     /* 3-qubit gates */
      {"ccx", Gates::mcx},      // Controlled-CX gate (Toffoli)
      {"cswap", Gates::mcswap}, // Controlled SWAP gate (Fredkin)
-     // Multi-qubit controlled gates
+     /* Multi-qubit controlled gates */
      {"mcx", Gates::mcx},       // Multi-controlled-X gate
      {"mcy", Gates::mcy},       // Multi-controlled-Y gate
      {"mcz", Gates::mcz},       // Multi-controlled-Z gate
diff --git a/src/simulators/tensor_network/tensor_net.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/tensor_network/tensor_net.hpp
index 3f2cf4e6..f4f94343 100644
--- a/src/simulators/tensor_network/tensor_net.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/tensor_network/tensor_net.hpp
@@ -62,7 +62,7 @@ protected:
   int32_t mode_index_;                                   // index of modes
   std::vector<std::shared_ptr<Tensor<data_t>>> tensors_; // list of tensors
   std::vector<std::shared_ptr<Tensor<data_t>>> qubits_;  // tail tensor for
-                                                        // qubits
+                                                         // qubits
   std::vector<std::shared_ptr<Tensor<data_t>>>
       qubits_sp_;                        // tail tensor for super qubits
   std::vector<int32_t> modes_qubits_;    // tail mode index for qubits
diff --git a/src/simulators/tensor_network/tensor_net_state.hpp b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/tensor_network/tensor_net_state.hpp
index 72bcd2ef..44a6221c 100644
--- a/src/simulators/tensor_network/tensor_net_state.hpp
+++ b/home/mtreinish/git/qiskit/qiskit-aer/src/simulators/tensor_network/tensor_net_state.hpp
@@ -330,18 +330,18 @@ const stringmap_t<Gates> State<tensor_net_t>::gateset_(
      {"p", Gates::mcp},       // Parameterized phase gate
      {"sx", Gates::mcsx},     // Sqrt(X) gate
      {"sxdg", Gates::mcsxdg}, // Inverse Sqrt(X) gate
-     // 1-qubit rotation Gates
+     /* 1-qubit rotation Gates */
      {"r", Gates::mcr},   // R rotation gate
      {"rx", Gates::mcrx}, // Pauli-X rotation gate
      {"ry", Gates::mcry}, // Pauli-Y rotation gate
      {"rz", Gates::mcrz}, // Pauli-Z rotation gate
-     // Waltz Gates
+     /* Waltz Gates */
      {"u1", Gates::mcp},  // zero-X90 pulse waltz gate
      {"u2", Gates::mcu2}, // single-X90 pulse waltz gate
      {"u3", Gates::mcu3}, // two X90 pulse waltz gate
      {"u", Gates::mcu3},  // two X90 pulse waltz gate
      {"U", Gates::mcu3},  // two X90 pulse waltz gate
-     // 2-qubit gates
+     /* 2-qubit gates */
      {"CX", Gates::mcx},       // Controlled-X gate (CNOT)
      {"cx", Gates::mcx},       // Controlled-X gate (CNOT)
      {"cy", Gates::mcy},       // Controlled-Y gate
@@ -360,10 +360,10 @@ const stringmap_t<Gates> State<tensor_net_t>::gateset_(
      {"csx", Gates::mcsx},     // Controlled-Sqrt(X) gate
      {"csxdg", Gates::mcsxdg}, // Controlled-Sqrt(X)dg gate
      {"ecr", Gates::ecr},      // ECR Gate
-     // 3-qubit gates
+     /* 3-qubit gates */
      {"ccx", Gates::mcx},      // Controlled-CX gate (Toffoli)
      {"cswap", Gates::mcswap}, // Controlled SWAP gate (Fredkin)
-     // Multi-qubit controlled gates
+     /* Multi-qubit controlled gates */
      {"mcx", Gates::mcx},       // Multi-controlled-X gate
      {"mcy", Gates::mcy},       // Multi-controlled-Y gate
      {"mcz", Gates::mcz},       // Multi-controlled-Z gate
```

hhorii:
@mtreinish thank you for your checking.

`clang-format -i` sometimes generates codes that will not be passed with `clang-format -Werror -n`. For example,

```
      {"cswap", Gates::mcswap}, // Controlled SWAP gate (Fredkin)
      // Multi-qubit controlled gates
      {"mcx", Gates::mcx},       // Multi-controlled-X gate
```
will warns as
```
src/simulators/tensor_network/tensor_net_state.hpp:365:65: error: code should be clang-formatted [-Wclang-format-violations]
     {"cswap", Gates::mcswap}, // Controlled SWAP gate (Fredkin)
```
. Therefore, I manually modified the codes after applying clang-format as
```
      {"cswap", Gates::mcswap}, // Controlled SWAP gate (Fredkin)
      /* Multi-qubit controlled gates */
      {"mcx", Gates::mcx},       // Multi-controlled-X gate
```

Another disruptive change is order of include files. `clang-format` sorts include statements and sometimes Aer headers are not self-contained (unfortunately). Then I added some include files or some namespace (such as `complex_t`  -> `AER::complex_t`). 

This diff accidentally comes from my old formatting trials (no `RemoveBracesLLVM`). 
```
-  for (int_t l_row = 0; l_row < left_rows; l_row++)
-    for (int_t r_col = 0; r_col < right_columns; r_col++)
+  for (int_t l_row = 0; l_row < left_rows; l_row++) {
+    for (int_t r_col = 0; r_col < right_columns; r_col++) {
       result(l_row, r_col) = 0;
+    }
+  }
```

hhorii:
@mtreinish can we set 94e41a6 stable-backport-potential?

PR #1637, opened 2022-10-28
Implement GPU tensor network simulator by using cuTensorNet

doichanj:
### Summary
This is implementation of a new simulation method `tensor_network` for `device=GPU` accelerated by cuQuantum SDK.
`tensor_network` method is tensor network contraction based simulator for general quantum circuits.

### Details and comments
`tensor_network` method may used for quantum circuit simulation with large number of qubits, but has fewer circuit depth.
This simulation method simulates quantum circuits by contracting tensor network built from connections of gates of input circuit. Tensor network contraction can be parallelize by using slicing technique optimized in cuTensorNet SDK.

This is performance results of `tensor_network` method compared to `statevector` using QAOA circuit + save expectation value (using random connected graph, connection ratio = 0.1) accelerated by single NDIVIA Tesla V100 (32GB)
image

This PR requires shot branching implementation PR #1596
Currently there is no implementation for `device=CPU` 

`tensor_network` method uses density matrix in tensor network to get norm of statevector.
So it also supports operations for `density_matrix` (can simulate noise sampling in single shot)
Because using density matrix inside, number of qubits for output of contraction is limited to memory size similar to density matrix. 
(So we can not get large statevector/reduced density matrix output)

To simulate quantum circuits with large number of qubits, expectation values, measure or sampling can be used. (Maybe sampling is very slow for random results)

This simulator supports parallelization by multiple-GPUs or multiple nodes on cluster (MPI), tensor network contraction is parallelized.


test/terra/backends/aer_simulator/test_conditional.py
```diff
@@ -78,9 +82,11 @@ def test_conditional_gates_64bit(self, method, device):
     @supported_methods(SUPPORTED_METHODS)
     def test_conditional_gates_132bit(self, method, device):
         """Test conditional gate operations on 132-bit conditional register."""
-        shots = 100
```
hhorii, line 81: I believe that this line is necessary. `shots` can be used without assignment, and then tests fail.

src/simulators/state.hpp
```diff
@@ -229,6 +229,11 @@ class Base {
   //set creg bit counts before initialize creg
   virtual void set_num_creg_bits(uint_t num_memory, uint_t num_register){}
 
+  void has_statevector_ops(bool flg)
```
hhorii, line 232: This method name looks odd. `State` does not have operations (`ops`). Maybe `enable_save_statevec` is better (though `save_amp` is also enabled).

src/simulators/tensor_network/tensor.hpp
```diff
@@ -0,0 +1,277 @@
...
+{
+protected:
+  int base_;
+  int rank_;
+  std::vector<int64_t> extents_;
+  int64_t size_;
+  std::vector<std::complex<data_t>> tensor_;
+  std::vector<int32_t> modes_;
+  reg_t qubits_;
+  bool sp_tensor_;
```
hhorii, line 43: could you leave some comments on each member? Maybe `tensor_` is data of multi-dimensional array defined `rank_` and `modes_`.
Also, why `int64_t` and `int32_t` are used here? Can they be replaced with `uint64_t` and `uint32_t`? Or can they be negative?

src/simulators/state.hpp
hhorii: Hmmm, `has_statevector_ops_` is just for `AER::TensorNetwork::State`.  I think It should be a method of the class.

qiskit_aer/backends/aer_simulator.py
```diff
@@ -276,6 +283,20 @@ class AerSimulator(AerBackend):
...
+      (i.e. measure, reset, noises, etc.) is applied (Default: True).
+      This option can decrease runs of shots if there will be less branches
+      than number of shots.
+      This option is available for ``"statevector"``, ``"density_matrix"``
+      and ``"stabilizer"``.
+      For GPU, ``cuStateVec_enable`` is not supported for this option.
+
+    * ``runtime_noise_sampling_enable`` (bool): This option enables/disables
+      runtime noise sampling. This option is only enabled when
+      ``shot_branching_enable`` is also True. (Default: False).
```
hhorii, line 298: Are these comments for shot-branching?

hhorii:
This is a new feature. A release note is necessary. I think the note should include some simple samples to use `tensor_network` method. I believe it can be available only if the system has NVIDIA GPU and cuQuantum. These information should be documented in release note also.

I'm wondering what will be happen if `tensor_network` is specified in no cuQuantum environment. I guess an error will be raised. Right?

test/terra/backends/aer_simulator/test_conditional.py
doichanj: I removed remaining shot branching dependencies

src/simulators/state.hpp
doichanj: I moved this to state in tensor network and changed name `enable_density_matrix`

src/simulators/tensor_network/tensor.hpp
doichanj: comment added

Integer types defined in here is supported types in cuTensorNet, the same type is used to avoid data conversion.

qiskit_aer/backends/aer_simulator.py
doichanj: removed them

doichanj:
I added exception when `method=tensor_network`, and `device=CPU`, but this causes issue in getting available methods in `aer_simulator.py` because `available_methods` only tests methods on CPU. So I modified to search for all the supported devices to return `tensor_network` if GPU is available.

cmake/conan_utils.cmake
```diff
@@ -10,7 +10,7 @@ macro(setup_conan)
     # Right now every dependency shall be static
     set(CONAN_OPTIONS ${CONAN_OPTIONS} "*:shared=False")
 
-    set(REQUIREMENTS nlohmann_json/3.1.1 spdlog/1.9.2)
+    set(REQUIREMENTS nlohmann_json/3.1.1 spdlog/1.5.0)
```
hhorii, line 13: is there a reason of this downgrade?

doichanj: There is an inconsistency of version to https://github.com/doichanj/qiskit-aer/blob/24df5135cc06d9f9dd839d985b46bd91b2511232/cmake/dependency_utils.cmake#L27 and this causes build error on Power systems.

hhorii (approved):
Now this looks fine. I would like to show some tutorial to show good examples for this new method.

PR #1642, opened 2022-11-03
Add support for Python 3.11

mtreinish:
### Summary

Python 3.11.0 was released on 10-24-2022, this commit marks the start of support for Python 3.11 in qiskit. It adds the supported Python version in the package metadata and updates the CI configuration to run test jobs on Python 3.11 and build Python 3.11 wheels on release.

### Details and comments

### DIFF
--- a/tools/install_openblas_i686.sh
+++ b/tools/install_openblas_i686.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+wget https://github.com/xianyi/OpenBLAS/releases/download/v0.3.21/OpenBLAS-0.3.21.zip
+unzip OpenBLAS-0.3.21.zip
+cd OpenBLAS-0.3.21
+make TARGET=NORTHWOOD
+make TARGET=NORTHWOOD PREFIX=/usr/ install
--- a/tools/install_openblas_i686.sh
+++ b/tools/install_openblas_i686.sh
@@ -3,5 +3,5 @@
 wget https://github.com/xianyi/OpenBLAS/releases/download/v0.3.21/OpenBLAS-0.3.21.zip
 unzip OpenBLAS-0.3.21.zip
 cd OpenBLAS-0.3.21
-make
-make PREFIX=/usr/ install
+make TARGET=KATMAI
+make TARGET=KATMAI PREFIX=/usr/ install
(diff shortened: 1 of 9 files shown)


mtreinish:
This is blocked on qiskit-terra 0.22.2 which will drop the tweedledum dependency for Python 3.11

hhorii:
cd6bb7d and 0b486fa will fix issues in "Build / wheel (ubuntu-latest)". With two commits, OpenBLAS is installed manually in `manylinux2014_i686`.

tools/install_openblas_i686.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+wget https://github.com/xianyi/OpenBLAS/releases/download/v0.3.21/OpenBLAS-0.3.21.zip
+unzip OpenBLAS-0.3.21.zip
+cd OpenBLAS-0.3.21
+make TARGET=NORTHWOOD
```
mtreinish, line 6: Why northwood? What if I want to run on the AMD K6-2 which I had before I got my Pentium 4 back in 2003?

tools/install_openblas_i686.sh
```diff
@@ -0,0 +1,7 @@
...
+make TARGET=NORTHWOOD PREFIX=/usr/ install
```
mtreinish, line 7:
```suggestion
make TARGET=P2
make TARGET=P2 PREFIX=/usr/ install
```

tools/install_openblas_i686.sh
hhorii: At least, I needed to set `TARGET` to build to work around this error in OpenBLAS build on i686:
```
strmm_kernel_LN.o: No such file or directory
```
Maybe the `TARGET` should be old enough, like `P2`.

hhorii: I found at least KATMAI is needed to build openblas https://github.com/Qiskit/qiskit-aer/pull/1653/commits/f410260547086d7c1eea5fb1ad11c75c9d8518d8.

mtreinish: I guess we need SSE or MMX to run openblas. Hopefully nobody tries to run aer with a CPU >23 yrs old :)

mtreinish:
@hhorii I think this should be working now if you want to review, if it looks good to you we can merge this and include it for backport in 0.11.2. (also thanks for the help in setting up the build for i686)

mtreinish:
The other related thing we should do is also deprecate support for Python 3.7 since that will go EoL over the summer. I'll do that in a separate PR.

hhorii (approved):

Issue #1641, opened 2022-11-03
`AerStatevector(QuantumVolume(1))` raises an error


hhorii:
### Informations

- **Qiskit Aer version**: 0.11.1
- **Python version**: 3.8
- **Operating system**: MacOS

### What is the current behavior?

`AerStatevector(QuantumVolume(1))` raises an error.
This is because `QuantumVolume(1).data[0].operation.definition` returns an empty `QuantumCircuit`, which return `True` with `not` operation.

### Steps to reproduce the problem

```
from qiskit.circuit.library import QuantumVolume
from qiskit_aer.quantum_info.states import AerStatevector
state = AerStatevector(QuantumVolume(1))
```

### What is the expected behavior?

No error.

### Suggested solutions

https://github.com/Qiskit/qiskit-aer/blob/d13c2a516e6442941ef9f409d6ac2309d49f8279/qiskit_aer/quantum_info/states/aer_statevector.py#L207

This test expression should check `== None`.

This issue is reported by @derwind

hhorii:
#1643 resolved this.

Issue #1640, opened 2022-11-03
`AerState` uses instructions unsupported in `matrix_product_state` method


hhorii:
### Informations

- **Qiskit Aer version**: 0.11.1
- **Python version**: 3.8
- **Operating system**: MacOS

### What is the current behavior?

`AerState` uses `mc*` instructions, which are not supported in `matrix_product_state` method

### Steps to reproduce the problem

```
from qiskit_aer.quantum_info.states import AerStatevector
from qiskit_aer.quantum_info.states.aer_state import AerState
from qiskit.circuit.library import QFT

AerStatevector(QFT(4), method='statevector') #ok
AerStatevector(QFT(4), method='matrix_product_state') #ng
```

### What is the expected behavior?

Pass the above code.

### Suggested solutions

`AerState` supports `apply_u` and `apply_cx` and `_aer_evolve_circuit` checks basic gates of a method.

Note:
This bug is reported by @derwind.

hhorii:
#1643 resolved this.

PR #1643, opened 2022-11-04
Use more gates for AerState to call state_controller

hhorii:
### Summary

Solve #1640 and #1641 by updateing `_aer_evolve_instruction`.

### Details and comments

MC* instructions were mainly used in `AerStatevector`. However, these instructions are not supported in some methods, such as `matrix_product_state`, and an exception is thrown if they are called. This PR fixes this bug by using `X`. `Y`, `Z`, `CX`. `CY`, `CZ`, `H` and `U`.

### DIFF
--- a/qiskit_aer/quantum_info/states/aer_state.py
+++ b/qiskit_aer/quantum_info/states/aer_state.py
@@ -358,7 +358,7 @@ class AerState:
         self._native_state.apply_mcphase(control_qubits + [target_qubit], phase)
 
     def apply_h(self, target_qubit):
-        """apply a u operation."""
+        """apply a h operation."""
         self._assert_allocated_or_mapped()
         self._assert_in_allocated_qubits(target_qubit)
         # update state
@@ -371,21 +371,21 @@ class AerState:
         # update state
         self._native_state.apply_u(target_qubit, theta, phi, lamb)
 
-    def apply_cu(self, control_qubit, target_qubit, theta, phi, lamb):
-        """apply a u operation."""
+    def apply_cu(self, control_qubit, target_qubit, theta, phi, lamb, gamma):
+        """apply a cu operation."""
         self._assert_allocated_or_mapped()
         self._assert_in_allocated_qubits(control_qubit)
         self._assert_in_allocated_qubits(target_qubit)
         # update state
-        self._native_state.apply_cu([control_qubit, target_qubit], theta, phi, lamb)
+        self._native_state.apply_cu([control_qubit, target_qubit], theta, phi, lamb, gamma)
 
-    def apply_mcu(self, control_qubits, target_qubit, theta, phi, lamb):
+    def apply_mcu(self, control_qubits, target_qubit, theta, phi, lamb, gamma):
         """apply a mcu operation."""
         self._assert_allocated_or_mapped()
         self._assert_in_allocated_qubits(control_qubits)
         self._assert_in_allocated_qubits(target_qubit)
         # update state
-        self._native_state.apply_mcu(control_qubits + [target_qubit], theta, phi, lamb)
+        self._native_state.apply_mcu(control_qubits + [target_qubit], theta, phi, lamb, gamma)
 
     def apply_mcswap(self, control_qubits, qubit0, qubit1):
         """apply a mcswap operation."""
--- a/qiskit_aer/quantum_info/states/aer_state.py
+++ b/qiskit_aer/quantum_info/states/aer_state.py
@@ -377,7 +377,7 @@ class AerState:
         self._assert_in_allocated_qubits(control_qubit)
         self._assert_in_allocated_qubits(target_qubit)
         # update state
-        self._native_state.apply_cu([target_qubit], theta, phi, lamb)
+        self._native_state.apply_cu([control_qubit, target_qubit], theta, phi, lamb)
 
     def apply_mcu(self, control_qubits, target_qubit, theta, phi, lamb):
         """apply a mcu operation."""
--- a/qiskit_aer/quantum_info/states/aer_state.py
+++ b/qiskit_aer/quantum_info/states/aer_state.py
@@ -280,6 +280,21 @@ class AerState:
         # update state
         self._native_state.apply_diagonal(qubits, diag)
 
+    def apply_x(self, target_qubit):
+        """apply a x operation."""
+        self._assert_allocated_or_mapped()
+        self._assert_in_allocated_qubits(target_qubit)
+        # update state
+        self._native_state.apply_x(target_qubit)
+
+    def apply_cx(self, control_qubit, target_qubit):
+        """apply a cx operation."""
+        self._assert_allocated_or_mapped()
+        self._assert_in_allocated_qubits(control_qubit)
+        self._assert_in_allocated_qubits(target_qubit)
+        # update state
+        self._native_state.apply_cx([control_qubit, target_qubit])
+
     def apply_mcx(self, control_qubits, target_qubit):
         """apply a mcx operation."""
         self._assert_allocated_or_mapped()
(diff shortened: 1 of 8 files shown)


qiskit_aer/quantum_info/states/aer_state.py
```diff
@@ -312,6 +357,28 @@ def apply_mcphase(self, control_qubits, target_qubit, phase):
...
+        # update state
+        self._native_state.apply_u(target_qubit, theta, phi, lamb)
+
+    def apply_cu(self, control_qubit, target_qubit, theta, phi, lamb):
+        """apply a u operation."""
+        self._assert_allocated_or_mapped()
+        self._assert_in_allocated_qubits(control_qubit)
+        self._assert_in_allocated_qubits(target_qubit)
+        # update state
+        self._native_state.apply_cu([target_qubit], theta, phi, lamb)
```
jakelishman, line 380: This doesn't refer to the `control_qubit`.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -184,27 +184,56 @@ def _aer_evolve_circuit(aer_state, circuit, qubits):
...
+
+        def _assert_qubits(expected):
+            if len(qubits) != expected:
+                raise AerError(
+                    f"Cannot apply a {expected}-qubit instruction {inst} to qubits {qubits}"
+                )
+
+        def _assert_params(expected):
+            if len(params) < expected:
+                raise AerError(f"Need {expected} parameters, but actually {len(params)}")
```
jakelishman, line 198: Is there a reason not to be exact `==` here?

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -184,27 +184,56 @@ def _aer_evolve_circuit(aer_state, circuit, qubits):
...
             aer_state.apply_reset(qubits)
+        elif inst.name in ['barrier', 'id']:
+            pass
         else:
             definition = inst.definition
-            if definition is inst:
-                raise AerError('cannot decompose ' + inst.name)
-            if not definition:
+            if definition is inst or definition is None:
+                print(inst.definition)
```
jakelishman, line 236: `print` probably isn't the right choice here - is it a left-over debugging statement?

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -184,27 +184,56 @@ def _aer_evolve_circuit(aer_state, circuit, qubits):
...
-                                params[0], params[1], params[2])
-        elif inst.name in ['x', 'cx', 'ccx']:
-            aer_state.apply_mcx(qubits[0:len(qubits) - 1], qubits[len(qubits) - 1])
-        elif inst.name in ['y', 'cy']:
-            aer_state.apply_mcy(qubits[0:len(qubits) - 1], qubits[len(qubits) - 1])
-        elif inst.name in ['z', 'cz']:
-            aer_state.apply_mcz(qubits[0:len(qubits) - 1], qubits[len(qubits) - 1])
+            _assert_qubits(1)
+            _assert_params(3)
+            aer_state.apply_u(qubits[0], params[0], params[1], params[2])
```
jakelishman, line 203: If you're unpacking the qubits and parameters, you probably don't need to waste time doing the assertions - you'll just get an `IndexError` if they're out of bounds anyway, which I think is probably fine for something that's meant to be a bit higher performance?

It's up to you, really - I feel like we could go either way.

src/controllers/state_controller.hpp
```diff
@@ -858,6 +951,42 @@ void AerState::apply_mcphase(const reg_t &qubits, const std::complex<double> pha
...
+}
+
+void AerState::apply_u(const uint_t qubit, const double theta, const double phi, const double lambda) {
+  assert_initialized();
+
+  Operations::Op op;
+  op.type = Operations::OpType::gate;
+  op.name = "u";
+  op.qubits = {qubit};
+  op.params = {theta, phi, lambda, 0.0};
```
jakelishman, line 973: Does `u` have 4 parameters in Aer?

src/controllers/state_controller.hpp
```diff
@@ -858,6 +951,42 @@ void AerState::apply_mcphase(const reg_t &qubits, const std::complex<double> pha
...
+void AerState::apply_cu(const reg_t &qubits, const double theta, const double phi, const double lambda) {
+  assert_initialized();
+
+  Operations::Op op;
+  op.type = Operations::OpType::gate;
+  op.name = "cu";
+  op.qubits = qubits;
+  op.params = {theta, phi, lambda, 0.0};
+
+  buffer_op(std::move(op));
```
jakelishman, line 987: This doesn't seem to have an analogue in the Python components - is it meant to?  Also, Terra's `cu` takes 4 parameters like Aer, so it could accept the last one in the signature as well.

src/simulators/matrix_product_state/matrix_product_state.hpp
```diff
@@ -828,10 +828,10 @@ std::vector<reg_t> State::
     rnds_list.push_back(rands);
   }
 
-  #pragma omp parallel if (BaseState::threads_ > 1) num_threads(BaseState::threads_)
+  //#pragma omp parallel if (BaseState::threads_ > 1) num_threads(BaseState::threads_)
   {
     MPS temp;
-    #pragma omp for
+    //#pragma omp for
     for (int_t i=0; i<static_cast<int_t>(shots);  i++) {
```
jakelishman, line 835: Are these intentional?

jakelishman (commented):
A couple of minor things.  At some point we might want to consider better ways of handling the large `if/elif` chain; it's possible that that will start having annoying overhead if done in Python space. I'd be hesitant to jump to hashmap lookups in Python, but it might be worth considering in the future if there are better ways.

src/controllers/state_controller.hpp
jakelishman: Oh wait whoops sorry - it _does_ have a Python-space component, and I even commented on it.  The 4-parameter thing is still a question, though.

qiskit_aer/quantum_info/states/aer_statevector.py
hhorii: Maybe `IndexError` will be consistent with the other parts. I would like to take `_assert*`.

src/controllers/state_controller.hpp
hhorii: https://github.com/Qiskit/qiskit-aer/blob/cd8710f3ec83391dbbe7f67d0c185c2b0538b17f/src/simulators/statevector/statevector_state.hpp#L264-L269

Aer supports U4.

src/simulators/matrix_product_state/matrix_product_state.hpp
hhorii: Sorry. This is just for debugging. I need to revert it... Thanks!

qiskit_aer/quantum_info/states/aer_statevector.py
hhorii: sorry. it is just for debugging.

src/controllers/state_controller.hpp
jakelishman: Sorry, the thing I mean is that Terra's `cu` is a 4-parameter gate.  This definition in Aer only takes three parameters, so there seems to be a mismatch of expectations between them.

qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -184,27 +184,36 @@ def _aer_evolve_circuit(aer_state, circuit, qubits):
     @staticmethod
     def _aer_evolve_instruction(aer_state, inst, qubits):
         """Apply instruction into aer_state"""
+
         params = inst.params
+
         if inst.name in ['u', 'u3']:
```
jakelishman, line 190: Given other conversations, should `cu` be in this list? It doesn't seem to be (maybe I can't spot it), and I'm guessing it's not tested?

jakelishman (approved):
Looks good to me now, thanks for the fixes, and for looking into the CU thing.

PR #1649, opened 2022-11-07
Fix device noise model from BackendV2 not to have excessive quantum errors on measures/resets

### LINKED ISSUE #1648: Noise model created from BackendV2 contains excessive noises on measures.
### Informations
- **Qiskit Aer version**: 0.11.1

Noise model created from `BackendV2` contains excessive (relaxation and depolarizing) `QuantumError`s on measures while `ReadoutError`s should be sufficient for measures. This is due to a bug in `qiskit_aer.noise.device.models.basic_device_gate_errors` where it returns QuantumErrors on non-Gate objects incorrectly after #1484.

```
from qiskit.providers.fake_provider import FakeManilaV2
from qiskit_aer.noise.device.models import basic_device_gate_errors

gate_errors = basic_device_gate_errors(None, target=FakeManilaV2().target)
errors_on_measure = [name for name, _, _ in gate_errors if name == "measure"]
```

`basic_device_gate_errors` never return errors on non-Gate objects.

itoko:
### Summary
Fixes #1648 by updating device noise models not to have excessive quantum errors on non-gate instructions such as measures and resets when creating them from BackendV2.

### Details and comments
This commit fixes `qiskit_aer.noise.device.models.basic_device_gate_errors` not to generate `QuantumError`s (neither depolarizing nor relaxation) for `Measure` instruction and not to generate depolarizing errors for other non-Gate instructions when `Target` object is supplied.

### DIFF
--- a/qiskit_aer/noise/device/models.py
+++ b/qiskit_aer/noise/device/models.py
@@ -163,9 +163,6 @@ def basic_device_gate_errors(properties=None,
             temperature=temperature
         )
 
-    # Initilize empty errors
-    depol_error = None
-    relax_error = None
     # Generate custom gate time dict
     custom_times = {}
     relax_params = []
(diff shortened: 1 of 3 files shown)


itoko:
I might need to keep relaxation errors on resets (different from measures). I'll look into it.

chriseclectic (approved):

Issue #1647, opened 2022-11-06
Segfault when running certain circuits with the GPU statevector simulator, `AerPauliExpectation`, and certain noise models


HaoTy:
### Informations

- **Qiskit Aer version**:
```
qiskit                   0.39.2
qiskit-aer               0.11.1
qiskit-aer-gpu           0.11.1
qiskit-ibmq-provider     0.19.2
qiskit-optimization      0.4.0
qiskit-terra             0.22.2
```
- **Python version**: 3.9.12
- **Operating system**: CentOS Stream 8
- **GPU**: A100-SXM4-80GB
- **CUDA version**: 11.7.0

### What is the current behavior?
Running certain circuits (e.g. the QAOA expectation value of an 18-qubit `SKModel` instance, see below) with the GPU statevector simulator, `AerPauliExpectation`, and certain noise models (e.g. 2-qubit depolar error) causes segmentation fault. 
Using the CPU statevector simulator does not have this issue. 
Some other circuits, for example, the QAOA expectation value of the `Maxcut` problem, do not have this issue. 
Some other noise models, including ideal simulations, Pauli errors, and 1-qubit depolar errors, do not have this issue.
16 qubits or below do not have this issue (the exact number may depend on machine specs).

### Steps to reproduce the problem
```python
import numpy as np
import networkx as nx
from qiskit_aer import AerSimulator
from qiskit_aer.noise import NoiseModel
from qiskit_aer.noise.errors.standard_errors import depolarizing_error
from qiskit.algorithms import QAOA
from qiskit.algorithms.optimizers import SPSA
from qiskit.utils import QuantumInstance
from qiskit.opflow import AerPauliExpectation
from qiskit_optimization.applications import Maxcut, SKModel
from qiskit.utils import algorithm_globals
from docplex.mp.model import Model

n = 18
p = 1
seed = 0

# Maxcut does not produce segfault
# graph = nx.random_regular_graph(3, n, seed)
# problem = Maxcut(graph).to_quadratic_program()

# SKModel with n=18 or more produces segfault
problem = SKModel(n, seed).to_quadratic_program()

# Some other problems can also cause this issue
# number_set = np.random.default_rng(seed).integers(2 ** (n // 2), size=n) + 1
# mdl = Model(name="Number partitioning")
# x = {i: mdl.binary_var(name=f"x_{i}") for i in range(n)}
[...]

Issue #1646, opened 2022-11-06
Cuda invalid configuration error when using the statevector GPU simulator with a large number of qubits


HaoTy:
### Informations

- **Qiskit Aer version**:
```
qiskit                   0.39.2
qiskit-aer               0.11.1
qiskit-aer-gpu           0.11.1
qiskit-ibmq-provider     0.19.2
qiskit-optimization      0.4.0
qiskit-terra             0.22.2
```
- **Python version**: 3.9.12
- **Operating system**: CentOS Stream 8
- **GPU**: A100-SXM4-80GB
- **CUDA version**: 11.7.0

### What is the current behavior?
Cuda gives an invalid configuration error when using the statevector GPU simulator with a large number of qubits (more than 30 in my case).
```
qiskit.exceptions.QiskitError: 'Circuit execution failed: ERROR:  [Experiment 0] for_each: failed to synchronize: cudaErrorInvalidConfiguration: invalid configuration argument'
```
Note that `AerPauliExpectation` is used and not until n=36 the memory becomes insufficient:
```
qiskit.exceptions.QiskitError: 'Circuit execution failed: ERROR:  [Experiment 0] Insufficient memory to run circuit circuit-115-714 using the statevector_gpu simulator. Required memory: 1048576M, max memory: 515313M (Host) + 81104M (GPU)'
```

### Steps to reproduce the problem
```python
import numpy as np
import networkx as nx
from qiskit_aer import AerSimulator
from qiskit.algorithms import QAOA
from qiskit.algorithms.optimizers import SPSA
from qiskit.utils import QuantumInstance
from qiskit.opflow import AerPauliExpectation
from qiskit_optimization.applications import Maxcut
from qiskit.utils import algorithm_globals

n = 32
p = 1
seed = 0

graph = nx.random_regular_graph(3, n, seed)
problem = Maxcut(graph).to_quadratic_program()
H, offset = problem.to_ising()

algorithm_globals.random_seed = seed
backend = AerSimulator(method="statevector", device="GPU")
quantum_instance = QuantumInstance(
    backend=backend,
    seed_simulator=seed,
    seed_transpiler=seed,
)

algorithm = QAOA(
    SPSA(),
    reps=p,
    quantum_instance=quantum_instance,
    expectation=AerPauliExpectation(),
)
algorithm._check_operator_ansatz(H)
energy_evaluation= algorithm.get_energy_evaluation(H)
[...]

PR #1650, opened 2022-11-07
Fix for GPU simulator, memory allocation

doichanj:
### Summary
This is fix for issue #1646 and issue #1647 

### Details and comments
Aer tried to allocate chunks fit in the free memory on GPU, but it caused memory allocation error.
This fix allocates 80% of free memory of GPU and rest of chunks are allocated on CPU memory.

Another fix is allocation of matrix buffer on GPU.
For noise sampling, maximum matrix size was not correct (was set by using original circuit before noise sampling)

### DIFF
--- a/releasenotes/notes/fix_GPU_statevector-715da5ead0a59fb5.yaml
+++ b/releasenotes/notes/fix_GPU_statevector-715da5ead0a59fb5.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    This is fix for GPU statevector simulator.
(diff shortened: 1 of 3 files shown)


hhorii:
What does "80%" mean? Should it be configurable? Or is this fixed number enough for practical?

doichanj:
Thrust internally allocates work buffer on GPU, so some memory space should be remained for Thrust.
80% is for safe, but I think 512MB is enough. I changed 80% to 512MB by fixed definition for memory reservation for Thrust

hhorii (approved):

PR #1663, opened 2022-11-18
Set different seed for each sampling in AerStatevector

hhorii:
### Summary

Previously seed is not initialized in AerStatevector and then sampled results are always same. With this commit, a seed is initialized for each sampling and sampled results can be vary.

### Details and comments

Previously, sampled counts are always same in multiple calls because `AerStatevector` uses the same seed to generate random values which are used in processing `sample_counts()`. More specifically, `_seed` is not initialized in `state_controller` if `seed_simulator` is not set. For example, following two sampled values `count0` and `count1` are the same.
```
circ = QuantumVolume(5, seed=1111)
state = AerStatevector(circ)

shots = 1024
counts0 = state.sample_counts(shots, qargs=range(5))
counts1 = state.sample_counts(shots, qargs=range(5))
```
Even if `seed_simulator` is set, the value is reused to initialize `RngEngine` and sampled counts become same.

With this PR,  `_seed` is initialized with `std::random_device()` and renewed `seed_simulator` for each sampling, and then sampled counts are always different in the same `AerStatevector` instance.

### DIFF
--- a/qiskit_aer/quantum_info/states/aer_statevector.py
+++ b/qiskit_aer/quantum_info/states/aer_statevector.py
@@ -116,7 +116,7 @@ class AerStatevector(Statevector):
 
         configs = self._aer_state.configuration()
         if 'seed_simulator' in configs:
-            configs['seed_simulator'] += 1
+            configs['seed_simulator'] = int(configs['seed_simulator']) + 1
         self._aer_state = AerState(**configs)
 
         self._aer_state.initialize(self._data, copy=False)
--- a/qiskit_aer/quantum_info/states/aer_statevector.py
+++ b/qiskit_aer/quantum_info/states/aer_statevector.py
@@ -113,7 +113,12 @@ class AerStatevector(Statevector):
         else:
             qubits = np.array(qargs)
         self._aer_state.close()
-        self._aer_state = AerState(**self._aer_state.configuration())
+
+        configs = self._aer_state.configuration()
+        if 'seed_simulator' in configs:
+            configs['seed_simulator'] += 1
+        self._aer_state = AerState(**configs)
+
         self._aer_state.initialize(self._data, copy=False)
         samples = self._aer_state.sample_memory(qubits, shots)
         self._data = self._aer_state.move_to_ndarray()
--- a/qiskit_aer/quantum_info/states/aer_statevector.py
+++ b/qiskit_aer/quantum_info/states/aer_statevector.py
@@ -84,6 +84,13 @@ class AerStatevector(Statevector):
         self._result = None
         self._configs = configs
 
+    def seed(self, value=None):
+        """Set the seed for the quantum state RNG."""
+        if value is None or isinstance(value, int):
+            self._aer_state.set_seed(value)
+        else:
+            raise AerError(f'This seed is not supported: type={value.__class__}, value={value}')
+
     def _last_result(self):
         if self._result is None:
             self._result = self._aer_state.last_result()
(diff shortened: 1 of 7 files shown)


qiskit_aer/quantum_info/states/aer_statevector.py
```diff
@@ -113,7 +113,12 @@ def sample_memory(self, shots, qargs=None):
         else:
             qubits = np.array(qargs)
         self._aer_state.close()
-        self._aer_state = AerState(**self._aer_state.configuration())
+
+        configs = self._aer_state.configuration()
+        if 'seed_simulator' in configs:
+            configs['seed_simulator'] = int(configs['seed_simulator']) + 1
+        self._aer_state = AerState(**configs)
```
jakelishman, line 120: This seems like a slightly odd split in responsibilities.  Perhaps it would be more convenient for now, and for future development if `AerState` gains an `advance_rng` method or something like that mutates the `AerState` in place.  It'll be easier to re-use between the different Aer `quantum_info` classes you're planning that way, it gives clearer hints as to what's happening, and modifying it is kind of an `AerState` internal feature that we don't particularly want client classes to need to know about.

src/controllers/state_controller.hpp
```diff
@@ -326,7 +326,7 @@ class AerState {
   bool initialized_ = false;
   uint_t num_of_qubits_ = 0;
   RngEngine rng_;
-  int seed_;
+  int seed_ = std::random_device()();
```
jakelishman, line 329: What was happening before this change?

jakelishman (commented):
Is it _always_ going to be the case that a user will want the RNG to advance by calling `sample_memory`?  I'm thinking probably yes, but I'm just mentioning it as a question.

qiskit_aer/quantum_info/states/aer_statevector.py
hhorii: `QuantumState` advances RNG. So, `AerStatevector` also needs to advance it in `sample_memory()`.
https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/quantum_info/states/quantum_state.py#L269

However, instead of manually update of `seed_simulator` by increasing `1`, it is better to succeed the state of `mt19937` from the previous `AerState`.

hhorii:
@jakelishman With 4b80871, `AerStatevector`  uses a random number generator of `mt19937`. Also, `AerStatevector` overrides `seed()` to enable reproducing same returns from `sample_memory()` by calling `seed()` before each `sample_memory()`.

src/controllers/state_controller.hpp
```diff
@@ -661,13 +658,21 @@ void AerState::initialize() {
 
   state_->initialize_qreg(num_of_qubits_);
   state_->initialize_creg(num_of_qubits_, num_of_qubits_);
-  rng_.set_seed(seed_);
```
jakelishman, line 664: Do we still need to track the `seed_` member after this change?  I see it's still used in `clear_ops()`, but it's not 100% clear to me why that function should reset the RNG.  Admittedly, I'm not exactly clear on the circumstances in which it's called.

jakelishman (commented):
One comment aside, this looks like a much cleaner split of responsibilities to me, thanks!

src/controllers/state_controller.hpp
hhorii: https://github.com/Qiskit/qiskit-aer/blob/b5b0c4dc271e8a986b7cd7d1fe3d40e3862b1d51/src/controllers/state_controller.hpp#L1228

This `seed_` is to be returned as a member of a result. `AerState.last_result()` returns it. We may not need to construct a result as `AerSimulator` does for each call of methods, but it is not expensive to keep `seed_`, I think.

jakelishman (approved):

PR #1666, opened 2022-11-22
Fix data-flow analysis in control-flow jump/mark

### LINKED ISSUE #1665: Issue with nested if_test
### Informations

- **Qiskit version**: 0.38.0
- **Qiskit Aer version**: 0.11.0
- **Python version**: 3.10.8
- **Operating system**: Mac OS 12.6, M1 Max

Does not produce expected output when nested if_test statement is used. 

```
from qiskit import IBMQ, Aer, QuantumCircuit, QuantumRegister, ClassicalRegister, execute
import numpy as np  

backend_sim = Aer.get_backend('qasm_simulator')
pi = np.pi

q = QuantumRegister(3)
c = ClassicalRegister(3)
circuit = QuantumCircuit(q, c)
circuit.initialize('010', circuit.qubits)

circuit.measure(q[0], c[0])
circuit.measure(q[1], c[1])

with circuit.if_test((c[0], 1)):
    with circuit.if_test((c[1], 0)):
        circuit.x(q[2])

with circuit.if_test((c[0], 0)):
    with circuit.if_test((c[1], 1)):
        circuit.x(q[2])

circuit.measure(q, c)

job = backend_sim.run(circuit, backend_sim, shots = 100)
results = job.result().get_counts()
print(results)
```
Output: {'010': 100}

Expected output bit string is '110'

### Suggested solutions
N/A

jakelishman:
### Summary

During the inlining of control-flow into jump/mark instructions, the data-flow graph was losing information.  The tree-like form within `QuantumCircuit` implicitly creates a basic-block structure, which groups the whole control-flow body into a single block that must be executed atomicly (as seen by the outer circuit).  When inlining the body, this is no longer the case, and we have to take more care to ensure that topological iteration through the data-flow graph does not accidentally move jump/mark instructions into incorrect places, nor allow unrelated instructions to appear between jump/mark pairs if they were no originally in that control-flow block.

There are two components here; the former is solved by ensuring the jump/mark instructions span the full data width including clbits, and the latter is solved by placing full circuit width barriers around each control-flow logical component.

### Details and comments

Fix #1665.

### DIFF
--- a/qiskit_aer/backends/aer_compiler.py
+++ b/qiskit_aer/backends/aer_compiler.py
@@ -12,7 +12,10 @@
 """
 Compier to convert Qiskit control-flow to Aer backend.
 """
-from qiskit.circuit import QuantumCircuit
+
+import itertools
+
+from qiskit.circuit import QuantumCircuit, Clbit
 from qiskit.extensions import Initialize
 from qiskit.pulse import Schedule, ScheduleBlock
 from qiskit.circuit.controlflow import (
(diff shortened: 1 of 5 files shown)


hhorii (approved):
> During the inlining of control-flow into jump/mark instructions, the data-flow graph was losing information. 

Decomposing an instruction that has a sub-circuit is more complicated than I expected.  Thank you for your fix.

> When inlining the body, this is no longer the case, and we have to take more care to ensure that topological iteration through the data-flow graph does not accidentally move jump/mark instructions into incorrect places, 

This also makes sense to me.

jakelishman:
Yeah, I didn't think about it at first either.  It's on my mind now because I'd been discussing with Kevin (@kdk) about better data structures for representing control-flow graphs alongside data-flow ones, and this was one of the major problems we identified.

PR #1671, opened 2022-11-29
Deprecate Python 3.7 support in Qiskit Aer

mtreinish:
### Summary

This commit deprecates Python 3.7 support. During the deprecation window we will continue to fully support Python 3.7 and test it in CI, but it will emit a warning that we will be removing support for 3.7 in the future to inform users to upgrade to Python 3.8 or newer. The removal will occur during a future release at some point over the summer as the upstream Python EoL date is in June 2023. [1]

### Details and comments

[1] https://devguide.python.org/versions/

### DIFF
--- a/qiskit_aer/__init__.py
+++ b/qiskit_aer/__init__.py
@@ -52,14 +52,14 @@ Exceptions
    AerError
 """
 
-# https://github.com/Qiskit/qiskit-aer/issues/1
-# Because of this issue, we need to make sure that Numpy's OpenMP library is initialized
-# before loading our simulators, so we force it using this ugly trick
 import platform
 import sys
 import warnings
 
 
+# https://github.com/Qiskit/qiskit-aer/issues/1
+# Because of this issue, we need to make sure that Numpy's OpenMP library is initialized
+# before loading our simulators, so we force it using this ugly trick
 if platform.system() == "Darwin":
     import numpy as np
     np.dot(np.zeros(100), np.zeros(100))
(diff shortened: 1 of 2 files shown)


qiskit_aer/__init__.py
```diff
@@ -56,6 +56,10 @@
 # Because of this issue, we need to make sure that Numpy's OpenMP library is initialized
 # before loading our simulators, so we force it using this ugly trick
 import platform
+import sys
+import warnings
```
jakelishman, line 60: Given the above comment, we probably want to keep it in close proximity to the weirdo OpenMP initialisation (if it's still even needed) - maybe move these imports below that or move the comment to immediately above the `if` rather than the import?

mtreinish: I had the import below originally, pylint didn't like that so I updated it in https://github.com/Qiskit/qiskit-aer/pull/1671/commits/f1dc6b70f202dccb567ad65fed1c0ce161e981b1. I'll move the comment above the `if`. As for whether it's still needed or not I have no idea, it dates back to the very beginning of the project (and predates my involvement).

jakelishman: ugh pylint lol

jakelishman:
The ECR-gate commit has snuck in here by accident, I think.

mtreinish:
sigh, that's I get for writing this before morning coffee. Let me fix my branch

mtreinish:
Ok, it should be fixed now

jakelishman (approved):

Issue #1638, opened 2022-10-30
Kraus decompostion of phase amplitude damping error is not appropriate


smiserman:
### What is the expected behavior?
The `phase_amplitude_damping_error(param_amp, param_phase, excited_state_population=0, canonical_kraus=True)` in Qiskit allows only the parameter range $p_a + p_p \le 1$, although all parameters $p_a, p_p \in [0,1]$ must be valid and this restriction is not necessary. Furthermore one should relate the phase parameter $p_p$ to the pure phase relaxation time $T_\phi$ by $p_p = 1- e^{t/T_\phi}$ with $T_\phi = T_1 T_2/(2T_1-T_2)$ where $T_1, T_2$ are the relaxation time constants of a specific qubit and the constants of the `thermal_relaxation_error(t1, t2, time, excited_state_population=0)`. All this can be achieved by choosing the appropriate Kraus matrices.

For a minimal and appropriate set please see[ my blog in stackexchange](https://quantumcomputing.stackexchange.com/questions/28720/qiskit-kraus-decompostion-of-phase-amplitude-damping-error-is-not-appropriate)

And furthermore there is a bug in the function `_excited_population(freq, temperature)`. The Boltzmann equilibrium distribution for the excited state is given by
$$p_1 =\frac{1}{1+e^{\Delta E/k_B T}} \simeq e^{-h f/k_B T}$$
where $h,\, k_B$ are the Planck and Boltzmann constants and $f=\Delta E/h,\, T$ are the frequency (energy gap) and temperature of the qubit. The source code shows an incorrect factor of 2, resulting in a population probability of $O(10^{-14})$ instead of $O(10^{-7})$ with $f \approx 5 \mathrm{Ghz}$ and $T \approx 15 \mathrm{mK}$. The error is certainly negligible in the simulations, but should still be corrected!

chriseclectic:
These are equivalent, it's just a difference in convention. The convention for Aer's combined phase and amplitude damping was taken from Joseph Emerson's lecture notes on Open Quantum Systems and is equivalent to `thermal_relaxation_error` if you set `param_amp = 1 - exp(-t/T1)` and `param_phase = exp(-t/T1) - exp(-2t/T2)`, or in terms of `T_phi`, `phase_param = exp(-t/T1) * (1-exp(-t/T_phi)`. If you want to use a different convention you can always write your own functions for your research.

[...]

PR #1672, opened 2022-12-01
Fix incorrect computation of excited state population

itoko:
### Summary
Fix a bug in computation of the excited population from qubit frequency and temperature that can be used in noise model construction. 

### Details and comments
As reported in #1638, there was an incorrect exponential factor of 2 and hence the excited population was incorrect (almost squared). This fix will affect users who generate a noise model using `NoiseModel.from_backend` with `temperature` option while it won't affect users of the default noise model that assumes no excitation (temperature=0).

### DIFF
--- a/releasenotes/notes/fix-temperature-a9c51c4599af3a49.yaml
+++ b/releasenotes/notes/fix-temperature-a9c51c4599af3a49.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixed a bug in computation of the excited population from qubit frequency and temperature
+    in noise model construction. Previously, there was an incorrect exponential factor of 2
+    and hence the excited population was incorrectly squared. This fix will affects users
+    who generate a noise model using :meth:`NoiseModel.from_backend` with ``temperature``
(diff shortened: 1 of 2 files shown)


hhorii (dismissed):

releasenotes/notes/fix-temperature-a9c51c4599af3a49.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixed a bug in computation of the excited population from qubit frequency and temperature
```
chriseclectic, line 4: I would reword to stress the error condition first eg:
> Fixed a bug in `NoiseModel.from_backend` where using the `temperature` kwarg with a non-default value would incorrectly compute the excited state population for the specified temperature. Previously, there was an additional factor of 2 in the Boltzman distribution calculation leading to an incorrect smaller value for the excited state population.

chriseclectic (commented):
Looks good. Suggestion for wording of release note

hhorii (approved):

PR #1674, opened 2022-12-01
Add ECR gate support to aer simulators

mtreinish:
### Summary

This commit adds support for the ECR (echo cross resonance) gate as described in [1] and [2] to aer. This is an entangling gate used on some IBM backends and being able for aer to simulate is import to be able to simulate those backends. Support for the gate is added to the density matrix, statvector, super operator, and unitary simulators.

### Details and comments

[1] https://arxiv.org/abs/2008.08571
[2] https://qiskit.org/documentation/stubs/qiskit.circuit.library.ECRGate.html

TODO: 

- [x] Add tests

### DIFF
--- a/qiskit_aer/backends/backend_utils.py
+++ b/qiskit_aer/backends/backend_utils.py
@@ -57,13 +57,13 @@ BASIS_GATES = {
         'rzz', 'rzx', 'ccx', 'cswap', 'mcx', 'mcy', 'mcz', 'mcsx',
         'mcp', 'mcphase', 'mcu', 'mcu1', 'mcu2', 'mcu3', 'mcrx', 'mcry', 'mcrz',
         'mcr', 'mcswap', 'unitary', 'diagonal', 'multiplexer',
-        'initialize', 'delay', 'pauli', 'mcx_gray'
+        'initialize', 'delay', 'pauli', 'mcx_gray', 'ecr'
     ]),
     'density_matrix': sorted([
         'u1', 'u2', 'u3', 'u', 'p', 'r', 'rx', 'ry', 'rz', 'id', 'x',
         'y', 'z', 'h', 's', 'sdg', 'sx', 'sxdg', 't', 'tdg', 'swap', 'cx',
         'cy', 'cz', 'cp', 'cu1', 'rxx', 'ryy', 'rzz', 'rzx', 'ccx',
-        'unitary', 'diagonal', 'delay', 'pauli',
+        'unitary', 'diagonal', 'delay', 'pauli', 'ecr',
     ]),
     'matrix_product_state': sorted([
         'u1', 'u2', 'u3', 'u', 'p', 'cp', 'cx', 'cy', 'cz', 'id', 'x', 'y', 'z', 'h', 's',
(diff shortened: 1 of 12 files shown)


src/framework/linalg/matrix_utils/matrix_defs.hpp
```diff
@@ -195,14 +196,20 @@ const cmatrix_t Matrix::SWAP =
                                    {{0, 0}, {1, 0}, {0, 0}, {0, 0}},
                                    {{0, 0}, {0, 0}, {0, 0}, {1, 0}}});
 
+const cmatrix_t Matrix::ECR =
+    Utils::make_matrix<complex_t>({{{0, 0}, {0, 0}, {1. / std::sqrt(2.), 0}, {0, 1. / std::sqrt(2.)}},
+                                   {{0, 0}, {0, 0}, {0, 1 / std::sqrt(2.)}, {1. / std::sqrt(2.), 0}},
+                                   {{1. / sqrt(2.), 0}, {0, -1. / std::sqrt(2.)}, {0, 0}, {0, 0}},
+                                   {{0, -1. / std::sqrt(2.)}, {1. / std::sqrt(2.), 0}, {0, 0}, {0, 0}}});
+
```
jakelishman, line 204: I think you've got an endianness difference in this matrix definition compared to Terra.

jakelishman: I just clicked through to Terra's documentation - they look to be flipped compared to how we actually implement the gate definition in code.

mtreinish: Sigh, I just copied it from the docs but I copied the matrix. I'll fix this in the morning.

jakelishman: Wait no, the docs are kind of right, they're just phrased a bit confusingly.  The top matrix is what Qiskit actually produces - the bottom one is what it would look like in big-endian.

hhorii:
Current Aer simulates an ECR gate by decomposing to RZX(pi/4), X, and RZX(-pi/4) via transpile. I guess one of main reasons to add it as a basis gate is for performance. But if we use gate fusion, maybe performance will be mostly same (because the current gate fusion does not fuse ECR gates, decomposition to RZX+X+RZX will be better in the current implementation).

mtreinish:
The primary motivation for adding it to the basis gate is for simulation of hardware, not really performance. As some IBM hardware is using ecr as the entangling gate we need to have it in the basis set so that we can simulate a circuit transpiled for those devices. Otherwise aer errors because ecr isn't a supported gate.

src/framework/linalg/matrix_utils/matrix_defs.hpp
chriseclectic: ECR gate looks wrong in qiskit. It should implement an RZX(pi/2) matrix, the echoing part only matters if you are decomposing.

chriseclectic (changes requested):
Terra looks to have incorrect gate def for ECR, do you want to wait till this issue is fixed in Terra? https://github.com/Qiskit/qiskit-terra/issues/9234

I think the correct matrix def should be
```
qi.Operator(RZXGate(np.pi/2)).data

[[ 0.70710678+0.j          0.        +0.j         -0.        -0.70710678j  0.        +0.j        ]
 [ 0.        +0.j          0.70710678+0.j          0.        +0.j          0.        +0.70710678j]
 [-0.        -0.70710678j  0.        +0.j          0.70710678+0.j          0.        +0.j        ]
 [ 0.        +0.j          0.        +0.70710678j  0.        +0.j          0.70710678+0.j        ]]
```

chriseclectic (approved):

qiskit_aer/backends/backend_utils.py
```diff
@@ -57,13 +57,13 @@
...
     ]),
     'density_matrix': sorted([
         'u1', 'u2', 'u3', 'u', 'p', 'r', 'rx', 'ry', 'rz', 'id', 'x',
         'y', 'z', 'h', 's', 'sdg', 'sx', 'sxdg', 't', 'tdg', 'swap', 'cx',
         'cy', 'cz', 'cp', 'cu1', 'rxx', 'ryy', 'rzz', 'rzx', 'ccx',
-        'unitary', 'diagonal', 'delay', 'pauli',
+        'unitary', 'diagonal', 'delay', 'pauli', 'ecr',
     ]),
     'matrix_product_state': sorted([
         'u1', 'u2', 'u3', 'u', 'p', 'cp', 'cx', 'cy', 'cz', 'id', 'x', 'y', 'z', 'h', 's',
```
hhorii, line 69: Because `matrix_product_state` supports `rzx` and `x`, it can also support `ecr`. I would like to cover it with other PR

hhorii (approved):

PR #1677, opened 2022-12-05
Specify package_data explicitly

### LINKED ISSUE #1457: `python setup.py install` should only copy relevant files
### Informations

- **Qiskit Aer version**: 0.10.3
- **Python version**: 3.10.2
- **Operating system**: Arch Linux

I'm trying to create a `PKGBUILD` for Arch Linux and I found some issues when building from source.

If I run

```
python setup.py build
python setup.py install --root="${pkgdir}/" --optimize=1 --skip-build
```

I would expect only files in `/usr/lib/python3.10/site-packages/qiskit/`. However, all `cmake`, `contrib` and even `src` folders are copied. This is not consistent with other python packages, and I guess is not intentional.

```
curl -O https://github.com/Qiskit/qiskit-aer/archive/refs/tags/0.10.3.tar.gz
tar -xf 0.10.3.tar.gz
cd qiskit-aer-0.10.3 && mkdir package
DISABLE_CONAN=ON python setup.py install --root=./package
ls -al package/usr
```

`pyton setup.py install` should only copy relevant files (i.e. `/usr/lib/python3.10/site-packages/qiskit/`)

### Suggested solutions

wshanks:
### Summary

Changes setup.py to avoid `include_package_data` because it results in extraneous files being included in the wheel.

### Details and comments

Currently, there are non-Python files needed in the qiskit_aer package: `VERSION.txt` and `instructions_table.csv`. These are included in the wheel through the use of the `include_package_data=True` option to `setup()`, but scikit-build, in a deviation from setuptools' behavior, ends up including everything in `MANIFEST.in` when this option is used, as reported
[here](https://github.com/scikit-build/scikit-build/issues/590). This behavior results in many extraneous files being included in the wheel. This PR works around this quirk by explicitly including the necessary files with the `package_data` argument of `setup()` instead of using `include_package_data`.

Closes #1457


mtreinish (approved):
This LGTM, thanks for fixing this. I was worried that the sdist would be missing the manifest files like this but the sdist jobs seemed to succeed without issue. My only concern moving forward is remembering to update this if we add any data files to the python package but that's a minor concern and not something that happens frequently

PR #1690, opened 2022-12-26
Remove deprecated module import

king-p3nguin:
### Summary
- Remove the import statement of the deprecated module in test.

### Details and comments
- The use of `qiskit.test.mock` is deprecated.


jakelishman (approved):
Thanks - this was replaced by `qiskit.providers.fake_provider`, so it looks like this import just got forgotten when the `FakeQuito` line two lines above turned up.

PR #1693, opened 2023-01-06
Remove setup_requires and forced pip installs in setup.py

### LINKED ISSUE #1688: setup_requires is deprecated
### Informations

- **Qiskit Aer version**: 0.11.2
- **Python version**: 3.11.1
- **Operating system**: Linux

qiskit-aer's setup.py specifies `setup_requires` and expects scikit-build to install its build requirements if they are missing. scikit-build does this but provides no easy way to opt out. This behavior makes it difficult for some other environments (like conda) to avoid invoking pip and installing packages outside of its packaging system. Ideally, all of qiskit-aer's build requirements would be preinstalled to avoid scikit-build trying to install anything with pip, but `cmake` is a special case. scikit-build/setuptools try to install it with pip if it is not installed even if the real `cmake` command is already available on the `PATH`.

Run `pip install --no-build-isolation .` in a build environment with all of qiskit-aer's build dependencies already installed except for the Python `cmake`  package (but with `cmake` installed on the `PATH`) and `PIP_NO_INDEX=1` set.

scikit-build successfully builds the package.

### Suggested solutions

In [this comment](https://github.com/scikit-build/scikit-build/issues/829#issuecomment-1358318265), one of the maintainers of scikit-build recommends moving the `setup_requires` packages to `build-system.requires` in `pyproject.toml` because `setup_requires` is deprecated by scikit-build and by setuptools. In this case, pip has a better mechanism for bypassing the installation of the build requirements. It is also mentioned that the next version of scikit-build will handle the cmake dependency better (not installing it with pip if it is already on the `PATH`).

I ran into this trying to make a conda package for qiskit-aer (https://github.com/conda-forge/staged-recipes/pull/21404). It is not major issue for that case -- I am just patching `setup_requires` out of `setup.py`. It is always nicer not to carry extra patches though.

jakelishman:
### Summary

Executing manual `pip install` commands from within `setup.py` is somewhat frowned upon from a Python packaging perspective.  By far the preferred mechanism now is to clearly state the build dependencies in a non-code format (`pyproject.toml`), and let a PEP-517-compatible builder handle virtual-environment isolation and dependency management.

We already had the dependencies declared statically, this commit just removes the residual code in `setup.py`.

### Details and comments

Fix #1688.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,87 +3,24 @@
 """
 Main setup file for qiskit-aer
 """
-import importlib
-import inspect
 import os
-import setuptools
-import subprocess
-import sys
-from pkg_resources import parse_version
 import platform
 
-
-def strtobool(val):
-    val_lower = val.lower()
-    if val_lower in ('y', 'yes', 't', 'true', 'on', '1'):
-        return True
-    elif val_lower in ('n', 'no', 'f', 'false', 'off', '0'):
-        return False
-    else:
-        raise ValueError(f'Value: "{val}" not recognizes as True or False')
-
-
-PACKAGE_NAME = os.getenv('QISKIT_AER_PACKAGE_NAME', 'qiskit-aer')
-_DISABLE_CONAN = strtobool(os.getenv("DISABLE_CONAN", "OFF"))
-_DISABLE_DEPENDENCY_INSTALL = strtobool(os.getenv("DISABLE_DEPENDENCY_INSTALL", "OFF"))
-
-
-
-def install_needed_req(import_name, package_name=None, min_version=None, max_version=None):
-    if package_name is None:
-        package_name = import_name
-    install_ver = package_name
-    if min_version:
-        install_ver += '>=' + min_version
-    if max_version:
-        install_ver += '<' + max_version
-
-    try:
-        mod = importlib.import_module(import_name)
-        mod_ver = parse_version(mod.__version__)
-        if ((min_version and mod_ver < parse_version(min_version))
-                or (max_version and mod_ver >= parse_version(max_version))):
-            raise RuntimeError(f'{package_name} {mod_ver} is installed '
-                               f'but required version is {install_ver}.')
-
-    except ImportError as err:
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -3,87 +3,24 @@
...
 
-if not _DISABLE_CONAN:
-    setup_requirements.append('conan>=1.22.2')
-
-requirements = common_requirements + ['qiskit-terra>=0.21.0', 'scipy>=1.0']
+requirements = [
+    'qiskit-terra>=0.21.0',
+    'numpy>=1.16.3',
+    'scipy>=1.0',
+]
```
hhorii, line 23: I just wondered that Aer uses `requirements.txt`as qiskit-terra does.

setup.py
```diff
@@ -3,87 +3,24 @@
...
-# also build time/setup requirements and will be added to both lists
-# of requirements
-common_requirements = [
-    'numpy>=1.16.3',
-]
-
-setup_requirements = common_requirements + [
-    'scikit-build>=0.11.0',
-    'cmake!=3.17,!=3.17.0',
-    'pybind11>=2.6',
```
hhorii, line 76: I guess that `pybind11` is missing after this change.

jakelishman: These are build requirements only, not install requirements.  They're specified more properly in `pyproject.toml`, which is where they're largely actually getting pulled in from, except in cases where the build process deliberately manually overrides that, like the Conda channel that Will manages in the linked issue.

If they're here as well, `setuptools` decides to try and install them itself, which isn't helpful.

setup.py
jakelishman: I doubt there's much of a reason.  In my own projects I tend to put this information in `pyproject.toml` these days, but I wasn't trying to completely overhaul the packaging.

jakelishman:
I need to go through and update some docs, and to fix the testing pipelines to properly use a PEP-517 builder instead of manual calls to `setup.py`.

hhorii (approved):

Issue #1695, opened 2023-01-10
CI fails in integration tests (Neko)


hhorii:
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?

Qiskit Neko Integration Test always fail because of `ModuleNotFoundError: No module named'qiskit_aer.backends.controller_wrappers'`.

### Steps to reproduce the problem

Run CI.

### What is the expected behavior?

Pass CI tests.

### Suggested solutions

This issue was happened previously because python tries to load `qiskit_aer` from the current working directory and not from the installed directory.

hhorii:
This issue was fixed as written in https://github.com/Qiskit/qiskit-aer/pull/1699.

PR #1699, opened 2023-01-11
Fix recent CI errors in Neko and python 3.11

### LINKED ISSUE #1698: CI fails in tests of python 3.11
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

https://github.com/Qiskit/qiskit-terra/issues/9291 was fixed with https://github.com/Qiskit/qiskit-terra/commit/0344a1c94b8eb8206da18cc74d9bdc70ad203c9a (https://github.com/Qiskit/qiskit-terra/pull/9310) which has not been merged into the pypi latest. Therefore, Aer tests encounter the same issues with https://github.com/Qiskit/qiskit-terra/issues/9291 was fixed with https://github.com/Qiskit/qiskit-terra/commit/0344a1c94b8eb8206da18cc74d9bdc70ad203c9a.

Run tests with python 3.11

Do not block other tests with fails of python 3.11.

### Suggested solutions

Change the github workflow configuration.

hhorii:
### Summary
Tests for python 3.11.1 are failed until the next terra release. Neko is failed because `python -e .` fails. This PR provides workaround for them.

### Details and comments

This PR pins minor version of python 3.11 to 3.11.0 to pass all the tests.
Install Aer in Neko with `python .` (without `-e` option) and remove `qiskit_aer` directory to correctly refer the installed package. 

Resolve #1698 and #1695.

### DIFF
--- a/.github/workflows/neko.yml
+++ b/.github/workflows/neko.yml
@@ -16,7 +16,5 @@ jobs:
           test_selection: backend
           repo_install_command: |
             mkdir ~/pip_tmp/
-            pip install -v .
+            TMPDIR=~/pip_tmp pip install -v .
             rm -rf qiskit_aer
-        env:
-          TMPDIR: ~/pip_tmp/
--- a/.github/workflows/neko.yml
+++ b/.github/workflows/neko.yml
@@ -15,5 +15,8 @@ jobs:
         with:
           test_selection: backend
           repo_install_command: |
+            mkdir ~/pip_tmp/
             pip install -v .
             rm -rf qiskit_aer
+        env:
+          TMPDIR: ~/pip_tmp/
--- a/.github/workflows/neko.yml
+++ b/.github/workflows/neko.yml
@@ -15,5 +15,5 @@ jobs:
         with:
           test_selection: backend
           repo_install_command: |
-            pip install .
+            pip install -v .
             rm -rf qiskit_aer
--- a/.github/workflows/neko.yml
+++ b/.github/workflows/neko.yml
@@ -15,6 +15,5 @@ jobs:
         with:
           test_selection: backend
           repo_install_command: |
-            pip install -r requirements-dev.txt
             pip install .
             rm -rf qiskit_aer
--- a/.github/workflows/neko.yml
+++ b/.github/workflows/neko.yml
@@ -14,4 +14,7 @@ jobs:
       - uses: Qiskit/qiskit-neko@main
         with:
           test_selection: backend
-          repo_install_command: pip install -e .
+          repo_install_command: |
+            pip install -r requirements-dev.txt
+            pip install .
+            rm -rf qiskit_aer
--- a/.github/workflows/neko.yml
+++ b/.github/workflows/neko.yml
@@ -15,6 +15,5 @@ jobs:
         with:
           test_selection: backend
           repo_install_command: |
-            mkdir ~/pip_tmp/
-            TMPDIR=~/pip_tmp pip install -v .
-            rm -rf qiskit_aer
+            pip install conan
+            pip install -v .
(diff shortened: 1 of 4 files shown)


mtreinish:
I'm not quite sure what the goal is here, the 3.11 jobs are still failing because of their use of 3.11. Duplicating the job definition doesn't remove the jobs or change anything that I can tell. Is the intent here to stop running 3.11 tests? You would have to remove 3.11 from the list to do that. The primary workaround you can do while waiting on terra 0.22.4 (which should be out on Monday or Tuesday) is to set the version to 3.11.0 the error was introduced by 3.11.1 so explicitly setting the patch version  should avoid this. 

The other option is to decouple the base aer test class from qiskit's base test class. We arguably shouldn't be inheriting a test class like this across a library boundary exactly for reasons like this. In general test code you want to be as flat and explicit as possible and defining a bunch of core behavior outside the local test suite will just cause trouble in the long run. (it's also been on my list for a while to stop exposing this as a public interface in terra, see: https://github.com/Qiskit/qiskit-terra/issues/6862 )

hhorii:
@mtreinish Thanks for your comments. I pinned python version to 3.11.0.

.github/workflows/neko.yml
```diff
@@ -14,4 +14,7 @@ jobs:
       - uses: Qiskit/qiskit-neko@main
         with:
           test_selection: backend
-          repo_install_command: pip install -e .
+          repo_install_command: |
+            pip install -r requirements-dev.txt
```
mtreinish, line 18: This shouldn't be needed, neko's tests have their own dev requirements (which get installed by the action definition) so we should only need to install aer from source to be able to run the tests

releasenotes/notes/fix_ci_fails_due_to_python311-367ea7ca0abbfc2a.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fix the minor version of Python 3.11 in CI to work around bugs of terra for 3.11.1.
+    Also, change installation of Qiskit-Aeer in Neko to correctly install `qiskit_aer` package.
```
mtreinish, line 5: We don't need a release note for this change, CI changes aren't user facing (which is the intended audience of release notes).

.github/workflows/neko.yml
hhorii: @mtreinish as shown in https://github.com/hhorii/qiskit-aer/actions/runs/3912462608/jobs/6687134955, installation of qiskit-aer fails because neko does not include `conan` in its `requirements-dev.txt`, I guess.

hhorii: @mtreinish I confirmed that `conan` command is installed but cmake does not recognize it as executable. In my local env, following logs are appeared in `pip install -v .` but not in [this CI](https://github.com/Qiskit/qiskit-aer/actions/runs/3928098538/jobs/6715355225).
```
    changing mode of /tmp/pip-build-env-hya3x7mj/overlay/bin/conan to 755
    changing mode of /tmp/pip-build-env-hya3x7mj/overlay/bin/conan_build_info to 755
    changing mode of /tmp/pip-build-env-hya3x7mj/overlay/bin/conan_server to 755
```
Usually, `pip` builds these commands in a temporary directory. I tried to change `TMPDIR` in aer installation but errors remained.

I would like install `conan` manually before qiskit-aer installation in neko. Could you give your thoughts?

mtreinish: I think this is pointing to a wider issue in building from source. It looks like pip installing conan from pyproject.toml in that log but then cmake is not able to find it for some reason.

hhorii: The current pyproject.toml install `conan` correctly in my local docker environment. I guess that it does not work well only in github actions for `conan`.

PR #1701, opened 2023-01-18
Fix multi-shots batched execution on GPU

doichanj:
### Summary
Unit test run stalls when using batched shot execution on GPU.

### Details and comments
Batched shots execution uses classical registers on GPU memory, and memory space for creg is reserved before allocating chunks. To reserve this, bits count for classical register should be set before allocation, but bits count was not set correctly. This caused infinite loop to find available memory on GPU. 

This fix set bits count of classical register before the allocation for batched shots execution.


hhorii (approved):
Looks good to me.

hhorii:
@doichanj could you write a release note?

hhorii (approved):

PR #1703, opened 2023-01-24
Add a work around for Aer installation issue in Neko

hhorii:
### Summary

This PR can resolve CI fail of a Neko with a work around

### Details and comments

`pip install .` fails in Neko's github action though it is successful in local environment.
conan commands can not be found in the action though it is installed.
This PR is a work around to fix the current issue in Neko by manually installing conan.

This is from https://github.com/Qiskit/qiskit-aer/pull/1699.


mtreinish (approved):
I don't like this as a solution. This is masking something that I think is wrong with aer's build configuration. Conan is getting installed by pip (as it's listed in the pyproject.toml as a build requirement), so if it's not being found that's likely an issue with our cmake files. That being said I don't have sufficient time to dig into what's going wrong here, so we can do this as a short term fix to unblock ci. But before we merge this can you open an issue to track the underlying issue so we can remove this hack when that's fixed.

PR #1705, opened 2023-01-25
Add release note for #1685

derwind:
### Summary

This PR is to add a release note for #1685.


derwind:
The test report says:

> The following tests exited without returning a status
> and likely segfaulted or crashed Python:
>
>	* test.terra.backends.test_config_pulse_simulator.TestConfigPulseSimulator.test_from_backend_system_model

in 3.10, macOS-latest env.

On the other hand, that is a part that I have not touched and do not understand the contents of. What would you suggest I do?

hhorii (approved):

Issue #1694, opened 2023-01-09
MPS simulator hangs taking a measurement with limited bond dimension


mhodson-rigetti:
### Informations

- **Qiskit Aer version**: 0.11.2
- **Python version**: 3.8.10
- **Operating system**: Linux (Ubuntu 20.04.1)

### What is the current behavior?

MPS simulator hangs during processing of a measurement instruction in specific cases reproduced below, and in the case that (a) the bond dimension is limited; and (b) the `mps_apply_measure` method is used for measurement.

The hang is caused by an infinite loop in the C++ `csvd` code because the value of `eps` is zero and none of the exit conditions are satisfied. The value of `eps` is zero because the input matrix is zero. The input matrix is zero because this is the result of the `MPS_Tensor::Decompose` method, executed in the context of `MPS::move_all_qubits_to_sorted_ordering()`.

References:

https://github.com/Qiskit/qiskit-aer/blob/6b0e9996e907b48a3788b15ac2010aa8e1f32ff7/src/simulators/matrix_product_state/svd.cpp#L157

https://github.com/Qiskit/qiskit-aer/blob/6b0e9996e907b48a3788b15ac2010aa8e1f32ff7/src/simulators/matrix_product_state/matrix_product_state_tensor.hpp#L653

https://github.com/Qiskit/qiskit-aer/blob/6b0e9996e907b48a3788b15ac2010aa8e1f32ff7/src/simulators/matrix_product_state/matrix_product_state_internal.cpp#L976

### Steps to reproduce the problem

Issue is reproduced with the following Python script. 

The problematic behavior is seen in this test case only if the measurement outcome is 1. This is selected stochastically by `AerSimulator` and cannot be seeded as far as I can determine. If the measurement outcome is 0 the program completes.

When the issue arises, the program hangs in the context of processing instruction `measure q[0] -> c[0];`.

```
from qiskit import QuantumCircuit, transpile
from qiskit_aer import AerSimulator

INPUT = """OPENQASM 2.0;
include "qelib1.inc";
qreg q[4];
creg c[4];
h q[3];
h q[1];
cx q[1],q[2];
cx q[3],q[0];
measure q[0] -> c[0];
measure q[1] -> c[1];
measure q[2] -> c[2];
measure q[3] -> c[3];"""

# Construct quantum circuit
circuit = QuantumCircuit.from_qasm_str(INPUT)

simulator = AerSimulator(
    method='matrix_product_state',
[...]

PR #1713, opened 2023-01-31
MPS: Changed the apply_measure algorithm

merav-aharoni:
### Summary
Changed the `MPS::apply_measure` method so that it does not move the qubits to sorted ordering before measuring.

### Details and comments
This PR addresses issue #1694.
 Changed the `apply_measure` algorithm so that it does not move the qubits to the sorted positions before measuring. Instead, we measure the qubits in the order of the current `MPS` structure and then sort the measurement outcomes as a post-process. Beside fixing the issue, this is expected to improve performance and enable more accurate representation of entangled states. For example, the circuit provided in issue #1694 cannot be represented accurately when the `MPS` structure is in sorted ordering, but can be represented accurately with the new `apply_measure`.
Note that this modification does not actually fix the bug. It is possible that there is still a bug somewhere in the code. This should be addressed separately.


merav-aharoni:
A note for the reviewers: `AerSimulator` actually sorts the qubits to be measured and removes the qubits that are not measured. Therefore, I could have assumed this in the call to `apply_measure_internal` and we could have avoided some of the internal sorting processes. Nevertheless, I preferred not to rely on this assumption. This adds some complication to the code, but makes it more robust.

src/simulators/matrix_product_state/matrix_product_state_internal.cpp
```diff
@@ -1573,6 +1579,36 @@ void MPS::propagate_to_neighbors_internal(uint_t min_qubit, uint_t max_qubit,
...
+             sub_ordering[next] = qubit_ordering_.order_[i];
+             next++;
+             break;
+          }
+      }
+  }
+  return sorted_qubits;
+}
+
+reg_t MPS::sort_measured_values(reg_t input_outcome, reg_t& sub_ordering) {
```
yaelbh, line 1598:
```suggestion
reg_t MPS::sort_measured_values(const reg_t& input_outcome, reg_t& sub_ordering) {
```

yaelbh (approved):
It was tough to review it, and follow all the manipulations that all these arrays go through: `sorted_qubits, sub_ordering, outcome_vector, sorted_outcome_vector`. I did it slowly with an example, and it looks correct. It will be nice if you find a way to somehow facilitate the code reading.

merav-aharoni:
@hhorii , Can you please review this PR? @yaelbh already reviewed and approved. I made the changes she requested.

hhorii (approved):
Change looks good to me.

Issue #1683, opened 2022-12-13
Develop own `assemble` to directly generate `Circuit` struct from `QuantumCircuit`


hhorii:
### What is the expected behavior?

Currently Qiskit-Aer uses `QObj` as the common representation of simulation targets.
However, as #1660 mentioned, recent `QObj` does not support latest features of `QuantumCircuit`.
Also, converting `QObj` is an apparent overhead because it is converted into `Circuit` struct of C++ in pybind.
By providing own `assemble()` to convert `QuantumCircuit` to `Circuit` struct of C++, we will get performance and flexibility for its specification.
This new feature will also resolve an issue of #1652.

adekusar-drl:
@hhorii are there any plans when this enhancement can be implemented?

hhorii:
@adekusar-drl this will be a new feature in 0.12.0, which will be released in middle of this February.

hhorii:
#1717 was merged.

Issue #1660, opened 2022-11-17
Running circuit with qubits added directly fails


kevinsung:
### Informations

- **Qiskit Aer version**: 0.11.0
- **Python version**: 3.10.8
- **Operating system**: Arch Linux

### What is the current behavior?
```
Simulation failed and returned the following error message:
ERROR: Failed to load qobj: Invalid Qobj experiment: not enough memory slots.
Result(backend_name='aer_simulator', backend_version='0.11.0', qobj_id='', job_id='e612fa58-f665-4d7f-979e-00f59d727f5f', success=False, results=[], date=2022-11-17T07:50:39.243329, status=ERROR: Failed to load qobj: Invalid Qobj experiment: not enough memory slots., header=None, metadata={}, time_taken=0.00040602684020996094)
```

### Steps to reproduce the problem
```python
from qiskit import ClassicalRegister, QuantumCircuit, QuantumRegister
from qiskit.circuit import CircuitInstruction
from qiskit.circuit.library import Measure
from qiskit import Aer

qubits = QuantumRegister(2)
clbits = ClassicalRegister(2)

circuit = QuantumCircuit()
circuit.add_bits(qubits)
circuit.add_bits(clbits)
for q, c in zip(qubits, clbits):
    circuit.append(CircuitInstruction(Measure(), [q], [c]))

backend = Aer.get_backend("aer_simulator")
result = backend.run(circuit).result()
print(result)
```

### What is the expected behavior?
Should work. Terra can handle it.

t-imamichi:
FYI: a circuit with `add_register` instead of `add_bits` works fine.

```python
from qiskit import ClassicalRegister, QuantumCircuit, QuantumRegister
from qiskit.circuit import CircuitInstruction
from qiskit.circuit.library import Measure
from qiskit import Aer

qubits = QuantumRegister(2)
clbits = ClassicalRegister(2)

circuit = QuantumCircuit()
circuit.add_register(qubits)
circuit.add_register(clbits)
for q, c in zip(qubits, clbits):
    circuit.append(CircuitInstruction(Measure(), [q], [c]))

backend = Aer.get_backend("aer_simulator")
result = backend.run(circuit).result()
print(result.get_counts())
```

output
```
{'00': 1024}
```

jakelishman:
@t-imamichi: sure, but that's not the point of the bug being reported.  Loose bits should work.

[...]

PR #1652, opened 2022-11-09
[perf] Cache circuits

ikkoham:
### Summary

2nd hack to improve the performance of primitives.

(1st version: https://github.com/Qiskit/qiskit-aer/pull/1620)

### Details and comments

Performance (VQE with HF)

main
image

this PR

image

4 PRs (Terra 2 + Aer 2)
image

PR #1717, opened 2023-02-07
Simulate `QuantumCircuit` without QObj

hhorii:
### Summary

This PR produces a way to simulate circuits without calling `assemble()`

### Details and comments

As written in https://github.com/Qiskit/qiskit-aer/issues/1683, Aer has some issues (#1660, #1652) caused by calling `assemble()`. This PR will resolve them.

1. **`AerCircuit`**
First, this introduces new classes to bind C++ structure: `AerCircuit` which is mapped to `AER::Circuit` class in C++. `AerCircuit` is generated from `QuantumCricuit` in `aer_compiler`.

2. **`execute()` for `std::vector<AerCircuit>` in `aer_controller_execute`**
Next, this provides `execute(std::vector<AER::Circuit> &circuits, Noise::NoiseModel &noise_model, json_t& config)` interface for python to start simulation. It returns the same result with that `aer_controller_execute()(qobj)` returns.
In `execute()`, parameterizations and circuit initialization are performed as `qobj.hpp` does.

3. **`AerBackend._run_circuits()`, `_execute_circuits()`, and `AerJob`**
Finally, a job `AerJob` is submitted in `_run_circuits()` and `AerBackend._execute_circuits_job()` is called in the method with `QuantumCircuit`, `NoiseModel`, and config(`dict`). The `QuantumCircuit` is converted to `AerCircuit` in `_execute_circuits_job()` and then `AerCircuit`s are simulated in `_execute_circuits()`. Existing path to simulate QObj still remains by using `_run_qobj()`, `_execute_qobj_job()`, and `_execute_qobj()`.

Note that, in DASK execution, `qobj` is still necessary. Therefore, in this PR, if `executor` is specified, simulators generate `qobj` and start simulation with the old path.


qiskit_aer/backends/unitary_simulator.py
```diff
@@ -260,12 +264,18 @@ def _execute(self, qobj):
         Returns:
             dict: return a dictionary of results.
         """
-        # Make deepcopy so we don't modify the original qobj
-        qobj = copy.deepcopy(qobj)
```
hhorii, line 264: Oops. We still need deep copy if Aer accepts qobj. At least, this code deletion should be in #1724. I will revert this code change.

adekusar-drl:
Thanks for the good job! I'd like to share some numbers I obtained running some Qiskit Machine Learning performance tests. I've built qiskit-aer from sources and this PR. Terra is 0.23.1. Then I ran my quantum kernel tests on 4 qubits only, just to check out what kind of improvements we may get. And these are the numbers I've got:
- Aer primitives: 7.9898 sec
- Ref primitives: 3.8325 sec

Aer primitives are still slower than the reference ones. In this case it is `Sampler`.

On other hand these are the number from the previous run in December without any updates in Aer.
- Aer primitives: 82.1253 sec
- Ref primitives: 3.7455 sec

Certainly, now Aer Sampler is better, but there's a room for improvements. 

I also collected profiler information, it can be handy.
image

hhorii:
@adekusar-drl Thank you for your comments. I'm glad to know this will improve performance. I think overheads to invoke simulation (C++ <-> Python) will be dominant for 4-qubit simulation, but we can reduce the overheads more.

@ikkoham has already implemented some optimization for `convert_circuit_binds` in a past closed PR. He will extract the optimization and send a new PR, soon.

hhorii:
Anyway, here is performance improvement with this PR. I used QuantumVolume (that does not have parameters). 

image

qiskit_aer/backends/aerbackend.py
```diff
@@ -209,7 +256,7 @@ def run(self,
         if isinstance(experiments, list):
             aer_job = AerJobSet(self, job_id, self._run, experiments, executor)
         else:
-            aer_job = AerJob(self, job_id, self._run, experiments, executor)
+            aer_job = AerJob(self, job_id, self._run, qobj=experiments, executor=executor)
```
mtreinish, line 259: This is a pre-existing bug so we don't need to fix it here, but for the case where circuits or schedules are passed to `run()` we probably want to be storing the input circuit objects on the job and not the qobj. Since for most people the qobj isn't a meaningful interface.

qiskit_aer/backends/aerbackend.py
```diff
@@ -345,6 +392,65 @@ def _run(self, qobj, job_id='', format_result=True):
...
 
+    def _run_direct(self, circuits, noise_model, config, job_id='', format_result=True):
+        """Run a job"""
+        # Start timer
+        start = time.time()
+
+        # Take metadata from headers of experiments to work around JSON serialization error
+        metadata_list = []
+        for idx, circ in enumerate(circuits):
+            if hasattr(circ, "metadata") and circ.metadata:
```
mtreinish, line 403: Can this `hasattr` ever be false, `QuantumCircuit.metadata` should always exist, it just might be `None`

qiskit_aer/backends/aerbackend.py
```diff
@@ -345,6 +392,65 @@ def _run(self, qobj, job_id='', format_result=True):
...
+                metadata = circ.metadata
+                metadata_list.append(metadata)
+                circ.metadata = {}
+                if "id" in metadata:
+                    circ.metadata["id"] = metadata["id"]
+                circ.metadata["metadata_index"] = idx
```
mtreinish, line 409: Where do these come from? I'm not sure I understand what is setting `id` and why we need to set the metadata index on the circuit object here.

qiskit_aer/backends/pulse_simulator.py
```diff
@@ -278,6 +278,11 @@ def _execute(self, qobj):
         qobj.config.qubit_freq_est = self.defaults().qubit_freq_est
         return pulse_controller(qobj)
 
+    def _execute_direct(self, circuits, noise_model, config):
+        """Execute circuits on the backend.
+        """
+        raise AerError("not implemented yet")
```
mtreinish, line 284: Will this ever get called? `_execute_direct` isn't a required part of inheriting from `AerBackend` and nothing in the pulse path as defined in `AerBackend` should ever call this, so I'd say lets just remove this.

qiskit_aer/backends/wrappers/bindings.cc
```diff
@@ -12,155 +12,24 @@ DISABLE_WARNING_POP
...
-#ifdef TEST_JSON // Convert input qobj to json to test standalone data reading
-        return AerToPy::to_python(AER::controller_execute<T>(json_t(qobj)));
-#else
-        return AerToPy::to_python(AER::controller_execute<T>(qobj));
-#endif
-    }
-};
+Operations::Op create_aer_operation(const py::handle &qasmQobjInst) {
+  return Operations::input_to_op(qasmQobjInst);
+}
```
mtreinish, line 23: Is this used anywhere? I didn't see anything that needed this.

qiskit_aer/circuit/aer_circuit.py
```diff
@@ -0,0 +1,291 @@
...
+                    params = copy(params)
+                    copied = True
+                params[i] = 0.0
+
+        if name in ('ccx', 'ccz', 'cp', 'cswap', 'csx', 'cx', 'cy', 'cz', 'delay', 'ecr',
+                    'h', 'id', 'mcp', 'mcphase', 'mcr', 'mcrx', 'mcry', 'mcrz', 'mcswap',
+                    'mcsx', 'mcu', 'mcu1', 'mcu2', 'mcu3', 'mcx', 'mcx_gray', 'mcy', 'mcz',
+                    'p', 'r', 'rx', 'rxx', 'ry', 'ryy', 'rz', 'rzx', 'rzz', 's', 'sdg', 'swap',
+                    'sx', 'sxdg', 't', 'tdg', 'u', 'x', 'y', 'z', 'u1', 'u2', 'u3',
+                    'cu', 'cu1', 'cu2', 'cu3'):
```
mtreinish, line 93: A set here will be about ~10x faster because a tuple will need to be iterated over for each lookup:

```suggestion
        if name in {'ccx', 'ccz', 'cp', 'cswap', 'csx', 'cx', 'cy', 'cz', 'delay', 'ecr',
                    'h', 'id', 'mcp', 'mcphase', 'mcr', 'mcrx', 'mcry', 'mcrz', 'mcswap',
                    'mcsx', 'mcu', 'mcu1', 'mcu2', 'mcu3', 'mcx', 'mcx_gray', 'mcy', 'mcz',
                    'p', 'r', 'rx', 'rxx', 'ry', 'ryy', 'rz', 'rzx', 'rzz', 's', 'sdg', 'swap',
                    'sx', 'sxdg', 't', 'tdg', 'u', 'x', 'y', 'z', 'u1', 'u2', 'u3',
                    'cu', 'cu1', 'cu2', 'cu3'}:
```

qiskit_aer/circuit/aer_circuit.py
```diff
@@ -0,0 +1,291 @@
...
+from qiskit.circuit import QuantumCircuit, Clbit, ParameterExpression
+from qiskit.tools.parallel import parallel_map
+from qiskit.providers.options import Options
+from qiskit.qobj import QobjExperimentHeader
+from qiskit_aer.aererror import AerError
+# pylint: disable=import-error, no-name-in-module
+from qiskit_aer.backends.controller_wrappers import AerCircuit_
+
+
+class AerCircuit:
```
mtreinish, line 26: I'm a bit uncertain about the need for this class. Couldn't we just go from `QuantumCircuit` to `AerCircuit_` directly in a for loop over `QuantumCircuit.data`. The only use for `AerCircuit` I saw was to call `AerCircuit.native_gates` which then gets passed to the execute function. I feel like we can remove a layer in the hierarchy here without too much effort. Just basically make `generate_aer_circuit` be the equivalent of `AerCircuit.native_gates` and return a  `Aercircuit_`  objects directly instead of wrapping it in a class.

releasenotes/notes/own_assembler-0c76e67a054bd12c.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    Aer simulates :class:`~qiskit.circuit.QuantumCircuit` without serialization to
+    :class:`~qiskit.qobj.QasmQobj`. Instead, Aer converts a circuit to
+    a :class:`.AerCircuit` instance that a wrapper of internal circuit structure
+    of Aer C++. With this change, because overheads of serialization is reduced,
+    performance of simulation especially for low-qubit circuts is improved.
+    Note that pulse simualation and DASK-based simulation still use Qobj and
+    :class`.AerJob` does not return Qobj otherwise.
```
mtreinish, line 10: We probably should have an additional note in this file for the addition of `AerJob.circuits` since that's also a new API.

qiskit_aer/backends/aer_simulator.py
```diff
@@ -730,6 +731,13 @@ def configuration(self):
         config.backend_name = self.name()
         return config
 
+    def _execute_direct(self, circuits, noise_model, config):
+        """Execute circuits on the backend.
+        """
+        aer_circuits = generate_aer_circuits(circuits)
+        ret = cpp_execute_direct(self._controller, aer_circuits, noise_model, config)
```
mtreinish, line 738: Maybe call this `cpp_execute_aer_circuit` instead of direct? Not sure, but I've been getting a bit confused with some of the function naming. Like execute vs run and now the direct variants. It's getting a bit hard for me to follow because we have at least 5 things now named execute or run in these classes.

mtreinish (commented):
I've made a quick first pass through most of the python code and the c++ wrappers. I still have to go through the bulk of the C++ changes in `src/` though but I wanted to leave some early comments before I did. Overall I'm very excited for this, this is how we should have originally built the Python interface in Aer as this will have much less overhead than have intermediate transforms and serialization formats. M big high level comment at this point is I think this still has one too many layers in the hierarchy I think we can remove the Python `AerCircuit` object and go straight from `QuantumCircuit` to the pybind `AerCircuit`. This will simplify things a bit and also reduce a tiny amount of overhead.

qiskit_aer/backends/aer_simulator.py
hhorii: I agree with you :-)

I would like to rename methods as follows:
```
run() --> _run_qobj()     -> [ThreadPool] -> __execute_qobj()     -> _execute_qobj()
      --> _run_circuits() -> [ThreadPool] -> __execute_circuits() -> _execute_circuits()
```

qiskit_aer/backends/aerbackend.py
hhorii: https://github.com/Qiskit/qiskit-aer/blob/ce9a8605545c671b653c07562d8283ca98fc951b/qiskit_aer/jobs/utils.py#L59-L88

`id` is set when "qobj" is divided for DASK execution. However, yes, here is not qobj execution and we can take it (until DASK supports circuit execution).

qiskit_aer/circuit/aer_circuit.py
hhorii: I'm thinking that `backend.run()` can take `AerCircuit` for repeatedly call the same circuit with different parameters in Aer (such as by `Estimator`). I guess that python class is better for some future utilities.
Another reason is serielization. pybind11 supports [pickling](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support) but a bit complex to handle `AER::Circuit`. Instead, if we just pickle current python `AerCircuit`, only pickling `QuantumCircuit`s (and some other fields) is necessary.

qiskit_aer/backends/pulse_simulator.py
hhorii: `_execute_circuits` (renamed from `_execute_direct`) is defined as `@abstractmethod` in `AerBackend`. So, we need to implement it for `PulseSimulator` or take `@abstractmethod`  from `AerBackend`. But if we take `@abstractmethod`, lint [errors](https://github.com/Qiskit/qiskit-aer/actions/runs/4239315616/jobs/7367249914) call of `_execute_direct` because  it does not return any. Of cause `_execute_direct` can return `None` and this lint error will be worked around, but I think keeping `@abstractmethod` and implementing `_execute_circuits` in deprecating `PulseSimulator` is reasonable.

```
************* Module qiskit_aer.backends.aerbackend
qiskit_aer/backends/aerbackend.py:432:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
```

hhorii:
@mtreinish I added `AerConfig` to reduce json overheads. Previously json was used for configuration. Aer uses `nlohmann/json.hpp` which produce overheads in memory allocation. By introducing `AerConfig`, we can reduce their overheads. I think Aer becomes always better than reference in above @adekusar-drl 's benchmark.

test/terra/backends/test_parameterized_qobj.py
```diff
@@ -67,57 +67,6 @@ def parameterized_qobj(
...
-            job = backend.run(qobj, **self.BACKEND_OPTS)
-            result = job.result()
-            success = getattr(result, 'success', False)
-            num_circs = len(result.to_dict()['results'])
-            self.assertTrue(success)
-
-            for j in range(num_circs):
-                statevector = result.get_statevector(j)
-                np.testing.assert_array_almost_equal(
-                    statevector, statevec_targets[j].data, decimal=7)
```
mtreinish, line 119: Why are these tests deleted? Shouldn't these still work as is because the input qobj type won't trigger the new code.

test/terra/backends/aer_simulator/test_circuit.py
```diff
@@ -0,0 +1,150 @@
...
+        qubits = QuantumRegister(3)
+        clbits = ClassicalRegister(3)
+
+        circuit = QuantumCircuit(qubits, clbits)
+        circuit.h(qubits[0])
+        circuit.cx(qubits[0], qubits[1])
+        circuit.cx(qubits[0], qubits[2])
+
+        for q, c in zip(qubits, clbits):
+            circuit.append(CircuitInstruction(Measure(), [q], [c]))
```
mtreinish, line 45: Is there a reason to do this instead of just?:

```suggestion
        for q, c in zip(qubits, clbits):
            circuit.measure(q, c)
```

test/terra/backends/aer_simulator/test_auto_method.py
```diff
@@ -216,18 +216,20 @@ def test_auto_method_nonclifford_circuit_and_kraus_noise(self):
...
-        qc_2 = QuantumVolume(5)
+        qc_2 = QuantumCircuit(50)
+        qc_2.h(range(50))
         qc_2.measure_all()
         circuits.append(qc_2)
         circuits.append(qc)
         backend = self.backend()
         shots = 100
-        result = backend.run(circuits, shots=shots).result()
+        result = backend.run(circuits, shots=shots, method='statevector').result()
```
mtreinish, line 232: Isn't this test specifically to run without a method set? (the method name is `auto_method`)

src/transpile/fusion.hpp
```diff
@@ -359,17 +359,28 @@ class NQubitFusion : public Fuser {
...
+  if (config._fusion_enable_n_qubits_2.has_value()) active = config._fusion_enable_n_qubits_2.value();
+#elif N == 3
+  if (config._fusion_enable_n_qubits_3.has_value()) active = config._fusion_enable_n_qubits_3.value();
+#elif N == 4
+  if (config._fusion_enable_n_qubits_4.has_value()) active = config._fusion_enable_n_qubits_4.value();
+#elif N == 5
+  if (config._fusion_enable_n_qubits_5.has_value()) active = config._fusion_enable_n_qubits_5.value();
+#endif
+  // if (JSON::check_key(activate_prop_name, config))
+  //   JSON::get_value(active, activate_prop_name, config);
```
mtreinish, line 383: Is this needed?

qiskit_aer/backends/wrappers/aer_controller_binding.hpp
```diff
@@ -0,0 +1,461 @@
...
+
+    py::class_<Config> aer_config(m, "AerConfig");
+    aer_config.def(py::init());
+    aer_config.def_readwrite("shots", &Config::shots);
+    aer_config.def_readwrite("method", &Config::method);
+    aer_config.def_readwrite("device", &Config::device);
+    aer_config.def_readwrite("precision", &Config::precision);
+    // executor=None,
+    // max_job_size=None,
+    // max_shot_size=None,
```
mtreinish, line 107: Are these needed? Or is it just to document the python side fields we're not specifying here, if so can you put a comment above that says "missing options from python" or something.

mtreinish (commented):
Thanks for the updates, overall this looks much better to me. I left some comments inline nothing major though. I also skimmed the C++ nothing really stood out to me, but there are a lot of small details and I might have missed something (especially all the boilerplate in pybind used to call the inner functions), hopefully we have sufficient test coverage.

qiskit_aer/backends/aer_compiler.py
```diff
@@ -325,3 +332,225 @@ def compile_circuit(circuits, basis_gates=None, optypes=None):
...
+        n_qubits=num_qubits,
+        qreg_sizes=qreg_sizes,
+        memory_slots=num_memory,
+        creg_sizes=creg_sizes,
+        name=circuit.name,
+        global_phase=global_phase,
+    )
+
+    if hasattr(circuit, "metadata") and circuit.metadata:
+        header.metadata = copy(circuit.metadata)
```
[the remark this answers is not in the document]
hhorii: It is not necessary now. Copy may become necessary if we need to add some additional data to it.

qiskit_aer/backends/aer_compiler.py
```diff
@@ -325,3 +332,225 @@ def compile_circuit(circuits, basis_gates=None, optypes=None):
...
+            qc.h(0)
+            qc.cx(0, 1)
+            qc.measure_all()
+            # Generate AerCircuit from the input circuit
+            aer_qc_list = assemble_circuits(circuits=[qc])
+    """
+    # generate aer circuits
+    # TODO parallel_map will improve performance for multi circuit assembly.
+    # However, it calls pickling AerCircuit in Linux environment. Until AerCircuit
+    # supports pickle, circuits are assembleed sequentially in a single thread
```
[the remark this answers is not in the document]
hhorii: I agree. Pickling of pybind object includes many calls of connection between C++ and python and they are expensive also in general.

qiskit_aer/backends/aerbackend.py
```diff
@@ -188,9 +191,62 @@ def run(self,
...
+        only_pulse = True
+        for circ in circuits:
+            only_circuits &= isinstance(circ, QuantumCircuit)
+            only_pulse &= isinstance(circ, (ScheduleBlock, Schedule))
+
+        if only_circuits and not only_pulse:
+            executor = run_options.get('executor', None)
+            if executor is None and 'executor' in self.options.__dict__:
+                executor = self.options.__dict__.get('executor', None)
+            if executor or validate:
```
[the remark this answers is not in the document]
hhorii: `validate` is still in arguments of `backend.run()`. Let's deprecate it in this PR.

test/terra/backends/aer_simulator/test_circuit.py
hhorii: Because other codes in this test file call the same way :-) I will update whole measurement in this test file.

test/terra/backends/aer_simulator/test_auto_method.py
hhorii: I do not why this test was modified :-)

test/terra/backends/aer_simulator/test_auto_method.py
```diff
@@ -216,18 +216,19 @@ def test_auto_method_nonclifford_circuit_and_kraus_noise(self):
...
         qc_2.measure_all()
         circuits.append(qc_2)
         circuits.append(qc)
         backend = self.backend()
         shots = 100
         result = backend.run(circuits, shots=shots).result()
-        self.assertEqual(result.status, 'PARTIAL COMPLETED')
-        self.assertTrue(hasattr(result.results[1].data, 'counts'))
-        self.assertFalse(hasattr(result.results[0].data, 'counts'))
+        result = backend.run(circuits, shots=shots, method='statevector').result()
```
mtreinish, line 234: Also this test is not asserting anything anymore. Before we merge can you please update this so it's back it's original state.

mtreinish (approved):
Overall this LGTM now. I'm a bit concerned about the size of this PR and whether I missed something. It's hard for me to feel confident that everything looks correct at ~3k lines especially as there is a lot of code that is just mapping things from python to c++, or from json to config which is easy to typo. Hopefully our test coverage will protect against this (although it's not ever going to be perfect).

For future reference it would be really good if you could try to split this up into smaller logical changes and leave unrelated refactors out (for example the `format()` to f-string). Like for me a good work flow would have been to have a series of PRs like:

1. introduce the c++ Config class
2. Add pybind AerConfig class
3. Add pybind AerCircuit class
4. Update AerBackend (and subclasses) to use AerCircuit during `run()`

Smaller logical changes like this would have made this far easier to review.

hhorii:
> For future reference it would be really good if you could try to split this up into smaller logical changes and leave unrelated refactors out (for example the `format()` to f-string). Like for me a good work flow would have been to have a series of PRs like:
> 
> 1. introduce the c++ Config class
> 2. Add pybind AerConfig class
> 3. Add pybind AerCircuit class
> 4. Update AerBackend (and subclasses) to use AerCircuit during `run()`
> 
> Smaller logical changes like this would have made this far easier to review.

I'm sorry that I made this PR too big... I totally agree with you.

test/terra/backends/aer_simulator/test_auto_method.py
```diff
@@ -216,18 +216,21 @@ def test_auto_method_nonclifford_circuit_and_kraus_noise(self):
...
         qc_2 = QuantumVolume(5)
+        qc_2 = QuantumCircuit(50)
+        qc_2.h(range(50))
         qc_2.measure_all()
         circuits.append(qc_2)
         circuits.append(qc)
         backend = self.backend()
         shots = 100
-        result = backend.run(circuits, shots=shots).result()
+        result = backend.run(circuits, shots=shots, method='statevector').result()
```
mtreinish, line 233:
```suggestion
        result = backend.run(circuits, shots=shots).result()
```

mtreinish (approved):
I think this LGTM now, given the caveats in my previous review message. I think we can move forward on merging this now. Thanks for the quick updates.

Issue #1720, opened 2023-02-07
Remove qobj supports as an argument of `AerSimulator.run()`


hhorii:
### What is the expected behavior?

`AerSimulator` should not support `QasmQobj` and `PulseQobj` to run simulation.

They are deprecated in 0.9.0.

https://github.com/Qiskit/qiskit-aer/blob/b9f35590c51bcb1deb9f6021352507bea0b6a4b9/qiskit_aer/backends/aerbackend.py#L162-L168

hhorii:
Even If we remove qobj supports in arguments, qobj will be still necessary to execute pulse simulation and DASK-based qasm simulation. That is, `assemble()` will be internally called in `run()`. To generate qobj for pulse, arguments for `assemble()`  is necessary. For example, `verify_wheel.py` executes pulse simulation as follows
```
    qobj = assemble([schedule],
                    backend=backend_sim,
                    qubit_lo_freq=[5.0],
                    meas_level=1,
                    meas_return='avg',
                    shots=1)
    results = backend_sim.run(qobj, system_model=system_model).result()
```
Note that pulse simulation requires different arguments for `assemble()` and `run()`. If we remove qobj supports, the same arguments will be specified and unfortunately pulse simulation doesn't work correctly.

I think that qobj should be removed after unsupport of pulse simulation (https://github.com/Qiskit/qiskit-aer/issues/1722).

Issue #1722, opened 2023-02-07
Deprecate pulse simulation


hhorii:
### What is the expected behavior?

[Qiskit-Dynamics](https://github.com/Qiskit/qiskit-dynamics) support pulse simulation. We can deprecate pulse simulator in Aer.

PR #1724, opened 2023-02-15
Remove qobj run in AerBackend

hhorii:
### Summary

Remove qobj support in `backend.run()`, which was deprecated 2 years ago.

### Details and comments

#1720

### DIFF
--- a/qiskit_aer/backends/aerbackend.py
+++ b/qiskit_aer/backends/aerbackend.py
@@ -18,7 +18,6 @@ import datetime
 import logging
 import time
 import uuid
-import warnings
 from abc import ABC, abstractmethod
 
 from qiskit.circuit import QuantumCircuit, ParameterExpression, Delay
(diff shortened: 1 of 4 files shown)


hhorii:
Even If we remove qobj supports in arguments, qobj will be still necessary to execute pulse simulation and DASK-based qasm simulation. That is, `assemble()` will be internally called in `run()`. To generate qobj for pulse, arguments for `assemble()`  is necessary. For example, `verify_wheel.py` executes pulse simulation as follows
```
    qobj = assemble([schedule],
                    backend=backend_sim,
                    qubit_lo_freq=[5.0],
                    meas_level=1,
                    meas_return='avg',
                    shots=1)
    results = backend_sim.run(qobj, system_model=system_model).result()
```
Note that pulse simulation requires different arguments for `assemble()` and `run()`. If we remove qobj supports, the same arguments will be specified and unfortunately pulse simulation doesn't work correctly.

I think that qobj should be removed after unsupport of pulse simulation (https://github.com/Qiskit/qiskit-aer/issues/1722).

PR #1726, opened 2023-02-16
[perf] improve _convert_circuit_binds

ikkoham:
### Summary

https://github.com/Qiskit/qiskit-aer/pull/1717#issuecomment-1432595770

### Details and comments

Sorry, this PR does not affect quantum kernel benchmark because conditional branches do not go in this improvement.
However, this PR improves VQE.
My simple benchmark for UCCSD ansatz:

image

### DIFF
--- a/qiskit_aer/backends/aerbackend.py
+++ b/qiskit_aer/backends/aerbackend.py
@@ -84,20 +84,26 @@ class AerBackend(Backend, ABC):
 
     def _convert_circuit_binds(self, circuit, binds):
         parameterizations = []
-        for index, inst_tuple in enumerate(circuit.data):
-            if inst_tuple[0].is_parameterized():
-                for bind_pos, param in enumerate(inst_tuple[0].params):
+        for index, instruction in enumerate(circuit.data):
+            if instruction.operation.is_parameterized():
+                for bind_pos, param in enumerate(instruction.operation.params):
                     if param in binds:
-                        parameterizations.append([[index, bind_pos], binds[param]])
+                        parameterizations.append([(index, bind_pos), binds[param]])
                     elif isinstance(param, ParameterExpression):
                         # If parameter expression has no unbound parameters
                         # it's already bound and should be skipped
                         if not param.parameters:
                             continue
-                        local_binds = {k: v for k, v in binds.items() if k in param.parameters}
-                        bind_list = [dict(zip(local_binds, t)) for t in zip(*local_binds.values())]
+                        len_vals = len(next(iter(binds.values())))
+                        bind_list = [
+                            {
+                                parameter: binds[parameter][i]
+                                for parameter in param.parameters & binds.keys()
+                            }
+                            for i in range(len_vals)
+                        ]
                         bound_values = [float(param.bind(x)) for x in bind_list]
-                        parameterizations.append([[index, bind_pos], bound_values])
+                        parameterizations.append([(index, bind_pos), bound_values])
         return parameterizations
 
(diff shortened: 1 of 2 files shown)


qiskit_aer/backends/aerbackend.py
```diff
@@ -84,20 +84,26 @@ def __init__(self,
 
     def _convert_circuit_binds(self, circuit, binds):
         parameterizations = []
-        for index, inst_tuple in enumerate(circuit.data):
-            if inst_tuple[0].is_parameterized():
-                for bind_pos, param in enumerate(inst_tuple[0].params):
+        for index, instruction in enumerate(circuit.data):
+            if instruction.operation.is_parameterized():
+                for bind_pos, param in enumerate(instruction.operation.params):
```
jakelishman, line 89: ❤️

qiskit_aer/backends/aerbackend.py
```diff
@@ -84,20 +84,26 @@ def __init__(self,
...
-                        parameterizations.append([[index, bind_pos], binds[param]])
+                        parameterizations.append([(index, bind_pos), binds[param]])
                     elif isinstance(param, ParameterExpression):
                         # If parameter expression has no unbound parameters
                         # it's already bound and should be skipped
                         if not param.parameters:
                             continue
-                        local_binds = {k: v for k, v in binds.items() if k in param.parameters}
-                        bind_list = [dict(zip(local_binds, t)) for t in zip(*local_binds.values())]
+                        len_vals = len(next(iter(binds.values())))
```
jakelishman, line 97: Have we checked that `binds` is non-empty before now?  This will fail with a hard-to-understand error if it's empty, which seems like a mistake a user could fairly easily make in a big script file.  If it's not too much overhead, we might also want to test that they're all the same length too, before we start indexing into them.

ikkoham: Yes. This line guarantee the dictionary is not empty. https://github.com/Qiskit/qiskit-aer/blob/01b32ce7100164c81e2ffb7e1752bc733c6aacec/qiskit_aer/backends/aerbackend.py#L388

jakelishman: Sure, but I don't think that's what we're calling `next` on here; we're doing that on sequences contained within `parameter_binds`, I think?  So we could still be in the position where the user has (erroneously) passed us something like
```python
parameter_binds = [{}, {}, {}]
```
`bool(parameter_binds)` is true, but `binds` in this function is bound to (e.g.) `parameter_binds[0]`, at which point a `next(iter(binds.values()))` will fail with a fairly opaque error.  I haven't tested that though, so I could still be wrong.

ikkoham: Ah...You are right. My bad.
image

But current implementation. Kernel is dead..
image

We should have a clear message. Thank you.

jakelishman: Ah yeah, the current implementation won't have any checks at all, so it'll be getting passed down to C++ and dying with some out-of-bounds access in there.  Your new way is definitely better, having the check in Python space, let's just make sure it has a more clear error (or handles zero parameter binds gracefully).

jakelishman (commented):
This seems sensible to me for large numbers of parameters needing to be bound.  Since we found that bug about empty sequences in parameter binds, could we add a quick test of it to prevent a regression, then this seems good to go to me.

jakelishman (approved):
Looks good to me, thanks!

PR #1727, opened 2023-02-17
Fix AerError in Estimator to warn

ikkoham:
### Summary

I realized that raising error here is API breaking. This PR changes it to warning.

### DIFF
--- a/qiskit_aer/primitives/estimator.py
+++ b/qiskit_aer/primitives/estimator.py
@@ -344,10 +344,6 @@ class Estimator(BaseEstimator):
     def _compute_with_approximation(
         self, circuits, observables, parameter_values, run_options, seed
     ):
-        if self._abelian_grouping:
-            warn(
-                UserWarning,
-            )
         # Key for cache
         key = (tuple(circuits), tuple(observables), self.approximation)
         parameter_binds = []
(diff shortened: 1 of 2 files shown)


qiskit_aer/primitives/estimator.py
```diff
@@ -344,7 +345,11 @@ def _compute_with_approximation(
         self, circuits, observables, parameter_values, run_options, seed
     ):
         if self._abelian_grouping:
-            raise AerError("approximation and abelian_grouping cannot be True at the same time.")
+            warn(
```
woodsp-ibm, line 348: Maybe an alternative would be to state in the parameter docstring for abelian_grouping that this setting is ignored when approximation is True. Abelian grouping is True by default (I cannot see most people wanting to disable grouping) hence if you just turn approximation True you will get this warning so it maybe annoying to people to have to go set grouping to False to avoid the warning.

hhorii (approved):

hhorii (commented):
Can you add a test case that `_abelian_grouping` is ignored?

ikkoham:
Thank you. I added the test. (It's hard to confirm that they are in fact being ignored.)

hhorii (approved):

PR #1729, opened 2023-02-22
Pin the version of conan with 1.X to work around API changes in conan 2.0

hhorii:
### Summary

This is just to check CI is correctly working.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
   "setuptools",
   "wheel",
-  "conan==1.59.0",
+  "conan<2.0.0",
   "scikit-build>=0.11.0",
   "cmake!=3.17.1,!=3.17.0",
   "ninja",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
   "setuptools",
   "wheel",
-  "conan>=1.40.0",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 requires = [
   "setuptools",
   "wheel",
-  "conan>=1.40.0",
+  "conan==1.59.0",
```
mtreinish, line 5: Since I doubt we're going to fix 2.0.0 compatibility before the 0.12.0 release. I would make this and the other pins `<2.0.0` so users aren't tied to a particular release of conan to build from source.

mtreinish (commented):
I assume we're planning on releasing 0.12.0 without fixing conan 2.0.0 compatibility. If so we should document this limitation in a `issue` release note regarding aer isn't compatible with conan 2.0.0 and link it to a tracking issue.

mtreinish (approved):
The only open question is do we need to make any changes on the c++ build side for standalone builds since conan is used there too. I'm fine if the answer is just documenting it as a known limitation this release is sufficient, but wanted to ask.

hhorii:
In standalone build, `<2.0.0` is also effective because it uses `conan` installed in `pip install -r requirements-dev.txt`.

PR #1731, opened 2023-02-23
Fix for BLA_VENDOR when compiling from source

acastanedam:
### Summary

At the moment the BLA_VENDOR is always set to OpenBLAS and other blas
libraries are only set if OpenBLAS is not found. Setting to OpenBLAS
only when a BLA_VENDOR is not specified allows for a blas library
defined by the user.

### Details and comments

The suggested change has been made on top of #1729, to avoid conflicts with conan>=2.0


hhorii:
Thank you for your contribution. This fix makes sense to me. Only thing I would like to request for this PR is adding a release note that inform Aer respects `BLA_VENDOR` if it is available.

acastanedam:
Thanks for checking it. I've just added a short (feature) release note.

PR #1676, opened 2022-12-03
[Do not merge] Implement AerDensityMatrix

derwind:
### Summary

This is a test PR to implement AerDensityMatrix on the top of #1616, #1662 and #1663. PRs that have not yet been merged into main branch are tentatively merged in this test PR.

This test PR is also a part of deliverables of QAMP fall 22: [Enhancement of Aer-based quantum_info #23](https://github.com/qiskit-advocate/qamp-fall-22/issues/23)

### Details and comments

Please do not merge this. There are still several portions left to be discussed.

PR #1732, opened 2023-02-27
Implement AerDensityMatrix

derwind:
### Summary

This is a PR to implement AerDensityMatrix, which is a fine-tuned version of #1676.

### Details and comments

When this PR is approved, #1676 should be closed at the same time.

### DIFF
--- a/qiskit_aer/quantum_info/states/aer_densitymatrix.py
+++ b/qiskit_aer/quantum_info/states/aer_densitymatrix.py
@@ -1,0 +1,354 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2017, 2019, 2020, 2021, 2022.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+DensityMatrix quantum state class.
+"""
+import copy
+import numpy as np
+
+from qiskit.circuit import QuantumCircuit, Instruction
+from qiskit.exceptions import QiskitError
+from qiskit.quantum_info.states import DensityMatrix
+from qiskit.quantum_info.operators.predicates import is_hermitian_matrix
+
+from qiskit_aer import AerSimulator
+from .aer_statevector import AerStatevector
+from .aer_state import AerState
+from ...backends.aerbackend import AerError
+from ...backends.backend_utils import BASIS_GATES
+
+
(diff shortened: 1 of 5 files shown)


qiskit_aer/quantum_info/states/aer_densitymatrix.py
```diff
@@ -0,0 +1,354 @@
...
+            else:
+                raise AerError(f'Input data is not supported: type={data.__class__}, data={data}')
+
+            self._aer_state = aer_state
+
+        super().__init__(data, dims=dims)
+
+        self._result = None
+        self._configs = configs
+
```
hhorii, line 102:
```suggestion

    def seed(self, value=None):
        """Set the seed for the quantum state RNG."""
        if value is None or isinstance(value, int):
            self._aer_state.set_seed(value)
        else:
            raise AerError(f'This seed is not supported: type={value.__class__}, value={value}')

```
seed method is missing.

qiskit_aer/quantum_info/states/aer_densitymatrix.py
```diff
@@ -0,0 +1,354 @@
...
+    def sample_memory(self, shots, qargs=None):
+        if qargs is None:
+            qubits = np.arange(self._aer_state.num_qubits)
+        else:
+            qubits = np.array(qargs)
+
+        aer_state = AerState(**self._aer_state.configuration())
+        aer_state.initialize(self._data, copy=False)
+        samples = aer_state.sample_memory(qubits, shots)
+        aer_state.close()
```
hhorii, line 184:
```suggestion
        self._aer_state.close()

        self._aer_state.renew()
        self._aer_state.initialize(self._data, copy=False)
        samples = self._aer_state.sample_memory(qubits, shots)
        self._data = self._aer_state.move_to_ndarray()
```

To keep a random generator in `AerState`, `sample_memory` needs the same `AerState` instance.

hhorii (commented):
To get consistent results of `sample_memory()`, `seed()` needs to be overwritten and `self._aer_state` needs to be used for this method. 

https://github.com/Qiskit/qiskit-aer/commit/a71f17f6b89951542f5d55979c763c7cca23bf1d

qiskit_aer/quantum_info/states/aer_densitymatrix.py
derwind: I had overlooked this. So Terra's QuantumState.seed is called and the effect of the call to seed() in test_aer_densitymatrix.py is inadequate. OK, I will fix this as you suggested.

qiskit_aer/quantum_info/states/aer_densitymatrix.py
derwind: Thank you. I will correct it according to your suggestion and #1663.

hhorii (approved):
Now this looks good to me. Thank you for your contribution.

PR #1733, opened 2023-02-28
Use qiskit_sphinx_theme for API doc

hhorii:
### Summary

Enable `qiskit_sphinx_theme `.

### Details and comments

This makes Aer to follow the other Qiskit components regarding documentation theme.


requirements-dev.txt
```diff
@@ -15,3 +15,5 @@ reno>=3.4.0
 ddt>=1.2.0,!=1.4.0
 matplotlib>=3.3
 seaborn>=0.9.0
+qiskit_sphinx_theme>=1.10
```
mtreinish, line 18: Do we still need the sphinx-rtd-theme above? Similarly do we need `sphinx-automodapi` and `sphinx-tabs` anymore? I don't see modapi in the extension list in conf.py and also while sphinx-tabs is in the list I don't think anythin in the docs source is using it.

hhorii: That's true.
- `sphinx-rtd-theme`
- `sphinx-automodapi`, and 
- `sphinx-tabs`
are not necessary now.

docs/_static/gallery.css
```diff
@@ -0,0 +1,195 @@
+/*
+Sphinx-Gallery has compatible CSS to fix default sphinx themes
```
mtreinish, line 2: Are we using sphinx gallery anywhere? If not then we probably don't need this file.

docs/conf.py
```diff
@@ -43,11 +43,21 @@
...
 # The full version, including alpha/beta/rc tags
 release = '0.12.0'
 
+rst_prolog = """
+.. |version| replace:: {0}
+""".format(release)
+
+html_static_path = ['_static']
+templates_path = ['_templates']
+html_css_files = ['style.css']
```
mtreinish, line 59: This shouldn't be needed as we can use the style sheet from the qiskit_sphinx_theme package.
```suggestion
```

docs/conf.py
```diff
@@ -43,11 +43,21 @@
...
 # The short X.Y version
 version = ''
 # The full version, including alpha/beta/rc tags
 release = '0.12.0'
 
+rst_prolog = """
+.. |version| replace:: {0}
+""".format(release)
+
+html_static_path = ['_static']
```
mtreinish, line 57: The only thing in `_static` is the `style.css` file which shouldn't be necessary anymore so we can remove this and the `_static` directory.

```suggestion
```

mtreinish (commented):
Overall this LGTM, just a couple more comments inline.

docs/conf.py
```diff
@@ -43,11 +43,21 @@
...
+import qiskit_sphinx_theme
+
 # The short X.Y version
 version = ''
 # The full version, including alpha/beta/rc tags
 release = '0.12.0'
 
+rst_prolog = """
+.. |version| replace:: {0}
+""".format(release)
```
jakelishman, line 55: why do this instead of populating the `version` variable above?

docs/conf.py
```diff
@@ -135,7 +140,10 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'sphinx_rtd_theme'  # use the theme in subdir 'theme'
+
+html_theme = "qiskit_sphinx_theme"
+
+html_theme_path = [".", qiskit_sphinx_theme.get_html_theme_path()]
```
jakelishman, line 146: I don't think you need to set this (nor to `import qiskit_sphinx_theme` at the top) - Sphinx should automatically find it through its plugin interface.

docs/conf.py
```diff
@@ -146,7 +154,11 @@
     'display_version': True,
     'prev_next_buttons_location': 'bottom',
     'style_external_links': True,
-    'style_nav_header_background': '#212121',
+    #'style_nav_header_background': '#212121',
```
jakelishman, line 157: can probably just delete this line

jakelishman (commented):
a couple of bits of cleanup, and I agree with Matthew's comments

PR #1738, opened 2023-03-09
Add documentation for density matrix

hitomitak:
### Summary
#1732 added `AerDensityMatrix` but its docs are not maintained.
This PR adds docs that sphnix can include this class in Aer's docs.

### Details and comments

`AerDensityMatrix` included the same description with `DensityMatrix` for the same method. Because `AerDensityMatrix` inherits `DensityMatrix`, such description is not necessary.

In addition, `qiskit_aer/quantum_info/__init__.py` did not list `AerDensityMatrix` and this PR adds it.

### DIFF
--- a/qiskit_aer/quantum_info/__init__.py
+++ b/qiskit_aer/quantum_info/__init__.py
@@ -24,6 +24,7 @@ States
    :toctree: ../stubs/
 
    AerStatevector
+   AerDensityMatrix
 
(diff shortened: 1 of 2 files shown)


jakelishman (approved):
Looks good to me, thanks - it's good to clear out the docstrings when they're just being inherited anyway, yeah.

PR #1740, opened 2023-03-09
Bump version string post release

mtreinish:
### Summary

Now that qiskit-aer 0.12.0 is out the door we should bump the version string on main to show that the version we're installing is newer than what's been released. This commit opens development for 0.13.0 on the main branch.

### Details and comments


jakelishman (approved):
well, we should do the black/clang format before we officially accept more PRs, but yeah.

PR #1741, opened 2023-03-10
Remove redundant wheel dep from pyproject.toml

thesamesam:
### Summary

Remove the redundant `wheel` dependency, as it is added by the backend automatically.

### Details and comments

Listing it explicitly in the documentation was a historical mistake and has been fixed since, see: [pypa/setuptools@f7d30a9](https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a).


jakelishman (approved):
This is good to know, thanks!

thesamesam:
Thanks!

PR #1747, opened 2023-03-14
Improve flexibility and performance of AER::Config

### LINKED ISSUE #1746: Performance regression from 0.11.2 to 0.12.0
### Informations

- **Python version**: 3.9
- **Operating system**: MacOS

Performance degradation due to #1717 is happened in 0.12.0.

#1717 provides pybind for `AER::Config`. This class will help simulation of circuits that use `param_binds` (such as in Primitive), but not for general circuits. `AER::Config` lists all items for configuration even though which are not used in a specified simulation method.

Run circuits with small qubits. 

image

The above results were measured with following codes:

```
from time import time
from qiskit.circuit.library import QuantumVolume

backend = AerSimulator(method='statevector')

print('num_qubits, iteration, elapsed')
for num_qubits in range(3, 15):
	circ = QuantumVolume(num_qubits)
	circ.measure_all()
	circ = transpile(circ, backend)
	
	start_ts = time()
	for j in range(100):
		ret = backend.run(circ).result()
	elapsed = time() - start_ts
	print(num_qubits, 100, elapsed)
```

No regression.

### Suggested solutions

Not list all configuration items.

hhorii:
### Summary

Resolve #1746.

### Details and comments

`AER::Config` has members only for frequently-used configuration items.
`AER::Config.options` covers other items.

All of not-membered configurations are stored in `AER::Config.options` . Therefore, it is easy to enable new configurations (If they are frequently called, we will consider making them members of the class).


qiskit_aer/backends/aer_compiler.py
```diff
@@ -337,12 +337,16 @@ def compile_circuit(circuits, basis_gates=None, optypes=None):
 
 
 def generate_aer_config(
-    circuits: List[QuantumCircuit], backend_options: Options, **run_options
+    circuits: List[QuantumCircuit],
+    default_options: Options,
+    backend_options: Options,
+    **run_options,
```
jakelishman, line 343: Strictly this is a breaking API change in a public method.  You could avoid the worst of the breaks by doing it as
```python
def generate_aer_config(circuits, backend_options, default_options = None, **run_options):
    default_options = default_options if default_options is not None else {}
    ...
```

src/controllers/aer_controller.hpp
```diff
@@ -418,49 +415,39 @@ void Controller::set_config(const Config &config) {
...
-  if (config.blocking_qubits.has_value())
-    cache_block_qubit_ = config.blocking_qubits.value();
+  config.get_option(cache_block_qubit_, "blocking_qubits");
 
   // enable batched multi-shots/experiments optimization
-  batched_shots_gpu_ = config.batched_shots_gpu;
-  batched_shots_gpu_max_qubits_ = config.batched_shots_gpu_max_qubits;
+  config.get_option(batched_shots_gpu_, "batched_shots_gpu");
+  config.get_option(batched_shots_gpu_max_qubits_,
+                    "batched_shots_gpu_max_qubits");
```
jakelishman, line 446: This doesn't necessarily need changing in this PR, but more just a comment: the top message suggests that the infrequently used options are just stored in the `options`, not directly looped through in Python space to bind.

If that's the case, how much time are we spending now binding these options to the instance members in C++ space?  If they're unlikely to actually be accessed, perhaps we should consider just storing the `config` as a member on the class, and turning all the individual options members into getter methods that do the parsing?

It would have the downside of configuration errors getting raised later in the work flow, potentially, but if the binding is a performance bottleneck, that might be worth it?

src/simulators/density_matrix/densitymatrix_state.hpp
jakelishman, line 1: There appears to be a lot of unrelated formatting changes in this file that's making it difficult to review.

src/simulators/statevector/statevector_state.hpp
```diff
@@ -333,54 +333,54 @@ class State : public QuantumState::StateChunk<statevec_t> {
...
+     {"rx", Gates::mcrx},       // Pauli-X rotation gate
+     {"ry", Gates::mcry},       // Pauli-Y rotation gate
+     {"rz", Gates::mcrz},       // Pauli-Z rotation gate
+                                // Waltz Gates
+     {"u1", Gates::mcp},        // zero-X90 pulse waltz gate
+     {"u2", Gates::mcu2},       // single-X90 pulse waltz gate
+     {"u3", Gates::mcu3},       // two X90 pulse waltz gate
+     {"u", Gates::mcu3},        // two X90 pulse waltz gate
+     {"U", Gates::mcu3},        // two X90 pulse waltz gate
+                                // 2-qubit gates
```
jakelishman, line 361: Here's some more unrelated formatting changes, but these ones have a problem too: some of the semantic meaning from the comment positioning is getting lost.

src/simulators/statevector/statevector_state.hpp
```diff
@@ -551,19 +546,17 @@ void State<statevec_t>::set_config(const Config &config) {
...
-  omp_qubit_threshold_ = config.statevector_parallel_threshold;
+  config.get_option(omp_qubit_threshold_, "statevector_parallel_threshold");
 
   // Set the sample measure indexing size
-  if (config.statevector_sample_measure_opt) {
-    int index_size = config.statevector_sample_measure_opt;
-    for (int_t i = 0; i < BaseState::qregs_.size(); i++) {
+  int index_size;
+  if (config.get_option(index_size, "statevector_parallel_threshold")) {
+    for (int_t i = 0; i < BaseState::qregs_.size(); i++)
```
jakelishman, line 558: I don't think that this is a safe change - I think that previously if `index_size` was 0, the `if` wasn't entered, whereas now it's entered provided that `statevector_parallel_threshold` is set to anything, including zero.

qiskit_aer/backends/aer_compiler.py
hhorii: Thanks. Though this method is an internal method, setting default value looks reasonable.

src/simulators/density_matrix/densitymatrix_state.hpp
hhorii: Sorry. I would like to reset these changes.

src/controllers/aer_controller.hpp
hhorii: Though I need to get profile data again, CPU was used to copy `AER::Config` in the last experiment. This is because fields of its class are many.

qiskit_aer/backends/aerbackend.py
```diff
@@ -79,6 +79,9 @@ def __init__(
         if backend_options is not None:
             self.set_options(**backend_options)
 
+        # Set default option to identify diffs
+        self._system_default_options = self._default_options()
+
```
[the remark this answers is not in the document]
hhorii: I found that `self._default_options()` is not correct. Default values of `AER::Config` is `AerSimulator._default_options()`. On the other hand, `self._default_options()` can be changed based on options for `__init__()`. It should be `AerSimulator._default_options()` and move in `generate_aer_config()`.

Anyway, in my Macbook, there is no performance gain by caching default options. I would like to reduce this cache in the next commit.

qiskit_aer/backends/aer_compiler.py
hhorii: As written in https://github.com/Qiskit/qiskit-aer/pull/1747#discussion_r1141580819, `default_options ` is constant and should not be given as an argument.

src/simulators/statevector/statevector_state.hpp
hhorii: This is my mistake. It should be "statevector_sample_measure_opt" as a value of `int index_size`.

src/simulators/statevector/statevector_state.hpp
```diff
@@ -556,14 +556,13 @@ void State<statevec_t>::set_config(const Config &config) {
...
 
   // Set the sample measure indexing size
-  if (config.statevector_sample_measure_opt) {
-    int index_size = config.statevector_sample_measure_opt;
-    for (int_t i = 0; i < BaseState::qregs_.size(); i++) {
+  int index_size;
+  if (config.get_option(index_size, "statevector_sample_measure_opt")) {
+    for (int_t i = 0; i < BaseState::qregs_.size(); i++)
       BaseState::qregs_[i].set_sample_measure_index_size(index_size);
-    }
```
jakelishman, line 566: This still doesn't look exactly correct to me (or maybe it's a different instance of a bug I commented on elsewhere).  `config.get_option` returns a `bool` that indicates whether the option exists, not the _value_ of the option, which is stored into the `index_size` reference - if `index_size` is 0, then in the old code the `if` wouldn't be entered, whereas here it will be.

jakelishman (commented):
I'm still confused by the use of JSON in this PR, but in principle the separation of common vs non-common operations seems ok to me.

src/framework/config.hpp
```diff
@@ -67,448 +67,198 @@ bool get_value(T &var, const std::string &key, const json_t &js) {
...
+    } else {
+      return false;
+    }
+  }
+
+  template <typename T>
+  bool get_option(T &var, const std::string &key) const {
+    return _get_option(options, var, key) ||
+           _get_option(__AER_SIMULATOR_DEFAULTS, var, key);
+  }
```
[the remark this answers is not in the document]
hhorii: A reason why we wanted to reduce JSON is its memory allocation cost and accessing its values. Especially, when lots of parameter values are set in Primitive, most of CPU used for malloc and free to construct and release `nlohmann::json` instances.

Setting `PyObject * handle` sounds great. But maybe we need to leave `nlohmann::json` option for standalone option that does not have dependency to python. I would like to rewrite this PR to use template as containers of configuration values.

src/simulators/statevector/statevector_state.hpp
hhorii: Thanks... You are right. Checking value of `index_size` is necessary here.

hhorii:
@jakelishman sorry. I re-investigated that where is the biggest overheads. I found that `AER::Transpile::CircuitOptimization.set_config()` copies `AER::Config` redundantly and then performance regression happens since 0.11.2.
In 0.11.2, `json_t` is used instead of `AER::Config`. This json has few fields in most cases and then the copy overheads are few.
Now, `AER::Config` has many fields and then copy overheads are increased. I tried to reduce this overheads by reducing not-frequently-used fields of `AER::Config`. But I found that just stopping this redundant copy makes performance better.

Now, this PR becomes very simple :-).

jakelishman (approved):
Ha, that's a pretty funny fix in the end!

PR #1748, opened 2023-03-16
deploy documentation in qiskit.org/ecosystem

1ucian0:
### Summary

The Qiskit Ecosystem lives in qiskit.org/ecosystem and new place for qiskit related projects that are not technically Qiskit. This PR adds that destination for documentation deployment.

### Details and comments

This is the first project for deploying in qiskit.org/ecosystem and we can go slowly. For now, a copy of the same docs in ecosystem would allow to have both q.o/documentation and q.o/ecosystem for testing that everything is fine. Once that's done, the metapackage can create the redirects.


mtreinish (approved):
LGTM this makes sense

PR #1749, opened 2023-03-18
Fix how to reference to config.cuStateVec_enable

derwind:
### Summary

Currently it seems the cuQuantum support version of Qiskit Aer cannot be built.

### Details and comments

This PR fixes the build error.
I attach a Dockerfile to build a fixed Qiskit Aer with cuQuantum support for reference: [Dockerfile.zip](https://github.com/Qiskit/qiskit-aer/files/11007658/Dockerfile.zip)
This PR branch is cloned and CUDA Compute Capability of 7.5 which is compatible with NVIDIA T4 is used. The Docker image should be approximately 10.7 GB in size.
Example codes used inside a container to verify the operation are as follows:
```python
>>> from qiskit import QuantumCircuit
>>> from qiskit_aer.quantum_info import AerStatevector as Statevector
>>> qc = QuantumCircuit(2)
>>> qc.h(0)
>>> qc.cx(0, 1)
>>> sv = Statevector(qc, device='gpu', custatevec_enable=True)
```

### DIFF
--- a/releasenotes/notes/fix-cuStateVec_enable-0936f2269466e3be.yaml
+++ b/releasenotes/notes/fix-cuStateVec_enable-0936f2269466e3be.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed an error when building Qiskit Aer with cuQuautum support.
+    CPU and normal GPU-enabled build are fine, but there was an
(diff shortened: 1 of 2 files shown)


hhorii:
This is a very great catch. Could you add release note as a fix of a bug?

hhorii (approved):
Thank you. This LGTM.

PR #1739, opened 2023-03-09
Prepare 0.12.0 release

mtreinish:
### Summary

This commit prepares for the 0.12.0 release. It moves the release notes to a self contained directory for the release and updates the wording and details of the release notes. Once this commit merges it should be the tagged commit for the release.

### Details and comments

PR #1750, opened 2023-03-18
Remove a release note in an invalid location

derwind:
### Summary

Remove a release note (which was for 0.12.0 release) in an invalid location.

### Details and comments

Essentially equivalent content is already incorporated in #1739.


hhorii (approved):
This is another good catch. Thanks a lot.

PR #1760, opened 2023-03-24
Defer gathering backends until they are needed

### LINKED ISSUE #1736: Log messages emitted on imports
### Informations

- **Qiskit Aer version**:
- **Python version**: Python 3.8.10
- **Operating system**: Linux (Ubuntu 22.04.1 LTS)

Importing any qiskit.aer module will log several messages of the form:

```
INFO:qiskit.compiler.assembler:Total Assembly Time - 0.13256 (ms)
```

Run the following Python program:

```
import logging
logging.basicConfig(level="DEBUG")
import qiskit.providers.aer
```

Importing a module ideally would not log any messages.

### Suggested solutions
The easiest option here would be to change [the backend detection code](https://github.com/Qiskit/qiskit-aer/blob/main/qiskit_aer/aerprovider.py#L33-L61) to run when the list of backends is requested, rather when in the constructor - happy to put a PR together for that.

There was an alternative solution proposed in https://github.com/Qiskit/qiskit-terra/issues/7498#issuecomment-1012416337 which suggested (if I understand correctly) removing the assembling of circuits in `available_methods`/`available_devices` and replacing it with some metadata on the binary.

SquidDev:
### Summary

This defers gathering the available backends in `AerProvider` until a backend is actually requested.

### Details and comments

This is largely written to fix #1736, though avoiding doing this extra work on import seems advantageous anyway.

I put this as a fix in the release notes, though wasn't entirely sure if this was the best place.


SquidDev:
Apologies for the lint failure there - I had run black locally, but not pylint.

hhorii (approved):
Thank you for your contribution. I added `@staticmethod` to `AerProvider._get_backend()` because it touches only class fields.

PR #1766, opened 2023-03-29
Adding circuit executor classes and shot-branching

doichanj:
### Summary
This PR restructures parallel simulation classes that were implemented in `StateChunk` class.
Instead of `StateChunk` this PR introduces `CircuitExecutor` classes outside of `State` classes.  

This PR also implements shot-branching revised from PR #1606 
Now the implementation is simplified in `CircuitExecutor::MultiStateExecutor` class

### Details and comments

image


src/controllers/aer_controller.hpp
```diff
@@ -917,18 +565,7 @@ Result Controller::execute(std::vector<Circuit> &circuits,
```
hhorii, line 565: I think `Controller::execute(std::vector<Circuit> &circuits, ..)` method can be

1. Determine methods (calling `simulation_methods()`)
2. Construct `std::vector<std::shared_ptr<Executor>>`
3. Determine experiment parallelization (`set_parallelization_experiments()`) by using `Executor.required_memory_mb()`
4. Call `Executor. run_circuit()` in parallel or serial

Construction of `Executor` can be via a static method in `simulators.hpp`.

hhorii (commented):
I would like to leave comments on `aer_controller.hpp` first.

src/simulators/circuit_executor.hpp
```diff
@@ -0,0 +1,1174 @@
...
+  // results are stored independently in each process if true
+  bool accept_distributed_results_ = true;
+
+  uint_t myrank_;               // process ID
+  uint_t nprocs_;               // number of processes
+  uint_t distributed_rank_;     // process ID in communicator group
+  uint_t distributed_procs_;    // number of processes in communicator group
+  uint_t distributed_group_;    // group id of distribution
+  int_t distributed_proc_bits_; // distributed_procs_=2^distributed_proc_bits_
+                                // (if nprocs != power of 2, set -1)
```
hhorii, line 87:
```
  uint_t myrank_;               // process ID
  uint_t nprocs_;               // number of processes
  uint_t distributed_rank_;     // process ID in communicator group
  uint_t distributed_procs_;    // number of processes in communicator group
  uint_t distributed_group_;    // group id of distribution
  int_t distributed_proc_bits_; // distributed_procs_=2^distributed_proc_bits_
                                // (if nprocs != power of 2, set -1)
```
Theses values are only for MPI, I think. Can we put them in `#ifdef AER_MPI`. Maybe we set some codes in `aer_controller.hpp` in it.

src/simulators/circuit_executor.hpp
```diff
@@ -0,0 +1,1174 @@
...
+  int num_process_per_experiment_ = 1;
+
+#ifdef AER_MPI
+  // communicator group to simulate a circuit (for multi-experiments)
+  MPI_Comm distributed_comm_;
+#endif
+
+  // settings for cuStateVec
+  bool cuStateVec_enable_ = false;
```
hhorii, line 96: This also can be in `#ifndef AER_THRUST_CPU` I guess.

src/simulators/circuit_executor.hpp
```diff
@@ -0,0 +1,1174 @@
...
+    // set configuration
+    method_ = method;
+    sim_device_ = device;
+
+    set_config(config);
+    set_parallelization(circ, noise);
+
+    // Rng engine (this one is used to add noise on circuit)
+    RngEngine rng;
+    rng.set_seed(circ.seed);
```
hhorii, line 489: `rng.set_seed(circ.seed);` is called in later (`run_circuit_with_sampling` or `run_circuit_shots `). It is better to use different seeds for them.

src/simulators/circuit_executor.hpp
```diff
@@ -0,0 +1,1174 @@
...
+    state.set_global_phase(circ.global_phase_angle);
+
+    state.set_distribution(1);
+    state.set_max_matrix_qubits(max_bits);
+
+    RngEngine rng;
+    rng.set_seed(circ.seed);
+    run_with_sampling(circ, state, result, rng, circ.shots);
+  } else {
+    // Vector to store parallel thread output data
```
hhorii, line 649: I'm confused that `run_circuit_with_sampling`  needs to consider about `parallel_shots_ `. If we use `sampling`, `state` will be calculated once and then get bitstrings from the `state`.

src/simulators/circuit_executor.hpp
```diff
@@ -0,0 +1,1174 @@
...
+      } else {
+        par_results[i].save_count_data(state.creg(), save_creg_memory_);
+      }
+    }
+    state.add_metadata(par_results[i]);
+  };
+  Utils::apply_omp_parallel_for((par_shots > 1), 0, par_shots,
+                                run_circuit_lambda);
+
+  // gather cregs on MPI processes and save to result
```
hhorii, line 776: The below can be in `#AER_MPI` (`gather_creg_memory ` is also), I guess.

hhorii (commented):
Comments for `circuit_executor.hpp` is here:

src/controllers/aer_controller.hpp
doichanj: I create `Controller::make_circuit_executor` function and removed `required_memory_mb` and `validate_state` from `Controller`.

src/simulators/circuit_executor.hpp
doichanj: There are many depending codes to these params, so I would like to keep them available for non MPI environment. But I added `#ifdef AER_MPI `to the codes we can separate.

src/simulators/circuit_executor.hpp
doichanj: I added `#ifdef AER_CUSTATEVEC` to this param.

src/simulators/circuit_executor.hpp
doichanj: Now initial rng used for noise sampling is reused in `run_circuit_xxx` functions.
I think this is the strategy used in the older code.

src/simulators/circuit_executor.hpp
doichanj: removed unnecessary functions and codes

src/simulators/circuit_executor.hpp
doichanj: `#ifdef AER_MPI` inserted

src/simulators/multi_state_executor.hpp
```diff
@@ -0,0 +1,758 @@
...
+    state_t tmp;
+    return tmp.required_memory_mb(circuit.num_qubits, circuit.ops);
+  }
+
+  uint_t get_process_by_chunk(uint_t cid);
+
+protected:
+  void set_config(const Config &config) override;
+
+  void set_distribution(uint_t num_states);
```
hhorii, line 106: Could you add comments to explain this method? Originally, I found the following comment in`state_chunk.hpp` to the method of the same name excepting its argument name `uint_t nprocs`.

> // set number of processes to be distributed

src/controllers/aer_controller.hpp
```diff
@@ -1042,809 +636,128 @@ Result Controller::execute(std::vector<Circuit> &circuits,
...
-  }
+    return std::make_shared<
+        CircuitExecutor::Executor<ExtendedStabilizer::State>>();
+  } break;
   case Method::matrix_product_state: {
-    MatrixProductState::State state;
-    return state.required_memory_mb(circ.num_qubits, circ.ops);
-  }
+    return std::make_shared<
+        CircuitExecutor::Executor<MatrixProductState::State>>();
```
hhorii, line 741: Why not `MultiStateExecutor` is used for MPS and others?

hhorii (commented):
Reviewed Executors. I'm wondering `MultiStateExecutor` is effective if shot-branching improves any methods.

src/controllers/aer_controller.hpp
doichanj: Only a simulation method who supports shot-branching inherits `MultiStateExecutor` at this time.

src/controllers/aer_controller.hpp
```diff
@@ -948,12 +535,14 @@ Result Controller::execute(std::vector<Circuit> &circuits,
     result.metadata.add(max_memory_mb_, "max_memory_mb");
     result.metadata.add(max_gpu_memory_mb_, "max_gpu_memory_mb");
 
+#ifdef AER_MPI
```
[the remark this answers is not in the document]
doichanj: merged ifdef blocks

src/framework/results/experiment_result.hpp
```diff
@@ -217,19 +211,24 @@ void ExperimentResult::save_data_average(const ClassicalRegister &creg,
...
 void ExperimentResult::save_data_pershot(const ClassicalRegister &creg,
                                          const std::string &key, const T &datum,
-                                         OpType type, DataSubType subtype) {
+                                         OpType type, DataSubType subtype,
+                                         const uint_t nshots) {
   switch (subtype) {
   case DataSubType::single:
-    data.add_single(datum, key);
+    for (int_t i = 0; i < nshots; i++)
+      data.add_single(datum, key);
```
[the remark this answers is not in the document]
doichanj: removed nshots and loop from `save_dat_pershot`

hhorii (approved):
Though I was not able to review the detail of implementation due to the size of changes, I believe that we can merge this PR and have time to refine implementation by the release of 0.13.0.

PR #1770, opened 2023-04-04
Add implicit cast of argument types

### LINKED ISSUE #1754: Simulators stopped accepting float input for shots
### Informations

- **Qiskit Aer version**: latest
- **Python version**:
- **Operating system**:

Previous versions of Aer allowed for setting shots values using things like `shots=1e4`.  The latest version stopped doing this and gives a cryptic error:

```
TypeError: (): incompatible function arguments. The following argument types are supported:
    1. (self: qiskit_aer.backends.controller_wrappers.AerConfig, arg0: int) -> None

Invoked with: <qiskit_aer.backends.controller_wrappers.AerConfig object at 0x7f7291bd0730>, 10000.0
```

Run a simulation, e.g. via the fake backends using a float representation of an int

This represents a breaking change, and should probably be fixed by casting.

hhorii:
### Summary

Fix #1754.

### Details and comments

Since 0.12.0, AerConfig is used to configure simulation, which is directly bound to a AER::Config object through pybind. This change requires application to specify strictly correct types of configuration options. This commit allows implicit casting to arguments if application specifies them with wrong types.


qiskit_aer/backends/aer_compiler.py
```diff
@@ -336,6 +337,83 @@ def compile_circuit(circuits, basis_gates=None, optypes=None):
...
+        return v
+    try:
+        ret = BACKEND_RUN_ARG_TYPES[k](v)
+        warn(
+            f'A type of an option "{k}" is {BACKEND_RUN_ARG_TYPES[k].__name__} '
+            "but {v.__class__.__name__} was specified."
+            "Implicit cast for an argument has been deprecated as of qiskit-aer 0.12.1.",
+            DeprecationWarning,
+            stacklevel=5,
+        )
```
jakelishman, line 409: This will warn a _little_ overzealously when scalar values are taken from Numpy arrays - `isinstance(numpy.int64(1), int)` is `False`, but should have no problems passing through the pybind11 boundary.  We might want to refactor the lookup table to include the coercion type (`int` / `bool` / `float`), and a tuple of permissible types (`(int, numpy.integer)` / `(bool, numpy.bool_)`, `(float, numpy.floating)`).

qiskit_aer/backends/aer_compiler.py
```diff
@@ -336,6 +337,83 @@ def compile_circuit(circuits, basis_gates=None, optypes=None):
...
+    "max_parallel_experiments": int,
+    "max_parallel_shots": int,
+    "max_memory_mb": int,
+    "fusion_enable": bool,
+    "fusion_verbose": bool,
+    "fusion_max_qubit": int,
+    "fusion_threshold": int,
+    "accept_distributed_results": bool,
+    "memory": bool,
+    # "noise_model": NoiseModel,
```
jakelishman, line 361: If the two comments are meant to be semi-documenting for consistency, it might be better to explain at the top why we don't include them - it's a bit non-obvious why they're here at the moment.

qiskit_aer/backends/aer_compiler.py
```diff
@@ -352,16 +430,32 @@ def generate_aer_config(
...
+        config.n_qubits = num_qubits
+        for key, value in backend_options.__dict__.items():
+            if hasattr(config, key) and value is not None:
+                value = _validate_option(key, value)
+                setattr(config, key, value)
+        for key, value in run_options.items():
+            if hasattr(config, key) and value is not None:
+                value = _validate_option(key, value)
+                setattr(config, key, value)
+        return config
```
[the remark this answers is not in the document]
hhorii: Sure. Let me check this cost.

qiskit_aer/backends/aer_compiler.py
hhorii: Good points.

qiskit_aer/backends/aer_compiler.py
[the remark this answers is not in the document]
hhorii: No performance impacts w/ vs w/o try-except. Also no performance impact optional and always validation. Maybe always do validation in try-except is fine.

jakelishman (commented):
Aside from a minor comment about how we're doing the checking, this seems ok to me to solve the bug as reported.

jakelishman (approved):

PR #1772, opened 2023-04-05
Do not use circuit metadata to internally manage simulation results

### LINKED ISSUE #1723: `AerBackend` edits circuits metadata inplace on `run`
### Informations

- **Qiskit Aer version**: 0.11.2
- **Python version**: 3.11
- **Operating system**: MacOS

Under `AerBackend.run`, input circuits get their metadata edited in place.

```python
from qiskit.providers.fake_provider import FakeNairobiV2

qc = QuantumCircuit(2)
qc.metadata = {"foo": "bar"}

circuits = [qc.copy() for _ in range(5)]

print("METADATA BEFORE EXECUTION:")
_ = [print(c.metadata) for c in circuits]

backend = FakeNairobiV2()
backend.run(circuits)

print("METADATA AFTER EXECUTION:")
_ = [print(c.metadata) for c in circuits]
```
ouputs:
```
METADATA BEFORE EXECUTION:
{'foo': 'bar'}
{'foo': 'bar'}
{'foo': 'bar'}
{'foo': 'bar'}
{'foo': 'bar'}
METADATA AFTER EXECUTION:
{'metadata_index': 0}
{'metadata_index': 1}
{'metadata_index': 2}
{'metadata_index': 3}
{'metadata_index': 4}
```
_(*) This is the behavior on a freshly initialized environment. The behavior seems somewhat erratic nonetheless if the kernel in my jupyter notebook is not restarted, occasionally preserving the metadata intact._

Metadata should not be edited in place (i.e. by reference).

### Suggested solutions
Make a copy of the metadata before handling it.

hhorii:
### Summary

This fixes `AerSimulator` to use circuit metadata to maintain mapping from input and output of an executor call. 
This resolves an issue https://github.com/Qiskit/qiskit-aer/issues/1723.

### Details and comments

Previously metadata is replaced with metadata defined by `AerSimulator`. This is because metadata may include unserializable object (see #1435). Since Aer can manage metadata since `AerCircuit` was added, Aer no longer needs to serialize metadata for simulation.

With this PR, each experiment result has a link to `AerCircuit` that generates the result and then `AerBackend` maintains metadata from a circuit corresponding to the `AerCircuit`. `QuantumCircuit.metadata` is not modified while simulation.


hhorii:
@jakelishman I have updated this PR for experiment result to have a link to the `AerCircuit` that produced the result. I did not use `py::handle` to directly set metadata to the result because I wanted to minimize dependencies to pybind11 in C++.

jakelishman (approved):
This looks reasonable to me, and the method of storing the metadata pointer in the C++ side looks like a reasonable compromise to support non-Python-based setups as well.

jakelishman (approved):

PR #1775, opened 2023-04-07
Fix wrong set of parameters in circuits with barriers

### LINKED ISSUE #1774: AerSimulator does not work for the circuit with barriers and parameters.
### Informations

- **Qiskit Aer version**: 0.12 (the following works for 0.11)
- **Python version**:
- **Operating system**:

AerSimulator does not work for the circuit with barriers and parameters.

```python
from qiskit.circuit.library import RealAmplitudes

qc1 = RealAmplitudes(1, insert_barriers=True)
AerSimulator().run(qc1.decompose(), parameter_binds=[{k: [0] for k in qc1.parameters}])
```
raises an error:
```
ERROR: Failed to load circuits: Invalid parameterized qobj: instruction position out of range
```

No error

hhorii:
`AerCircuit` is created from a circuit by iterating its instrucitons while skpping barrier. This leads inconsistency with positions of parameter bindings. This commit adds barrier instruction to the class and keeps positions of parameter bindings.

### Summary

This fixes #1774 

### Details and comments

`AerCircuit` is created while skipping barrier instructions. However, skipping barrier instructions makes positions of parameter bindings wrong. This fix use `barrier()` a new method of `AerCircuit` and then keeps positions of parameter bindings consistent.

### DIFF
--- a/test/terra/backends/test_parameterized_qobj.py
+++ b/test/terra/backends/test_parameterized_qobj.py
@@ -421,11 +421,12 @@ class TestParameterizedQobj(common.QiskitAerTestCase):
         circuit.barrier()
         circuit.measure_all()
 
-        parameter_binds = [{theta: [pi/2], phi: [pi/2]}]
+        parameter_binds = [{theta: [pi / 2], phi: [pi / 2]}]
         res = backend.run([circuit], shots=1024, parameter_binds=parameter_binds).result()
 
         self.assertSuccess(res)
         self.assertEqual(res.get_counts(), {"111": 1024})
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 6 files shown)


test/terra/backends/test_parameterized_qobj.py
```diff
@@ -404,6 +404,29 @@ def test_run_empty(self):
...
+        circuit.rx(theta, 2)
+        circuit.barrier()
+        circuit.rx(phi, 0)
+        circuit.rx(phi, 1)
+        circuit.rx(phi, 2)
+        circuit.barrier()
+        circuit.measure_all()
+
+        parameter_binds = [{theta: [pi / 2], phi: [pi / 2]}]
+        res = backend.run([circuit], shots=1024, parameter_binds=parameter_binds).result()
```
jakelishman, line 425: I think the `shots=2000` on line 409 is accidentally left in - it's never used.  (I'm a little surprised there's no linter that complains about it.)

jakelishman (commented):
Aside from a super minor nitpick in the test, this looks good to me, thanks.

jakelishman (approved):

PR #1776, opened 2023-04-07
Release 0.12.1

hhorii:
### Summary

Prepare release of 0.12.1


hhorii:
Note that we need to wait for approval of limit request in pypi to release new versions: https://github.com/pypi/support/issues/2677

jakelishman (commented):
We probably need a "prelude" release note for Aer 0.12.1 itself - on Terra, we usually call that note something like `prepare-aer-0.12.1`, and for a bugfix release we just put a sentence saying "Qiskit Aer 0.12.1 is a minor bugfix release" or something like that.  If there's a _super_ major bug, we sometimes call it out in the prelude, otherwise we just let the rest of the release notes speak for themselves.

jakelishman (approved):
The version numbers appear to be bumped correctly in the places I expect, and it's got a prelude release note that looks fine, but I didn't look further than that (since the PR is +37k,-30k lines haha).

PR #1778, opened 2023-04-10
Support `SwitchCaseOp` that will be released in qiskit-terra 0.24.0

### LINKED ISSUE #1759: Support `switch` statement in dynamic-circuits simulations

Terra 0.24 is in the process of adding a `switch` statement representation (see Qiskit/qiskit-terra#9833).  We'll want Aer to support that in its dynamic-circuits simulations as well.  This issue is just a tracker for that.

hhorii:
### Summary

Add support of `SwitchCaseOp`

### Details and comments

Until 0.24.0 is released, `SwitchCaseOp` defined in `qiskit_aer/library/control_flow_instructions/switch_case.py` is used. Once 0.24.0 is released, the second commit should be reset.

This will resolve #1759.


qiskit_aer/backends/aer_compiler.py
```diff
@@ -328,6 +347,95 @@ def _inline_if_else_op(self, instruction, continue_label, break_label, parent, b
...
+            bit_map[c]
+            for c in (
+                (
+                    {instruction.operation.target}
+                    if isinstance(instruction.operation.target, Clbit)
+                    else set(instruction.operation.target)
+                )
+                - set(instruction.clbits)
+            )
+        )
```
jakelishman, line 371: Minor, but I think this becomes a lot simpler if you just make `mark_cargs` a set right from the start.  It becomes one later on anyway (with what appears to be some duplicated calculations), and it doesn't need any ordering.

qiskit_aer/backends/aer_compiler.py
```diff
@@ -328,6 +347,95 @@ def _inline_if_else_op(self, instruction, continue_label, break_label, parent, b
...
+        case_labels = []
+        case_args_lists = []
+        case_bit_maps = []
+        case_bodies = []
```
jakelishman, line 375: These four lists are hard for me to reason about, and I think they get out-of-sync; if `CASE_DEFAULT` is in a set of labels, we don't append anything to `case_args_lists`.  It might be simpler to make a single list whose scalar type is something like a
```python
case_object = collections.NamedTuple("case_object", ("label", "args", "bit_map", "body"))
```

jakelishman (commented):
Thanks Hiroshi, it's great to have the PR for this support so quickly!

hhorii:
@jakelishman Thanks. With you comments, codes became much simpler, I think.

jakelishman (commented):
In principle I think this looks fine.  We should probably wait until Terra 0.24 is released properly before merge, though, so we don't need the Aer temporaries.

hhorii:
@jakelishman thank you for your review. terra 0.24.0 was released and I changed this PR to use its `SwitchCaseOp`.

PR #1786, opened 2023-04-20
Remove REQUIRED to find PythonLibs in FindPybind11

hhorii:
### Summary

Remove `REQUIRED` to find `PythonLibs` in FindPybind11

### Details and comments

As @henryiii mentioned [here](https://github.com/Qiskit/qiskit-aer/pull/1786#issuecomment-1526151867), `REQUIRED` is not necessary to find `PythonLibs`. By removing this, Aer's CI will be back from the following error:

```
Could NOT find PythonLibs (missing: PYTHON_LIBRARIES) (found version "3.7.15")
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
   "setuptools",
   "conan<2.0.0",
-  "scikit-build>=0.11.0",
+  "scikit-build==0.17.1",
   "cmake!=3.17.1,!=3.17.0",
   "ninja",
   "pybind11>2.6",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
   "setuptools",
   "conan<2.0.0",
-  "scikit-build==0.17.1",
+  "scikit-build<0.17.2",
   "cmake!=3.17.1,!=3.17.0",
   "ninja",
   "pybind11>2.6",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 requires = [
   "setuptools",
   "conan<2.0.0",
-  "scikit-build>=0.11.0",
+  "scikit-build==0.17.1",
```
mtreinish, line 5: I would make this a version cap but not restrict it to explicitly the only known working version. Something like: 
```suggestion
  "scikit-build<0.17.2",
```

mtreinish (commented):
This was a good find, it looks like it's been reported upstream already too with: https://github.com/scikit-build/scikit-build/issues/957 so we can pin the version until there is a fix upstream. Just one comment inline about the mechanics

henryiii:
The problem is this:

https://github.com/Qiskit/qiskit-aer/blob/c3070fbecd1fdbd903abb5a43691f9eba5920818/cmake/FindPybind11.cmake#L1

There are no Python libs in the manylinux image, and it's a bug to link to it on UNIX-like systems - it will cause possible segfaults if you use a different patch version of Python. You only need to link to it on Windows.

You can removed the "REQUIRED" there and that should fix it. Though pybind11 actually has shipped an excellent pybind11Config.cmake since around 2.6, so I'd recommend just deleting this file and using `find_package(pybind11 CONFIG REQUIRED)` instead.

hhorii:
@henryiii Thank you for your comments. I removed `REQUIRED` and build was succeeded.

mtreinish (approved):

PR #1792, opened 2023-04-24
fix fail of qobj run without run_options

### LINKED ISSUE #1771: Duplicate parameter_binds
### Informations

- **Python version**:
- **Operating system**:

Qobj is being deprecated so it's probably not worth fixing. But if I `assemble()` a circuit with an Aer backend, then call its `backend.run()`, the job would fail with 

```
_run_qobj() got multiple values for argument 'parameter_binds'
```

because `assemble` inserts `parameter_binds` into `circuits.config`.

```
from qiskit import transpile, assemble
from qiskit.test.reference_circuits import ReferenceCircuits
from qiskit_aer import Aer

sim = Aer.get_backend("aer_simulator")
bell = ReferenceCircuits.bell()
qobj = assemble(transpile(bell, sim), backend=sim)
sim.run(qobj)
```

I can still run a qobj with Aer.

hhorii:
### Summary

resolve #1771.

### Details and comments

This PR is to avoid duplicated arguments to call `backend. _run_qobj()`.

### DIFF
--- a/qiskit_aer/backends/aerbackend.py
+++ b/qiskit_aer/backends/aerbackend.py
@@ -186,8 +186,8 @@ class AerBackend(Backend, ABC):
                 for key, value in circuits.config.__dict__.items():
                     if key not in run_options and value is not None:
                         run_options[key] = value
-                if "parameter_binds" in run_options:
-                    parameter_binds = run_options.pop("parameter_binds")
+            if "parameter_binds" in run_options:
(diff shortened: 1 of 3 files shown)


mtreinish (approved):
LGTM, this seems straightforward enough. I was thinking it was surprising that we didn't have coverage for this, but then I realized all the existing coverage wasn't explicitly building a qobj.

PR #1794, opened 2023-04-26
Attempt to pin manylinux image version in wheel jobs

mtreinish:
### Summary

Recently the linux wheel build jobs are failing because it is unable to find the python 3.7 libraries to link against when building wheels under cibuildwheel. It appears there were new manylinux docker images released around the same time the jobs started failing. This commit pins the docker image version we use in the jobs to try and fix the failure. If this doesn't work the next thing ot try is bumping the cibuildwheel version we use in the job.

### Details and comments

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,8 +12,8 @@ requires = [
 build-backend = "setuptools.build_meta"
 
 [tool.cibuildwheel]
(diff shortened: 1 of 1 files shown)


mtreinish:
This isn't the underlying cause of the error. #1786 is the real fix

PR #1801, opened 2023-05-08
Fix library path to cuQuantum and cuTENSOR

doichanj:
### Summary
This PR fixes library path to cuQuantum SDK defined in CMakeLists.txt

### Details and comments
Because the latest cuQuantum and cuTENSOR have multiple libraries for different CUDA versions.
So this fix uses CUDA version number returned from FindCUDA to set library path.


hhorii (approved):
LGTM.

PR #1802, opened 2023-05-08
Activate jQuery in docs

Eric-Arellano:
### Summary

Sphinx 6 no longer activates jQuery by default. The qiskit_sphinx_theme still uses it, so we now have the theme include jQuery. But, you have to add the theme to `extensions` in `conf.py` for that to take effect.


hhorii:
@Eric-Arellano Thanks. This Lint error is from qiskit-terra 0.24 that removes `close()` from `Sampler` with its docstring. I created #1803 and @ikkoham is resolving it with #1804.

hhorii (approved):
LGTM.

PR #1808, opened 2023-05-10
Fix setting MPI processes and ranks

doichanj:
### Summary
This fix sets MPI processes and ranks correctly to enable MPI parallelization.

### Details and comments
Because qobj is not used, `Aer::Controller::execute` function with qobj is not called and MPI processes and ranks are not set correctly.
So this fix calls `MPI_Comm_size` and `MPI_Comm_rank` in another `execute` function.


hhorii (approved):
LGTM

PR #1812, opened 2023-05-15
Fix cuQuantum static linking

doichanj:
### Summary
This is fix for static link of cuQuantum API when building with `-DCUQUANTUM_STATIC=true`

### Details and comments
This fix corrects required static libraries for cuQuantum


hhorii (approved):
I confirmed that this works well: https://gist.github.com/hhorii/548c3d0afb50b7b06a3136152524d274

Issue #1795, opened 2023-04-26
Aer Estimator wrongly reads global phase


A-tA-v:
### Informations

- **Qiskit Aer version**: 0.12.0
- **Python version**: 3.10
- **Operating system**: Mac OS Ventura, Intel

### What is the current behavior?
Aer Estimator crashes while running simulation.

### Steps to reproduce the problem
Script below

### What is the expected behavior?
I create a parameter placeholder at the beginning.
After prepending the circuit with ZFeatureMap and several compositions
the final circuit obtains a non-zero global phase represeneted as a linear
combination of placeholder's entries. I do *NOT* change the global phase
myself, rather this is the outcome of aforementioned operations.
At the time of evaluation, the linear combination of placeholder's entries
cannot not be cast to a float value causing an exception.

In the simple example below, I reproduced the behaviour by explicitely 
expressing the global phase as a sum of two placeholder entries.
Note, the default estimator works without a problem.

### Suggested solutions
Checking of global phase before evaluation is needed.

### Script reproducing the issue:
```
import time
import numpy as np
from qiskit import QuantumCircuit
from qiskit.circuit import ParameterVector
from qiskit.primitives import Estimator as BackendEstimator
from qiskit.providers.fake_provider import FakeKolkataV2
from qiskit.quantum_info import SparsePauliOp
from qiskit.utils import algorithm_globals
from qiskit_aer.primitives import Estimator as AerEstimator


def report():
    num_qubits = 10
    algorithm_globals.random_seed = 1234567
    np.random.seed(algorithm_globals.random_seed)

    # Create a circuit with parameter placeholder.
    params = ParameterVector(name="r", length=num_qubits - 1)
    qc = QuantumCircuit(num_qubits)
    for i in range(num_qubits - 1):
        qc.cx(i, i + 1)
        qc.rz(params[i], i)
    qc.global_phase = params[0] + params[1]  # <<<<< M A K E S  T H E  P R O B L E M

    params = np.random.rand(len(params))
    observable = SparsePauliOp.from_list([("Z" * qc.num_qubits, 1)])

    tic = time.perf_counter()
    backend_options = {"seed_simulator": algorithm_globals.random_seed, "shots": 4096}
    use_aer = 1  # <<<<< choose estimator here
    if use_aer != 0:
        transpile_options = {"seed_transpiler": algorithm_globals.random_seed}
        run_options = {"shots": 1000}
        estimator = AerEstimator(
            backend_options=backend_options,
[...]

Issue #1781, opened 2023-04-14
Problems with the extended_stabilizer method


acastanedam:
### Informations

- **Qiskit AER version**: main (@0.13.0) 
- **Operating system**: Linux

### What is the current behavior?
Aersimulator run and estimator do not work when method=extended_stabilizer

### Steps to reproduce the problem
This small script shows what is happening:

``` python
import numpy as np
from qiskit_aer import AerSimulator
from qiskit.circuit.library import EfficientSU2
from qiskit_aer.primitives import Estimator as aer_estimator
from qiskit import transpile

backend = AerSimulator(method="extended_stabilizer")

ansatz = EfficientSU2(1)
print(ansatz.decompose())

num_parameters = ansatz.num_parameters
parameters = np.random.rand(num_parameters)
parameters = np.reshape(parameters, (-1, num_parameters)).tolist()
batch_size = len(parameters)

# RUN
try:
# This does not work ...
    _ansatz = transpile(ansatz, backend=backend, basis_gates=backend.configuration().basis_gates, optimization_level=2)
    print(_ansatz.decompose())
    job = backend.run(_ansatz, parameter_binds=[{k: [0] for k in _ansatz.parameters}])
    print(job.result())
except:
# ... however this does
    print(ansatz.decompose())
    job = backend.run(ansatz.decompose(), parameter_binds=[{k: [0] for k in _ansatz.parameters}])
    print(job.result())

# PRIMITIVE
estimator = aer_estimator(
    backend_options=backend.options._fields,
    transpile_options={"optimization_level": 2},
    run_options=None,
    approximation=False,
    skip_transpilation=False,
)

try:
    print("Estimator 1")
    # This does not work ...
    operator = "X"
    job = estimator.run(batch_size * [ansatz], batch_size * [operator], parameters)
    print(job.result())
except:
    print("Estimator 2")
    # neither this ...
    job = estimator.run(batch_size * [ansatz.decompose()], batch_size * [operator], parameters)
    print(job.result())
finally:
    # nor this ...
    print("Estimator 3")
    estimator = aer_estimator(
        backend_options=backend.options._fields,
        transpile_options={"optimization_level": 3},
        run_options=None,
        approximation=False,
        skip_transpilation=True,
    )
    _ansatz = transpile(ansatz, backend=backend, basis_gates=backend.configuration().basis_gates, optimization_level=2)
    job = estimator.run(batch_size * [_ansatz], batch_size * [operator], parameters)
[...]

PR #1814, opened 2023-05-17
Support parameterized global phase

### LINKED ISSUE #1798: Sampler and Estimator with noise model doesn't work consistently for parametrized circuits
### Informations

- **Python version**: 3.11
- **Operating system**: osx

Running parametrized circuits using Sampler or Estimator with noise model fails. Although it works when the noise model is built using certain fakebackends like FakeCambridge. 

The following code works when `backend2` is passed to the noise model but fails with `backend1`.
Same for Estimator as well. 

```python
from qiskit import *
from qiskit.circuit import ParameterVector
from qiskit.providers.fake_provider import FakeCambridge, FakeKolkata
from qiskit_aer.noise import NoiseModel
from qiskit_aer.primitives import Sampler

backend1, backend2 = FakeCambridge(), FakeKolkata()
noise = NoiseModel.from_backend(backend2)

run_opt = {'shots':10000}
backend_opt = {'noise_model':noise}

samp = Sampler(backend_options=backend_opt, run_options=run_opt)

ang = ParameterVector('x', 3)
qc = QuantumCircuit(1)
qc.u(*ang, [0])
qc.measure_all()

res = samp.run(qc, parameter_values=[1.3, 2.1, 0.7]).result()
print(res.quasi_dists)
```

It should work consistently regardless of what fakebackend is used for noise model.

hhorii:
### Summary

Though `ParameterExpression` can be set to `QuantumCircuit.global_phase`, Aer does not bind parameter values to it in simulation phase. 
This PR fixes this problem by resolving values of `global_phase` with specified `parameter_binds` in `AerSimulator.run` as well as for parameter values of gates.

### Details and comments

This PR resolves #1798, #1781, and #1795.


src/controllers/controller_execute.hpp
```diff
@@ -65,7 +65,7 @@ Result controller_execute(std::vector<Circuit> &input_circs,
   //    i is the instruction index in the experiment
   //    j is the param index in the instruction
   //    pars = [par0, par1, ...] is a list of different parameterizations
-  using pos_t = std::pair<uint_t, uint_t>;
+  using pos_t = std::pair<int_t, int_t>;
```
doichanj, line 68: I think it is not necessary to change `unit_t` to `int_t` to point this params are global phases, and use pre-defined value such as `GLOBAL_PHASE_POS` instead of using `-1`

doichanj (changes requested):
I think it is better to define value to identify global phase rather than using `-1`

qiskit_aer/backends/aer_compiler.py
```diff
@@ -465,7 +465,13 @@ def assemble_circuit(circuit: QuantumCircuit):
 
     qreg_sizes = []
     creg_sizes = []
-    global_phase = float(circuit.global_phase)
+    if (
+        isinstance(circuit.global_phase, ParameterExpression)
+        and len(circuit.global_phase.parameters) > 0
+    ):
+        global_phase = 1.0
```
jakelishman, line 472: Very unimportant, but I'd think that the more natural default is 0, not 1 - `QuantumCircuit.global_phase` is the angle, not the unitary value.

jakelishman (commented):
This largely looks fine to me.  Needing to overload special values into the `uint_t` isn't super ideal, but doesn't look that bad to me - I think it'd be fine to either do it with `GLOBAL_PHASE_POSITION == UINT_MAX` or something as suggested, or to just continue with the switch to signed integers that's done here.  In terms of bit patterns, they're not dissimilar - you're just assigning special meaning to values in a certain range.

doichanj (approved):

Issue #1773, opened 2023-04-05
Aer simulates parameterized circuits without parameter binds


hhorii:
### Informations

- **Qiskit Aer version**: 0.12.0
- **Python version**:
- **Operating system**:

### What is the current behavior?

Aer simulates parameterized circuits without parameter binds

### Steps to reproduce the problem

```
from qiskit import transpile
from qiskit.circuit.library import EfficientSU2
from qiskit_aer import AerSimulator
from numpy.random import random
from time import time

simulator = AerSimulator()

circ = EfficientSU2(5, reps=1)
circ.measure_all()
circ = transpile(circ, simulator)

simulator.run([circ]).result().get_counts()
```

### What is the expected behavior?

The above code raises an error.

### Suggested solutions

Check length of `parameter_binds`.

PR #1817, opened 2023-05-18
Check parameter_binds arguments for parameterized circuits

hhorii:
### Summary

If `QuantumCircuit` has `parameters` and `parameter_binds` argument does not exist, 
`AerSimulator` raises `AerError` with a message:`"circuits have parameters but parameter_binds is not specified."`

### Details and comments

Since 0.12.0, parameter values in circuits are temporarily replaced with constant values
and parameter values are assigned in C++ library. Therefore, if `parameter_binds` is specified,
simulator returns results with the constnat values as paramter values. With this PR,
Aer raises an error if `parameter_binds` is not specified though circuits have parameters.

This will resolve https://github.com/Qiskit/qiskit-aer/issues/1773.


jakelishman (dismissed):
This seems sensible to me.

jakelishman (approved):

Issue #1815, opened 2023-05-17
Initializing Noise model from backend gives TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'


kevinsung:
### Informations

- **Qiskit Aer version**: 0.12.0
- **Python version**: 3.10.11
- **Operating system**: Arch Linux

### Steps to reproduce the problem
```python
from qiskit_aer.noise import NoiseModel
from qiskit_ibm_provider import IBMProvider

hub = "ibm-q-internal"
group = "deployed"
project = "default"
provider = IBMProvider(instance=f"{hub}/{group}/{project}")
backend = provider.get_backend("ibm_sherbrooke")
noise_model = NoiseModel.from_backend(backend)
```
```
Traceback (most recent call last):
  File "/home/kjs/projects/noise-conversion/bug-aer.py", line 9, in <module>
    noise_model = NoiseModel.from_backend(backend)
  File "/usr/local/venvs/noise-conversion-Sq5ilF1M/lib/python3.10/site-packages/qiskit_aer/noise/noise_m
odel.py", line 366, in from_backend
    gate_errors = basic_device_gate_errors(
  File "/usr/local/venvs/noise-conversion-Sq5ilF1M/lib/python3.10/site-packages/qiskit_aer/noise/device/
models.py", line 152, in basic_device_gate_errors
    return _basic_device_target_gate_errors(
  File "/usr/local/venvs/noise-conversion-Sq5ilF1M/lib/python3.10/site-packages/qiskit_aer/noise/device/
models.py", line 252, in _basic_device_target_gate_errors
    relax_error = _device_thermal_relaxation_error(
  File "/usr/local/venvs/noise-conversion-Sq5ilF1M/lib/python3.10/site-packages/qiskit_aer/noise/device/
models.py", line 337, in _device_thermal_relaxation_error
    t2 = _truncate_t2_value(t1, t2)
  File "/usr/local/venvs/noise-conversion-Sq5ilF1M/lib/python3.10/site-packages/qiskit_aer/noise/device/
models.py", line 351, in _truncate_t2_value
    if t2 > 2 * t1:
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

### What is the expected behavior?
Should work.

hhorii:
I think this is the same issue with #1779. Some parameters of noise (such as T1) are `None` which is not expected. Probably no specification about `None` value so Aer will handle it as ideal (no noise) gates and qubits.

[...]

PR #1818, opened 2023-05-18
Fix handling of None in noise model construction from BackendV2

### LINKED ISSUE #1779: AerSimulator.from_backend does not handle faulty-qubits
### Informations

- **Qiskit Aer version**: latest
- **Python version**:
- **Operating system**: OSX

Trying to run `AerSimulator.from_backend` on a backend with `faulty_qubits` leads to errors.

It builds a noise model for all qubits not in the `faulty_qubits` list

### Suggested solutions
Ignore qubits that are in the `faulty_qubits` list

itoko:
### Summary
Fixes `NoiseModel.from_backend` not to fail when calling it with a V2 backend having faulty qubits such that T1 and T2 are None in the `target.qubit_properties`. Fixes #1779 and #1815.

### Details and comments
Previously, `NoiseModel.from_backend`  had no handling of None case for T1 or T2 value defined in a `target.qubit_properties` and raised a `TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'` as reported in #1779 and #1815. This commit updates `basic_device_gate_errors` function and `NoiseModel.form_backend` method so that both handle None for T1 or T2 without failure. After this commit, in the case when T1 and T2 values are None for a qubit, an identity QuantumError (i.e. effectively no thermal relaxation error) will be added to the qubit for all gates acting on qubits including the qubit.

### DIFF
--- a/qiskit_aer/noise/device/models.py
+++ b/qiskit_aer/noise/device/models.py
@@ -107,11 +107,12 @@ def basic_device_gate_errors(
         properties (BackendProperties): device backend properties.
         gate_error (bool): Include depolarizing gate errors (Default: True).
         thermal_relaxation (Bool): Include thermal relaxation errors (Default: True).
-                If ``T1`` or ``T2`` value is not supplied via ``target`` (i.e. None) for some qubits,
-                no thermal relaxation errors are added to the qubits even if this flag is True.
-                If ``frequency`` is not defined (i.e. None) for a qubit, no excitation is considered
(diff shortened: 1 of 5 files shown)


hhorii (approved):
Setting `np.inf` to `t1` and `t2` sounds reasonable for the current IBM backends. They return `None` of `t1` and `t2` if their qubits are not available. If transpiler works for circuits, these qubits will not be used and noise based on `np.inf` will not be applied.

PR #1823, opened 2023-05-25
Warn if approximation=False and shots=None in Estimator

### LINKED ISSUE #1820: Estimator handles shots=None incorrectly
### Informations

- **Python version**: 3.10.11
- **Operating system**: Fedora 38

If you run the Estimator with `shots=None` the result has `shots=1024`.

```python
from qiskit import QuantumCircuit, QuantumRegister, transpile
from qiskit_aer.primitives import Estimator

hamiltonian = SparsePauliOp("X")
qubits = QuantumRegister(1)
circuit = QuantumCircuit(qubits)
circuit.measure_all()

estimator = Estimator()
job = estimator.run(circuit, hamiltonian, shots=None)
result = job.result()
result.metadata
```
```
[{'shots': 1024,
  'variance': (0.9972190856933594+0j),
  'simulator_metadata': [{'noise': 'ideal',
    'batched_shots_optimization': False,
    'measure_sampling': True,
    'parallel_shots': 1,
    'remapped_qubits': False,
    'active_input_qubits': [0],
    'num_clbits': 1,
    'parallel_state_update': 8,
    'sample_measure_time': 0.000191442,
    'num_qubits': 1,
    'device': 'CPU',
    'input_qubit_map': [[0, 0]],
    'method': 'stabilizer',
    'fusion': {'enabled': False}}]}]
```

`shots` should be None. I'm not sure what `variance` here means but if it depends on the number of shots then it is also incorrect.

ikkoham:
### Summary

Resolve #1820


kevinsung:
What does approximation=True or False mean?

kevinsung:
How about automatically setting `approximation=True` instead if `shots=None`. That seems more likely to be aligned with the user's intent.

ikkoham:
I think it's confusing since Aer simulator has the default number of shots. Some users may use this spec.

ikkoham:
@hhorii What do you think?

ikkoham:
I added some docs based on Horii-san's summary:
image

hhorii (approved):
This looks good to me.

PR #1825, opened 2023-05-26
Set the number of qubits with the coupling map

### LINKED ISSUE #1769: Simulator method of FakeBackends changes num qubits in configuration
### Informations

- **Qiskit Aer version**: 0.12
- **Python version**: 3.9
- **Operating system**: MacOs

When using `AerSimulator.from_backend` information such as the coupling map is taken from the backend, however if the simulation method is changed to "stabilizer", the number of qubits is set to a value larger than the coupling map.

```python
backend = AerSimulator.from_backend(FakeMontreal(), method=method)
backend.configuration().num_qubits
```
Returns 10000 for stabilizer, 63 for extended stabilizer, 32 for statevector, 16 for density matrix.

In this case the fake backend has 27 qubits in its coupling map.

The number of qubits in the configuration should be set to the number of qubits in a coupling map if the coupling map is not None.

hitomitak:
### Summary
Resolve #1769

### Details and comments
- Added lines to set the number of qubit by the coupling map


hhorii:
@hitomitak could you add some test cases and release note?

hitomitak:
@hhorii I added the test and release notes.

hhorii (approved):

PR #1831, opened 2023-05-26
Fix REAM #1806: qiskit.providers.ibmq is deprecated; ibmq_athens is not available

davidlearn:
### Summary
- In REAME, use IMProvider instead qiskit.providers.ibmq  to creare provider
- Use NoiseModel.from_backend_properties to create noise model since IBMQ,get)backend is not longer available
- Add noise_model to AerSimulator
- Tested in IBM-Q

### Details and comments
The same as above,

PR #1832, opened 2023-05-27
update readme example code

### LINKED ISSUE #1806: Renew example codes in README
Current example codes in README prints deprecation warnings and does not work.

- `qiskit.providers.ibmq` is deprecated
- `ibmq_athens` is not available

We need to update the code.

andre-a-alves:
### Summary
Updates example code in README

### Details and comments
Closes #1806

This PR updates the README so that it runs and does not print deprecation warnings. `qiskit.providers.ibmq` has been replaced with `qiskit_ibm_provider.IBMProvider`, and `ibmq_athens` has been replaced with `ibmq_manila`.

The example code was successfully in a freshly-installed environment run both in the command line and in a Jupyter Notebook, and the output comments have also been updated to reflect the output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,8 @@ $ python
 
 ```python
 import qiskit
-from qiskit import IBMQ
 from qiskit_aer import AerSimulator
+from qiskit_ibm_provider import IBMProvider
 
 # Generate 3-qubit GHZ state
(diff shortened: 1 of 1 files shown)


hhorii:
Thank you for your contribution. I would like to use #1831 which was posted a bit earlier than this PR.

Issue #1824, opened 2023-05-25
"approximation" argument of Estimator is poorly named


kevinsung:
### What is the expected behavior?

It's somewhat confusing that setting `approximation=True` is necessary to get the _exact_ expectation value. For example, consider this sentence from the [documentation](https://qiskit.org/ecosystem/aer/stubs/qiskit_aer.primitives.Estimator.html):
> If None and approximation is True, it calculates the exact expectation values.

I suggest renaming the argument to `simulate_sampling_error` or just `simulate_error`.

athanzxyt:
Hi @hhorii this PR is ready for review.

ikkoham:
Yes. This requires some `DeprecationWarning`. I'll redesign overall with `options` (with flexible `dataclass`) align with `qiskit-ibm-runtime`.

PR #1835, opened 2023-06-02
rename approximation argument of Estimator to simulate_error

athanzxyt:
### Summary

- Close Issue #1824 
- Change Estimator argument "approximation" to "simulate_error" as suggested.
- Change documentation appropriately such that "approximation" is replaced with "simulate_error"

### DIFF
--- a/qiskit_aer/primitives/estimator.py
+++ b/qiskit_aer/primitives/estimator.py
@@ -41,7 +41,7 @@ class Estimator(BaseEstimator):
 
     :Run Options:
         - **shots** (None or int) --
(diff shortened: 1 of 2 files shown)


hhorii:
Changing argument name is an API change and we need a process of Qiskit. We can not change it with a single PR (at least, we need to show deprecation warning for the obsolete usage in three months).

As discussed in https://github.com/Qiskit/qiskit-aer/pull/1823, @ikkoham is planning to re-design arguments for Estimator in 0.13. Please let me close this PR.

PR #1839, opened 2023-06-06
Fix the grouping index bug in Estimator

### LINKED ISSUE #1837: Aer Estimator returns incorrect values in version 12.0
### Informations

- **Qiskit Aer version**: 12
- **Python version**:
- **Operating system**:

I am attempting to rerun a notebook that worked back in Dec. of 2022: https://quantum-enablement.org/events/skku-2022/Runtime-3.html

For Aer version 12, the VQE problem on the simulator repeatedly converges to an incorrect value hovering in the range of ~-0.57, whereas the true value is ~-0.70.  Downgrading Aer to the previous version, e.g 0.11.2, once again returns values consistent with previous results, eg. -0.698, and expectation of using a noise-less simulator.

Re-run notebook section using Aer Estimator

The Estimator in version 12 should yield results inline with the previous version and the known result.

ikkoham:
### Summary

Fix #1837 

### Details and comments

I misunderstand the order of list of operators from `PauliList.group_commuting`, so the indices are incorrect.

### DIFF
--- a/qiskit_aer/primitives/estimator.py
+++ b/qiskit_aer/primitives/estimator.py
@@ -321,8 +321,7 @@ class Estimator(BaseEstimator):
 
         result_index = 0
         for _circ_ind, basis_map in exp_map.items():
-            for _basis_ind in range(len(basis_map)):
-                param_vals = basis_map[_basis_ind][1]
+            for _basis_ind, (_, (_, param_vals)) in enumerate(basis_map.items()):
                 if circ_ind == _circ_ind and basis_ind == _basis_ind:
                     result_index += param_vals.index(param_val)
                     return result_index
--- a/qiskit_aer/primitives/estimator.py
+++ b/qiskit_aer/primitives/estimator.py
@@ -145,7 +145,11 @@ class Estimator(BaseEstimator):
                 self._observable_ids[_observable_key(observable)] = len(self._observables)
                 self._observables.append(observable)
         job = PrimitiveJob(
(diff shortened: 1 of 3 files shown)


hhorii (approved):

qiskit_aer/primitives/estimator.py
```diff
@@ -317,14 +321,12 @@ def _calculate_result_index(circ_ind, obs_ind, term_ind, param_val, obs_maps, ex
 
         result_index = 0
         for _circ_ind, basis_map in exp_map.items():
-            for _basis_ind, (_, param_vals) in basis_map.items():
+            for _basis_ind, (_, (_, param_vals)) in enumerate(basis_map.items()):
```
ikkoham, line 324: Only this line is changed in this PR.

qiskit_aer/primitives/estimator.py
```diff
@@ -317,14 +321,12 @@ def _calculate_result_index(circ_ind, obs_ind, term_ind, param_val, obs_maps, ex
...
                 if circ_ind == _circ_ind and basis_ind == _basis_ind:
                     result_index += param_vals.index(param_val)
                     return result_index
                 result_index += len(param_vals)
-        raise AerError(
-            "Bug. Please report from isssue: https://github.com/Qiskit/qiskit-aer/issues"
-        )
+        raise AerError("Bug. Please report from issue: https://github.com/Qiskit/qiskit-aer/issues")
```
ikkoham, line 329: tiny typo fix

PR #1845, opened 2023-06-13
Remove `AerCircuit` from result of `backend.run()`

hhorii:
### Summary

Since #1772, results of C++ library include `AerCircuit` as metadata to identify a template circuit. This metadata is internally delivered to users by calling `backend.run().result()`. If a user copies the metadata, errors are happened because `AerCircuit` is not serializable.

### Details and comments

Remove `AerCircuit` objects from any results of `backend.run()`

### DIFF
--- a/releasenotes/notes/remove_aer_circuit_from_metadata-e4fe09029c1a3a3c.yaml
+++ b/releasenotes/notes/remove_aer_circuit_from_metadata-e4fe09029c1a3a3c.yaml
@@ -1,5 +1,5 @@
 ---
 fixes:
   - |
-    Results of `backend.run()` were not serializable because they include `AerCircuit`s.
-    This commit makes the results serializable by removing `AerCircuit`s from metadata.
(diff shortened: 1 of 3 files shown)


jakelishman (dismissed):

releasenotes/notes/remove_aer_circuit_from_metadata-e4fe09029c1a3a3c.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Results of `backend.run()` were not serializable because they include `AerCircuit`s.
+    This commit makes the results serializable by removing `AerCircuit`s from metadata.
```
jakelishman, line 5:
```suggestion
    Results of ``backend.run()`` were not serializable because they include :class:`.AerCircuit`\ s.
    This commit makes the results serializable by removing :class:`.AerCircuit`\ s from metadata.
```

jakelishman (approved):

Issue #1807, opened 2023-05-10
Remove Python 3.7 from Github actions


hhorii:
### What is the expected behavior?

Python 3.7 is no longer supported in qiskit-terra and Aer does not need to support the version. We do not need to run tests and to create packages in the version for 0.13.

tungbq:
Hi @hhorii do you need any help? May I pick up this task?

devilkiller-ag:
@hhorii Can I work on this?

natestemen:
@devilkiller-ag bounties are open until the associated issue is closed, so yes you can still work on this! It does seem like #1819 might resolve the issue, but it hasn't seen much attention. @hhorii or @1ucian0 can you comment on the status of the linked PR?

devilkiller-ag:
Thanks @natestemen

hhorii:
maybe due to configuration of this repository or organization, I cannot assign contributors to this issue. I would like to assign @tungbq. I will merge #1819 in near future (I recently worked for release of 0.12.1).

robotAstray:
was this issue completed succesfully? @hhorii

PR #1819, opened 2023-05-20
Remove Python 3.7 from Github actions

tungbq:
### Summary
Resolves: #1807 

### Details and comments
Remove Python 3.7 from Github actions

PR #1847, opened 2023-06-13
removed python 3.7 from github workflows

devilkiller-ag:
### Summary
This PR is for resolving the issue #1807 which needs python 3.7 from the github actions.

### Details and comments
What actully meant by the statement "We do not need to run tests and to create packages in the version for 0.13." ? Do I also need to add changes to tests?


hhorii:
Thank you for your PR. 
We will take https://github.com/Qiskit/qiskit-aer/pull/1819 and let me close this PR.

PR #1851, opened 2023-06-16
Fix param positions for circuits with conditionals

### LINKED ISSUE #1849: bug in Estimator for parameterized dynamic circuits
### Informations

- **Qiskit Aer version**: 0.12.1
- **Python version**: 3.8
- **Operating system**: mac

The following circuit has two angle parameters `s` and `t`, and it fails in the estimator with:
`ERROR: Failed to load circuits: Invalid parameterization: instruction param position out of range`

```py
from qiskit.circuit import QuantumCircuit, ClassicalRegister, Parameter
from qiskit_aer.primitives import Estimator

estimator = Estimator(run_options={'shots': 1000})
t = Parameter('t')
s = Parameter('s')

# dynamic circuit
circ2 = QuantumCircuit(4)
circ2.add_register(ClassicalRegister(2, 'cr'))
circ2.cx(0, 2)
circ2.cx(1, 2)
circ2.rz(t, 2)
circ2.h(2)
circ2.measure(2, 0)
circ2.z(0).c_if(0, 1)
circ2.z(1).c_if(0, 1)
circ2.barrier()
#circ2.x(2).c_if(0, 1)
circ2.h(0)
circ2.cx(0, 3)
circ2.cx(1, 3)
circ2.rz(s, 3)
circ2.h(3)
circ2.measure(3, 0)
circ2.z(0).c_if(0, 1)
circ2.z(1).c_if(0, 1)
circ2.barrier()
print(circ2.draw(fold=60))

obs2 = SparsePauliOp('IIZY')
values2 = [[1.57, 1.56]]
res2 = estimator.run(circ2, obs2, values2).result()
print(res2)
```

But it works when the parameters are removed and replaced with concrete angles.

This may be related to how the instructions and parameters of the circuit are parsed with Aer. In terra this should be trivial to bind a circuit with a list of parameters so I'm not sure why it fails in Aer / if it's trying to do something custom?

hhorii:
### Summary

Resolve #1849

### Details and comments

When conditional gates exist, parameter positions are changed because `bfunc` operations are injected.
It is the root cause of #1849 that reports wrong parameter bindings with circuits that have conditional gates.

With this commit, positions of Qiskit instructions and  Aer operations are mapped and Aer's parameter bindings are based on this mapping.

- [ ] Add tests

### DIFF
--- a/qiskit_aer/backends/aer_compiler.py
+++ b/qiskit_aer/backends/aer_compiler.py
@@ -595,6 +595,8 @@ def assemble_circuit(circuit: QuantumCircuit):
     aer_circ.num_memory = num_memory
     aer_circ.global_phase_angle = global_phase
 
+    num_of_aer_ops = 0
+    index_map = []
     for inst in circuit.data:
         # To convert to a qobj-style conditional, insert a bfunc prior
         # to the conditional instruction to map the creg ?= val condition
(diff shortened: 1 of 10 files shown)


ajavadia:
I can confirm this solves #1849 so I think after adding a test it can be merged

test/terra/backends/test_parameterized_qobj.py
```diff
@@ -441,22 +461,28 @@ def test_check_parameter_binds_exist(self):
 
     def test_global_phase_parameters(self):
         """Test parameterized global phase"""
-        backend = AerSimulator()
+        backend = AerSimulator(method="extended_stabilizer")
```
doichanj, line 464: Is it enough only testing `extended_stabilizer`?

doichanj (approved):

PR #1869, opened 2023-07-11
Add test case to serialize result of simulation

hhorii:
### Summary

Add a test case to check serialization of simulation result

### Details and comments

This PR adds a test case to check serialization of simulation result work well.


jakelishman (commented):
This looks like a good idea, thanks.  Could we add a test that the result object can roundtrip through `pickle` as well, while we're at it?  I imagine that anything that can roundtrip JSON can roundtrip pickle, but it'd be good to be explicit.

hhorii:
@jakelishman sorry that my response is lazy. I added pickle test case.

doichanj (approved):

PR #1871, opened 2023-07-11
set numpy version constraint only in dev

hhorii:
### Summary

Set lower version of numpy than 1.25 only in development

### Details and comments

Qiskit-Aer itself does not have constraint about numpy versions currently.


requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
+numpy<1.25
 stestr>=2.5.0
 cmake!=3.17.1,!=3.17.0
 conan<2.0.0
```
jakelishman, line 4: Since this is just a CI constraint, it'd be more normal to put `numpy<1.25` or `numpy==1.24.4` into `constraints.txt` rather than `requirements-dev.txt`.

hhorii: I encountered following error.
```
ollecting wheel
  Using cached wheel-0.40.0-py3-none-any.whl (64 kB)
ERROR: Cannot install numpy==1.24.4 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested numpy==1.24.4
    The user requested (constraint) numpy>=1.16.3

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

Notice:  A new release of pip is available: 23.0.1 -> 23.1.2
Notice:  To update, run: pip install --upgrade pip
```
https://github.com/Qiskit/qiskit-aer/actions/runs/5526181345/jobs/10080712985?pr=1871#step:5:17

This is because `setup.py` sets ">1.16".

doichanj (approved):

PR #1875, opened 2023-07-18
Support BackendV2

### LINKED ISSUE #1681: Update Aer backends to BackendV2

The BackendV2 interface is the next version of the abstract Backend interface used for describing backends in Qiskit. It was introduced in qiskit-terra 0.19.0 and it's now the primary version used for representing backends in provider packages. We
should update Aer's backends `AerSimulator`, `StatevectorSimulator`, etc to be based on `BackendV2` instead of `BackendV1`.

hitomitak:
### Summary
Resolve #1681 

### Details and comments
Add method in aerbackend class to create Target and return name.


qiskit_aer/primitives/estimator.py
```diff
@@ -301,17 +301,23 @@ def _create_meas_circuit(self, basis: Pauli, circuit_index: int):
...
     @staticmethod
     def _combine_circs(circuit: QuantumCircuit, meas_circuits: list[QuantumCircuit]):
         circs = []
         for meas_circuit in meas_circuits:
-            new_circ = circuit.copy()
-            for creg in meas_circuit.cregs:
-                new_circ.add_register(creg)
-            new_circ.compose(meas_circuit, inplace=True)
-            _update_metadata(new_circ, meas_circuit.metadata)
+            if circuit.num_qubits >= meas_circuit.num_qubits:
```
ikkoham, line 312: This should not be `False`. If you get it, it is lack of validation.

releasenotes/notes/support_backend_v2-9eeb4690199b01d1.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Update Aer Backend to BackendV2. Refer to `#1681<https://github.com/Qiskit/qiskit-aer/issues/1681>`.
```
mtreinish, line 4: This needs a bit more detail to document the user upgrade implications. The only one that I can think of is `backend.name` is now a string attribute not a method.

test/terra/backends/aer_simulator/test_algorithms.py
```diff
@@ -123,7 +123,8 @@ def test_extended_stabilizer_sparse_output_probs(self):
         for i in range(1, nqubits):
             circ.cx(i - 1, i)
         circ.measure_all()
-        circ = transpile(circ, backend)
+
+        # circ = transpile(circ, backend)
```
mtreinish, line 127: Why is this commented out?

test/terra/backends/aer_simulator/test_options.py
```diff
@@ -46,7 +46,7 @@ def test_seed_simulator_option_measure(self, method, device):
         qc.h([0, 1])
         qc.reset(0)
         qc.measure_all()
-        qc = transpile(qc, backend)
+        qc = transpile(qc, backend, optimization_level=0)
```
mtreinish, line 49: Why do we need to use level 0 here?

mtreinish (approved):
Overall this LGTM. I left a couple of inline questions and a comment about the release notes that need a little expansion, but not worth blocking on any of these. Sorry it took me so long to look at this. Using the v1 -> v2 converter is a good path to supporting the v2 interface in the short term and makes this migration fairly simple and straightforward.

Longer term we'll need to start moving away from using configuration, properties, and defaults as they'll be deprecated in qiskit in the near future (probably not until April next year).

The neko test failures do point to some compatibility issues with using v2 here (for example it looks like `QuantumInstance` assume aer is always backendv1). How we handle those is up to @doichanj as documenting them as a breaking change is a viable path forward, but then we need to make sure there is a release note explaining this in greater depth.

mtreinish:
Actually on reflection, I'm a bit concerned about the neko failures, because as soon as aer 0.13.0 releases that will cause neko to fail on all other projects unless they pin to `qiskit-aer < 0.13.0`. So we might need to get to the bottom of it before releasing

test/terra/backends/aer_simulator/test_options.py
doichanj: with transpile>0, H gate is converted to P gate in extended_stabilizer method that takes very long time to run this circuit. In backendV1, H gate was not converted

doichanj:
I created fix to solve backend.name() error https://github.com/Qiskit/qiskit/pull/11065

releasenotes/notes/support_backend_v2-9eeb4690199b01d1.yaml
hitomitak: Add the detailed description to the release note.

test/terra/backends/aer_simulator/test_algorithms.py
hitomitak: Because the simulation time becomes so long that the timeout occurs. This transpile method adds non-clifford gate to the circuit so extended stabilizer simulation requires long simulation time. If the time until timeout is enough long and the simulation is completed, the test is passed without this comment out.

test/terra/backends/aer_simulator/test_options.py
mtreinish: I remember this now, the issue tracking this in qiskit is: https://github.com/Qiskit/qiskit/issues/10568

doichanj (approved):
Thank you, we are ready to merge it

PR #1876, opened 2023-07-18
Support u3 gate application in Aer runtime API

ibm-wakizaka:
### Summary

Add a support for u3 gate to the runtime API of aer.

### Details and comments

Currently the API does not support the u3 gate even though an aer state supports it.
I add a support to apply u3 gate by just calling `AerState::apply_u`.


hhorii (commented):
Could you add a release note, like "`aer_apply_u3` is added to `aer_runtime_api.h`"?

ibm-wakizaka:
Added a release note.

I also applied clang-format to `aer_runtime.cpp` to pass the test.

Issue #1873, opened 2023-07-12
Parallelisation over experiments stops working after submitting unparallelised experiment


LSticher:
### Informations

- **Qiskit Aer version**: 0.12.1
- **Python version**: 3.8.17
- **Operating system**: Ubuntu 22.04.2
- **CPU**: AMD Ryzen 5600x 6C/12T

### What is the current behavior?

I am trying to parallelise the execution of simultaneously submitted circuits, that is focusing on parallelisation set by the ```max_parallel_experiments``` option. I've run into two issues. The first issue is that the execution is not parallelised when a low, but still parallelisable, number of experiments is submitted. As example I've once submitted 6 circuits simultaneously and a second time I've submitted 12 circuits simultaneously. The 6 circuit experiment is not parallelised at all and is even slower to execute than the 12 circuit experiment.  

```python
import time
import qiskit
import qiskit_aer as qk_aer

serial_sim = qk_aer.AerSimulator(max_parallel_threads=1, max_parallel_experiments=1, max_parallel_shots=1)
parallel_sim = qk_aer.AerSimulator(max_parallel_threads=0, max_parallel_experiments=0, max_parallel_shots=0)

def get_circ():
    circ = qiskit.transpile(qiskit.circuit.library.QuantumVolume(num_qubits=20, depth=5), basis_gates=['u','cx'])
    circ.measure_all()
    return circ

circs6 = [get_circ() for i in range(6)]
circs12 = [get_circ() for i in range(12)]

t1 = time.time()
t = serial_sim.run(circs12).result()
t2 = time.time()
print(t2-t1)

t1 = time.time()
t = serial_sim.run(circs6).result()
t2 = time.time()
print(t2-t1)

t1 = time.time()
t = parallel_sim.run(circs12).result()
t2 = time.time()
print(t2-t1)

t1 = time.time()
t = parallel_sim.run(circs6).result()
t2 = time.time()
print(t2-t1)
```

On my machine this prints:

```python
2.516629219055176
1.2833096981048584
0.5285968780517578
1.3140149116516113
```
This should exemplify the non-parallelisation of the 6 circuit execution. While a bit weird, this is not my main problem. I assume there is some internal logic to decide when parallelisation is worth it/should actually be applied and the 6 circuit execution is (somewhat erroneously in this case) considered to be not worth parallelising.

However, my main problem is best exemplified by running the code below:

```python
import time
import qiskit
import qiskit_aer as qk_aer

[...]

PR #1880, opened 2023-07-26
Fix OpenMP nested parallel

doichanj:
### Summary
This is fix for issue #1873 

### Details and comments
OpenMP parallel nested was not correctly set when number of input circuits is less than number of threads.
Also parallel state update was not correctly set if number of input circuits is more than 1.

For single circuit with multiple-shots run, when nested parallel is used `omp_nested=True` was not set in the metadata of result for a circuit, so this is added.


hhorii (approved):
LGTM.

Issue #1878, opened 2023-07-24
Segfault when simulating circuits with specific number of qubits


g33kex:
### Informations

- **Qiskit Aer version**: 0.12.2
- **Python version**:  3.11.3
- **Operating system**: Archlinux

### What is the current behavior?

The following code leads to a segfault when running the simulator:
```python
from qiskit import ClassicalRegister, QuantumRegister, QuantumCircuit
from qiskit import transpile
from qiskit_aer.backends import AerSimulator

n=39
#n=52 # Doesn't segfault with n=52
A = QuantumRegister(n)
B = QuantumRegister(n)
m = ClassicalRegister(len(A))
qc = QuantumCircuit(A, B, m)
qc.ccx(B[0], A[0], A[1])
#qc.ccx(A[2], A[0], A[1]) # Using this gate will not segfault if n=39 but will segfault if n=81
qc.measure(A, m)
backend_sim = AerSimulator()
transpiled = transpile(qc, backend_sim)
print("Circuit transpiled.")
job_sim = backend_sim.run(transpiled, shots=1) # This triggers a segmentation fault
print("No segfault!")
```

When n=39 the circuit has 78 qubits, 39 classical bits and a single Toffoli gate. The simulation step segfaults. 
However if n=52, the circuit runs without segfault. This also happens with different circuits. The process doesn't run out of memory.
Also for n=39, doing the Toffoli gate between qubits from only register A doesn't result in segfault, but using qubits from both registers results in a segfault. But if we now set n=81, it segfaults in both cases. But if we set n=96, it doesn't segfault anymore.

### Steps to reproduce the problem

Run code above. It will print "Circuit transpiled." and then segfault. I tried it on ArchLinux and Windows.

### What is the expected behavior?

The program should not segfault.

### Suggested solutions

I have no idea where this bug comes from, it is really weird.

Here is the core dump from the system logs:
```
Jul 24 10:58:46 computer systemd-coredump[79394]: [🡕] Process 79345 (python3) of user 1000 dumped core.
                                                 
                                                 Module /test/.venv/lib/python3.11/site-packages/scipy/spatial/_qhull.cpython-311-x86_64-linux-gnu.so without build-id.
                                                 Module /test/.venv/lib/python3.11/site-packages/scipy/spatial/_qhull.cpython-311-x86_64-linux-gnu.so
[...]

PR #1881, opened 2023-07-26
Fix required_memory_mb

doichanj:
### Summary
This is fix for issue #1878 

### Details and comments

requried_memory_mb function for statevector returns wrong value when number of qubits is very large because of overflow of 64 bits integer. 
Now it returns SIZE_MAX value when number of qubits is too large so that Qiskit Aer can know memory is not sufficient for statvector/unitary/density matrix methods.


releasenotes/notes/fix_required_memory_mb-7aeafa0fe553b85a.yaml
```diff
@@ -0,0 +1,65 @@
+---
```
hhorii, line 1: need to modify this release note.

Issue #1888, opened 2023-08-06
Aer stabilizer simulator measurement outcomes depend on whether other measurements are done concurrently for large circuits


sam-craig-smith:
### Informations

- **Qiskit Aer version**: 0.12.2
- **Python version**: 3.9.6
- **Operating system**:
    - Name of the operating system: posix
    - Name of the OS system: Darwin
    - Version of the operating system: 22.5.0

### What is the current behavior?

I have been using large Clifford circuits (65 qubits total, 33 data qubit register, 32 ancilla qubit register, and one 32 bit classical register). The circuits consist of hadamards, cnots, resets, and measurement. I have found some strange behaviour in the measurement statistics that I see when run with the Aer stabilizer simulator.

In particular, the statistics from a measurement of one ancilla qubit depend on whether other measurements are performed concurrently on other ancilla qubits. This should not be the case. I see some ancilla qubit measurements deterministally give a 0 outcome when other measurements are performed concurrently, and if the other measurements are removed from the circuit then a deterministic 1 outcome is seen. (The 0 outcomes is the one that I expect). 

### Steps to reproduce the problem

I have an example of a circuit that demonstrates this behaviour. The circuit is a large example, I when I look at smaller circuits this behaviour disappears, so this is the smallest example I have. 

Here is a circuit when all ancilla are measured 

[circ_measure_all_ancilla.pdf](https://github.com/Qiskit/qiskit-aer/files/12270322/circ_measure_all_ancilla.pdf)

And here is a plot of the outcomes seen for each ancilla measurement over 10000 counts: 

[counts_measure_all_ancilla.pdf](https://github.com/Qiskit/qiskit-aer/files/12270324/counts_measure_all_ancilla.pdf)

Now, here is the same circuit but with some measurements removed: 

[circ_do_not_measure_all_ancilla.pdf](https://github.com/Qiskit/qiskit-aer/files/12270325/circ_do_not_measure_all_ancilla.pdf)

And here is the plot of the outcomes for the second circuit, again over 10000 counts 

[counts_do_not_measure_all_ancilla.pdf](https://github.com/Qiskit/qiskit-aer/files/12270328/counts_do_not_measure_all_ancilla.pdf)

The code to produce this output is: 

```
from qiskit import qasm3, Aer, transpile, ClassicalRegister 
from qiskit import ClassicalRegister
import copy 
from matplotlib import pyplot as plt 
import numpy as np 

[...]

PR #1895, opened 2023-08-16
Fix measure in stabilizer method

doichanj:
### Summary
This is fix for the issue of measurement in stabilizer simulator reported in issue #1888

### Details and comments
There was wrong bit masking code in deterministic measurement, this fix corrects bit mask


hhorii:
We need test cases that validate this change is correct and the issue will not be reproduced in the future releases.

hhorii (approved):
LGTM

PR #1902, opened 2023-08-22
Add Eclipse IDE project files to .gitignore.

sfantao:
### Summary
Add entries to `.gitignore` related with Eclipse IDE files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,8 @@ docs/api/
 
 # Ignore DASK temporary files
 dask-worker-space/*
+
+#Eclipse synchronized project
+.ptp-sync-folder
+.ptp-sync/
+.pydevproject


mtreinish (commented):
Thanks for opening a PR. In this case I'm not sure we want to move forward with this. There are a myriad of editors out there and they all have local state stored in different ways. It's not sustainable to continually add to the per repo gitignore file for every editor that someone is using. 

Instead git supports using global gitignore for your local environment. For example see: 
https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer 

You can create a `.gitignore` file in your home directory with the contents you're adding here and then configure git to use that file for all repositories. This is what I do because my editors cache files aren't in qiskit-aer's .gitignore.

sfantao:
@mtreinish, thank you, that's fair enough. I'll skip this PR then.

Issue #1908, opened 2023-08-23
Random segfault on existing tests due to data alignment.


sfantao:
### Informations

- **Qiskit Aer version**: tip of the trunk
- **Python version**: 3.9
- **Operating system**: Linux x86_64

### What is the current behavior?
Segfault happens randomly during some of the tests. Tested on `test.terra.states.test_aer_state.TestAerState.test_appply_diagonal`. The segfault comes from a load intrinsic in `qv_avx2.cpp`:
```
    for (int64_t k = 0; k < END; k += 1) {
      const auto base = k << (batch + 1);
      const auto until = base + (1UL << (batch + 1));
      for (auto i = base; i < until; i += 2) {
        auto tgt_qv_data =
            _mm256_load(reinterpret_cast<double *>(&(qv_data[i])));
        auto input_data = _load_diagonal_input(input_vec, double_tmp, i, qregs,
                                               qregs_size, q0_mask);
        _mm_complex_multiply<double>(tgt_qv_data, input_data);
        _mm256_store(reinterpret_cast<double *>(&(qv_data[i])), tgt_qv_data);
      }
    }
```
The issue, I think, is that in the faulty instance `&(qv_data[0])` is not 32-byte aligned. Depending on the alignment the tests succeeds or fails. I see that in:
```
template <typename data_t>
void QubitVector<data_t>::allocate_mem(size_t data_size) {
  // Free any currently assigned memory
  free_mem();
  // Allocate memory for new vector
  if (data_ == nullptr) {
#if !defined(_WIN64) && !defined(_WIN32)
    void *data = nullptr;
    posix_memalign(&data, 64, sizeof(std::complex<data_t>) * data_size);
    data_ = reinterpret_cast<std::complex<data_t> *>(data);
#else
    data_ = reinterpret_cast<std::complex<data_t> *>(
        malloc(sizeof(std::complex<data_t>) * data_size));
#endif
  }
}
```
there is an effort to force the main allocator do be 64-byte aligned being std::complex only 8-byte aligned, but in this instance the `data_` array might be initialized from other buffer and not allocated. I don't have a good understanding of the tests to track the origin well. This is the faulty instruction and backtrace:
```
(gdb) frame                                                                                                                                                                                         
[...]

PR #1909, opened 2023-08-24
Make vector allocation aligned to 64-bytes

sfantao:
### Summary
Make vector allocation aligned to 64-bytes as that is the same alignment requirement used by QubitVector data.

### Details and comments
`AER::Vector` is in some cases used to initialize `QubitVector` data and the code assumes that that storage is at least 32-byte aligned so that mm256 instructions can be used. This PR makes `AER::Vector` storage have the same alignment requirements (64-byte) as `QubitVector` so that one doesn't get into segfaults when trying to access this storage with vector intrinsics.

This fixes https://github.com/Qiskit/qiskit-aer/issues/1908.


doichanj (approved):

doichanj (changes requested):
Please add release note for this fix

sfantao:
Added release note for the fix.

sfantao:
@doichanj , can you help understand how this PR is breaking the Windows builds? It shouldn't change anything there. I only see the compiler running out of memory.

doichanj:
Windows build sometimes fails because of heap memory shortage. Only we can do is to rerun until it will not fail

doichanj (approved):

PR #1910, opened 2023-08-24
Define environment variable to allow Qiskit-Aer to be built without CUDA requirements

sfantao:
### Summary
Define environment variable to allow Qiskit-Aer to be built without CUDA requirements

### Details and comments
It is convenient to sometimes be able to leverage an existing system CUDA installation as opposed to load redundant python CUDA packages that provide the same functionality. This enables having smaller footprint in the Qiskit-Aer installation and facilitates development and testing with CUDA. This PR enables an environment variable to control this at build time.


doichanj (changes requested):
Please add release note for this change

sfantao:
Added release note for the new feature.

doichanj (approved):

PR #1913, opened 2023-08-24
Add support to compile with Intel MKL + Use it in MPS simulation

Patataman:
### Summary
Hello, this PR add the capability of compiling against Intel MKL, and using its functions in the MPS simulation method.

It is build over the LAPACK SVD (https://github.com/Qiskit/qiskit-aer/pull/1897).

I understand that you may not like it as this will imply contemplate both, OpenBLAS and Intel MKL when developing. So, I am prepared for a full rejection of the PR.

### Details and comments
1. Intel MKL overhead is a bit bigger than OpenBLAS, but performance in general is better.
2. To enable it, must pass the -DMKL flag in compiling time. E.g: ``python ./setup.py bdist_wheel -DMKL=True``
3. While developing I got some segmentation faults in the SVD function. They happened because I was not using malloc/free at the time, so it should not happen again. However, I might missed something.
4. All the tests pass, but I have only benchmarked against MPS simulation.

And the desired results:

Server specs

| | |
| ------------- | ------------- |
| CPU | Intel Xeon Gold 6148 |
| # sockets  | 2  |
| # cores | 20 |
| RAM | 192GB  |
| GPU | None |
| OS | Ubuntu 22.04.1 LTS |
| Python | 3.10 |
| OpenBLAS/LAPACK | 0.3.21 |
| gcc | v11.3 |

image

And, the average time (in seconds) from 5 different executions:

 Depth | LAPACK QR | MKL QR | LAPACK D&C | MKL D&C |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| 1 | 0.04652729 | 0.035822439 | 0.049358368 | 0.053574371 |
| 3 | 0.109769773 | 0.05801568 | 0.126575661 | 0.060559988 |
| 5 | 0.148080826 | 0.108965158 | 0.16601491 | 0.101443863 |
| 10 | 28.21881118 | 21.28681765 | 21.8562469 | 18.65257401 |
| 12 | 429.5337416 | 219.2043834 | 208.0082648 | 168.9656805 |

If I have time, I will add also results for depth 15.


merav-aharoni:
Best to review this after #1897 is merged, because the changes here include those.
@doichanj - the main question here is for you, whether you want to include the MKL package in Aer.

doichanj:
I think MKL should be optional that will not be included in distribution package of Aer. In this PR, MKL is enabled by setup option, so users who want to use MKL can build from the source

PR #1915, opened 2023-08-24
Support `Expr` that Terra 0.24 newly introduced

hhorii:
### Summary

Add support of simulation of classical expressions in Terra 0.24

### Details and comments

This PR adds experimental feature to support classical expression introduced in Terra 0.24.

The supported expression is for test expression in `switch`, `while` and `if else`. `Expr` class and its sub-classes are converted newly defined C++ classes in `operations.hpp`, which are bound to python. `AerCompiler` supports conversion from branch conditions in `switch` `while`, and `if else` into `Expr` and then assembles the circuit to `AerCircuit` with newly introduced C++ classes for `Expr`.


jakelishman (commented):
I haven't had and likely won't have time to review in full depth - I think we're probably going to need to move to a model where that's more the responsibility within the Aer team, but I've just run through quickly and made a couple of notes on the high-level implementation.

In C++ land, I was vaguely thinking that a more efficient simulator/interpreter would want to convert the tree representation of Qiskit's `Expr` into a linear executable bytecode rather than retaining the tree structure into C++ land (better cache locality of the data, no recursion needed during execution, potential for JIT optimisations while writing out the bytecode, etc).  I'm not suggesting you change to that in this PR if you don't think it's necessary, I'm just mentioning it as the path I'd thought made a little more sense for a simulator.

src/framework/operations.hpp
```diff
@@ -287,7 +610,8 @@ struct Op {
   bool conditional = false; // is gate conditional gate
   uint_t conditional_reg; // (opt) the (single) register location to look up for
                           // conditional
-  RegComparison bfunc;    // (opt) boolean function relation
+  BinaryOp binary_op;     // (opt) boolean function relation
```
doichanj, line 613: `Op.bfunc` is replaced with `Op.binary_op` but `OpTyp::bfunc` is still remaining as is. Is it OK?

qiskit_aer/backends/aer_compiler.py
```diff
@@ -366,11 +410,15 @@ def _inline_switch_case_op(self, instruction, continue_label, break_label, paren
...
-        ) - set(instruction.clbits)
+
+        target_clbits = []
+        if isinstance(instruction.operation.target, Clbit):
+            target_clbits = [bit_map[instruction.operation.target]]
+        elif isinstance(instruction.operation.target, Expr):
+            self._list_clbit_from_expr(bit_map, target_clbits, instruction.operation.target)
+        else:
+            target_clbits = [bit_map[c] for c in instruction.operation.target]
+        mark_cargs = set(cargs + target_clbits) - set(instruction.clbits)
```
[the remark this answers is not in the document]
jakelishman: Here is the promised design document: Qiskit/rfcs#50

qiskit_aer/backends/aer_compiler.py
```diff
@@ -199,12 +215,43 @@ def _inline_circuit(self, circ, continue_label, break_label, bit_map=None):
...
+                _convert_clbit_index(expr.left)
+                _convert_clbit_index(expr.right)
+            elif isinstance(expr, Var):
+                if isinstance(expr.var, Clbit):
+                    expr.var = bit_map[expr.var]
+                else:
+                    expr.var = [bit_map[clbit] for clbit in expr.var]
+            return expr
+
+        return (_convert_clbit_index(deepcopy(cond_tuple[0])), cond_tuple[1])
```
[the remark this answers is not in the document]
hhorii: I created `_convert_clbit_index()` to convert cbit indices with `bit_map` in the given `Expr`. Do you mean that this conversion is not necessary?

qiskit_aer/backends/aer_compiler.py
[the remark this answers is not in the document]
hhorii: I believe that backward compatibility needs to be maintained regarding `Var` instantiation that takes `Clbit` or `ClassicalRegister`. So, in the new design, there are three types of `Var`:
- based on `Clbit`
- based on `ClassicalRegister`
- based on specific type

In simulation, `Var` needs to be assigned a specific memory area. I will use the existing memory management in Aer and implicitly create `ClassicalRegister` for the third type of `Var`s before simulation.

qiskit_aer/backends/aer_compiler.py
```diff
@@ -578,7 +637,11 @@ def assemble_circuit(circuit: QuantumCircuit):
     for creg in circuit.cregs:
         creg_sizes.append([creg.name, creg.size])
 
-    is_conditional = any(getattr(inst.operation, "condition", None) for inst in circuit.data)
+    is_conditional = any(
+        getattr(inst.operation, "condition_expr", None)
+        or getattr(inst.operation, "condition", None)
+        for inst in circuit.data
+    )
```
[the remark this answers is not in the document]
hhorii: `AerCompiler` will add `conditiona_expr` to instruction that Aer newly inserted. Therefore, deep copy will not be happened.

qiskit_aer/backends/aer_compiler.py
```diff
@@ -621,16 +685,114 @@ def assemble_circuit(circuit: QuantumCircuit):
...
+            _assemble_unary_operator(expr.op), _assemble_expr(circ, expr.operand)
+        )
+    elif isinstance(expr, Binary):
+        aer_expr = AerBinaryExpr(
+            _assemble_binary_operator(expr.op),
+            _assemble_expr(circ, expr.left),
+            _assemble_expr(circ, expr.right),
+        )
+    else:
+        raise AerError(f"unknown expression: {expr.__class__}")
```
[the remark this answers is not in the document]
hhorii: Yes. `ExprVisitor` looks much better than the current my impl.

src/framework/operations.hpp
hhorii: Thanks. It should be `binary_op`.

qiskit_aer/backends/aer_compiler.py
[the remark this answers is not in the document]
jakelishman: It's more that I can't immediately see how control could pass to the definition of that function - I think the `if/elif/elif` above it is exhaustive.

src/framework/operations.hpp
hhorii: Hmmm, we still use `Operations::OpType::bfunc` and keeping `apply_bfunc` is reasonable.

qiskit_aer/backends/aer_compiler.py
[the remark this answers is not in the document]
hhorii: @jakelishman I updated this PR to use `ExprVisitor`. I think this class does not have `visit()` so I created it. Maybe we can use `accept()` instead of creating this new method.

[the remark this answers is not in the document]
jakelishman: Yeah `ExprVisitor` doesn't have a `visit` method because you call it as `expr.accept(visitor)`; the `Expr` subclasses handle choosing which method to call, rather than the `ExprVisitor` doing magic inference.

doichanj (approved):

doichanj (dismissed):

qiskit_aer/backends/aer_compiler.py
```diff
@@ -199,12 +215,78 @@ def _inline_circuit(self, circ, continue_label, break_label, bit_map=None):
...
+            node.operand.accept(self)
+            return node
+
+        def visit_binary(self, node, /):
+            node.left.accept(self)
+            node.right.accept(self)
+            return node
+
+        def visit_generic(self, node, /):
+            raise AerError(f"unsupported expression is used: {node.__class__}")
```
jakelishman, line 249: It feels like this visitor should construct the corresponding `AerExpr` tree, rather than deepcopying the Terra one and mutating it?  In particular, mutating `Var.var` is super unsafe, especially because it's being changed to a value of an incorrect type here.  (I also want to be a bit sneaky in Terra 1.0 and make `expr.Var` and `expr.Value` immutable; they logically always should have been.)

Can you re-use `_AssembleExprImpl` here?  That does the clbit remapping already, I think?

qiskit_aer/backends/aer_compiler.py
```diff
@@ -199,12 +215,78 @@ def _inline_circuit(self, circ, continue_label, break_label, bit_map=None):
...
+            elif isinstance(expr, Var):
+                if isinstance(expr.var, Clbit):
+                    expr.var = bit_map[expr.var]
+                else:
+                    expr.var = [bit_map[clbit] for clbit in expr.var]
+            return expr
+
+        left = deepcopy(cond_tuple[0]).accept(AerCompiler._ClbitConverter(bit_map))
+        right = cond_tuple[1]
+        return (left, right)
```
jakelishman, line 277: I have a previous review comment about this up top, but in what situations do we reach line 260?  As far as I know, a condition can _only_ be an `Expr` or a `tuple[Clbit | ClassicalRegister, int]`, so the if statements on lines 253 to 239 should be exhaustive, right?  In what situations might we get through them?

qiskit_aer/backends/aer_compiler.py
```diff
@@ -199,12 +215,78 @@ def _inline_circuit(self, circ, continue_label, break_label, bit_map=None):
...
+        if isinstance(expr, Unary):
+            self._list_clbit_from_expr(bit_map, clbits, expr.operand)
+        elif isinstance(expr, Binary):
+            self._list_clbit_from_expr(bit_map, clbits, expr.left)
+            self._list_clbit_from_expr(bit_map, clbits, expr.right)
+        elif isinstance(expr, Var):
+            if isinstance(expr.var, Clbit):
+                clbits += bit_map[expr.var]
+            else:
+                clbits += [bit_map[clbit] for clbit in expr.var]
```
jakelishman, line 289: There's a helper in Terra for iterating only over the `Var` instances: `qiskit.circuit.classical.expr.iter_vars`.  You could implement this like
```python
from qiskit.circuit.classical import expr

def _clbits_from_expr(self, bit_map, node):
    out = set()
    for var in expr.iter_vars(node):
        if isinstance(var.var, Clbit):
            out.add(bit_map[var.var])
        elif isinstance(var.var, ClassicalRegister):
            out.update(bit_map[bit] for bit in var.var)
        else:
            # not handled
    return out
```
and you don't need to manually handle all the possible entries in the tree (for example, your current implementation forgets `Cast`), and Terra will automatically handle any new `Expr` subclasses that get added for you.

qiskit_aer/backends/aer_compiler.py
```diff
@@ -672,19 +877,22 @@ def _assemble_op(aer_circ, inst, qubit_indices, clbit_indices, is_conditional, c
...
     elif name == "initialize":
         _check_no_conditional(name, conditional_reg)
         aer_circ.initialize(qubits, params)
     elif name == "roerror":
         _check_no_conditional(name, conditional_reg)
         aer_circ.roerror(qubits, params)
     elif name == "multiplexer":
-        aer_circ.multiplexer(qubits, params, conditional_reg, label if label else name)
+        aer_circ.multiplexer(qubits, params, conditional_reg, aer_cond_expr,
+                             label if label else name)
```
jakelishman, line 893: I'm surprised that this passes the lint CI, because this line splitting isn't compatible with `black`, which I thought Aer was using.

releasenotes/notes/support_classical_expr-dd621e5c0fd23a15.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    Support ``qiskit.circuit.classical.expr.Expr`` introduced in Qiskit Terra 0.24.
+    If a circuti includes unsupported expression which will be introduced in Qiskit 0.25
+    or later, ``AerError`` is thrown. Objects of sub-classes of ``Expr`` are converted to
+    corresponding sub-classes of ``AER::Operations::CExpr`` in C++ and evaluated. 
+    In addition, ``AER::Operations::RegComparison`` in ``bfunc`` operation is replaced
+    with corresponding binary operations.
```
jakelishman, line 9:
```suggestion
    Support ``qiskit.circuit.classical.expr.Expr`` introduced in Qiskit 0.44.
    If a circuit includes unsupported expression which will be introduced in Qiskit 0.45
    or later, ``AerError`` is thrown. Objects of sub-classes of ``Expr`` are converted to
    corresponding sub-classes of ``AER::Operations::CExpr`` in C++ and evaluated. 
    In addition, ``AER::Operations::RegComparison`` in ``bfunc`` operation is replaced
    with corresponding binary operations.
```

Terra refers to its using "Qiskit" version numbers now, so I guess we should here as well.

jakelishman (commented):
I've looked over the Python-space components again, and they look generally fine to me.  I can't do the C++ components, sorry - I don't have the time to be getting that deep into Aer any more, but given that the tests pass, I'm guessing that they work ok.

I left a couple of small suggestions for tidying things up using some additional Terra helper methods that already exist / potentially re-using code, but it's not critical; it could be fixed later if necessary.

qiskit_aer/backends/aer_compiler.py
hhorii: Thanks. `Var` is possible to come here. I rewrite this part.

qiskit_aer/backends/aer_compiler.py
hhorii: Thanks. I changed the code as you suggested.

qiskit_aer/backends/aer_compiler.py
hhorii: This class is no longer used in the latest code. Now, Only `Var` instance is copied.

doichanj (approved):
I think Expr implementation looks fine and I confirm there is no side effect by this change

Issue #1925, opened 2023-09-04
Aer runtime api (from contrib) exposes the wrong sx operation


aromanro:
### Informations

- **Qiskit Aer version**:

0.13.0 - from VERSION.txt

- **Python version**:

3.11.4

- **Operating system**:

Windows 11

### What is the current behavior?

Currently it's implemented with an rx operation like this:

```
// sqrt(NOT) gate
void aer_apply_sx(void *handler, uint_t qubit) {
  AER::AerState *state = reinterpret_cast<AER::AerState *>(handler);
  state->apply_mcrx({qubit}, -M_PI / 4.0);
};
```

This introduces a different global phase than the sx gate exposed to python (or described in qiskit documentation), for example. It's not a big issue, but I think it should have the same definition.

### Steps to reproduce the problem

Compare results obtained with the normal qiskit sx gate and those obtained with using the aer runtime 'aer_apply_sx'.

### What is the expected behavior?

Identical results.

### Suggested solutions

Replace implementation with:

```
void aer_apply_sx(void *handler, uint_t qubit) {
  AER::AerState *state = reinterpret_cast<AER::AerState *>(handler);
  state->apply_unitary({ qubit }, AER::Linalg::Matrix::SX);
};
```

I'll make a pull request for this soon.

PR #1926, opened 2023-09-04
Fix for https://github.com/Qiskit/qiskit-aer/issues/1925 (Aer runtime…

aromanro:
… api (from contrib) exposes the wrong sx operation)

### Summary

Replaced the runtime api sx implementation to the one expected from documentation (and functionality in other places in qiskit aer).

### Details and comments

Issue: https://github.com/Qiskit/qiskit-aer/issues/1925

The sx gate was implemented in the api using a rotation, which added a global phase compared with the regular implementation in qiskit aer. Now it's replaced with the unitary AER::Linalg::Matrix::SX.

### DIFF
--- a/contrib/runtime/aer_runtime.cpp
+++ b/contrib/runtime/aer_runtime.cpp
@@ -139,7 +139,7 @@ void aer_apply_tdg(void *handler, uint_t qubit) {
 // sqrt(NOT) gate
 void aer_apply_sx(void *handler, uint_t qubit) {
   AER::AerState *state = reinterpret_cast<AER::AerState *>(handler);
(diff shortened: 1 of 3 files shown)


doichanj (changes requested):
I think it is better to implement `AerState::apply_mcx` function in `src/controller/state_controller.hpp` and just call it from aer_runtime

```
void AerState::apply_mcsx(const reg_t &qubits) {
  assert_initialized();

  Operations::Op op;
  op.type = Operations::OpType::gate;
  op.name = "mcsx";
  op.qubits = qubits;

  buffer_op(std::move(op));
}
```

doichanj (approved):

Issue #1927, opened 2023-09-06
extended_stabilizer simulator returns max_memory_mb error even for very small circuits


epelofske-LANL:
### Information

- **Qiskit Aer version**: `qiskit-aer==0.12.2`
- **Python version**: Python 3.9.17
- **Operating system**: macOS

### What is the current behavior?

Some circuit types, regardless of their size return the following error:
```
qiskit.exceptions.QiskitError: 'ERROR: a circuit requires more memory than max_memory_mb. ,  EMPTY'
```

When executed with the the `extended_stabilizer` simulator option. 

This suggests that this error is not actually being thrown because of the size of the circuit, since these circuits can be simulated very easily on for example the `qasm_simulator`. 

### Steps to reproduce the problem
[ex.txt](https://github.com/Qiskit/qiskit-aer/files/12529372/ex.txt) is a circuit with 5 qubits which throws this error. 

```
from qiskit.providers.aer import AerSimulator
from qiskit import qasm3
simulator = AerSimulator(method='extended_stabilizer')

qc = qasm3.load("ex.txt")
simulator.run(qc, shots=10).result().get_counts()
```

Note that issue #1921 also mentions this error; this error occurs for circuits without classical control instructions. This qasm file gives an example circuit (with 36 active qubits) where this happens: [example.txt](https://github.com/Qiskit/qiskit-aer/files/12529423/example.txt)

doichanj:
I think this should be fixed by PR 1933 and option commented in issue #1921

Issue #1921, opened 2023-08-30
Aer function to check the memory requirement upfront when using simulators


patdenim:
### What is the expected behavior?
It would be nice to know what is the memory required before executing a circuit using a simulator.  
It is impossible to exactly know the memory needed when using the **extended_simulator**.  Looking at the code [extended_stabilizer_state.hpp](https://github.com/Qiskit/qiskit-aer/blob/main/src/simulators/extended_stabilizer/extended_stabilizer_state.hpp). The formula `required_chi()` used by the extended simulator does not seem to be only a function of the exact gate counts in the circuit 
```
size_t State::required_memory_mb(uint_t num_qubits,
                                 const std::vectoroperations::Op &ops) const {
  size_t required_chi = compute_chi(ops.cbegin(), ops.cend());
  // 5 vectors of num_qubits*8byte words
  // Plus 2*CHSimulator::scalar_t which has 3 4 byte words
```

Here is an example that shows this would help if knowing the exact memory required per circuit.  
Consider 2 different circuits, with the following gates (**Circuit** **A** is having less gates than **Circuit** **B**) :

**Circuit A** 
OrderedDict([
             ('u1', 272),
             ('sx', 133),
             ('cx', 114),
             ('h', 36),
             ('measure', 36)])

**Circuit B**
OrderedDict([
             ('u1', 480),
             ('cx', 228),
             ('sx', 216),
             ('h', 36),
             ('measure', 36)])

Then execute those 2 circuits on the `extended_stabilizer`

```
extended_stabilizer_simulator = AerSimulator(method='extended_stabilizer')
# Transpile circuit for backend
tcirc = transpile(circuit, extended_stabilizer_simulator)
extended_stabilizer_result = extended_stabilizer_simulator.run(tcirc, shots=1).result()
print('This succeeded?: {}'.format(extended_stabilizer_result.success))

```

Circuit A is failing with message 
```
Simulation failed and returned the following error message:
ERROR: a circuit requires more memory than max_memory_mb.
```
while Circuit B works. (even though Circuit B is much bigger than Circuit A)

doichanj:
I found issue in calling `required_memory_mb` function that `State::set_config` is not called before calculating required memory. In extended stabilizer, `extended_stabilizer_approximation_error` was not set from config parameters. This parameter can control size requirement when setting lerger error acceptance

doichanj:
[...]

Issue #1920, opened 2023-08-29
Matrix product state simulator aborts with wrong error message


smiserman:
### Informations

- **Qiskit Aer version: 0.12.1**:
- **Python version: 3.10.9**:
- **Operating system: Win11, 16 GB RAM**:

### What is the current behavior?
The `AerSimulator(method='matrix_product_state')` aborts with an error message if the number of qubits of the circuit is greater than the maximum memory / 32 MB, regardless of the size of a generated MPS. With 16 GB RAM the number of qubits is limited to 504, which is in any case not necessary with a moderate bond dimension. 

### Steps to reproduce the problem
```python
num_qubits = 505
circ = QuantumCircuit(num_qubits)
circ.h(0)
for i in range (0, num_qubits-1):
    circ.cx(i, i+1)
circ.measure_all()

simulator = AerSimulator(method='matrix_product_state')
res = simulator.run(circ).result()
```
gives
![image](https://github.com/Qiskit/qiskit-aer/assets/117047381/7bbbd746-b9d5-4b80-aa75-10f53df4d0f7)

### What is the expected behavior?
The simulation of the circuit above works fine in 1-2 sec for 504 qubits. So it should not abort with 505 qubits. Furthermore the required memory for the MPS is only about 0.3 MB and not  $505 \cdot 32 = 16160$ MB! 

### Suggested solutions
You should keep track of the size of the generated MPS, which is bounded above by approximately $N (32 d^2 + 8 d)$ bytes, where $N$ is the number of qubits and $d$ is the maximum bond dimension. There is some overhead for list, tuple, and Ndarray structures that can be neglected for larger $d$. In my own simulation, run in Numpy, I get $N (32 d^2 + 8 d + 440)$ and which, interestingly, is several seconds faster than the qiskit simulation for larger $d.$ So one should be able to comfortably simulate 1000 qubit circuits with d = 64 (correspond to 6 entanglement blocks in a VQE ansatz) and even more so with d=2 on a 16 GB RAM device.

**Another remark**
You should adapt your [matrix product state simulation method tutorial](https://[qiskit.org/documentation/stable/0.42/tutorials/simulators/7_matrix_product_state_method.html) in qiskit: *We can handle more qubits than this, but execution may take a few minutes. Try running a similar circuit with 500 qubits! Or maybe even 1000 (you can get a cup of coffee while waiting).*
[...]

PR #1933, opened 2023-09-08
Fix required_memory_mb for MPS and extended stabilizer

doichanj:
### Summary
This is the fix for `required_memory_mb` that returned wrong (larger) memory size for MPS and extended stabilizer.
Fix for issues #1920  , #1921 and #1927 

### Details and comments
`required_memory_mb` for MPS returned only initial size and returned size in byte not MB so it was very large. I added size estimator to calculate max bond dimension and now it returns reasonable memory size.
This size estimation only calculates size of tensors from the input operators and it takes ignorable time, but I think it is not good to call this multiple times. Before this fix, `required_memory_mb` was called from some functions. I changed to call this once at the beginning of simulation.

`required_memory_mb` for extended stabilizer did not refer to the configurable option `extended_stabilizer_approximation_error` becuase `State::set_config` was not called before calling `required_memory_mb`.
The issue #1921 and #1927 are not resolved as is, but we can decrease required memory size by setting larger `extended_stabilizer_approximation_error` option.


src/simulators/circuit_executor.hpp
```diff
@@ -204,6 +215,8 @@ template <class state_t>
 Executor<state_t>::Executor() {
   max_memory_mb_ = 0;
   max_gpu_memory_mb_ = 0;
+  required_memory_mb_ = 0;
+  set_required_memory_mb_ = false;
```
hhorii, line 219: I know that calling `required_memory()` multiple time is inefficient. However, `CircuitExecutor` should be stateless about `Circuit` and `Config` to be simulated. If we have these cache, `circuit` and `config` need to be `const` members of `CircuitExecutor`.

doichanj: Now required_memory_mb is ccalled once in controller, but called several times in executor to make it stateless

hhorii (approved):
Now LGTM.

PR #1840, opened 2023-06-07
Batch QuantumCircuit with multiple parameter_binds

ikkoham:

PR #1901, opened 2023-08-22
Implementing runtime parameter binding

doichanj:
### Summary
This is new feature to optimize GPU simulation for single circuit with multiple parameters.
Before this PR, multiple circuits were given by binding parameters from a single circuit. This PR bind parameters to each gates at runtime on a single circuit with multiple shots of simulations, so we can speed up by combining batched execution on GPU.
image

### Details and comments
Runtime parameter binding puts list of parameters to each operation and parameter is selected for each shot at runtime.
This option changes parallelization from multi-experiments to multi-shots.
image

PR #1935, opened 2023-09-12
Improve performance of sampler experiment

ikkoham:
### Summary

Same idea with https://github.com/Qiskit/qiskit-aer/pull/1840

### Details and comments

Benchmark code:
```python
import numpy as np

from qiskit.circuit.library import EfficientSU2
from qiskit_aer.primitives import Sampler

rng = np.random.default_rng(15)

qc = EfficientSU2(3)
qc.measure_all()

num_circuits = 1000
sampler = Sampler()
parameter_values = rng.random((num_circuits, qc.num_parameters))
circuits = [qc] * num_circuits

%timeit sampler.run(circuits, parameter_values).result()
```
0.12.2: 716 ms ± 1.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
This PR: 534 ms ± 4.33 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

### DIFF
--- a/qiskit_aer/primitives/sampler.py
+++ b/qiskit_aer/primitives/sampler.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 from collections.abc import Sequence
 
+import numpy as np
 from qiskit.circuit import ParameterExpression, QuantumCircuit
 from qiskit.compiler import transpile
 from qiskit.exceptions import QiskitError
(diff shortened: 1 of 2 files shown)


doichanj (approved):
I confirm this PR will improve performance with runtime parameter binding #1901

Issue #1928, opened 2023-09-06
Support half-integer Pi angle RZ gates in "stabilizer" and "clifford" methods


chriseclectic:
### What is the expected behavior?

The qiskit transpiler has a lot of issues wrt to Clifford basis circuits which makes testing larger circuits on the clifford or 
stabilizer simulator difficult.

Probably the most straightforward way to "fix" this is to allow "rz" basis gates in Aer for the "stabilizer" when the parameter is a half-integer multiple of pi, (k * pi /2), possible with some modification of terras transpiler passes to make sure 1Q synthesis and basis decomposition keeps RZ gates as Cliffords whenever possible.

This validation could happen when the circuit is loaded in to the simulator (if we want to be able to choose the stabilizer method for the "automatic" case). Another option to just make this work with the stablizer/clifford methods would be runtime validation whenever the simulator goes to apply an RZGate, to check the angle is mod 2pi is half integer multiple of Pi, and then apply the appropriate I, Z, S, or Sdg gate based on the value (or raise an exception of it is an invalid value).

ikkoham:
Do you have a plan to support Rx and Ry as well? RzGates come from transpilation for IBM Quantum processor, but qiskit is backend agnostic.

mtreinish:
FWIW, the `Target` class supports specifying supported gates with fixed angle values. For example: https://github.com/Qiskit/qiskit/blob/main/qiskit/providers/fake_provider/fake_mumbai_v2.py#L441 the transpiler probably doesn't fully work with a target specified like this as this feature is undertested. But it at least in theory lets aer advertise to users and the transpiler that it only supports multiple fixed angle values for a rotation gate.

PR #1938, opened 2023-09-14
Add rotation gates to stabilizer

doichanj:
### Summary
This PR adds rotation gates (rx, ry and rz gates) to stabilizer if input theta is multiple of pi/2.
#1928 

### Details and comments
stabilizer method is automatically selected by `method=automatic` when all the input thetas of rotation gates are multiple of pi/2. Or if `method=stabilizer  is set, Aer raises and exception when invalid thetas are passed.


hhorii (approved):
LGTM

PR #1939, opened 2023-09-15
Fix incorrect excited state population when building noise model from BackendV2

### LINKED ISSUE #1937: wrong excited state population when building noise model from backend - frequency unit
### Informations

- **Qiskit Aer version**: 0.12.2
- **Python version**: 3.10
- **Operating system**: WSL - Ubuntu

When building a noise model from a backend with a non zero temperature, the excited state population is not correctly computed.
The Maxwell Boltzmann distribution is computed in qiskit_aer.noise.device.models._excited_population(freq, temperature) where the frequency and temperature are normally in GHz and mK, but the frequency is passed in Hertz, yielding a 1e9 error in exp_param and an incorrect population.

The frequency is taken from relax_params in qiskit_aer.noise.device.models._device_thermal_relaxation_error, called in qiskit_aer.noise.device.models_basic_device_target_gate_errors, where the params are taken from Target.qubit_properties, that is an IBMQubitProperties object, where the frequency is stored in Hz.

Build a noise model from backend with non zero temperature : 
```
from qiskit_aer.noise import NoiseModel
provider = IBMProvider()
machine='ibmq_jakarta'
model = NoiseModel.from_backend(provider .get_backend(machine), temperature=10)

```
On my end this yields a warning : 
`qiskit_aer/noise/device/models.py:394: RuntimeWarning: overflow encountered in exp  exp_param = exp((47.99243 * freq) / abs(temperature))`

And then look at freq in qiskit_aer.noise.device.models._excited_population, it's in Hz.

The frequency should be expected in Hz, and a factor of 1e-9 should be added to the exp_param computation.

### Suggested solutions

Change the function to accept freq in Hz and then add a 1e-9 factor in exp_param.

itoko:
### Summary
Fixed a bug where thermal relaxation errors with incorrect excited state population are created when building a noise model from BackendV2 with non-zero temperature.
Fixed #1937 .

### Details and comments
Fixed an oversight in the handling of frequency units when building a noise model from BackendV2. For that, as suggested in #1937, I changed the `qiskit_aer.noise.device.models._excited_population` to expect the frequency argument in Hz (from GHz) so that the path for BackendV2 works correctly without any updates, and, instead, updated the path for BackendV1.

### DIFF
--- a/qiskit_aer/noise/device/models.py
+++ b/qiskit_aer/noise/device/models.py
@@ -171,12 +171,15 @@ def basic_device_gate_errors(
         )
 
     # Generate custom gate time dict
+    # Units used in the following computation: ns (time), Hz (frequency), mK (temperature).
     custom_times = {}
     relax_params = []
     if thermal_relaxation:
         # If including thermal relaxation errors load
-        # T1, T2, and frequency values from properties
+        # T1 [ns], T2 [ns], and frequency [GHz] values from properties
         relax_params = thermal_relaxation_values(properties)
+        # Unit conversion: GHz -> Hz
+        relax_params = [(t1, t2, freq * 1e9) for t1, t2, freq in relax_params]
         # If we are specifying custom gate times include
         # them in the custom times dict
         if gate_lengths:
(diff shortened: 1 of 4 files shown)


hhorii (approved):
LGTM

hhorii (changes requested):
Please add release note.

hhorii (approved):

Issue #1941, opened 2023-09-19
Run tests with unreleased version of Qiskit-related dependencies


1ucian0:
Aer depends on `qiskit-terra`. As `qiskit 1.0` is approaching, removal of deprecated code is getting more aggressive, consider running `qiskit-aer` tests with the the code in qiskit main branch. You can do that running this command in your environment and running tests with it. 
```sh
pip install -U git+https://github.com/Qiskit/qiskit-terra
```

If any of the test fail or find `DeprecatedWarning`, proactively fix that code (by deprecation policy, you should have a clear alternative for the removed code). Otherwise, your code will stop working in coming releases. 

Ideally, it would be great if a regular (say, weekly?) CI job makes this check for you. Similar to https://github.com/Qiskit/qiskit-ibm-runtime/blame/main/.github/workflows/unit-tests-terra-main.yml

PR #1950, opened 2023-09-29
Add test for unreleased Qiskit

doichanj:
### Summary
This PR is preparation for unreleased Qiskit to implement changes and remove deprecations (#1941 )

### Details and comments
New test case `unit-tests-latest-qiskit-terra` is used to detect changes in the latest Qiskit and if any deprecation error found, this PR should be updated

### 
Change logs:
`bind_parameters` is replaced with `assign_parameters`
`QuantumCircuit.i()` is replaced with `QunatumCircuit.i()` by PR #1945


.github/workflows/tests.yml
```diff
@@ -280,3 +280,54 @@ jobs:
...
           stestr run --slowest
         shell: bash
+  unit-tests-latest-qiskit-terra:
+    runs-on: ${{ matrix.os }}
+    needs: [sdist, lint]
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        python-version: ["3.11"]
+        os: ["ubuntu-latest"]
```
mtreinish, line 290: Are you thinking we should run this on every commit? Or was the eventual plan to run this job periodically when it's no longer a `WIP`? We can set this up in a new workflow file and create a nightly cron job to run this test if we want to do it periodically, see: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

.github/workflows/tests.yml
```diff
@@ -280,3 +280,54 @@ jobs:
...
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-${{ matrix.python-version}}-pip-test-${{ hashFiles('setup.py','requirements-dev.txt','constraints.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.python-version}}-pip-test-
+            ${{ runner.os }}-${{ matrix.python-version}}-pip-
+            ${{ runner.os }}-${{ matrix.python-version}}-
+      - name: Install Deps
+        run: |
+          python -m pip install -U -c constraints.txt -r requirements-dev.txt wheel
+          pip install -U git+https://github.com/Qiskit/qiskit-terra.git
```
mtreinish, line 316: 国王崩御、国王万歳

```suggestion
          pip install -U git+https://github.com/Qiskit/qiskit.git
```

mtreinish (commented):
This is a great idea, we had this coverage in the past where tests ran with terra main on each job. But blocking PRs on terra main was a bit too disruptive at the time (just from a dependency management PoV). So adding a dedicated job back to ensure we have some coverage of upcoming changes to terra is a good idea, especially if we do it in a non-blocking manner. The open question I have is mostly about the long term strategy for how/when we run this job.

.github/workflows/tests.yml
doichanj: I want to use this test case for testing deprecations and changes of Qiskit. I will remove this test case and only merge fixes in this PR.

1ucian0:
> The open question I have is mostly about the long term strategy for how/when we run this job.

Other components run it daily:
 * `qiskit-ibm-provider`: https://github.com/Qiskit/qiskit-ibm-provider/actions/workflows/unit-tests-terra-main.yml
 * `qiskit-ibm-runtime`: https://github.com/Qiskit/qiskit-ibm-runtime/actions/workflows/unit-tests-terra-main.yml

mtreinish (approved):

PR #1953, opened 2023-10-06
Release 0.13.0

doichanj:
### Summary
Preparation PR for Aer 0.13.0

### Details and comments
Required PRs merged for release 0.13.0 are

- #1875 
- #1915  
- #1950


mtreinish:
I just found out via testing that the `pypa/gh-action-pypi-publish` action only works on linux. So we'll need to combine the existing `upload-artifact` action with a `download-artifact` in a second stage that publishes to pypi for the non-linux jobs (see: https://github.com/pypa/gh-action-pypi-publish/tree/unstable/v1#non-goals for the reference to this).

doichanj:
I added separate job in deploy.yml to publish on Linux environment by using upload/download artifacts. I hope this works fine.
(I'm testing on other repository to publish on TestPyPI)

mtreinish:
Yeah, that looks better. Although, FWIW you don't need to link up the GPU jobs or any of the custom architecture linux jobs. I personally would keep them separate because those builds can be very slow relative to the linux x86 builds, macOS, and windows builds (because of the qemu emulation or CUDA). If you keep them separate you don't block the upload of the primary platforms for the slower builds. I had a similar concern on Rustworkx, you can see what I did here: https://github.com/Qiskit/rustworkx/pull/1001/files

doichanj:
I fixed `wheel-build` and `wheel-arm64-macos` only use shared publisher

woodsp-ibm:
I note in the requirements it says qiskit-terra >= 0.21
```
requirements = [
    "qiskit-terra>=0.21.0",
    "numpy>=1.16.3",
    "scipy>=1.0",
]
```
Is this something that should be changed - elsewhere, with the change over to qiskit that has been used now by many dependents e.g
```
qiskit>=0.44
```
which I think is the earliest version that had the change done.

doichanj:
I would like to rename test/terra in qiskit-aer 0.14.0

setup.py
```diff
@@ -22,7 +22,7 @@
 extras_requirements = {"dask": ["dask", "distributed"]}
 
 requirements = [
-    "qiskit-terra>=0.21.0",
+    "qiskit>=0.25.0",
```
mtreinish, line 25: I think this should  either be `qiskit-terra>=0.25.0` or `qiskit>=0.44.0`

hhorii (commented):
Prelude will be helpful for release notes. Here is a case of 0.12: https://github.com/mtreinish/qiskit-aer/blob/b6f8e606bd36bbe66cda1246f748515a6a203c3d/releasenotes/notes/0.12/prepare-0.12-0da477fc0492ca5d.yaml

doichanj:
I'm sorry but I broke this PR so I opened new PR #1964 for release 0.13.0

PR #1954, opened 2023-10-06
[rebranding] Qiskit Aer is now Aer

1ucian0:
With the removal of the notion of _elements_ in Qiskit, Qiskit Aer is now rebranded as Aer. This does not implies the change of the package or repo rename.


doichanj:
Will we rebrand Aer when Qiskit 1.0 will be released or now?

1ucian0:
I think the sooner, the better. Is there any cross dependency with Qiskit 1.0 you think we should consider? Remember this change does not change the module name or the package. Just the way we refer to it as a component in the Qiskit ecosystem.

doichanj (approved):

PR #1958, opened 2023-10-11
Move methods into class pages for docs

arnaucasau:
### Summary

Moved all the methods and attributes into class pages to speed up the building process of the documentation with the new ecosystem sphinx theme. This change unblocks this [PR](https://github.com/Qiskit/qiskit-aer/pull/1946).

See https://github.com/Qiskit/qiskit/pull/10455 for more information.

PR #1961, opened 2023-10-19
Fix templates documentation

arnaucasau:
### Summary
This PR creates another template for the classes without inherited members simplifying the `class.rst` template file.

Also, In the original `class.rst`, we have two consecutive for, one for all methods, and another one for only the inherited_members. With this two for, we are duplicating the entries of the inherited methods (they appear two times in the page). In a previous [pull request](https://github.com/Qiskit/qiskit-aer/pull/1958), I added an extra if but that removed the inherit_members that we had before. This PR fixes that problem.


doichanj:
could you check license setting

doichanj (approved):

PR #1963, opened 2023-10-24
Deprecate approximation option in Estimator

ikkoham:
### Summary

Deprecate approximation option. `approximation=True` will be the default in the future.
If users want to use `approximation=False`, they should use `BackendEstimator` instead.

### DIFF
--- a/qiskit_aer/primitives/estimator.py
+++ b/qiskit_aer/primitives/estimator.py
@@ -32,6 +32,7 @@ from qiskit.providers import Options
 from qiskit.quantum_info import Pauli, PauliList
 from qiskit.quantum_info.operators.base_operator import BaseOperator
 from qiskit.result.models import ExperimentResult
+from qiskit.utils import deprecate_arg, deprecate_func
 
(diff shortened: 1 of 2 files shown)


ikkoham:
Needs a release note

doichanj (approved):

qiskit_aer/primitives/estimator.py
```diff
@@ -100,7 +107,15 @@ def __init__(
         self._transpile_options = Options()
         if transpile_options is not None:
             self._transpile_options.update_options(**transpile_options)
-        self.approximation = approximation
+        if not approximation:
+            warn(
+                "Option approximation=False is deprecated as of qiskit-aer 0.13. "
+                "It will be removed no earlier than 3 months after the release date. "
+                "Instead, use BackendEstmator from qiskit.primitives.",
```
woodsp-ibm, line 114: I just happened to notice this PR - in looking over it I see that BackendEst***i***mator is missing the ***i*** in this warning - same in the one below. Pretty minor but I thought I'd point it out.

Issue #1957, opened 2023-10-11
`qiskit.test` is going to be removed in `qiskit 1.0`


1ucian0:
Qiskit exposes test utilities as part of its deliverable package. This is not consider good practice and [its removal is a long overdue issue](https://github.com/Qiskit/qiskit/issues/6862). Qiskit teams's intention is [to clean that up for `qiskit 1.0`](https://github.com/Qiskit/qiskit/pull/10998) (planned to be released on Jan 31st, 2024).  The module will be available in the `0.*` versions (EoL, mid 2024) and [raise a deprecation warning](https://github.com/Qiskit/qiskit/pull/11001).

Aer uses `qiskit.test` in, at least, the following file:
https://github.com/Qiskit/qiskit-aer/blob/64a969656e841e9bbd5b76c7e19ae9a475762cd9/test/terra/common.py#L33

Consider stop importing it from qiskit and move the relevant testing-supporting code to Aer codebase.

1ucian0:
Is this issue solved by https://github.com/Qiskit/qiskit-aer/pull/1971 ?

doichanj:
yes, Aer is now independent from qiskit.test

1ucian0:
Awesome. Closing via https://github.com/Qiskit/qiskit-aer/pull/1971 in that case.

PR #1971, opened 2023-10-27
Remove qiskit.test from test/common.py

doichanj:
### Summary
Removed importing qiskit.test that will be deprecated  #1957 

### Details and comments
some classes in qiskit.test are copied to test/common.py and test/decorators.py

### DIFF
--- a/test/terra/common.py
+++ b/test/terra/common.py
@@ -18,6 +18,7 @@ import inspect
 import logging
 import os
 import warnings
+import unittest
 from enum import Enum
 from itertools import repeat
 from math import pi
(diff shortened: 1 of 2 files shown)


wshanks:
I noticed that this PR changed the inheritance of the `QiskitAerTestCase` so that it no longer depends on `testtools.TestCase`, going straight to `unittest.TestCase`. This is why the Python 3.12 tests now generate warnings about `addDuration` like [here](https://github.com/Qiskit/qiskit-aer/actions/runs/7564004257/job/20597524765#step:8:15). Switching from `unittest.TestCase` to `testtools.TestCase` would avoid this. I opened a bug about the warning upstream [here](https://bugs.launchpad.net/subunit/+bug/2045171).

doichanj:
I would like to keep as is until warnings will be errors, because there are some differences between 2 tools that we have to modify many test codes

Issue #1977, opened 2023-10-30
Initializing the zero state and applying a circuit returns a different statevector than applying a circuit without initializing the zero state.


JamesB-1qbit:
### Informations

- **Qiskit Aer version**: 0.13.0
- **Python version**: 3.9.16
- **Operating system**: MacOS

### What is the current behavior?
When using method="statevector" with AerSimulator, a circuit with an `initialize` of |00> (i.e. [1,0,0,0]) returns a different statevector than the same circuit without explicitly initializing the |00> state.

### Steps to reproduce the problem
```python3
import qiskit
from qiskit_aer import AerSimulator

circuits = dict()

circuits["zero"] = qiskit.QuantumCircuit(2)
circuits["zero"].initialize([1., 0., 0., 0.])

circuits["X"] = qiskit.QuantumCircuit(2)
circuits["X"].x(0)

circuits["zero+X"] = circuits["zero"].compose(circuits["X"])

def transpile_run_and_return_statevector(circ):
    aersim = AerSimulator(method="statevector")
    transpiled_circ = qiskit.transpile(circ, aersim)
    transpiled_circ.save_statevector()
    sim_results = aersim.run(transpiled_circ).result()
    print(f"Circuit:\n{transpiled_circ}")
    print(f"Statevector:\n{sim_results.get_statevector(transpiled_circ)}")

for name, circ in circuits.items():
    print(f"For circuit {name}\n")
    transpile_run_and_return_statevector(circ)
    print("\n\n")
```
results in 
```
For circuit zero

Circuit:
     ┌──────────────────────┐ statevector 
q_0: ┤0                     ├──────░──────
     │  Initialize(1,0,0,0) │      ░      
q_1: ┤1                     ├──────░──────
     └──────────────────────┘      ░      
Statevector:
Statevector([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j],
            dims=(2, 2))



For circuit X

Circuit:
global phase: π/2
     ┌────────────────┐ statevector 
q_0: ┤ U3(π,-π/2,π/2) ├──────░──────
     └────────────────┘      ░      
q_1: ────────────────────────░──────
                             ░      
Statevector:
Statevector([3.74939946e-33+6.123234e-17j, 1.00000000e+00+0.000000e+00j,
             0.00000000e+00+0.000000e+00j, 0.00000000e+00+0.000000e+00j],
            dims=(2, 2))



For circuit zero+X

Circuit:
global phase: π/2
[...]

PR #1980, opened 2023-11-01
Applying global phase multiplication to initialize operation

doichanj:
### Summary
This PR is fix for issue #1977 

### Details and comments
The global phase variable was not multiplied after `initialize` operation, this fix multiplies the global phase to the state in `initialize` operation.
Also `matrix_product_state` method did not use global phase, so this PR adds support of global phase in MPS method

### DIFF
--- a/src/simulators/statevector/statevector_executor.hpp
+++ b/src/simulators/statevector/statevector_executor.hpp
@@ -104,7 +104,7 @@ protected:
   // computing the tensor product with the new state |psi>
   // /psi> is given in params
   void apply_initialize(const reg_t &qubits, const cvector_t &params,
-                        RngEngine &rng);
+                        RngEngine &rng, bool recursive = false);
 
   void initialize_from_vector(const cvector_t &params);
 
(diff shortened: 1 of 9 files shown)


src/simulators/statevector/statevector_executor.hpp
```diff
@@ -1260,9 +1260,17 @@ std::vector<reg_t> Executor<state_t>::sample_measure(const reg_t &qubits,
 template <class state_t>
 void Executor<state_t>::apply_initialize(const reg_t &qubits,
                                          const cvector_t &params,
-                                         RngEngine &rng) {
+                                         RngEngine &rng, bool recursive) {
```
hhorii, line 1263: I think we do not need to change this method signature:
```
template <class state_t>
void Executor<state_t>::apply_initialize(const reg_t &qubits,
                                         const cvector_t &params_,
                                         RngEngine &rng) {
  auto sorted_qubits = qubits;
  std::sort(sorted_qubits.begin(), sorted_qubits.end());

  cvector_t tmp;
  // apply global phase here
  if (Base::states_[0].has_global_phase()) {
    tmp.resize(params.size());
    for (int_t i = 0; i < params.size(); i++) {
      tmp[i] = params[i] * Base::states_[0].global_phase();
    }
  }
  const cvector_t& params = tmp.empty() ? params_: tmp;
```

`omp parallel` maybe necessary if `param` exceeds a threshold.

hhorii (approved):

PR #1984, opened 2023-11-02
Fix name map of aerbackend

doichanj:
### Summary
Fix name maps that causes error in Qiskit's controlled gate testcases.

### Details and comments
Some custom multi controlled gates are caused test failure in `test.python.circuit.test_controlled_gate` in Qiskit because control qubits are sets only when the custom gates are MCU1Gate, MCPhaseGate and MCXGate, but float type value is passed,
so I added type cast to integer to avoid failure


doichanj:
I would like to close this one and fix in PR #1995

PR #1994, opened 2023-11-08
handle T2 > 2*T1 and clarify terminology

dtmcclure:
### Summary
- replaces the NoiseError when T2 > 2*T1 with a warning and a line that caps T2 at the 2*T1 limit
- clarified some terminology issues i happened to notice while doing the above

### Details and comments

Since T1 and T2 are typically measured at different times, it's not that uncommon for the most recently reported T2 to exceed the most recently reported T1. Rather than treating this as an error, it seems better to just cap T2 at the physical limit and emit a warning.

While at it, I replaced some terms to clarify that T2 is the overall coherence time, not the relaxation time (T1) or the dephasing time (T_phi).

### DIFF
--- a/qiskit_aer/noise/errors/standard_errors.py
+++ b/qiskit_aer/noise/errors/standard_errors.py
@@ -14,7 +14,7 @@ Standard quantum computing error channels for Aer.
 """
 
 import itertools as it
-
+import logging
 import numpy as np
 
 from qiskit.circuit import Reset
@@ -28,6 +28,7 @@ from qiskit.quantum_info.operators.predicates import is_unitary_matrix
 from .quantum_error import QuantumError
 from ..noiseerror import NoiseError
 
+logger = logging.getLogger(__name__)
 
 def kraus_error(noise_ops, canonical_kraus=False):
     """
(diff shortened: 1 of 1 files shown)


itoko:
Thanks, @dtmcclure (sorry for slow response). I'm all for improving terminology.
Also the upgrade in handling of T2 value makes sense to me.
In current API, callers of `thermal_relaxation_error` has responsibility to truncate T2 value and a private utility function [_truncate_t2_value](https://github.com/Qiskit/qiskit-aer/blob/7adacaad36ba6fbbc2f28b23fc8765dd7e5dfcd6/qiskit_aer/noise/device/models.py#L368) is provided. Raising errors in the case of T2 > 2 * T1 may be a bit beneficial (error-pruning) for users who provide T2 value by hand while the internal truncation (proposed in this PR) may be convenient for users who pass T2 values experimentally measured (not always need to write `thermal_relaxation_error(t1, _truncate_t2_value(t1, t2), time)`). I now lean to weight the latter benefit. I agree the change but it should be well documented.

Could you update docstring to mention the new spec (truncation of T2 to 2 * T1) and add a release note to tell the upgrade (raise an error -> silent truncation in the case of T2 > 2 * T1)?

Issue #1988, opened 2023-11-06
`AerSimulator.coupling_map` is not properly generated


t-imamichi:
### Informations

- **Qiskit Aer version**: 0.13.0
- **Python version**: 3.10.13
- **Operating system**: macOS 14.1

### What is the current behavior?

`AerSimulator` becomes `BackendV2` but `AerSimulator.coupling_map` does not seem properly generated.

### Steps to reproduce the problem

```python
from qiskit_aer import AerSimulator

backend = AerSimulator(coupling_map=[[0,1]])
print('backend.coupling_map:', backend.coupling_map, type(backend.coupling_map))
print('backend.configuration().coupling_map:', backend.configuration().coupling_map)
```

output
```
backend.coupling_map: None <class 'NoneType'>  # expected `CouplingMap`
backend.configuration().coupling_map: [[0, 1]]
```

### What is the expected behavior?

`AerSimulator.coupling_map` returns the coupling_map instead of None.

### Suggested solutions

`AerSimulator.set_options` allows users modify coupling map on-demand. It might be good to overload `BackendV2.coupling_map` to generate a coupling map dynamically.

Issue #1987, opened 2023-11-04
 AerSimulator.from_backend() does not work with qiskit_aer 0.13.0


mmartinsantamaria:
### Informations

- **Qiskit Aer version**: 0.13.0
- **Python version**: 3.11.4
- **Operating system**: macOS 11.7.10

### What is the current behavior?

When the next code is executed:

```
from qiskit_ibm_provider import IBMProvider
from qiskit_aer import AerSimulator

provider=IBMProvider()
backend=provider.get_backend("ibm_nairobi")
backend=AerSimulator.from_backend(backend)

```
The next error appears:

```
KeyError                                  Traceback (most recent call last)
File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/qiskit/providers/models/backendconfiguration.py:383, in QasmBackendConfiguration.__getattr__(self, name)
    382 try:
--> 383     return self._data[name]
    384 except KeyError as ex:

KeyError: 'description'

The above exception was the direct cause of the following exception:

AttributeError                            Traceback (most recent call last)
Cell In[3], line 3
      1 #Consigo imitar las características del computador real
      2 #utilizando el método from_backend
----> 3 backend_aer=AerSimulator.from_backend(backend)

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/qiskit_aer/backends/aer_simulator.py:843, in AerSimulator.from_backend(cls, backend, **options)
    840         options["noise_model"] = noise_model
    842 # Initialize simulator
--> 843 sim = cls(configuration=configuration, properties=properties, **options)
    844 return sim

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/qiskit_aer/backends/aer_simulator.py:706, in AerSimulator.__init__(self, configuration, properties, provider, **backend_options)
    702 # Cache basis gates since computing the intersection
    703 # of noise model, method, and config gates is expensive.
    704 self._cached_basis_gates = self._BASIS_GATES["automatic"]
--> 706 super().__init__(
    707     configuration, properties=properties, provider=provider, backend_options=backend_options
    708 )

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/qiskit_aer/backends/aerbackend.py:75, in AerBackend.__init__(self, configuration, properties, defaults, backend_options, provider)
     70 configuration.simulator = True
     71 configuration.local = True
     72 super().__init__(
[...]

Issue #1982, opened 2023-11-01
Transpiler transpiles unnecessary gate optimization


doichanj:
### Informations

- **Qiskit Aer version**: 0.13.0
- **Python version**:
- **Operating system**:

### What is the current behavior?
From Aer 0.13.0 with BackendV2, transpiler returns circuit with unnecessary gate conversions.

For example H gate is transpiled to P gate, X gate is transpiled to U3 gate reported in #1977 

For extended stabilizer method, converting H to P is bad strategy, because H is Clifford gate but P is not that takes very long time to simulate.

### What is the expected behavior?
For the simulator backends, gate conversion is not necessary and should be ignored even when transpiler's optimization level is higher (if some of the backend model is not applied)

### Suggested solutions
This can be avoided by setting some parameters in backend target class (e.g. error rates) as default parameters for simulator backends.

PR #1995, opened 2023-11-10
Fix AerBackend issues caused by upgrading BackendV2

doichanj:
### Summary
This PR fixes BackendV2 implementations in Aer compatible for Qiskit 0.45

### Details and comments
This PR fixes issues :
- #1987 
- #1988 
- #1982 

Also this PR fixes test failure in Qiskit `test.python.circuit.test_controlled_gate` it requires fix in Qiskit https://github.com/Qiskit/qiskit/pull/11172

For #1987 if input backend does not have description, it fails in Qiskit, so this fix add description.

For #1988, Aer did not build coupling map from option, so this fix check additional options and build coupling map

For #1982, added custom transpiler pass to rebuild basis gates including only gates appears in the input circuit, that prevent unnecessary change of gate. This is enabled when `optimization_level < 2`


qiskit_aer/backends/aer_simulator.py
```diff
@@ -518,6 +518,7 @@ class AerSimulator(AerBackend):
                 "continue_loop",
                 "reset",
                 "switch_case",
+                "barrier",
```
mtreinish, line 521: Barrier typically isn't needed in a target because qiskit always assumes barrier is valid for a backend since it's primarily a compiler directive. Was there something specific that was failing

qiskit_aer/backends/name_mapping.py
```diff
@@ -250,33 +267,79 @@ def __init__(self, num_ctrl_qubits, ctrl_state=None):
...
+    "tdg": TdgGate,
+    "swap": SwapGate,
+    "cx": CXGate,
+    "cy": CYGate,
+    "cz": CZGate,
+    "ccx": CCXGate,
+    "csx": CSXGate,
+    "cp": CPhaseGate,
+    "cu": CUGate,
+    "cu1": CU1Gate,
```
mtreinish, line 309: For the standard gates they're covered by default from https://github.com/Qiskit/qiskit/blob/main/qiskit/circuit/library/standard_gates/__init__.py#L49 the custom name mapping passed to `convert_to_target()` is additive, the code takes `get_standard_gate_name_mapping()`'s output and runs `.update(NAME_MAPPING)` on it. Was there a reason you had to add these here too? Also in general if you pass in a gate class instead of an instance like this those instructions will be treated as variadic instructions which doesn't seem correct for standard library gates.

qiskit_aer/backends/plugin/aer_backend_plugin.py
```diff
@@ -0,0 +1,116 @@
...
+    def _add_ops(self, dag, ops):
+        num_unsupported_ops = 0
+        opnodes = dag.op_nodes()
+        if opnodes is None:
+            return num_unsupported_ops
+
+        for node in opnodes:
+            if isinstance(node.op, ControlFlowOp):
+                for block in node.op.blocks:
+                    num_unsupported_ops = num_unsupported_ops + self._add_ops(
```
mtreinish, line 47:
```suggestion
                    num_unsupported_ops += self._add_ops(
```

qiskit_aer/backends/plugin/aer_backend_plugin.py
```diff
@@ -0,0 +1,116 @@
...
+                        circuit_to_dag(block), ops
+                    )
+            if node.name in self.config.target:
+                if node.name not in ops:
```
mtreinish, line 51: Doesn't this have a linear search overhead. Could you make `ops` a set instead I think everything else will work fine if it's a set.

qiskit_aer/backends/plugin/aer_backend_plugin.py
```diff
@@ -0,0 +1,116 @@
...
+        self.config.target._gate_name_map.clear()
+        self.config.target._qarg_gate_map.clear()
+        self.config.target._global_operations.clear()
+
+        # rebuild gate sets from circuit
+        for name in ops:
+            if name not in self.config.target:
+                if name != "measure":
+                    self.config.target.add_instruction(NAME_MAPPING[name], name=name)
+        self.config.target.add_instruction(Measure())
```
mtreinish, line 83: TBH, this feels a bit unsafe in general. My initial thought was if you need to modify the target you should just create a new object and copy over the bits you want. But then that wouldn't work because you're relying on an internal detail that the transpiler passes the `target` by reference to all the passes. I'm worried that this isn't a explicit guarantee in the pass manager and could be fragile.

The bigger concern around this is that not all pases need to use the target. Since qiskit is still in the middle of it's transition to using `Target` everywhere (see https://github.com/Qiskit/qiskit/issues/9256 for the tracking issue), some external passes might not ever look at the target. Especially in the plugin interface external plugins may not look at the target and will only look at `PassManagerConfig.basis_gates` which is generated at pass manager initialization time. I think to be safe you probably also want to do:

```suggestion
        self.config.target._gate_map.clear()
        self.config.target._gate_name_map.clear()
        self.config.target._qarg_gate_map.clear()
        self.config.target._global_operations.clear()

        # rebuild gate sets from circuit
        for name in ops:
            if name not in self.config.target:
                if name != "measure":
                    self.config.target.add_instruction(NAME_MAPPING[name], name=name)
        self.config.target.add_instruction(Measure())
        self.config.basis_gates.clear()
        self.config.basis_gates.extend(self.config.target.operations)
```
to handle this. But if you do need to go down this path I think you have to document this as internal for aer only and not for use outside of the preset pass managers.

qiskit_aer/backends/plugin/aer_backend_plugin.py
```diff
@@ -0,0 +1,116 @@
...
+from qiskit.transpiler.passes import UnitarySynthesis
+from qiskit.transpiler.passes import HighLevelSynthesis
+from qiskit.circuit.equivalence_library import SessionEquivalenceLibrary as sel
+from qiskit.circuit.measure import Measure
+from qiskit.circuit import ControlFlowOp
+from qiskit.converters import circuit_to_dag
+from qiskit_aer.backends.name_mapping import NAME_MAPPING
+
+
+class AerBackendRebuildGateSetsFromCircuit(TransformationPass):
```
mtreinish, line 27: I think this should actually be an `AnalysisPass` you're not actually transforming the circuit, but are updating the target instead.

qiskit_aer/backends/plugin/aer_backend_plugin.py
```diff
@@ -0,0 +1,116 @@
...
+        self.config.target._gate_map.clear()
+        self.config.target._gate_name_map.clear()
+        self.config.target._qarg_gate_map.clear()
+        self.config.target._global_operations.clear()
+
+        # rebuild gate sets from circuit
+        for name in ops:
+            if name not in self.config.target:
+                if name != "measure":
+                    self.config.target.add_instruction(NAME_MAPPING[name], name=name)
```
mtreinish, line 82: I'm not sure this will do the correct thing in the case of a noise model. If you have a noise model won't this remove all the connectivity constraints caused by that? (this also connects with my comment below, you'd need to update the coupling map for any optimization or scheduling plugins).

qiskit_aer/backends/plugin/aer_backend_plugin.py
doichanj: Because there are only translation and scheduling plugins, I used translation plugin to analyze before going to optimization pass

qiskit_aer/backends/aer_simulator.py
doichanj: I removed barrier and add barrier inside plugin.
Adding barrier is needed because barrier can not be found in any name mapping and plugin returns with unsupported operation

qiskit_aer/backends/name_mapping.py
doichanj: removed standard gates from NAME_MAPPING

qiskit_aer/backends/aerbackend.py
```diff
@@ -351,17 +342,16 @@ def max_circuits(self):
...
     @property
     def target(self):
-        self._target = convert_to_target(
-            self.configuration(), self.properties(), self.defaults(), self._mapping
-        )
-        return self._target
+        if self._target is not None:
+            return self._target
+
+        return convert_to_target(self.configuration(), self.properties(), None, NAME_MAPPING)
```
hhorii, line 348: This is just a minor comment. Previously, converted `target` was cached. With this change, it will not be cached. I think frequent access of `target` property will produce performance regression. Introducing caching may help performance (though setting `method` may need clear cache of `target`).

hhorii (approved):
Though I added a minor comment, I believe that it is not so important.

qiskit_aer/backends/aerbackend.py
doichanj: because if target is cached and used for transpilation to multiple circuits, basis gates for the first circuit is reused that can be a problem if the second circuit has gate that the first one does not have

PR #1983, opened 2023-11-02
Release 0.13.1

doichanj:
### Summary
This PR is preparation for Release 13.

### Details and comments
This PR will contain:

- #1965
- #1976 
- #1980 
- #1989
- #1990
- #1996
- #1995 
- #2000 
- #2001
- #2002 
- #2003 
- #1993

PR #1996, opened 2023-11-13
Remove use of opflow in `Estimator`

ElePT:
### Summary
This PR removes the only `opflow` dependency in Aer (supporting `PauliSumOp` as an input to the Estimator). The `opflow` module was deprecated and is on its way of being removed from Qiskit, which means it will no longer be available for downstream projects (see https://github.com/Qiskit/qiskit/pull/11111). Once the linked PR is merged, importing `opflow` in the Aer primitives will break the neko integration tests (see https://github.com/Qiskit/qiskit/actions/runs/6848167193/job/18617803285?pr=11111#step:2:1073), so if I am not mistaken, the optimal path would be to remove the use of opflow first in Aer and then in Qiskit.

### Details and comments
I have added an upgrade release note documenting the removal of the supported type.

### DIFF
--- a/qiskit_aer/primitives/estimator.py
+++ b/qiskit_aer/primitives/estimator.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2022.
+# (C) Copyright IBM 2022, 2023.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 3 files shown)


mtreinish:
@doichanj do you think we should backport this for 0.13.1 (#1983)? Since it's just a type hint I think it's not really breaking, because if someone is using qiskit<1.0 they'll still be able to pass in a `PauliSumOp` and it should all work like before.

ikkoham (approved):
I agree with this direction. I wait for @doichanj decision.

doichanj (approved):
I agree with this removal and I would like to include release 0.13.1

PR #2003, opened 2023-11-21
Reverse ordering to read out error in sampling measure

doichanj:
### Summary
This PR fixes order of applying read out error in sampling measures in the circuit executor classes,
 that causes different results from older version of Aer.

### Details and comments
The order to apply read out error in sampling measure was reverse ordering by popping from back of vector in older version of Aer. So this fix reorder loop to apply consistent random values to read out error as older version.
This PR fixes failure of Qiskit's test case `test.python.primitives.test_backend_estimator.TestBackendEstimator.test_layout`

Also this PR includes fix of checking if batch execution on GPU can be applied or not.


hhorii (approved):
LGTM.

hhorii (approved):

Issue #1686, opened 2022-12-18
Any explicit or implicit import of Qiskit Aer would initialize all GPUs on the system if the CUDA support is built


leofang:
### Informations

- **Qiskit Aer version**: 0.11.2 
- **Python version**: 3.9.15
- **Operating system**: Ubuntu 20.04

### What is the current behavior?
Importing Qiskit Aer either implicitly or explicitly, as shown below, would get all GPUs on the system initialized, as evidenced by monitoring `nvidia-smi` (there are other tools to check this, but `nvidia-smi` is the simplest).

### Steps to reproduce the problem
1. Install `qiskit-aer-gpu` from PyPI (or build from source; how it's installed is irrelevant as long as the CUDA support is built) 
2. Run any of the following command to get Aer imported
   - `python -i -c "import qiskit_aer"`
   - `python -i -c "import qiskit.providers.aer"`
   - `python -i -c "from qiskit.providers.aer import AerSimulator"`
   - `python -i -c "import qiskit; print(qiskit.__qiskit_version__)"`
3. While the Python interpreter is idle waiting for input (due to the interactive prompt `-i`), check `nvidia-smi`. On a multi-GPU system, it is clear that the CUDA context is initialized on all GPUs:
```
$ nvidia-smi 
Sat Dec 17 19:05:35 2022       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 530.06       Driver Version: 530.06       CUDA Version: 12.1     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA RTX A6000    On   | 00000000:21:00.0 Off |                  Off |
| 30%   53C    P2    88W / 300W |    267MiB / 49140MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
|   1  NVIDIA RTX 6000...  On   | 00000000:22:00.0 Off |                  Off |
| 30%   58C    P2    82W / 300W |    431MiB / 49140MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
[...]

PR #2012, opened 2023-11-30
Avoid accessing GPUs when using CPU only

doichanj:
### Summary
This is fix for issue #1686 

### Details and comments
Aer queried available methods and devices by testing with simple circuit whether it successfully runs or not everytime when AerSimulator is initialized. So in the initialization phase, Aer accesses GPUs even if users only need to run on CPU.
In this PR, devices are queried to C++ binary if it is built for GPU or not.
And methods are listed in Python side.

Aer accessed all the available GPUs at the initialization phase. In this PR, only GPUs set in `target_gpus` options are accessed when using `device=GPU` option.


hhorii (approved):
Getting available devices in C++ controller looks reasonable.

PR #2023, opened 2024-01-08
Remove usage of ``qiskit.extensions``

Cryoris:
### Summary

The ``qiskit.extensions`` module has been pending deprecation since Qiskit 0.45. It is deprecated for 0.46 and removed in 1.0.

This is blocking Qiskit/qiskit#11488 and Qiskit/qiskit#11498.

### Details and comments

This removal concerns the import location of ``UnitaryGate`` and the usage of the ``ExtensionError``. The first is easily fixed, whereas for the latter I tried using ``ValueError`` or ``TypeError``, which seemed to cover the error meanings. Technically, this is a breaking change and we could introduce an intermediary class that inherits from the deprecated ``ExtensionError`` and the new choice of error. However, since we will soon be changing to 1.0 and we also skipped this in Qiskit Terra (as we thought it highly unlikely that users are actually relying on this error type) it might be fine to just change the error type.

### DIFF
--- a/qiskit_aer/backends/aer_compiler.py
+++ b/qiskit_aer/backends/aer_compiler.py
@@ -24,7 +24,7 @@ import numpy as np
 from qiskit.circuit import QuantumCircuit, Clbit, ClassicalRegister, ParameterExpression
 from qiskit.circuit.classical.expr import Expr, Unary, Binary, Var, Value, ExprVisitor, iter_vars
 from qiskit.circuit.classical.types import Bool, Uint
-from qiskit.extensions import Initialize
+from qiskit.circuit.library import Initialize
 from qiskit.providers.options import Options
 from qiskit.pulse import Schedule, ScheduleBlock
 from qiskit.circuit.controlflow import (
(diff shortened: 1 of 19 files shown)


jakelishman (commented):
Changing the error types is _technically_ not stable - what did we do for any `ExtensionError`s being raised by the same objects during the move on the Terra side?  If we can switch these over to match whatever we did there, that's probably the safest.  If not, maybe let's make it `QiskitError`, since the chances are that _if_ there's anybody attempting to catch these errors, that's what they'll be using.

Cryoris:
Yeah that's what I tried to explain above; we didn't do anything and just changed the type: https://github.com/Qiskit/qiskit/pull/10725#discussion_r1341169002. I can add an intermediary type, though that'll require a bunch of filtering warnings for the 0.46 release and is not compatible with Qiskit/qiskit#11488 since there we remove `ExtensionError` completely, right?

jakelishman:
Oh sorry, the GitHub notification just took me to the diff and I missed that I hadn't already read the top comment of this one.  Yeah, in that case let's just bite the bullet and accept it here too - if we did it on Terra, it'd be tricky to hold Aer to a higher standard.

jakelishman (approved):

jakelishman:
(I'll leave unmerged for the Aer team to look.)

PR #2028, opened 2024-01-12
Support `Store` and storage `Var`

hhorii:
### Summary

Support `Store` and `Var` as storage, which will be released in Qiskit 1.0

### Details and comments

This PR adds support of `Store` and `Var` as storage. This PR requires Qiskit 1.0 and will not be merged until its release.

`AerStore` is an ad-hoc solution for transpilation issue of `Store`. Currently, `Store` is not supported `transpile()` and can be moved to an inappropriate position after transpilation. By using an custom instruction `AerStore`, this PR prevents from moving `Store`.


hhorii:
I found that assignment of `Var` with `Var` fails.

hhorii:
> I found that assignment of `Var` with `Var` fails.

this was resolved.

doichanj (approved):

PR #2029, opened 2024-01-12
Universal CPU Build Support for Qiskit Aer with CUDA and ROCm

basnijholt:
…

This PR introduces an update to the `qiskit_aer` build process, ensuring that the CPU variant of `controller_wrappers` is always built, alongside CUDA and ROCm variants. This ensures that builds targeting CUDA or ROCm can still operate on systems that only have CPU capabilities.

Key Changes:
1. **CPU Build as a Default**: The CPU backend (`controller_wrappers_cpu`) is now always built. This is independent of whether CUDA (`controller_wrappers_cuda`) or ROCm (`controller_wrappers_rocm`) backends are being built.

2. **Conditional CUDA and ROCm Builds**: Compilation of CUDA and ROCm backends remains conditional based on the `AER_THRUST_BACKEND` setting. This ensures that these backends are built only when specified, without affecting the CPU build.

3. **Organized Build Process**: The build process in `CMakeLists.txt` is organized to handle CPU, CUDA, and ROCm backends separately. Each backend has its specific compiler flags, definitions, and dependencies, ensuring a clean and manageable build process.

This update significantly improves the flexibility and applicability of Qiskit Aer, enabling it to run on a broader range of systems, from specialized CUDA and ROCm environments to general-purpose CPU-only machines.

I my case this is useful because we distribute a single environment / container that run both on CPU and GPU machines. ATM `qiskit-aer` is our only dependency that doesn't allow this mode of operation.


qiskit_aer/backends/controller_wrappers.py
```diff
@@ -0,0 +1,84 @@
+from contextlib import suppress
```
basnijholt, line 1: A perhaps less readable but DRYer version is
```python
__all__ = [
    "AerConfig",
    "aer_controller_execute",
    "AerUnaryExpr",
    "AerUnaryOp",
    "AerBinaryExpr",
    "AerBinaryOp",
    "AerUintValue",
    "AerBoolValue",
    "AerUint",
    "AerBool",
    "AerCast",
    "AerVar",
    "AerCircuit",
    "AerConfig",
]

def try_import_backend(backend_module_name):
    try:
        backend_module = __import__(backend_module_name, globals(), locals(), __all__, 0)
        return backend_module
    except ImportError:
        return None

IMPORTED_BACKEND = None
BACKENDS = ['controller_wrappers_cuda', 'controller_wrappers_rocm', 'controller_wrappers_cpu']

for backend in BACKENDS:
    backend_module = try_import_backend('.' + backend)
    if backend_module:
        IMPORTED_BACKEND = backend.split('_')[-1]
        globals().update({name: getattr(backend_module, name) for name in backend_module.__all__})
        break

if IMPORTED_BACKEND is None:
    raise ImportError("No backend found for qiskit-aer.")
```

basnijholt: Changed it to this.

qiskit_aer/backends/controller_wrappers.py
```diff
@@ -0,0 +1,29 @@
...
+    try:
+        return importlib.import_module(module_name, "qiskit_aer.backends")
+    except ImportError:
+        return None
+
+
+IMPORTED_BACKEND = None
+BACKENDS = ["cuda", "rocm", "cpu"]
+
+for backend_suffix in BACKENDS:
```
basnijholt, line 15: One could also add a environment variable to for example force the CPU implementation.

But that can be done in a followup PR.

doichanj:
Can you make this PR to build both CPU and CUDA controller_wrapers at the same time by specifying some parameters to cmake so that we can include both CPU/CUDA binaries in the single wheel package for distribution at deployment?
I'm also considering to do so because we can not distribute GPU package separatery on PyPI, but I can not make both binaries at the same time

basnijholt:
@doichanj, thanks a lot for looking at this PR!

This problem is still very annoying for us because we're maintaining 2 locked environments for CUDA and non-CUDA now. It would be awesome to be able to use one installation that works both on CPU and CUDA.

doichanj:
I could not build with this PR when setting `-DAER_THRUST_BACKEND=CUDA`, it seems CPU build is not correctly configured and failed

PR #2041, opened 2024-01-29
Remove importing qiskit.execute

doichanj:
### Summary
Removed importing qiskit.execute from test

### Details and comments
qiskit.execute was deprecated, but unused import was remained in tests so removed


hhorii (approved):
LGTM

PR #2046, opened 2024-02-02
remove qiskit-terra before installing Qiskit/qiskit.git@main in CI

1ucian0:
Currently the tests against `Qiskit/qiskit.git@main` are failing because a Qiskit 1.0 namespace is tricky. @jakelishman wrote  https://qisk.it/1-0-packaging-migration covering exactly this situation and I'm blindly following his advice.

Exactly this form test probably will change in the future. This just fixes current situation.


jakelishman:
Aer shouldn't have any dependencies that _require_ qiskit-terra, so the failings at the moment should just be because the versions of dependencies, Aer and `qiskit@main` are being resolved separately.  I can push a commit to this branch that simplifies things (i.e. uses an example of the first type of Action I wrote in the packaging guide), but also I can add a test to the unit-test matrix to run Aer against 1.0.0rc1 in regular CI runs too, since Aer should already be 90% of the way there, I think.

jakelishman:
I've replaced this PR with #2047, which I made Luciano a co-author on.

Issue #2021, opened 2024-01-03
Copy results to python data structure in`measure_sampler` is not parallelized


hhorii:
### Informations

- **Qiskit Aer version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?

`measure_sampler` has big sequential part though sampling is parallelized.

https://github.com/Qiskit/qiskit-aer/blob/180a0b6431498998c236927aa53d3324b4819860/src/simulators/circuit_executor.hpp#L1090-L1113

### Steps to reproduce the problem

```
from qiskit_aer import AerSimulator
from qiskit.circuit.library import QuantumVolume
from qiskit import transpile

simulator = AerSimulator()
c = QuantumVolume(20)
c.h(c.qubits[:])
c.measure_all()

c = transpile(c, simulator)

for shots in [10, 100, 1000, 10000, 100000, 1000000, 10000000]:
    res = simulator.run(c, shots=shots).result()
    print("shots=", shots, res.metadata)
```

Most of execution time of `shots==10000000` is serial execution.

### What is the expected behavior?

Most of sampling should be parallelized.

### Suggested solutions

Parallelize the above codes.

PR #2049, opened 2024-02-05
Parallelize sampling measure

doichanj:
### Summary
This PR optimizes performance of sampling measure #2021 by 

- decreasing size of array for counts
- parallelizing loops by OpenMP

### Details and comments

This PR adds new basic class `BitVector` to store bit counts obtained by sampling measure.
Bits counts were stored as array of 64-bits integer previously that wastes large memory space and memory bandwidth.
`BitVector` has a variable length bit string that requires fewer memory.
(`BitVector` can be used for stabilizer method, but not included in this PR)

There were some loops not parallelized before:

- conversion to integer index to bit count
- saving counts in results

are now parallelized by OpenMP.


hhorii (approved):
LGTM. `BitVector` looks very reasonable than `reg_t`, indeed.

hhorii:
Just FYI. The reason why `reg_t` is used is Aer may support higher level than qubit, such as qutrit.

hhorii (approved):
LGTM

PR #2062, opened 2024-02-19
[WIP] Add support > 63 qubits for extended stabilizer

doichanj:
### Summary
This PR breaks limit of 64 qubits of extended stabilizer by using variable length bit string.

### Details and comments

Removed transformed caches to save memory storage for unlimited number of qubits.

I will add some more optimizations for larger number of qubits.


doichanj:
I would like to close this one because something has been broken

PR #2067, opened 2024-02-20
Remove unsupported qiskit-ibmq-provider from docs job

mtreinish:
### Summary

The tox docs job definition was previously installing the qiskit-ibmq-provider which is no longer supported and the project is orphaned. This commit fixes this by removing the unnecessary dependency.


mtreinish:
I integrated this into #2064 to fix all the CI issues stemming from the qiskit 1.0 release in a single PR. So we can close this and just continue on #2064

Issue #2074, opened 2024-02-26
CostBasedFusion aggregate_operations crash


aromanro:
### Informations

- **Qiskit Aer version**: Latest dev main branch sources
- **Python version**: Irrelevant, as I'm using it from a c++ project
- **Operating system**: Windows 11

### What is the current behavior?

During some tests I do, `CostBasedFusion::aggregate_operations` crashes.

### Steps to reproduce the problem

I'm executing some randomly generated circuits, I'm not yet sure how to reproduce it, but I think I could point out to a reason and a solution.

This is the problem code:
```
        double estimated_cost =
            estimate_cost(ops, (uint_t)j,
                          i) // fusion gate from j-th to i-th, and
            + (j == 0 ? 0.0 : costs[j - 1 - fusion_start]); // cost of (j-1)-th
```

With debug, I end up with j being 2, fusion_start also 2 and costs size being 2. This results in a costs[-1], resulting in a crash.

### What is the expected behavior?

Not crashing, obviously :)

### Suggested solutions

Replace the `j == 0` check with `j == fusion_start`.

aromanro:
More info:

I think this commit introduced the issue: https://github.com/Qiskit/qiskit-aer/commit/a5e3aa265ccd0f1b3fafadbe0207c185a6a30406

hhorii:
We believe that #2075 fixes this issue.

PR #2075, opened 2024-02-27
Fix fusion CostBasedFusion::aggregate_operations

doichanj:
### Summary
This is fix for #2074 

### Details and comments
fixes overrun of cost array


aromanro:
If I recall correctly the code, j cannot get lower than fusion_start, so a check for inequality is of no use. == should be a little bit faster. Not that it matters much.

hhorii (dismissed):

hhorii (approved):

PR #2076, opened 2024-02-27
Support `bit_not` for `Bool` type

hhorii:
### Summary

This PR is to support `bit_not` for `Bool` values.

### Details and comments

`bit_not` is defined for `Uint` values. However, in other language, depending hardware, `Bool` values can be applied as `logical_not` (`False` is `0` and `True` is `1`). This PR supports such implicit operations for only "`bit_not`".


jakelishman:
Ed Chen pointed out to me that `bit_xor` (and I would guess `bit_or`) has the same problem - not supported on Terra's pseudo Bool/bit type in Aer.  Can we make `bit_xor`, `bit_or` and `bit_and` work on bit/bools as well?

hhorii:
@jakelishman Thank you for your comments.
I have added a test case to check `bit_xor` works `Bool ^ Bool -> Bool` and `Uint ^ Uint -> Uint` cases (I believe that `Bool ^ Uint` is not allowed in Qiskit). I think current implementation covers these operation. Could you tell what cases Ed Chen wants?

https://github.com/Qiskit/qiskit-aer/pull/2076/files#diff-d43cca4f2db1daf52abcc25296f734a5af2a4f30b6423f9665d33bf217efc936R1180-R1238

doichanj (approved):

PR #2079, opened 2024-03-04
Updated First Tutorial for `AerSimulator`

MonitSharma:
### Summary

Since Qiskit has been updated to a newer version, the basic tutorial to work with `qiskit-aer` was failing for many. So corrected the first tutorial to work with the latest version of Qiskit. Followed the previous tutorial and kept most of the sections same.

### Details and comments

Some definition are copied from the previous tutorial, but the code part is correct and works for the latest version of qiskit. Since in Qiskit slack, and also in stack overflow many people are complaining about the Aer simulator not working, this basic tutorial may help them get started with the updated qiskit.


doichanj (approved):
Thank you, this is helpful for many users

PR #2088, opened 2024-03-26
Add EstimatorV2

ikkoham:


ikkoham:
I simplified because it is confusing. As documented, there is only a normal distribution approximation

qiskit_aer/primitives/estimator_v2.py
```diff
@@ -0,0 +1,146 @@
...
+        for pauli in paulis:
+            circuit.save_expectation_value(
+                Pauli(pauli), qubits=range(circuit.num_qubits), label=pauli
+            )
+        result = self._backend.run(
+            circuit, parameter_binds=[parameter_binds], **self.options.run_options
+        ).result()
+
+        # calculate expectation values (evs) and standard errors (stds)
+        rng = np.random.default_rng(self.options.run_options.get("seed_simulator"))
```
hhorii, line 129: I'm wondering that `rng` should be shared in a ` _run()` call.

qiskit_aer/primitives/estimator_v2.py
```diff
@@ -0,0 +1,146 @@
...
+                raise ValueError(
+                    f"The {i}-th pub has precision less than 0 ({pub.precision}). ",
+                    "But precision should be equal to or larger than 0.",
+                )
+
+    def _run(self, pubs: list[EstimatorPub]) -> PrimitiveResult[PubResult]:
+        return PrimitiveResult([self._run_pub(pub) for pub in pubs])
+
+    def _run_pub(self, pub: EstimatorPub) -> PubResult:
+        circuit = pub.circuit.copy()
```
hhorii, line 104: I think `pub.circuit` should be compiled to `AerCompiler` here to avoid copy. This will be an optimization in next releases.

qiskit_aer/primitives/estimator_v2.py
doichanj: I think implementation of rng is as same as that of qiskit.primitives https://github.com/Qiskit/qiskit/blob/090b2b193d6de88017e26d99c0ac379b85382395/qiskit/primitives/statevector_estimator.py#L142

ikkoham: Yes. This is the point where we decided to make this specification for maintainability when we implemented SamplerV2.

doichanj (approved):

PR #2089, opened 2024-03-28
Remove unnecessary numpy build dependency

wshanks:
numpy C bindings are handled through pybind11 which does not depend on numpy. numpy is otherwise only a runtime dependency of qiskit-aer. This change just saves a download during build.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,8 +6,6 @@ requires = [
   "cmake!=3.17.1,!=3.17.0",
   "ninja",
   "pybind11>2.6",
-  "oldest-supported-numpy; python_version>'3.7' or platform_machine=='aarch64' or platform_python_implementation=='PyPy'",
-  "numpy==1.16.3; python_version<='3.7' and platform_machine!='aarch64' or platform_python_implementation=='PyPy'",
 ]
 build-backend = "setuptools.build_meta"
 


wshanks:
pybind11's most recent release supports numpy 2.0. I was curious if building with that was enough to enable numpy 2.0 support qiskit-aer, but I found it too hard to test completely because not all dependencies have versions that support numpy 2.0 yet (matplotlib, sympy, pandas, even qiskit and rustworkx). I was able to run almost all the tests though and everything that could run passed except some visualization tests that failed because of https://github.com/sympy/sympy/issues/26094 (and some visualization tests failed to run because the version of matplotlib I used did not work with numpy 2.0).

jakelishman:
Will: you might be interested in Qiskit/qiskit#11999 - I was able to do a full test against Numpy 2.0 successfully. Fwiw, since that PR, Scipy released a 1.13.0rc1 that has Numpy 2.0 support, and the other components are pretty easy to build from source to test - I found them all sufficiently compatible with Numpy 2.0 to not cause problems for Terra.

wshanks:
Thanks @jakelishman. I built matplotlib's and sympy's `main` branches locally and was able to see all tests pass with numpy 2.0! I tried install numpy 2.0 into an environment with qiskit-aer 0.13 and see lots of errors, so it seems like a new release built against the new pybind11 should take care of numpy 2.0 support for qiskit-aer.

doichanj (approved):

PR #2098, opened 2024-04-15
Fix noise sampling on shot-branching

doichanj:
### Summary

This PR is fix for noise sampling with shot-branching

### Details and comments

Runtime noise sampling used for shot-branching/batch execution on GPU has issue when noise is applied on labeled gates.

[WIP] There is issue remaining in shot-branching + runtime parameter binding with noises


hhorii (approved):
I understood that main changes of this PR is runtime noise-sampling. An operation is kept and noise operations are injected before and after it (originally, the operation is replaced with a sequence of operations).

PR #2101, opened 2024-04-15
Use sphinx.ext.linkcode for more precise source code links

melechlapson:
### Summary
This is a follow-up to https://github.com/Qiskit/qiskit/pull/11851 and addresses https://github.com/Qiskit/documentation/issues/517 by switching out the sphinx.ext.viewcode for the sphinx.ext.linkcode extension which allows our GitHub links in the documentation to be linked to the specific lines of code, not just the file. This was tested successfully in https://github.com/Qiskit/qiskit_sphinx_theme/pull/589 so now we want to implement it in the qiskit, runtime, and provider repos.

Example links generated in this PR build:

a class definition https://github.com/Qiskit/qiskit-aer/blob/main/qiskit_aer/backends/aer_simulator.py#L40-L1046
a method https://github.com/Qiskit/qiskit-aer/blob/main/qiskit_aer/noise/noise_model.py#L849-L910
a function https://github.com/Qiskit/qiskit-aer/blob/main/qiskit_aer/utils/noise_transformation.py#L141-L232

### Details and comments
Since some references in the API documentation actually refer to methods from the inherited `qiskit` classes, it was necessary to add a check at the end to change the base url if the method was from the `qiskit` class and link ot the qiskit repo.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -31,8 +31,13 @@
 Sphinx documentation builder
 """
 
-import os
 import datetime
+import importlib
+import inspect
+import os
+import re
+from pathlib import Path
+
 # Set env flag so that we can doc functions that may otherwise not be loaded
 # see for example interactive visualizations in qiskit.visualization.
 os.environ['QISKIT_DOCS'] = 'TRUE'
@@ -64,7 +69,7 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.autosummary',
     'sphinx.ext.mathjax',
-    'sphinx.ext.viewcode',
+    "sphinx.ext.linkcode",
     'sphinx.ext.extlinks',
     'jupyter_sphinx',
     'reno.sphinxext',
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -146,3 +150,86 @@
...
+        linespec = ""
+    else:
+        ending_lineno = lineno + len(source) - 1
+        linespec = f"#L{lineno}-L{ending_lineno}"
+
+    github_branch = determine_github_branch()
+    if "qiskit/" in file_name:
+        base_url = "https://github.com/Qiskit/qiskit-aer/tree/"
+    else:
+        base_url = "https://github.com/Qiskit/qiskit/tree/"
```
Eric-Arellano, line 234: Are these flipped incorrectly?

releasenotes/notes/switch_source_links_to_link_directly_to_github-bc554dee4f145c7b.yaml
```diff
@@ -0,0 +1,5 @@
+
+features:
+  - |
+    Source links in the API documentation will now link to the exact lines in the GitHub repository.
+
```
Eric-Arellano, line 5: Imo you can delete this because it doesn't change the code itself

tox.ini
```diff
@@ -10,6 +10,9 @@ setenv =
   LANGUAGE=en_US
   LC_ALL=en_US.utf-8
   SETUPTOOLS_ENABLE_FEATURES=legacy-editable
+passenv=
+  GITHUB_REF_NAME,
```
Eric-Arellano, line 14:
```suggestion
  GITHUB_REF_NAME
```

docs/conf.py
```diff
@@ -146,3 +150,86 @@
...
+
+    try:
+        source, lineno = inspect.getsourcelines(obj)
+    except (OSError, TypeError):
+        linespec = ""
+    else:
+        ending_lineno = lineno + len(source) - 1
+        linespec = f"#L{lineno}-L{ending_lineno}"
+
+    github_branch = determine_github_branch()
```
Eric-Arellano, line 230: You already define GITHUB_BRANCH above

docs/conf.py
```diff
@@ -146,3 +150,86 @@
...
+
+    try:
+        full_file_name = inspect.getsourcefile(obj)
+    except TypeError:
+        return None
+    if full_file_name is None:
+        return None
+    try:
+        relative_file_name = Path(full_file_name).resolve().relative_to(REPO_ROOT)
+        file_name = re.sub(r"\.tox\/.+\/site-packages\/", "", str(relative_file_name))
```
Eric-Arellano, line 218: I think the Tox thing was only happening because of running CI in Azure. So we should be able to simplify this, like Runtime and Provider

```suggestion
        file_name = Path(full_file_name).resolve().relative_to(REPO_ROOT)
```

Eric-Arellano (commented):
Thanks!

tox.ini
melechlapson: no commas needed?

docs/conf.py
melechlapson: yep forgot about that

tox.ini
Eric-Arellano: The commas break things

docs/conf.py
Eric-Arellano: You also can make this code easier to read by defining the variable as `repo_name = "qiskit" if "qiskit/" in file_name else "qiskit-aer", then inline `base_url` into the final `return` rather than it being a variable

tox.ini
melechlapson: ok. FWIW I followed what i did in qiskit_ibm_runtime and there are commas there

docs/conf.py
melechlapson: sure

tox.ini
Eric-Arellano: It looks like Runtime is fine:

https://github.com/Qiskit/qiskit-ibm-runtime/blob/a2a803e37c08746fe5c02138bd6896414ae1e7a9/tox.ini#L13-L15

FYI you would use commas if they were all on the same line. (INI sucks - not a very good config format)

Eric-Arellano (approved):
Thanks! Let's double check the artifact is good in CI once that runs, but otherwise this looks good.

doichanj (approved):

Eric-Arellano:
FYI @doichanj we realized that the PR description and title were misleading - our bad! This change does not help with making things more precise - `viewcode` already took you to the precise source code. The main benefit of this PR is that it takes you to GitHub, rather than the Sphinx-generated copy of your code. The downside is more complex config.

See https://github.com/qiskit-community/qiskit-algorithms/pull/170#issuecomment-2067065449 for discussion.

I don't have a strong opinion. Let us know if you prefer to keep this or revert the change.

Eric-Arellano:
@doichanj Qiskit Algorithms ended up rejecting this change to keep things simple: https://github.com/qiskit-community/qiskit-algorithms/pull/170#issuecomment-2095934597. Would you like us to revert this PR or keep as is?

doichanj:
I think it is OK to keep it as is

PR #2120, opened 2024-05-02
Fix init of EstimatorV2 and SamplerV2

### LINKED ISSUE #2117: Minor issue in EstimatorV2 __init__ - can't specify method in options
### Informations

- **Qiskit Aer version**: 0.14.1
- **Python version**: 3.11
- **Operating system**: mac

Minor issue in EstimatorV2 __init__ - can't specify method in options, 

```python
from qiskit_aer.primitives import EstimatorV2

options = {
    "backend_options": {
        "method" : "matrix_product_state"
    }
}
estimator = EstimatorV2(options=options)

```

Fails with 

```python
--->  self._backend = AerSimulator(method=method, **self.options.backend_options)

TypeError: qiskit_aer.backends.aer_simulator.AerSimulator() got multiple values for keyword argument 'method'
```

See code above

Should be able to specify the method at the options level

### Suggested solutions
Combine the dicts like this

```python
    def __init__(
        self,
        *,
        options: dict | None = None,
    ):
        """
        Args:
            options: The options to control the default precision (``default_precision``),
                the backend options (``backend_options``), and
                the runtime options (``run_options``).
        """
        self._options = Options(**options) if options else Options()
        method = "density_matrix" if "noise_model" in self.options.backend_options else "automatic"
        ops = dict(method=method) 
        ops.update(self.options.backend_options)
        self._backend = AerSimulator(**ops)
```

doichanj:
### Summary

Fixes  #2117 

### Details and comments

This PR fixes init function of EstimatorV2 and SamplerV2 to store `method` option in `self.options.backend_options` if `method` is not specified as input parameter


qiskit_aer/primitives/sampler_v2.py
```diff
@@ -98,6 +98,12 @@ def __init__(
         self._seed = seed
 
         self._options = Options(**options) if options else Options()
+        if "method" not in self.options.backend_options:
+            method = (
+                "density_matrix" if "noise_model" in self.options.backend_options else "automatic"
```
hhorii, line 103: does this mean that noise simulation always use "density_matrix" in sampler?

qiskit_aer/primitives/sampler_v2.py
```diff
@@ -98,6 +98,12 @@ def __init__(
...
+            )
+            ops = dict(method=method)
+            self.options.backend_options.update(ops)
```
hhorii, line 106:
```
self.options.backend_options.update({method: method})
```
may be simpler.

qiskit_aer/primitives/estimator_v2.py
```diff
@@ -70,12 +70,20 @@ def __init__(
...
-        method = "density_matrix" if "noise_model" in self.options.backend_options else "automatic"
-        self._backend = AerSimulator(method=method, **self.options.backend_options)
-
-    def from_backend(self, backend, **options):
-        """use external backend"""
-        self._backend.from_backend(backend, **options)
+        if "method" not in self.options.backend_options:
+            method = (
+                "density_matrix" if "noise_model" in self.options.backend_options else "automatic"
+            )
```
kevinsung, line 76: Why is this line necessary? Shouldn't we just let AerSimulator set the method if it's not specified? It should default to `automatic`, which should be smart enough to choose `density_matrix` if `noise_model` is specified.

doichanj: I removed this so now `method=automatic` is set by default and #2117 is fixed

hhorii (approved):

PR #2130, opened 2024-05-09
fix undefined behaviors

enum-class:
### Summary

fix undefined behaviors, and one minor bug


enum-class:
Hi @doichanj and @hhorii ,
What is the procedure of p.r.'s ?
Could you please help me with it ?

doichanj:
I think we have to use signed integer for OpenMP loop because it causes compilation error on Windows (it fails only on Windows...)

doichanj (approved):

PR #2142, opened 2024-05-17
Remove qiskit aer translation stage

### LINKED ISSUE #2141: Transpile with multiple circuits fails after qiskit 1.1
### Informations

- **Qiskit Aer version**: *
- **Python version**: *
- **Operating system**: *

Transpile with multiple circuits fails in MacOS and Windows. Details are in https://github.com/Qiskit/qiskit/issues/12425

### Suggested solutions

In qiskit 1.1 there is a behavior change in the transpile. In many cases (especially in MacOS and Windows) the transpiler disables multiprocessing feature because of performance reason. In the parallel transpile backend Target information is deep copied via dill, but now all pass manager run shares the internal state. So Target should not be mutated.

mtreinish:
### Summary

This commit removes the qiskit aer translation plugin. This was added as a workaround for a Qiskit/qiskit#11351. This has been fixed in Qiskit since 0.45.2 and is no longer necessary. The mechanism by which the workaround worked was unsound in practice as it was mutating the target and also explicitly using private attributes of the Target. This is causing real issues now as it only worked by assuming the target wasn't shared between passmanagers which is never guaranteed. Similarly the reliance on internal private attributes of the Target class will cause issues in the future when the target internals change (see Qiskit/qiskit#12292). This commit opts to remove the plugin in its entirity as it's no longer necessary and actively causing issues with Qiskit 1.1 and transpiling targeting aer backends with >1 circuit. As it's private internal detail there isn't a release note.

### Details and comments

Fixes Qiskit/qiskit#12425
Fixes #2141


doichanj (approved):
I will check Aer's test cases without plug-in

Issue #2111, opened 2024-04-30
Shot branching tests are unreliable on macOS


wshanks:
### Informations

- **Qiskit Aer version**: 0.14.1
- **Python version**: 3.8-3.12
- **Operating system**: macOS x86_64

### What is the current behavior?

Some of the shot branching tests are failing for macOS, for example [here](https://github.com/Qiskit/qiskit-aer/actions/runs/8888318423/job/24405081563). The tests get run as part of the conda build and two of them failed in three of the five Python versions run:

* `test.terra.backends.test_runtime_parameterization.TestRuntimeParameterization.test_dynamic_circuit_with_shot_branching_1___statevector____CPU__`
* `test.terra.backends.aer_simulator.test_shot_branching.TestShotBranching.test_shot_branching_reset_moving_qubits_1___statevector____CPU__`

The CI run with the failures is [here](https://github.com/conda-forge/qiskit-aer-feedstock/runs/24418450911).

### Steps to reproduce the problem

Run the tests (a few times?) under macOS x86_64.

### What is the expected behavior?

All tests pass reliably.

doichanj:
I'm recognizing this issue is caused by PR #2098 but I have not figure out the source of the issue yet

PR #2143, opened 2024-05-20
Remove test_shot_branching

doichanj:
### Summary

Testing shot_branching option is unstable on MacOS that causes CI fails. #2111 

### Details and comments

I'm still trying to resolve this issue but it takes some more time.
So I would like to remove test case for shot_branching tentatively to merge urgent PRs


test/terra/backends/aer_simulator/test_shot_branching.py
hhorii, line 1: use `@unittest.skip` if this change is temporal.

hhorii (approved):

Issue #2154, opened 2024-05-30
Improve support for Pauli error channels for utility experiment-like simulations


chriseclectic:
### What is the expected behavior?

To improve the efficiency of working with Pauli error channels in operators and noise models, particularly with a focus of large number of qubit simulations on stabilizer or MPS simulators.

I am working on 2 specific changes to Aer at the moment:

1. [x] #2157 Add a `PauliLindbladError` noise error class supported by Aer's noise model and C++ sampling to represent Pauli channels generated by Pauli lindblad dissipators. When these are sparse generators these can be sampled from efficiently for large numbers of qubit errors like is used in PEC and PEA mitigation experiments.
2. [x] #2156 Add a `PauliError` noise error class to more efficiently work with Pauli channels as operators (for composition, tensor product, etc) than the current QuantumError format which represents Pauli channels as QuantumCircuits. This new operator will use a PauliTable internally, but still serialize to the same structure as a pauli channel QuantumError for running on C++ simulators so that they don't need any additional changes.

PR #2155, opened 2024-05-30
Add `BaseQuantumError` class

chriseclectic:
### Summary

This refactors QuantumError to inherit from a `BaseQuantumError` class, and `NoiseModel` to allow adding BaseQuantumErrors rather than only QuantumErrors.

~Depends on #2158~

### Details and comments

This is a pre requisit for adding the optimized PauliLindbladError and PauliError operators proposed in #2154 
On its own this PR should have no visible changes to current users.

### DIFF
--- a/qiskit_aer/noise/noise_model.py
+++ b/qiskit_aer/noise/noise_model.py
@@ -21,9 +21,6 @@ from warnings import warn, catch_warnings, filterwarnings
 import numpy as np
 
 from qiskit.circuit import Instruction, Delay
-from qiskit.circuit import QuantumCircuit
-from qiskit.circuit import Reset
-from qiskit.circuit.library.generalized_gates import PauliGate, UnitaryGate
 from qiskit.providers import QubitProperties
 from qiskit.providers.exceptions import BackendPropertyError
 from qiskit.providers.models import BackendProperties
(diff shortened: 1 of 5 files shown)


qiskit_aer/noise/noise_model.py
```diff
@@ -953,108 +951,6 @@ def to_dict(self, serializable=False):
...
-                if all_gate_qubits is not None:
-                    for gate_qubits in all_gate_qubits:
-                        noise_model.add_readout_error(roerror, gate_qubits, warnings=False)
-                # Add all-qubit readout error
-                else:
-                    noise_model.add_all_qubit_readout_error(roerror, warnings=False)
-            # Invalid error type
-            else:
-                raise NoiseError("Invalid error type: {}".format(error_type))
-        return noise_model
```
doichanj, line 1056: Is it difficult to implement `from_dict` for new `BaseQuantumError` structure?
(I think `from_dict` is not so important, but I wonder if there are some users need this)

doichanj: I see

doichanj (approved):

qiskit_aer/noise/noise_model.py
qci-amos: For what it's worth, this is useful for people who want to run noisy simulations on a server but want to enable clients to pick their noise model. I'll also note that the associated `to_dict` method is still present.

I'm not using this currently and I'm not requesting it be restored. The reason I noticed is because a unit test of mine broke with qiskit-aer 0.15, but I'll just delete my test.

kt474: This removal also broke the unit tests in `qiskit-ibm-runtime` because we need to serialize/deserialize noise models to send to the server.

For now, we can pin the aer version while we deprecate and remove support for this on our side.

PR #2156, opened 2024-05-30
Add optimized `PauliError` quantum error operator class

chriseclectic:
### Summary

- [x] Depends on #2155 
- [x] Add unit tests
- [x] Add release note

This adds an optimized `PauliError` error operator class that uses a PauliTable rather than QuantumCircuits internally to more efficiently represent Pauli channels. 

### Details and comments

It satisfied the BaseOperator group (tensor, compose etc), and can be converted to exisiting QuantumError or QuantumChannel operators (though the quantum channel ones will be inefficient as Pauli channels).


doichanj (approved):
This is clear to me and I could understand by reading test cases

PR #2157, opened 2024-05-30
Pauli lindblad error

chriseclectic:
### Summary

- [x] Depends on #2155 
- [x] Depends on #2156
- [x] Add unit tests
- [x] Add release note

This adds an optimized `PauliLindbladError` error operator class that represents a Pauli channel generated by Pauli-Lindblad dissipator generators.

### Details and comments

This error op can be used to efficiently represent large N-qubit error channels that have sparse generators (like used in the utility paper and PEC / PEA mitigation experiments).

It satisfied the BaseOperator group (tensor, compose etc), and can be converted to existing QuantumError or PauliError channels (though this may be exponentially expensive if the generator has an exponential number of non-zero Pauli error probabilities). 

Adds support to the C++ QuantumError code to sample directly from the PauliLindblad generators by using the fact the generated channels commute and can be sampled from independently and composed to choose a Pauli error term to apply.


doichanj (approved):
I think this is clear to me, and I would like to add some more optimization on C++ side by using this technique

Issue #2121, opened 2024-05-03
EstimatorV2 extended_stabilizer fails" invalid instructions {"instructions": {save_expval}}


zlatko-minev:
### Informations

- **Qiskit Aer version**: 0.14.1
- **Python version**: 3.11
- **Operating system**: Mac

### What is the current behavior?

Can't run extended stabilizer with Estimator, see code:

### Steps to reproduce the problem
```python
from qiskit_aer.primitives import EstimatorV2 
qc = QuantumCircuit(2)
qc.sx(0)
qc.cx(0,1)
obs = ['ZI']

options = {
    "backend_options": {  
        "method": "extended_stabilizer",
    }
}
estimator = EstimatorV2(options=options)
print(estimator._backend)
job = estimator.run([(qc, obs)])

```
Return
```python
AerSimulator('aer_simulator_extended_stabilizer')
Simulation failed and returned the following error message:
ERROR:  [Experiment 0] Circuit circuit-394 contains invalid instructions {"instructions": {save_expval}} for "extended_stabilizer" method.Circuit circuit-394 contains invalid parameters  for "extended_stabilizer" method.
```

### What is the expected behavior?
Should run extended stabilizer with Estimator.

### Suggested solutions
NA

kevinsung:
So strange. When I run the code in @zlatko-minev 's repro, I get a different error. I'm using the same Aer version and Python 0.14.1, but my OS is Fedora Linux 40.

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], [line 13](vscode-notebook-cell:?execution_count=1&line=13)
      [6](vscode-notebook-cell:?execution_count=1&line=6) obs = ['ZI']
      [8](vscode-notebook-cell:?execution_count=1&line=8) options = {
      [9](vscode-notebook-cell:?execution_count=1&line=9)     "backend_options": {  
     [10](vscode-notebook-cell:?execution_count=1&line=10)         "method": "extended_stabilizer",
     [11](vscode-notebook-cell:?execution_count=1&line=11)     }
     [12](vscode-notebook-cell:?execution_count=1&line=12) }
---> [13](vscode-notebook-cell:?execution_count=1&line=13) estimator = EstimatorV2(options=options)
     [14](vscode-notebook-cell:?execution_count=1&line=14) print(estimator._backend)
     [15](vscode-notebook-cell:?execution_count=1&line=15) job = estimator.run([(qc, obs)])

[...]

PR #2161, opened 2024-06-03
[WIP] Add save_expval to extended_stabilizer

doichanj:
### Summary

This PR adds (enables) `save_expval `and `save_expval_var` to `extended_stabilizer` requested by #2121

### Details and comments

`save_expval `and `save_expval_var` in extended_stabilizer were disabled because there were bug in expectation values calculation.
I found 3 issues:

- `save_expval_var` was not included in `State::check_measurement_opt`
- Phase for Y gate was not updated
- The same random number generator was used in the parallel loop in `Runner::norm_estimation`

Now it works for cliffords cases, but it still does not work for non-cliffords cases


aromanro:
I can confirm - as I tried this in a separate repo - that it works on clifford circuits but not for non-clifford. Additionally I had to implement this:
```
double State::expval_pauli(const reg_t &qubits, const std::string &pauli) {
  RngEngine rng;
  rng.set_random_seed();
  return expval_pauli(qubits, pauli, rng)  ;
}
```
as it originally returned 0.

PR #2162, opened 2024-06-04
Fix example

chrisamirani:
### Summary
This doc which is live https://qiskit.github.io/qiskit-aer/getting_started.html has a wrong example code which I am fixing to run correctly with qiskit 1.0

### Details and comments
`qiskit.execute()` is deprecated and instead we should run the job on the simulator directly. The example provide on the get started page is wrong and can cause confusion. I am changing the example so we can run it directly on the aer sim.


doichanj:
Thank you, but I think `backend.run()` will be also deprecated, so could you please use SamplerV2 instead of using `backend.run()`

chrisamirani:
Hi @doichanj do you mean the one in `qiskit_ibm_runtime` ? I thought maybe we wanted to keep introductory examples as simple as possible.

doichanj:
No, I mean the example in readme.md, it uses `SamplerV2.run()` instead of using `AerSimulator.run()`

chrisamirani:
> No, I mean the example in readme.md, it uses `SamplerV2.run()` instead of using `AerSimulator.run()`

Oh yeah that's even better. Done!

chrisamirani:
Hi @doichanj Can we merge this?

doichanj (approved):

Issue #2167, opened 2024-06-06
Missing `cp39-cp39-macosx_11_0_arm64` wheels since `0.14.1`


burgholzer:
### Informations

- **Qiskit Aer version**: Starting from `0.14.1`
- **Python version**: CPython 3.9
- **Operating system**: macOS arm64

### What is the current behavior?

Since the release of `qiskit-aer` `0.14.1` no Python wheels for Python 3.9 on arm64 macOS have been uploaded to PyPI.

### Steps to reproduce the problem

Try to `pip install qiskit-aer` on arm64 macOS under Python 3.9. It will try to build from the sdist instead of downloading a pre-built wheel.

### What is the expected behavior?

Pre-built wheels have been available before `0.14.1`. So if there is no big reason for omitting the 3.9 arm64 macOS wheel, such a wheel should be uploaded to PyPI.

### Suggested solutions

Check the CD PyPI publishing pipeline, why the specific wheel is not produced anymore.

burgholzer:
> #2106 skips Python 3.8 and 3.9 because `macos-latest` has become causing CI fails, is there any better way to support 3.8 and 3.9 for arm64 macos ?

You might have been hit by https://github.com/actions/setup-python/issues/808 back then. However, this situation has been resolved and both, Python 3.8 and 3.9, are now available on the `macos-latest` runners via `setup-python` and work smoothly.

doichanj:
OK I will remove skipping them, thank you

PR #2169, opened 2024-06-07
Restore MacOS Arm64 for Python 3.8 and 3.9

doichanj:
### Summary

This PR restores distributions for Python 3.8 and 3.9 for MacOS on Arm64 #2167

### Details and comments

In `deploy.yml` and `build.yml`, we skipped Python 3.8 and 3.9, and we restore them in this fix

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -158,7 +158,6 @@ jobs:
       - name: Build Wheels
         env:
           CIBW_ARCHS_MACOS: arm64
-          CIBW_SKIP: "pp* cp38* cp39*"
         run: cibuildwheel --output-dir wheelhouse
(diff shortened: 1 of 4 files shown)


burgholzer:
If I read the logs here correctly, the failures you are seeing is because the install tries to build a rather old version of numpy (`numpy<1.17,>=1.16`) for which no arm64 wheels are available. That numpy version is requested because of the `cvxpy<1.1.15,>=1.0.0 (from -r requirements-dev.txt (line 6))` dependency, which only started to ship `arm64` wheels after `v1.2.1` (see https://pypi.org/project/cvxpy/1.2.1/).

Is there any particular reason for the upper bound on the `cvxpy` dependency?

hhorii (approved):

PR #2173, opened 2024-06-12
Fix from backend of V2 primitives to use existing AerSimulator

doichanj:
### Summary

This PR fixes `EstimatorV2.from_backend` and `SamplerV2.from_backend` to use pre-defined AerSimulator for simulation

### Details and comments

With this fix, user can define specific simulator object with backend options instead of passing options for V2 primitives.

for example, following script can use AerSimulator with statevector simulation and with noise model
```
sim = AerSimulator(method="statevector", noise_model=noise_model)
estimator = EstimatorV2.from_backend(sim)
```

This PR fixes `from_backend` that initialized backend object by calling `AerSimulator.from_backend` but  when input backend is AerSimulator just copy it.

### DIFF
--- a/qiskit_aer/primitives/estimator_v2.py
+++ b/qiskit_aer/primitives/estimator_v2.py
@@ -76,7 +76,10 @@ class EstimatorV2(BaseEstimatorV2):
     def from_backend(cls, backend, **options):
         """make new sampler that uses external backend"""
         estimator = cls(**options)
-        estimator._backend = AerSimulator.from_backend(backend)
+        if isinstance(backend, AerSimulator):
+            estimator._backend = backend
+        else:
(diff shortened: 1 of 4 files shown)


qiskit_aer/primitives/estimator_v2.py
```diff
@@ -130,6 +133,7 @@ def _run_pub(self, pub: EstimatorPub) -> PubResult:
         result = self._backend.run(
             circuit, parameter_binds=[parameter_binds], **self.options.run_options
         ).result()
+        print(result)
```
hhorii, line 136: is for debugging?

doichanj: I forgot to delete this one

hhorii (approved):

PR #2179, opened 2024-06-21
Fix legacy accesses of `CircuitInstruction`

jakelishman:
### Summary

Using Terra's `CircuitInstruction` like a tuple is a legacy access pattern.  The pattern of using named access attributes was introduced about two years ago, and the legacy format is about to be deprecated. This fixes up the last couple of uses.

### Details and comments

Terra 1.2 will likely deprecate this.


hhorii (approved):
LGTM.

PR #2187, opened 2024-07-04
Remove standalone and qobj

doichanj:
### Summary

This PR removes standalone and qobj

### Details and comments

standalone and qobj has been deprecated since #2050 and 3 month passes, so we can remove them


hhorii (commented):
I think we need to deprecate DASK support if we remove qobj support. Especially, we may need to deprecate
- executor
- max_job_size
- max_shot_size

Otherwise, `AerCircuit` needs to be serializable.

hhorii (approved):

PR #2196, opened 2024-07-18
Add apply_initialize to AerState

doichanj:
… _CUSTOM_INSTR

### Summary

This is fix for the daily test for latest Qiskit

### Details and comments

This PR adds `apply_initialize `to `AerState `and move initialize from `BASIS_GATES `to `_CUSTOM_INSTR `


hhorii (approved):
Great catch!!

PR #2198, opened 2024-08-02
Rewrite `CircuitInstruction` as iterable to avoid deprecation warning of Qiskit 1.2

t-imamichi:
### Summary

Rewrite some codes to avoid the following deprecation warning of Qiskit 1.2

> DeprecationWarning: Treating CircuitInstruction as an iterable is deprecated legacy behavior since Qiskit 1.2, and will be removed in Qiskit 2.0. Instead, use the operation, qubits and clbits named attributes.


jakelishman (approved):
Thanks for this - I thought we'd caught all of them proactively in #2179 and #2191, but I guess a couple still slipped through.

PR #2199, opened 2024-08-02
Derive from `BaseSamplerV1` rather than `BaseSampler`

garrison:
### Summary

This change is in accordance with the following warning that is triggered by the Qiskit 1.2 rc:

> qiskit_aer/primitives/sampler.py:69: DeprecationWarning: The class ``qiskit.primitives.base.base_sampler.BaseSampler`` is deprecated as of qiskit 1.2. It will be removed no earlier than 3 months after the release date. The `BaseSampler` class is a type alias for the `BaseSamplerV1` interface that has been deprecated in favor of explicitly versioned interface classes. It is recommended to migrate all implementations to use `BaseSamplerV2`. However, for implementations incompatible with `BaseSamplerV2`, `BaseSampler` can be replaced with the explicitly versioned `BaseSamplerV1` class.

### DIFF
--- a/qiskit_aer/primitives/sampler.py
+++ b/qiskit_aer/primitives/sampler.py
@@ -22,14 +22,14 @@ import numpy as np
 from qiskit.circuit import ParameterExpression, QuantumCircuit
 from qiskit.compiler import transpile
 from qiskit.exceptions import QiskitError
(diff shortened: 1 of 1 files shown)


doichanj:
I think we can also deprecate V1 based AerSampler and AerEstimator

doichanj (approved):

Issue #2207, opened 2024-08-20
EstimatorV2 not working properly with noise model and runtime_parameter_bind_enable both activated


arthurostrauss:
### Informations

- **Qiskit Aer version**: 0.15.0
- **Python version**: 3.10
- **Operating system**:  Mac OS Sonoma

### What is the current behavior?

Instantiating a custom noise model with AerSimulator while setting `runtime_parameter_bind_enable` to True and using the EstimatorV2.from_backend() yields inconsistent behaviour

### Steps to reproduce the problem

`from qiskit.primitives import BackendEstimatorV2, StatevectorEstimator
from qiskit_aer.primitives import EstimatorV2
from qiskit.quantum_info import SparsePauliOp
from qiskit_aer.noise import NoiseModel, coherent_unitary_error
from qiskit.circuit import QuantumCircuit, Parameter, ParameterVector
from qiskit_aer import AerSimulator
from qiskit import transpile
import numpy as np
from qiskit.quantum_info import Operator
from qiskit.circuit.library import RXGate

pi = np.pi
phi = pi
phi_in_degree = int(phi * 180 / pi)
custom_rx_label = "x"+ str(phi_in_degree)
gamma = 0.2  # Noise strength
theta = ParameterVector('θ', 1)
circuit_context = QuantumCircuit(1)
circuit_context.rx(phi, 0)
circ = QuantumCircuit(1)
circ.unitary(Operator(circuit_context), [0], label=custom_rx_label)
parametrized_gate = QuantumCircuit(1)
parametrized_gate.rx(pi + theta[0], [0])
circ.append(parametrized_gate.to_instruction(), [0])

observable = SparsePauliOp.from_list([('I', 1.), ('Z', 1.)])

statevector = StatevectorEstimator()
noise_model = NoiseModel(["unitary", 'u', 'h', 's', 'sdg'])
noise_model.add_all_qubit_quantum_error(coherent_unitary_error(Operator(RXGate(gamma*phi))), [custom_rx_label])

backend = AerSimulator(noise_model=noise_model, runtime_parameter_bind_enable=True)
backend_estimator = BackendEstimatorV2(backend = backend)
aer_estimator = EstimatorV2.from_backend(backend)

circ = transpile(circ, backend, optimization_level=0)

pub = [(circ, observable, np.array([[0.6913867 ],
        [0.6824007 ],
        [0.7085619 ],
        [0.72341967],
        [0.72430754],
        [0.72378087],
[...]

PR #2209, opened 2024-08-22
Fix parameter array initialization for runtime parameter binding

doichanj:
### Summary

This is fix for issue #2207 

### Details and comments

Runtime parameter binding had issue when a gate with multiple parameters did not have all the bound parameter but used default parameter values. 
This fix populates default parameter values to parameters in buffer

Also this PR optimizes runtime parameter binding with noise model when running on CPU by using runtime noise sampling technique not to copy circuits on every shots*parameters


hhorii (approved):
Are tests necessary?

hhorii (approved):

PR #2248, opened 2024-10-29
Fixed MPS save failure when memory limit is ignored

### LINKED ISSUE #2246: Aer fails to save MPS if it estimates (incorrectly) that there is insufficient memory
### Informations

- **Qiskit Aer version**: 0.15.1
- **Python version**: 3.11
- **Operating system**: Linux

First of all, I find that #2056 is still a problem even though it is marked as fixed.

Secondly, at https://github.com/Qiskit/qiskit-aer/pull/2059 it was mentioned that I can use `max_memory_mb=-1` to ignore the check of required memory.  However, even when that is passed, the `save_matrix_product_state` instruction will fail to create an MPS if the error would have been triggered.  Here's a full example:

```python
from qiskit.circuit.library import PauliEvolutionGate
from qiskit.synthesis import SuzukiTrotter

L = 100
hamiltonian = SparsePauliOp.from_sparse_list([], num_qubits=L)
for x in range(L - 1):
    hamiltonian += SparsePauliOp.from_sparse_list(
        [
            ("XX", (x, x + 1), 1),
            ("YY", (x, x + 1), 1),
        ],
        num_qubits=L,
    )
sim = AerSimulator(
    method="matrix_product_state",
    matrix_product_state_max_bond_dimension=10,
    max_memory_mb=-1,
)
qc = QuantumCircuit(L)
qc.append(
    PauliEvolutionGate(hamiltonian, synthesis=SuzukiTrotter(reps=100)), qargs=qc.qubits
)
qc = qc.decompose()
qc.save_matrix_product_state("mps")
sim.run(qc).result().data(0)["mps"]
```

fails with

```
  File "/var/home/garrison/Qiskit/aer-mps-fail.py", line 19, in <module>
    sim.run(qc).result().data(0)["mps"]
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
KeyError: 'mps'
```

However, if you change it to `L = 10`, it works (at least on my system; these numbers likely need to be tweaked slightly elsewhere).

Code example above

It should save the matrix product state, especially if given `max_memory_mb=-1`.  And ideally the memory check will be fixed so that passing `max_memory_mb=-1` is not necessary, especially when `matrix_product_state_max_bond_dimension` already provides an upper bound on the memory that can be used.

### Suggested solutions

I have not examined the code yet but may soon.

gadial:
### Summary
Fixes a bug preventing MPS runs when the estimated memory usage is too big, even if memory limit is disabled.

Fixes #2246 

### Details and comments
When `max_memory_mb=-1` the maximum memory amount check for MPS should be disabled; however, `Executor<state_t>::get_max_parallel_shots` still uses the estimated memory in its computation, which might result in the value `0`, disabling the run of the engine altogether. To prevent this, `get_max_parallel_shots` results to the `mem==0` case if the maximum memory amount check is disabled.

### DIFF
--- a/releasenotes/notes/fix_mps_save_without_memory_limit-64d927c4b4e1e2d4.yaml
+++ b/releasenotes/notes/fix_mps_save_without_memory_limit-64d927c4b4e1e2d4.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes a bug preventing the MPS simulator from running when its estimate
+    for the required memory is too large, even when `max_memory_mb=-1`.
--- a/src/simulators/circuit_executor.hpp
+++ b/src/simulators/circuit_executor.hpp
@@ -428,7 +428,7 @@ uint_t Executor<state_t>::get_max_parallel_shots(
     const Config &config, const Circuit &circ,
     const Noise::NoiseModel &noise) const {
   uint_t mem = required_memory_mb(config, circ, noise);
-  if (mem == 0)
+  if (mem == 0 || !check_required_memory_)
     return circ.shots * circ.num_bind_params;
 
   if (sim_device_ == Device::GPU && num_gpus_ > 0) {
--- a/test/terra/backends/aer_simulator/test_save_matrix_product_state.py
+++ b/test/terra/backends/aer_simulator/test_save_matrix_product_state.py
@@ -16,6 +16,9 @@ from ddt import ddt
 import math
 import numpy as np
 from qiskit import QuantumCircuit, transpile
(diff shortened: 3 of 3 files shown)


gadial:
Should we use this PR to fix the MPS memory usage estimation ([here](https://github.com/Qiskit/qiskit-aer/blob/9899a2946dd8ec7feefb7b82cac652537e294849/src/simulators/matrix_product_state/matrix_product_state_size_estimator.hpp#L68))? I am currently not familiar enough with MPS to see if there's an obvious bug leading to a wrong memory estimation.

PR #2265, opened 2024-11-24
Fix for save_expval truncation

### LINKED ISSUE #2249: AerEstimatorV2 fails to execute small circuits transpiled against larger backends
### Informations

- **Qiskit Aer version**: 0.15.1
- **Python version**:
- **Operating system**:

AerEstimatorV2 gives memory error when trying to execute small circuits that have been transpiled against larger backends.  Essentially it is the same issue as #2084 

Try this circuit and operator transpiled against an Eagle machine

```python

N = 12
qc = QuantumCircuit(N)

qc.x(range(N))
qc.h(range(N))

for kk in range(N // 2, 0, -1):
    qc.ch(kk, kk - 1)
for kk in range(N // 2, N - 1):
    qc.ch(kk, kk + 1)

op = SparsePauliOp('Z'*12)

```

I currently get

QiskitError: 'ERROR:  [Experiment 0] Insufficient memory to run circuit circuit-205 using the statevector simulator. Required memory: 18446744073709551615M, max memory: 127903M ,  ERROR: Insufficient memory to run circuit circuit-205 using the statevector simulator. Required memory: 18446744073709551615M, max memory: 127903M'

But the active number of qubits is only 12

The simulation should work if the number of active qubits is small enough to fit in memory

gadial:
### Summary
Fixes small problems preventing the usage of `save_expval` truncation.

Fixes #2249 

### Details and comments
PR #2216 added a mechanism for truncating `save_expval`, removing qubits for which the pauli measurement operator is `I`, since those qubits are not needed for the simulation of the measurement. This is crucial for running the simulator on small circuits compiled on large backends, since the `save_expval` is extended to all the qubits in the backend, meaning we might have a 127-qubit `save_expval` for a circuit with only 12 active qubits (e.g. in the added test case).

However, the truncation code was not activated in practice due to the `!qubitset_.empty()` check performed prior to the execution of the code which excludes `I` qubits. Since the `save_expval` command is the last one in the circuit when, e.g., running via an estimator (which adds the command to the circuit just before its run) and gates are traversed in reverse order at this stage, the `save_expval` command would always fail this test, defaulting to the `qubitset_.insert(op.qubits.begin(), op.qubits.end());` line which adds all its qubits.

This PR removes the `!qubitset_.empty()` check and fixes some problems arising from this:
1. In edge cases, e.g. a circuit consisting only of a measurement of the pauli op `I` the truncation stage results in a circuit without qubits, and so it is skipped and no expval result it obtained. Hence, we always add the a qubit to the qubit set if it is empty, even if it corresponds to `I`. This does not enlarge the qubit set enough to pose a real problem, although smarter logics can be implemented.
2. A `op.qubits.resize(qubitmap_.size());` command was missing in the remapping code (although present in the non-remapping truncation code in `set_params`); it was added. Without resizing, the `save_expval` command fails due to generation of invalid pauli mask data in the `expval_pauli` method.
3. When truncating the pauli string, the index of each pauli in the string is computed from the qubit index. However, in cases where the operator is only on a subset of the qubits, the qubit index might be too large. For example, if the pauli `ZZ` is used on qubits 1,2 and qubit 0 is truncated away, while creating the new pauli the code will attempt to access the pauli in index 2 (more precisely, in location `pauli.size()-1-2` since paulis are accessed in reverse order). This is fixed by using the index of the qubits inside `op.qubits` instead of their absolute index.


hhorii (dismissed):
LGTM

.github/workflows/build.yml
```diff
@@ -54,69 +54,3 @@ jobs:
...
-          CIBW_SKIP: "*-manylinux_i686 pp* cp36* cp37* cp38* cp39* cp310* cp311* *musllinux*"
-          CIBW_ENVIRONMENT: QISKIT_AER_PACKAGE_NAME=qiskit-aer-gpu-cu11 QISKIT_AER_CUDA_MAJOR=11 CMAKE_VERBOSE_MAKEFILE=true AER_THRUST_BACKEND=CUDA CUDACXX=/usr/local/cuda/bin/nvcc AER_CUDA_ARCH="7.0 7.2 7.5 8.0 8.6 8.7" AER_PYTHON_CUDA_ROOT=/opt/_internal AER_CIBUILD=true
-          CIBW_REPAIR_WHEEL_COMMAND: 'auditwheel repair --exclude libcudart.so.11.0 --exclude libcustatevec.so.1 --exclude libcutensornet.so.2 --exclude libcutensor.so.1 --exclude libcutensorMg.so.1 --exclude libcusolver.so.11 --exclude libcusolverMg.so.11 --exclude libcusparse.so.11 --exclude libcublas.so.11 --exclude libcublasLt.so.11 -w {dest_dir} {wheel}'
-        run: |
-          python -m cibuildwheel --output-dir wheelhouse
-      - uses: actions/upload-artifact@v4
-        with:
-          path: ./wheelhouse/*.whl
-          name: artifact-${{ github.job }}
-  wheel-gpu:
```
doichanj, line 90: Why did you remove GPU build wheel ?
I think we still need them for testing build

gadial: It's a temporary measure since those builds are failing (probably due to CUDA update) and this prevents us from merging this fix.

doichanj (approved):

PR #2273, opened 2024-12-10
Fix GPU build failure

doichanj:
### Details and comments

Limiting version of cuQUantum < 24.11.0 because the latest version contains wrong GLIBC version


doichanj:
This is merged to #2265 so I close this one

PR #2285, opened 2025-01-14
Adds unique name for upload-artifact jobs

gadial:
### Summary
This PR fixes an error in the github action for `deploy` preventing wheels from being uploaded.

### Details and comments
When using `actions/upload-artifact@v4` a unique name should be given for separate jobs, otherwise the name clash results in `Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run`.

This PR adds a unique name to the action.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./wheelhouse/*.whl
-          name: ${{ github.job }}-${{ matrix.os }}
+          name: aer-deploy-wheel-build-${{ matrix.os }}
   build_wheels_aarch64:
     name: Build wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
@@ -60,7 +60,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./wheelhouse/*.whl
-          name: ${{ github.job }}-${{ matrix.os }}
+          name: aer-deploy-build_wheels_aarch64-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
@@ -88,7 +88,7 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           path: ./wheelhouse/*.whl
-          name: publish-shared-wheels
+          name: aer-deploy-wheel-arm64-macos-${{ matrix.os }}
   sdist:
     name: Publish qiskit-aer sdist
     runs-on: ubuntu-latest
@@ -111,7 +111,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./dist/qiskit*
-          name: ${{ github.job }}-${{ matrix.os }}
+          name: aer-deploy-sdist-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
@@ -156,7 +156,7 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           path: ./wheelhouse/*.whl
-          name: ${{ github.job }}-${{ matrix.os }}
+          name: aer-deploy-gpu-build-cuda11-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
@@ -201,7 +201,7 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           path: ./wheelhouse/*.whl
-          name: ${{ github.job }}-${{ matrix.os }}
+          name: aer-deploy-gpu-build-cuda12-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
@@ -238,7 +238,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./wheelhouse/*.whl
-          name: ${{ github.job }}-${{ matrix.os }}
+          name: aer-deploy-build_wheels_s390x-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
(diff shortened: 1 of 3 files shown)


jakelishman (commented):
You'll need to make sure that the corresponding `download-wheels` action is also updated, and likely use a prefix that you can reliably know, to glob against.  I think `github.job` will be different for each of these runs, so you might struggle to write a usable glob.

(e.g. `name: dist-${{ 'whatever' }}-${{ 'whatever' }}` in all uploads, then
```yaml
- uses: actions/download-artifact@v4
  with:
    - pattern: 'dist-*'
    - merge-multiple: true
```

.github/workflows/deploy.yml
```diff
@@ -60,6 +60,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./wheelhouse/*.whl
+          name: aer-deploy-build_wheels_aarch64-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
```
jakelishman, line 65: Strictly, this set of artifacts (and a couple more noted below) gets uploaded in the line right below this one, so it doesn't need to be pulled down by the `download-artifact` run in the `publish_shared_wheels` job.  Perhaps we could have the prefix be `deploy-shared-`, then jobs that previously uploaded to the `publish-shared-wheels` artifact name can become (e.g.) `deploy-shared-x86_64-${{ matrix.os }}` and `deploy-shared-arm64-macos`, while the others become `deploy-separate-aarch64-linux` etc?  So then the glob is on `deploy-shared-*`, which won't attempt to re-download the wheels that aren't marked as job dependencies.

.github/workflows/deploy.yml
```diff
@@ -87,7 +88,7 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           path: ./wheelhouse/*.whl
-          name: publish-shared-wheels
+          name: aer-deploy-wheel-arm64-macos-${{ matrix.os }}
```
jakelishman, line 91: This is still `upload-artifact@v3` and needs bumping or the `download-artifact@v4` won't be able to find it.

This is the only other `upload-artifact` run (along with the top comment) that gets downloaded by the shared uploader - all the others below it in this file might not want to match the prefix exactly.

.github/workflows/deploy.yml
```diff
@@ -110,6 +111,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./dist/qiskit*
+          name: aer-deploy-sdist-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
```
jakelishman, line 116: Deployed right after - maybe give a `aer-deploy-separate-` prefix to avoid being pulled by the shared uploader?

.github/workflows/deploy.yml
```diff
@@ -154,6 +156,7 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           path: ./wheelhouse/*.whl
+          name: aer-deploy-gpu-build-cuda11-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
```
jakelishman, line 161: Deployed here.  Also, it's still `upload-artifact@v3.`

.github/workflows/deploy.yml
```diff
@@ -198,6 +201,7 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           path: ./wheelhouse/*.whl
+          name: aer-deploy-gpu-build-cuda12-${{ matrix.os }}
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
```
jakelishman, line 206: Deployed here.  Also, it's `upload-artifact@v3.`

jakelishman (commented):
The top comment explains a little more, but a final comment (from my perspective): technically, it's possible to set
```yaml
with:
  skip-existing: true
```
in the config of the PyPI upload step, which would mean that the race condition on the separate wheels accidentally getting pulled into the shared wheels upload wouldn't matter - anything already uploaded would just be silently skipped.  That's the historic behaviour of `twine`.

That said, it's probably best just to fix the prefixes to not need to bother - it's easy enough for us to just avoid the race condition entirely, and if we ever need to retrigger the job to rebuild, we can just modify it on-the-fly (since we'd have to be modifying the script anyway).

.github/workflows/deploy.yml
gadial: I agree. I changed this to "separate" in d9d7fb78 and set the upload prefix to handle "shared".

.github/workflows/deploy.yml
gadial: Bumped to v4 in d9d7fb78 and changed the name to include "shared". I was hesitant to change to v4 in my original PR since I am not aware of the reasons to the original bump to v4 and why only it affected only part of the file.

.github/workflows/deploy.yml
gadial: Added the prefix as you suggested in d9d7fb78

.github/workflows/deploy.yml
gadial: Bumped to v4 and added "separate" in d9d7fb78

.github/workflows/deploy.yml
gadial: Bumped to 4 and added "separate" in d9d7fb78

.github/workflows/deploy.yml
jakelishman: The change from v3 to v4 was a huge breaking change in the upload/download-artifact jobs.  You certainly always have to match (v4 download can't find v3 uploads, and so on), and I think GitHub will be pushing to deprecate and remove the old v3 system soon too - iirc, it was completely piggy-backed onto their logging framework, and never properly designed as an artifact store.

jakelishman (approved):
From what I can see, this looks right now, though debugging changes to CI/CD pipelines is always a bit slow, and it's unfortunate that Aer doesn't have a way of emulating "dry run" deployments to test them, like Terra does.

Thanks for the changes, Gadi.

PR #2287, opened 2025-01-15
Add Python3.13 support and remove Python 3.8 support

gadial:
### Summary
This PR adds Python 3.13 support for building wheels and removes Python 3.8 which was dropped by qiskit version 1.3.

### Details and comments
The version of cibuildwheel was updated to the latest 2.22 to ensure support for Python 3.13.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,7 +3,7 @@ cmake!=3.17.1,!=3.17.0
 conan<2.0.0
 scikit-build>=0.11.0
 asv
-cvxpy<1.5.0
+cvxpy<1.6.0
 pylint
 black[jupyter]~=24.1
 clang-format~=15.0.7
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,15 +15,15 @@ jobs:
         os: ["macos-13", "ubuntu-latest", "windows-2019"]
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Python Python '3.10'
+      - name: Set up Python Python '3.13'
         uses: actions/setup-python@v5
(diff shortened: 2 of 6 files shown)


jakelishman (commented):
This seems sensible on the surface of it - I guess let's watch CI.  This ought to come with a release note, since it's a public-facing change to Aer.

requirements-dev.txt
```diff
@@ -3,7 +3,7 @@ cmake!=3.17.1,!=3.17.0
 conan<2.0.0
 scikit-build>=0.11.0
 asv
-cvxpy<1.5.0
+cvxpy<1.6.0
```
jakelishman, line 6: cvxpy 1.5.4 also doesn't have Python 3.13 wheels (though it's just a test dependency, not a build one).  If 1.5.4 works, is there any chance that 1.6.0 works too, and we can just completely remove the upper bound?

jakelishman (commented):
One minor question.  I think we should have a release note for this to comment that Aer now supports Python 3.13 - it's of interest to users.

When we're ready to merge, we can modify the branch-protection rules to allow merging with the new CI setup, since GitHub's still currently waiting for 3.8, 3.12.0, which don't run now.

requirements-dev.txt
gadial: I'm reluctant to remove upper bounds in a project like Aer which is in maintenance mode, so the main pitfalls to handle are incompatible new version of libraries.

jakelishman: Yeah, that seems like a good idea for the dev requirement.

jakelishman (approved):
Thanks Gadi

Issue #2291, opened 2025-01-16
DiagonalFusion::aggregate_operations can crash in some circumstances


aromanro:
### Informations

- **Qiskit Aer version**: Latest development version
- **Python version**: Not relevant as qiskit aer is used from C++
- **Operating system**: Windows 11

### What is the current behavior?

I was testing the MPS simulator with some randomly generated circuits... with some I've got crashes.
The call chain (a part of it) of this reported crash is `AerState::apply_measure` -> `AerState::flush_ops` -> `AerState::transpile_ops` -> `Fusion::optimize_circuit` -> ... -> `DiagonalFusion::get_next_diagonal_end` -> `DiagonalFusion::is_diagonal_op`

The last call is with an invalid op. The problem occurs because `DiagonalFusion::get_next_diagonal_end` gets an invalid `from` parameter.
The `ops` list had 102 size and `next_diagonal_start` was 102.

### Steps to reproduce the problem

This would be probably hard to reproduce, I was generating some random circuits with more than 100 ops to test some things and maybe one in 100 circuits crashed (it's just an order of magnitude guesstimation), probably related with this (I also found this https://github.com/Qiskit/qiskit-aer/issues/2286 during the tests).

### What is the expected behavior?

No crash.

### Suggested solutions

`DiagonalFusion::get_next_diagonal_end` https://github.com/Qiskit/qiskit-aer/blob/582407a3f8ef5070c850fb4c417c53c1db1f5998/src/transpile/fusion.hpp#L578 could have this check added right at the beginning:

```
  if (ops.size() <= from)
    return -1;
```

PR #2299, opened 2025-01-25
Fix for https://github.com/Qiskit/qiskit-aer/issues/2291

aromanro:
### Summary

`DiagonalFusion::aggregate_operations` could crash in some circumstances

### Details and comments

Sometimes it could get an invalid `from` parameter. Now it's modified to check it and return -1 if it's not valid.
See https://github.com/Qiskit/qiskit-aer/issues/2291 for more information.


gadial (approved):
LGTM. Would be even better if we had a small test to capture the crash.

Issue #2292, opened 2025-01-16
MPS::apply_multi_qubit_gate can crash in some circumstances


aromanro:
### Informations

- **Qiskit Aer version**: Latest development version
- **Python version**: Not relevant as qiskit aer was used from C++ when noticing it
- **Operating system**: Windows 11

### What is the current behavior?

I was testing the MPS simulator with some randomly generated circuits... with some I've got crashes.
This one I noticed after fixing https://github.com/Qiskit/qiskit-aer/issues/2291 

The call chain was: `AerState::apply_measure` -> `AerState::flush_ops` -> `Base::apply_ops` -> `MatrixProductState::State::apply_op` -> `MPS::apply_diagonal_matrix` -> `MPS::apply_matrix` -> `MPS::apply_multi_qubit_gate`.

It crashed on this line: https://github.com/Qiskit/qiskit-aer/blob/582407a3f8ef5070c850fb4c417c53c1db1f5998/src/simulators/matrix_product_state/matrix_product_state_internal.cpp#L798

`is_diagonal` was `true` (as it should be by looking at the call chain), `mat` had 1 row and 8 columns (being a diagonal matrix, only the diagonal is given, so it's actually a vector)... but at that line it's used with `mat(row, col)` which is generating the crash.

### Steps to reproduce the problem

Again, this might be quite tough to reproduce, I was generating some random circuits with more than 100 ops to test some things and maybe one in 100 circuits crashed.

### What is the expected behavior?

No crash.

### Suggested solutions

The problem line should probably be replaced by something like:
```
new_mat(new_vec[row], new_vec[col]) = is_diagonal ? 0. : mat(row, col);
```

aromanro:
It seems that the code that follows expects the matrix to be in a single row format if `is_diagonal` is true, so probably a better fix would be:

```
  cmatrix_t new_mat(is_diagonal ? 1 : sidelen, sidelen);

  if (is_diagonal) {
    for (uint_t col = 0; col < sidelen; ++col) {
      new_mat(0, new_vec[col]) =
          mat.GetRows() == 1 ? mat(0, col) : mat(col, col); // this is just in case something passes a matrix
                                                            // instead of a vector even if is_diagonal is true
    }
  } else {
    for (uint_t col = 0; col < sidelen; ++col) {
      for (uint_t row = 0; row < sidelen; ++row) {
        if (row == col)
          new_mat(new_vec[row], new_vec[row]) = mat(row, row);
        else
[...]

PR #2300, opened 2025-01-25
Fix for https://github.com/Qiskit/qiskit-aer/issues/2292

aromanro:
### Summary

`MPS::apply_multi_qubit_gate` could crash in some circumstances

### Details and comments

If the `is_dagonal` parameter was `true` and the passed matrix was a vector, the passed vector was accessed as if it would be a square matrix, resulting in a crash. The code is now changed to avoid that and to construct a vector for the case of `is_diagonal` set to `true`.
For more informations see: https://github.com/Qiskit/qiskit-aer/issues/2292


gadial (approved):
Looks good, but it would have been even better if we could add a test with a small matrix that runs quickly and reproduces the crash.

PR #2301, opened 2025-01-25
Adjust QuantumError.to_dict() to handle lack of Instruction.condition

mtreinish:
### Summary

In Qiskit 2.0 the Instruction.condition attribute will be removed following it's deprecation in 1.3.0. This attribute has been superseded by the IfElseOp which covers the use case but offers more functionality. The removal is pending in Qiskit/qiskit#13506 and the use of qiskit-aer in Qiskit's tests is blocking progress on that PR. This commit makes the usage of .condition optional to maintain compatibility with Qiskit<2.0 but also work in >=2.0 with the attribute removed.


jakelishman (approved):
This looks straightforward enough thanks.

PR #2308, opened 2025-02-04
fixes #2042

### LINKED ISSUE #2042: estimator add a register named 'c' to the circuit which may clash with existing classical register
This line in the estimator should make sure to add a classical register which does not already exist in the circuit

https://github.com/Qiskit/qiskit-aer/blob/6f375e34183e0f5347b62c5eff0ddaddd437bbcc/qiskit_aer/primitives/estimator.py#L364

SyedMustafaAhmad:
Greetings,

This is my first issue.

### Summary
Fixed issue #2042 by adding a check for estimator adding duplicate classical registers in _combine_circs.

### Details and comments

I added a check to _combinecircs in qiskit_aer > primitives > estimator.py to prevent a clash of register names for estimator which happens if the same name is used e.g. 'c'.

### DIFF
--- a/qiskit_aer/primitives/estimator.py
+++ b/qiskit_aer/primitives/estimator.py
@@ -371,8 +371,11 @@ class Estimator(BaseEstimator):
         circs = []
         for meas_circuit in meas_circuits:
             new_circ = circuit.copy()
+            # Track existing classical register names in new_circ
+            existing_creg_names = {creg.name for creg in new_circ.cregs}
             for creg in meas_circuit.cregs:
-                new_circ.add_register(creg)
+                if creg.name not in existing_creg_names: # Prevent duplicate registers
+                    new_circ.add_register(creg)
             new_circ.compose(meas_circuit, inplace=True)
             _update_metadata(new_circ, meas_circuit.metadata)
             circs.append(new_circ)


qiskit_aer/primitives/estimator.py
```diff
@@ -371,8 +371,11 @@ def _combine_circs(circuit: QuantumCircuit, meas_circuits: list[QuantumCircuit])
         circs = []
         for meas_circuit in meas_circuits:
             new_circ = circuit.copy()
+            # Track existing classical register names in new_circ
+            existing_creg_names = {creg.name for creg in new_circ.cregs}
             for creg in meas_circuit.cregs:
-                new_circ.add_register(creg)
+                if creg.name not in existing_creg_names: # Prevent duplicate registers
+                    new_circ.add_register(creg)
```
gadial, line 378: This is not a good solution since we silently avoid adding some of the registers, leading to unexpected behavior for the user. We should either raise an error with clear description of the name clash or at least output a warning with the same description and rename the measurement register name (I prefer the first solution as I believe it lowers the chance for the user to create code with unexpected results).

gadial (changes requested):
Thank you for taking this on yourself. I added a comment on my preferred solution to the bug.

qiskit_aer/primitives/estimator.py
SyedMustafaAhmad: Thank you for the comment and since this is my first contribution I would highly appreciate it if I could get any feedback. I have written two solutions to this (with different approaches). Could you tell me which one is better to use in this repository?

# 1. Raising the Error
```python
for meas_circuit in meas_circuits:
    new_circ = circuit.copy()
    existing_creg_names = {creg.name for creg in new_circ.cregs}
    for creg in meas_circuit.cregs:
        # If there's a register name clash, raise a clear error
        if creg.name in existing_creg_names:
            raise QiskitError(
                f"Cannot add measurement register '{creg.name}' because its name "
                "already exists in the circuit. Please rename or remove duplicates."
            )
        new_circ.add_register(creg)
```

# 2. Auto-renaming with a Suffix
```python
for meas_circuit in meas_circuits:
    new_circ = circuit.copy()
    existing_creg_names = {creg.name for creg in new_circ.cregs}
    for creg in meas_circuit.cregs:
        new_name = creg.name
        # If there's a register name clash, rename and warn
        if new_name in existing_creg_names:
            warnings.warn(
                f"Register name '{new_name}' already exists. Automatically "
                f"renaming measurement register to avoid conflicts.",
                UserWarning,
            )
            suffix = 1
            while new_name in existing_creg_names:
                new_name = f"{creg.name}_{suffix}"
                suffix += 1
        new_circ.add_register(ClassicalRegister(creg.size, new_name))
```

PR #2313, opened 2025-02-06
Version update to 0.16.1

gadial:


gadial:
> Do you want to update: https://github.com/Qiskit/qiskit-aer/blob/main/docs/conf.py#L50 too?

Good catch, done.

mtreinish (approved):

Issue #2305, opened 2025-02-02
Code that never executes in MPS::sample_measure_single_qubit


aromanro:
### Informations

- **Qiskit Aer version**: Latest development version
- **Python version**: Doesn't matter, it's in the C++ source
- **Operating system**: All

### What is the current behavior?

There is some code in `src/simulators/matrix_product_state/matrix_product_state_internal.cpp' in `MPS::sample_measure_single_qubit` that never executes, more precisely, here: https://github.com/Qiskit/qiskit-aer/blob/1e86d65d9a6c70c7715ee42cad7861cfeeecf522/src/simulators/matrix_product_state/matrix_product_state_internal.cpp#L1686

### Steps to reproduce the problem

I was looking over the sources to learn more about the qiskit aer internals when I've spotted this. Might be harmless, but it looks suspicious, so I labelled it as a bug.

### What is the expected behavior?

I need to look more carefully over the code and perhaps over the referred paper (and also over 'Andrew J. Ferris and Guifre Vidal, “Perfect sampling with unitary tensor networks,” Phys. Rev. B 85, 165146 (2012)') to be sure...

### Suggested solutions

Either the 
```
    if (qubit != 0) // multiply mat by left lambda
      for (uint_t col = 0; col < mat.GetColumns(); col++)
        for (uint_t row = 0; row < mat.GetRows(); row++)
          mat(row, col) *= lambda_reg_[qubit - 1][row];
```
code should not be there or more probably, it should not be there but the 
```
      for (uint_t col = 0; col < mat.GetColumns(); col++)
        for (uint_t row = 0; row < mat.GetRows(); row++)
          mat(row, col) *= lambda_reg_[qubit - 1][row];
```
should be in the `else` branch.

aromanro:
Without yet looking over the referred article(s), it looks like the multiplication by left lambda should not be done, all lambdas are multiplied along into `mat' by the code that follows and multiplies the 'right lambda'.

This does not seem to be a bug, but simply a small enhancement by removing the useless code. Related with this, the `shots` parameter for the function is useless and it should be removed.
Also I think that the documentation states that the complexity is O(1) for a shot, which is not true as the matrices are multiplied together for each shot.

gadial:
[...]

PR #2314, opened 2025-02-08
Removed some code that is never executed and an unused parameter

aromanro:
### Summary

Removed the `shots` parameter for `MPS::sample_measure` as it wasn't used and a useless check and unreachable code in the same function.

### Details and comments

Related issue: https://github.com/Qiskit/qiskit-aer/issues/2305


releasenotes/notes/removed-useless-code-MPS-simulator-98ca4717e0612ac2.yaml
```diff
@@ -0,0 +1,14 @@
+---
```
gadial, line 1: I don't think release notes are needed here, as this change is not user-facing.

gadial (commented):
Looks good to me. My only comment is that I don't think we need the release notes.

releasenotes/notes/removed-useless-code-MPS-simulator-98ca4717e0612ac2.yaml
aromanro: I removed them.

gadial (approved):
Thank you.

PR #2326, opened 2025-03-05
Fix multi-controlled gates in name mapping

AlbertJP:
### Summary

`MCSXGate`, `MCYGate`, `MCZGate`, `MCRXGate`, `MCRYGate`, `MCRZGate`, `MCU3Gate`, `MCUGate`, and `MCSwapGate` can not be instantiated due to a mismatch in the number of qubits with the base gate:
```
CircuitError: 'The number of control qubits must be in `[1, num_qubits - base_gate.num_qubits]`.'
```

### Details and comments

I came across this issue while writing a unit test for Qiskit/qiskit#13952.

Commit https://github.com/doichanj/qiskit-aer/commit/4587384b63e894a60dc57fbd790b189c6be2d709 (pull #1995) changed the base gates of these gates from non-controlled gates into singly controlled gates, which means there is one qubit too much in the base gate.

An alternative solution would be to reduce `num_ctrl_qubits` by one when calling `ControlledGate.__init__()`, but then the multi-controlled gates can only be created for two or more control qubits and one would need to extract the last qubit from the control state to pass it to the base gate.

### DIFF
--- a/qiskit_aer/backends/name_mapping.py
+++ b/qiskit_aer/backends/name_mapping.py
@@ -20,19 +20,19 @@ from qiskit.circuit.store import Store
 from qiskit.circuit.library import (
     U2Gate,
     RGate,
-    CYGate,
-    CZGate,
-    CSXGate,
-    CU3Gate,
-    CSwapGate,
+    YGate,
+    ZGate,
+    SXGate,
+    U3Gate,
+    SwapGate,
     PauliGate,
     DiagonalGate,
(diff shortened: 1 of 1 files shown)


AlbertJP:
I have just changed some existing tests to use the `MCYGate` and `MCZGate` classes instead. I'll add tests for the others as well.

AlbertJP:
All gates from `name_mapping.py` are now in the tests as well.

PR #2327, opened 2025-03-05
Update BaseEstimator import to support qiskit 2.0

ElePT:
### Summary
https://github.com/Qiskit/qiskit/pull/13877 removes the non-versioned primitive V1 base class aliases. It looks like the `AerSampler` was already using the versioned base class, but `AerEstimator` wasn't. This PR updates the `BaseEstimator` import to `BaseEstimatorV1`.

### Details and comments

The PR has been extended to handle more incompatibilities with 2.0.


kt474 (approved):
LGTM, this is also causing issues in the `qiskit-ibm-runtime` unit tests

wshanks:
In my testing with Aer and Qiskit 2.0, this line is also problematic. Do you want to address it here as well?

https://github.com/Qiskit/qiskit-aer/blob/0f2ff7143a180edcfbc2bda6bd2cdad1f312642e/qiskit_aer/primitives/sampler.py#L27

The fix in this PR and the line I pointed out are both problems with trying to do `import qiskit_aer.primitives` with Qiskit 2.0.

wshanks:
Ah, there is a helpful test run that now picks up the Qiskit rc and encounters further issues:

```
   File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/qiskit_aer/primitives/sampler.py", line 24, in <module>
    from qiskit.circuit.bit import Bit
ModuleNotFoundError: No module named 'qiskit.circuit.bit'
```

ElePT:
A couple of unit tests with fractional gates fail because of https://github.com/Qiskit/qiskit/issues/14002. Given that this PR is blocking other efforts, I think that the best course of action is to not use fractional gates + multiprocessing in aer tests until the issue has been addresed.

test/terra/primitives/test_estimator.py
```diff
@@ -322,7 +322,7 @@ def test_result_order(self):
         qc2.ry(np.pi / 2 * param, 0)
         qc2.measure_all()
 
-        estimator = Estimator(approximation=True)
+        estimator = Estimator(approximation=True, skip_transpilation=True)
```
AlbertJP, line 325: Try `run_options = {'shots': None}` to get exact values for the circuits that result in 0.

Alternative: change qc2 to give -1 instead of 0, then you won't have shot noise either. (Shot noise from the Estimator is 0 when the expectation value is -1 or 1)

AlbertJP: This option is documented at https://github.com/Qiskit/qiskit-aer/blob/076fe00c69d1281bf921d5e56be7cf1dd306b4a9/qiskit_aer/primitives/estimator.py#L68-L75. Honestly, I don't understand why it did not work. I suspect a bug in the Estimator class in combination with `skip_transpilation=True`.

ElePT: It's likely. Something to look at in a follow-up.

test/terra/backends/aer_simulator/test_truncate.py
```diff
@@ -106,25 +106,13 @@ def test_truncate_non_measured_qubits(self):
...
-        ]
         noise_model = NoiseModel.from_backend(self.device_backend())
         backend = self.backend(noise_model=noise_model, enable_truncation=False)
-        circuit = transpile(self.create_circuit_for_truncate(), backend, coupling_map=coupling_map)
+        circuit = self.create_circuit_for_truncate()
 
         result = backend.run(circuit, shots=100).result()
         metadata = result.results[0].metadata
-        self.assertEqual(metadata["num_qubits"], 10)
+        self.assertEqual(metadata["num_qubits"], 4)
```
gadial, line 115: This seems to subtly change the test. Previously, there was a difference between the `num_qubits` value (which seems to have been determined by the now-extinct coupling map) and the number of active qubits - I believe this difference results from how truncation works in the CPP level. However, I did not manage to recreate this desired behavior yet when working with qiskit 2.0 so we might want to keep this fix as it is for now.

test/terra/backends/aer_simulator/test_shot_branching.py
```diff
@@ -595,6 +595,10 @@ def test_shot_branching_kraus_circuit_noise(self, method, device):
     # ---------------------------------------------------------------------
     # Test conditional
     # ---------------------------------------------------------------------
+    @unittest.skipUnless(
```
gadial, line 598: Can these tests be fixed to work with the new `if` mechanism in qiskit 2.0?

gadial (approved):
Thank you for this substantial work. I left a few small comments which may be addressed but I believe we can merge without handling them as well.

test/terra/backends/aer_simulator/test_shot_branching.py
ElePT: To some extent, yes. From what I have seen, the new `if` mechanism is only supported by some simulator methods (for example, `statevector` doesn't support it). I think that moving forward the tests would have to be refactored and only test conditionals with the supported simulator methods, but this is a bit tricky at the moment because we still want to support the "old" path, and there are already some tests for the new `if`, so I figured that the best intermediate solution would be the conditional skip. I can open an issue to keep track of this.

test/terra/backends/aer_simulator/test_truncate.py
ElePT: I actually think there is no difference in the way truncation works, the only actual difference I have seen is in the way the transpiler works between 1.4 and 2.0, but that is intentional because we added extra consistency checks between the inputs. The original test provided a backend with 5 qubits and a coupling map with 10, and this is something that `transpile` now processes differently. I actually tested transpiling only with the coupling map and the output metadata matches the original test. So maybe we could do that.

test/terra/backends/aer_simulator/test_shot_branching.py
gadial: An issue seems like the correct way to proceed, thank you.

gadial (approved):
LGTM

PR #2329, opened 2025-03-06
Update noise model for compatibility with  Qiskit 1.4.x and 2.x

ElePT:
### Summary
The noise_model.from_backend method should support both backend v1 and v2 for compatibility with pre-2.0 qiskit versions.

### Details and comments
I ran the qiskit unit test suite with 1.4.1 and this PR installed and tests ran without issues.

### DIFF
--- a/qiskit_aer/noise/noise_model.py
+++ b/qiskit_aer/noise/noise_model.py
@@ -22,6 +22,7 @@ import numpy as np
 
 from qiskit.circuit import QuantumCircuit, Instruction, Delay, Reset
 from qiskit.circuit.library.generalized_gates import PauliGate, UnitaryGate
+from qiskit.providers import QubitProperties
 from qiskit.transpiler import PassManager
 from qiskit.utils import apply_prefix
 from .device.models import _excited_population, _truncate_t2_value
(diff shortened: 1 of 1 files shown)


qiskit_aer/noise/noise_model.py
```diff
@@ -378,6 +379,27 @@ def from_backend(
                     UserWarning,
                 )
             dt = backend.dt
+        elif backend_interface_version <= 1:
+            # BackendV1 will be removed in Qiskit 2.0, so we will remove this soon
```
gadial, line 383: To finalize this notion of "soon" I think maybe we should add both to the comment and to the warning message a concrete version number which drops support for V1 backends (I suggest 0.17)

gadial (approved):
LGTM, with a single comment about making aer 0.17 officially drop BackendV1

qiskit_aer/noise/noise_model.py
ElePT: I think it would be beneficial to keep this around for the life cycle of qiskit 1.4. Given that we don't know how many aer releases will happen in this period, maybe it's better for us not to commit to a specific version, it will give us more flexibility to act upon the removal in the future.

gadial: Agreed, so I think we can merge.

PR #2345, opened 2025-06-04
Bump version to 0.17.1

gadial:


jakelishman (dismissed):
I think you might need to do `docs/conf.py` too, and potentially add a prelude release note?

I'll approve and you can merge if I'm wrong (I'm just on my phone).

gadial:
> I think you might need to do `docs/conf.py` too, and potentially add a prelude release note?
> 
> I'll approve and you can merge if I'm wrong (I'm just on my phone).

You're absolutely right about `conf.py` but I don't think a release note is needed for a minor version bump.

jakelishman (approved):

PR #2361, opened 2025-08-19
[WIP] Fix GPU builds and Tests

jmmshn:
### Summary

Qiskit 2.1 seems to fix the failing GPU builds.
Example of successful run on fork: 
https://github.com/jmmshn/qiskit-aer/actions/runs/17079326402

Two other changes caused test breaks:
- Removal of `sympy` from the dependency: https://github.com/Qiskit/qiskit/pull/14396
- Porting the `expr.Var` to Rust: https://github.com/Qiskit/qiskit/pull/14068

This PR aims to update the code and pipelines so that the GPU builds, and tests pass.

### Details and comments

For the version updates, this PR modifies the install setup slightly so we can test on a "strict" set of dependencies for more consistent CI/CD.

For the rust port, the did not port over the ability to initialize a classical Variable from a list.

So the following code works in 2.0.3 but not in 2.1.0.

```python
from qiskit import ClassicalRegister
from qiskit.circuit import Clbit
from qiskit.circuit.classical import expr, types

creg = ClassicalRegister(3, 'creg') # Classical register with 3 bits
creg_expr = expr.Var(creg, types.Uint(creg.size))
isinstance(creg_expr.var, ClassicalRegister) # True
res = expr.Var([creg_expr], creg_expr.type) # Fails in 2.1.0
res = expr.Var(creg_expr.var, creg_expr.type) # Works in 2.1.0
```

And the code will work in 2.1.0 if you remove the braces.  It looks like the data from all the variables are still 
Note that the ability to create a from a list was not documented in the old code so it might be better to remove it entirely
https://github.com/Qiskit/qiskit/blob/19eeb418b808cb3c6a2f43ee70ecb7314482636a/qiskit/circuit/classical/expr/expr.py#L141C1-L152C1


jmmshn:
@gadial, @1ucian0, would it be OK to remove the windows tests in order to get the gpu wheels built?

gadial:
> @gadial, @1ucian0, would it be OK to remove the windows tests in order to get the gpu wheels built?

As far as I'm concerned, yes. I was completely stumped as to how to fix the GPU builds, so it's a higher priority for me. I'll try fixing the Windows problems in a future PR (unless you manage to magically solve this as well).

jmmshn:
@gadial, Thanks!
I'll disable the windows tests.

> As for the classical var part - turns out what Aer currently does is invalid
That's what it looked like to me as well.

gadial:
Let's also fix the mergify issue. Here's what I think a fixed file should look like:

https://github.com/Qiskit/qiskit-aer/blob/8387c274407bebd64ae2bbe437d832b80797b49b/.mergify.yml

We remove the `method: squash` from the `pull_request_rules:` part and add `merge_method: squash` to the `queue_rules:` part.

jmmshn:
@mergifyio refresh

jmmshn:
@gadial, I still see the same mergify error no matter how I change the mergify yaml. I suspect mergify is someone looking only at main for a part of the workflow

gadial:
As it turned out, the problem was not with mergify; I simply had to click a button so the full workflow would run.

It seems there are now additional problems; if you don't want to deal with them I can attempt to try your PR and finalize it.

wshanks:
I opened an [issue](https://github.com/Qiskit/qiskit-aer/issues/2366) with some information about what I have seen with trying to build on Windows.

gadial:
I opened a new PR based on the changes in this one to try to address the additional problems:

https://github.com/Qiskit/qiskit-aer/pull/2368

Meanwhile, it seems a new problem with arm64 popped... I added a hack to address it.

I also think I found a solution to the `Var` problem (which fails the tests for this PR) - by doing
```
reg = ClassicalRegister(bits = [bit_map[clbit] for clbit in cond_tuple[0].var])
expr = Var(reg, cond_tuple[0].type)
```

I create a new "aliasing" register, which can be safely used to instantiate expressions, while pointing to the correct bits. I apologize that I did not think about this before.

PR #2368, opened 2025-09-09
Fix GPU builds and Tests (Qiskit>=2.1)

gadial:
A PR based on #2361 to address the various build issues we're facing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ["macos-13", "ubuntu-latest", "windows-2019"]
+        os: ["macos-13", "ubuntu-latest"] # disable Windows for now https://github.com/Qiskit/qiskit-aer/pull/2361#issuecomment-3227442312
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python Python '3.13'
--- a/qiskit_aer/backends/aer_compiler.py
+++ b/qiskit_aer/backends/aer_compiler.py
@@ -235,8 +235,6 @@ class AerCompiler:
         elif isinstance(cond_tuple[0], Var):
             if isinstance(cond_tuple[0].var, Clbit):
                 expr = Var(bit_map[cond_tuple[0].var], cond_tuple[0].type)
-            elif isinstance(cond_tuple[0].var, ClassicalRegister):
-                expr = Var([bit_map[clbit] for clbit in cond_tuple[0].var], cond_tuple[0].type)
             else:
                 raise AerError(
                     f"jump condition does not support this tyep of Var: {cond_tuple[0]}."
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,7 @@
+scipy
+numpy
+sympy
+qiskit>=2.1.0
 stestr>=2.5.0
 cmake!=3.17.1,!=3.17.0
 conan<2.0.0
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,7 @@ jobs:
     name: Build qiskit-aer wheels
     strategy:
       matrix:
-        os: ["macos-13", "ubuntu-latest", "windows-2019"]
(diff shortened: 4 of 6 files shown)


requirements-dev.txt
```diff
@@ -1,3 +1,7 @@
+scipy
+numpy
+sympy
+qiskit>=2.1.0
```
jakelishman, line 4: Are these dev-only requirements?  I think they're more likely runtime requirements - they should be wherever Aer specifies its runtime requirements if so.

I think requiring Qiskit 2.1 would be a big bump for Aer.

qiskit_aer/backends/aer_compiler.py
```diff
@@ -236,7 +236,8 @@ def _convert_jump_conditional(self, cond_tuple, bit_map):
             if isinstance(cond_tuple[0].var, Clbit):
                 expr = Var(bit_map[cond_tuple[0].var], cond_tuple[0].type)
             elif isinstance(cond_tuple[0].var, ClassicalRegister):
-                expr = Var([bit_map[clbit] for clbit in cond_tuple[0].var], cond_tuple[0].type)
+                reg = ClassicalRegister(bits=[bit_map[clbit] for clbit in cond_tuple[0].var])
+                expr = Var(reg, cond_tuple[0].type)
```
jakelishman, line 240: I think the real trouble is that this line should never _actually_ remap the bits at all; it's a trouble with Qiskit core's classical data model that individual clbits can be remapped at all.  Registers can't really be remapped, and "proper" `Var` nodes have a complete explicit representation of "closure" so don't have the same trouble.

Assuming this passes the tests, I think we can live with it, but really, I think unless each `clbit` just maps to itself, it's _probably_ a bust circuit incoming already.

.github/workflows/build.yml
```diff
@@ -49,6 +49,10 @@ jobs:
       - name: Build Wheels
         env:
           CIBW_ARCHS_MACOS: arm64
+          CIBW_BEFORE_BUILD: |
+            python -m pip install "conan<2.0.0"
+            conan config init --force
+            sed -i '' 's/"16.0"/"16.0", "17.0"/' ~/.conan/settings.yml
```
jakelishman, line 55: The idea that we have to patch the Conan settings file like this _feels_ weird to me, but I don't know enough to say whether it's sensible or not.  Perhaps at a minimum you could add a comment explaining what it is we're doing?

gadial: That's extremely weird to me (since is only happens for this one arm64 instance) but I think it's one of the unavoidable results of us still sticking with conan 1.x and not moving to 2 (which is not trivial; I tried). The long term solution would probably be moving to 2. I'll add a comment.

jakelishman: Fwiw, I can't even tell what package's version / whatever it is that we're patching here.  Judging by the number itself, I'm _guessing_ it's LLVM or Clang?  It'd be good to comment that too.

.github/workflows/build.yml
```diff
@@ -49,6 +49,10 @@ jobs:
       - name: Build Wheels
         env:
           CIBW_ARCHS_MACOS: arm64
+          CIBW_BEFORE_BUILD: | # conan<2.0.0 on arm64 (turns to be 1.66.0) complains about not supporting Clang 1.7
```
jakelishman, line 52:
```suggestion
          CIBW_BEFORE_BUILD: | # conan<2.0.0 on arm64 (turns to be 1.66.0) complains about not supporting Clang 17
```

requirements-dev.txt
```diff
@@ -1,3 +1,7 @@
+scipy
+numpy
+sympy
+qiskit>=2.0.0
```
jakelishman, line 4: `scipy`, `numpy` and `qiskit` should all already be direct dependencies of Aer - does everything work if these are removed from the list?

jmmshn: They are in the setup.py.
However, there might be some value in having a more strict set of testing dependencies compared to the required dependencies.  This can often be useful in tracking down build problems across time.

gadial: It seems removing `numpy` and `scipy` from the requirements doesn't have an adverse effect, so I'll keep them out for now.

PR #2370, opened 2025-09-14
Windows build error fix

gadial:
### Summary
Fix for the build error that arose when moving away from the `windows-2019` runner.

### Details and comments
After the `windows-2019` runner was discontinued, Aer switched to `windows-2022` and started hitting the error `fatal error C1060: compiler is out of heap space` while compiling.

This seems to be related to the templating system, and in particular to how `to_python` functions are instantiated. The failure arises when attempting to instantiate the function for `Aer::Data`, probably because of its many parents (while debugging it was observed that arbitrarily removing some parents results on the memory problem disappearing).

A quick solution is to drop templating for this specific instance of `to_python` completely. If this succeeds, perhaps we should rethink the use of templates altogether for the existing well-defined datatypes which require specific handling.

### DIFF
--- a/qiskit_aer/backends/aer_simulator.py
+++ b/qiskit_aer/backends/aer_simulator.py
@@ -370,7 +370,7 @@ class AerSimulator(AerBackend):
         is sparse. The overall runtime scales as Sn^{2}.
 
       - ``"norm_estimation"``: An alternative sampling method using
-        random state inner products to estimate outcome probabilites. This
+        random state inner products to estimate outcome probabilities. This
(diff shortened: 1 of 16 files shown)


jakelishman (approved):
From Slack from Gadi:

> After discussing with Matthew, we've agreed that for now we'll open an issue about the aer MCU problem, but skip the tests in order to put out a new release. Afterwards I'll try updating aer to conan 2.0 which may fix this problem as well (also, Jun Doi is looking at this)

and tracked in #2371.

With those caveats, I can approve and move forwards.

PR #2384, opened 2025-11-05
Update OS and Python-version support

### LINKED ISSUE #2378: Need binary distribution on PyPI for Python 3.14
Python 3.14 was released today, but the binary distributions of qiskit-aer are not yet available on pypi. Similar to the previous issue #2281, now it is required to go through the same drill for Python 3.14.

wshanks:
This change adds 3.14 to the lists of Python versions to test and to the list to build with cibuildwheel. It also marks the package as compatible with Python 3.13 and 3.14. It does not drop Python 3.9 or update the Python versions used in parts of the GitHub workflow that only use a single Python version, though doing those updates would also be a good idea.

Closes #2378


pyproject.toml
```diff
@@ -12,7 +12,7 @@ build-backend = "setuptools.build_meta"
 [tool.cibuildwheel]
 manylinux-x86_64-image = "manylinux2014"
 manylinux-i686-image = "manylinux2014"
-skip = "pp* cp36* cp37* cp38* *musllinux*"
+skip = "pp* cp36* cp37* cp38* *musllinux* cp??t-*"
```
jakelishman: I think we'd need `cp???t-*` really, I think?  The first valid match is `cp313t-*`, presumably.

wshanks: Thanks, yes, I meant `cp3??t-*`. I was trying to copy the suggestion in the documentation [here](https://cibuildwheel.pypa.io/en/stable/options/#build-skip-examples).

jakelishman:
Ignoring the GPU builds: the trouble with Python 3.14 looks like Numpy upped their minimum supported glibc to 2.27 in Numpy 2.3, which is also the first Numpy to have 3.14 wheels.  We're still implicitly building against glibc 2.17 because of the manylinux2014 image, so I guess we'll need to update our build image.

We can potentially update the build image only for Python 3.14, since Aer wouldn't be installable (other than from-source builds of Numpy) on glibc 2.17 _anyway_, or skip the tests on 3.14 only.  There's definitely options, but I'll take the question about the manylinux/glibc version over to Qiskit proper.

wshanks:
Okay, that sounds good. At least from the logs we have now it looks like the gpu builds worked okay and just ran into the same numpy issue as the cpu build.

wshanks:
It is maybe worth also checking if it is okay to drop Python 3.9 now. 3.13 is also used to install cibuildwheel. I didn't bump that to 3.14 but that would probably be fine.

jakelishman:
It's fine to drop Python 3.9 in a new Aer release - Qiskit is in the process of dropping it (Qiskit/qiskit#15139), it's just that we have some backwards-compat tests that need a rework before we can do it completely.  It'll be gone in Qiskit 2.3.

wshanks:
It looks like everything is okay moving up from Python 3.9 though we need to update the required checks in the repo settings.

We need to add a release note as well but I will wait to hear what the decision is about numpy and the build image before pushing again since it triggers so much CI time.

jakelishman:
man Aer's build system is ooooooooold

jakelishman:
Ok that's a frankly disgusting collection of hacks and changes, but it is at least a minimum of CI running again.  I would like to follow up with some build-system modernisation to get rid of the worst of this stuff.

The GPU builds will need further work before being reinstated, but I think that statement has been true at most points of Aer's development, so I think it's _probably_ right to skip them for now and try and fix them later.

I want another sign-off from a maintainer, especially around the changes to platform support, before merging this, though.

mtreinish (approved):
The thrust hack is the only thing that I'm not 100% on, but the rest of it looks fine. The conan stuff is unfortunate but until we upgrade the conan version to a new version I don't see an alternative. I do something similar when I build aer from source locally.

jakelishman:
fwiw on the Thrust hack: I'm fairly sure that C++ build systems are _supposed_ to set `_VSTD` to some potentially internally versioned variant of `std`, but something's up with the Apple Clang 17.0 configuration we're using now and it's not getting set.  Updating Thrust to the latest available on Conan didn't change matters either, so I'm more suspicious it's something in our configuration our side, or a side effect of the gross conan-config hack to pretend like it actually supports Apple Clang 17.0.

Clearly it builds and runs with that in place, and I looked into single line of one header file that uses it in Thrust, and it _seems_ pretty innocuous.

wshanks:
What do we need to do to go from this PR merged to Python 3.14 wheels on PyPI? A version bump PR and then a tag?

jakelishman:
This PR is a fairly breaking change to the support tiers, so if I do a complete release it feels a bit nicer to go in a full 0.18 minor, but there's only been a single bugfix since 0.17.2.  I might cherry-pick these changes into a temporary branch and issue a standalone 0.17.2 wheel with cp314 support when I get a chance (just trying to finish a couple of other things up).

There's been an unusual increase in activity on Aer in issues and PRs recently, and it would be nice to get some of those reviewed and merged in order to cut a _full_ 0.18.0 release, but I really don't know timescales for that.

PR #2390, opened 2025-11-21
Base x86 compiler flags on target rather than host system

wshanks:
`CMAKE_HOST_SYSTEM_PROCESSOR` represents the architecture used to run cmake while `CMAKE_SYSTEM_PROCESSOR` represents the system that the project is being compiled for. Previously, the host architecture and a special OSX variable were used to determine if x86 specific flags should be used. The OSX variable helped avoid using the x86 build flags for Apple Silicon cross builds and otherwise x86 was only used to build x86 binaries in CI (aarch64 is built with qemu emulation). However, cross-building for aarch64 from x86 would fail because the x86 specific flags would be used. This change uses the target architecture to guard those flags and removes the OSX variable usage as it is no longer needed.


jakelishman (approved):
Thanks Will, this seems entirely sensible.  I actually thought I'd approved and merged this last week, but I guess not.

PR #2397, opened 2026-01-22
Improve numerical robustness in noise transformation by assuming PSD for CVXPY optimization

emma6-62607015e-34:
### Summary
Skip explicit PSD check in CVXPY to improve numerical robustness

### Details and comments
In the method `_transform_by_operator_list()` inside the noise transformation utilities, the CVXPY optimization problem is solved by calling `prob.solve()`.  
Because the matrix involved is a Gram matrix by construction, it *should* be positive semidefinite (PSD). However, due to numerical noise, very small negative eigenvalues can appear, which can cause CVXPY’s default DCP check to fail. Using larger operator lists/basis sets might increase the numerical noise which is why this issue is often observed for more complex decompositions. Using `prob.solve(assume_PSD = True)` circumvents this problem by skipping the DCP check of the solver.

Rationale:
- The matrix in the quadratic form is a Gram matrix and therefore PSD by definition, therefore the small negative eigenvalues and failed DCP checks are numerical artifacts
- skipping the DCP checks avoids unnecessary numerical failures without changing the actual problem

Notes:
- the existing tests already cover noise transformation behaviour in relevant scenarios which is why I did not introduce a new test

### DIFF
--- a/qiskit_aer/utils/noise_transformation.py
+++ b/qiskit_aer/utils/noise_transformation.py
@@ -416,9 +416,6 @@ def _transform_by_operator_list(
         constraints=[cvxpy.sum(x) <= 1, x >= 0, source_fids.T @ x <= target_fid],
     )
     # solve quadratic program
-    # A is a Gram matrix and therefore PSD by construction
-    # numerical noise can induce small negative eigenvalues which can 
-    # lead cvxpy's DCP check to fail. This can be circumvented by assume_PSD = True
     prob.solve(assume_PSD=True)
     probabilities = x.value
     return probabilities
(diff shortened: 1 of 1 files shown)


jakelishman (commented):
Thanks for this!  I honestly don't know enough about the maths to comment on the merits, but I'll believe you're correct.  Could you add a small comment about the changed line explaining why the assumption holds, like you have in the PR comment?

At the moment, the CLA bot can't finalise the check because the first commit (the one you made locally) uses an email address that isn't linked to your GitHub account.  Can you either add it to GitHub, or rewrite the commit history to use an email address that _is_ connected?

jakelishman:
Oh haha - I see the original commit _did_ have the comment!  It'd be good to keep that inline in the code, so we can see the justification when reading it.

emma6-62607015e-34:
I think my first local commit didn't use any email address at all. I cannot seem to get it fixed right now which is why I'll close this pull request and open a new one later which also includes the comment in the code.

jakelishman:
The first commit had an email associated with your name and RWTH Aachen, if it helps.

emma6-62607015e-34:
Thank you, that definitely helped! I still decided to clean it up and add a now pull request right now.

PR #2398, opened 2026-01-22
Improve numerical robustness in noise transformation by assuming PSD for CVXPY optimization

emma6-62607015e-34:
### Summary
Skip explicit PSD check in CVXPY to improve numerical robustness

### Details and comments
In the method `_transform_by_operator_list()` inside the noise transformation utilities, the CVXPY optimization problem is solved by calling `prob.solve()`.  
Because the matrix involved is a Gram matrix by construction, it *should* be positive semidefinite (PSD). However, due to numerical noise, very small negative eigenvalues can appear, which can cause CVXPY’s default DCP check to fail. Using larger operator lists/basis sets might increase the numerical noise which is why this issue is often observed for more complex decompositions. Using `prob.solve(assume_PSD = True)` circumvents this problem by skipping the DCP check of the solver.

Rationale:
- The matrix in the quadratic form is a Gram matrix and therefore PSD by definition, therefore the small negative eigenvalues and failed DCP checks are numerical artifacts
- skipping the DCP checks avoids unnecessary numerical failures without changing the actual problem

Notes:
- the existing tests already cover noise transformation behaviour in relevant scenarios which is why I did not introduce a new test

### DIFF
--- a/qiskit_aer/utils/noise_transformation.py
+++ b/qiskit_aer/utils/noise_transformation.py
@@ -416,7 +416,10 @@ def _transform_by_operator_list(
         constraints=[cvxpy.sum(x) <= 1, x >= 0, source_fids.T @ x <= target_fid],
     )
     # solve quadratic program
-    prob.solve()
+    # the matrix A in the quadratic program is a Gram matrix and therefore by construction PSD
+    # due to numerical noise small negative eigenvalues can occur causing a DCPError with prob.solve()
+    # as the PSD check might fail; assume_PSD=True tells cvxpy to not perform this check
+    prob.solve(assume_PSD=True)
     probabilities = x.value
     return probabilities
 


jakelishman (approved):
Thanks for this!

I'll trigger CI, though given a couple of bug reports open on Qiskit/qiskit right now, it's possible it won't pass.  That will be unrelated to your PR, though.

jakelishman:
Ah no, the failures do seem related - at least a couple are cvxpy complaining about not recognising the option. Is it only valid for certain solvers?

emma6-62607015e-34:
I am sorry. I misplaced the assume_PSD=True option. It is actually placed inside cvxpy.quad_form(A, x, assume_PSD=True). I'll close this pull request again and open a new one. I am sincerely sorry for the inconvenience.

PR #2406, opened 2026-01-30
Improve sampling performance and MPI initialization safety

seanyeh77:
### Summary
This PR optimizes state string construction during sampling by replacing the slow approach. It achieves a 19x performance boost, reducing execution time from 25.8505s to 1.3570s (at shots = 10,000,000).
The other feature is fix the mpirun problem. The cause is that MPI was not properly finalized.
### Details and comments
Optimizes state string: Use the std::string instead of std::stringstream.
MPI problem: Add a check for MPI initialization and use std::atexit to Finalize the mpi.
Other feature: std::vector::reserve and emplace_back are used to optimize memory allocation and element insertion.


jakelishman (changes requested):
This appears to be LLM generated and a collection of unrelated changes, including simply commenting out exception-throwing code.

I won't put review effort into such a patch.

seanyeh77:
I apologize for using the LLM to generate the description of this PR. I will revise it as soon as possible.

seanyeh77:
I sincerely apologize for the previous PR description. I have now carefully revised the PR and would greatly appreciate it if you could take another look when you have the time. Thank you very much for your patience and guidance.

PR #2409, opened 2026-02-16
[BUG] Fix CUDA peer copy size and OpenMP init

Ghjk94522:
### Summary

Fix two bugs:
- Use byte counts for cudaMemcpyPeerAsync in device chunk container.
- Parallelize unitary matrix initialization to avoid redundant writes.

### Details and comments

This PR addresses two distinct issues:

1. Incorrect size handling in `copyIn`/`CopyOut` during P2P transfer  
   The buffer size calculation/setting was incorrect in the P2P memcpy path, which could lead to incomplete transfers or invalid memory access.  
   Verified and reproduced on dual RTX 4090 setup.  
   Fixed by properly calculating/aligning the transfer size.  

2. Race condition during unitary matrix initialization in ROCm Thrust & OpenMP backend  
   In AMD ROCm environments (using rocthrust + OMP parallelization), concurrent initialization of unitary matrices could cause data races / incorrect values.  
   This manifested as **random failures** in `test/terra/backend/aer_simulator/test_standard_gates`.  
   Root cause: insufficient synchronization during matrix population.  
   Fixed by adding proper synchronization (critical section / barrier).

### DIFF
--- a/src/simulators/statevector/chunk/device_chunk_container.hpp
+++ b/src/simulators/statevector/chunk/device_chunk_container.hpp
@@ -668,7 +668,7 @@ void DeviceChunkContainer<data_t>::CopyIn(Chunk<data_t> &src, uint_t iChunk) {
                       cudaMemcpyDeviceToDevice, stream(iChunk));
     } else {
       cudaMemcpyPeerAsync(chunk_pointer(iChunk), device_id_, src.pointer(),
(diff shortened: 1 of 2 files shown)


Ghjk94522:
Would appreciate a review whenever anyone have some time! 🙏

PR #2417, opened 2026-03-26
Switch from pybind11 enum_ to native_enum

wshanks:
The motivation for switching is that `enum_` triggers a bug in pybind11 when using gcc 14. In particular, this bug:

https://github.com/pybind/pybind11/issues/5565

which has already been patched in pybind11 but the patch has not yet been released. Code within pybind11 used for generating the special methods was triggering an ambiguous template instantiation.

Using `native_enum` requires pybind11 3.0 or higher so the build requirement was updated.

There are side effects to these changes but they should not affect the way these undocumented enums are used within the codebase (and a GitHub search reveals no external usage):

* The `arithmetic` tag was dropped since the current usage does not use Python mathematical operators on these classes. This change removes methods like `__xor__` from the Python classes.
* The `export_values` call was dropped. That promotes the enum members into the module namespace, but all current use references them as class members of the enum.
* The classes are standard Python `enum.Enum` instances rather than pybind11 classes following the same interface.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = [
   "scikit-build>=0.11.0",
   "cmake!=3.17.1,!=3.17.0",
   "ninja",
-  "pybind11>2.6",
+  "pybind11>=3.0",
 ]
 build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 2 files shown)


wshanks:
[Here](https://pybind11.readthedocs.io/en/latest/classes.html#native-enum) is the `native_enum` documentation. There is some discussion [here](https://github.com/pybind/pybind11/pull/5555) about migrating from `enum_` but it is a little hard to follow because the signature changed a bit from the description in the PR but is described further down.

ihincks (approved):
Thanks @wshanks, these changes look good to me. 

Since these enums are not public, and on top of that they almost certainly haven't changed in a way that will break anyone who does happen to be using them (and you checked, not on GitHub), I don't see a reason for a changelog, it will just be noise.

Issue #1742, opened 2023-03-10
Test failures with 0.12.0 (0.11.2 passes)


thesamesam:
### Informations

- **Qiskit Aer version**: 0.12.0
- **Python version**: Python 3.9.16
- **Operating system**: Gentoo Linux (I'm the maintainer of its package)

### What is the current behavior?

Various tests seem to fail with qiskit-aer-0.12.0, while qiskit-0.11.2 passed all its tests at the time of release and still passes them when I check now, which suggests it's not an environment issue.

The failures all look identical:
```
______________________________________________________________ TestNoiseModel.test_noise_model_from_lagos_v2 ______________________________________________________________
'NoneType' object is not iterable

During handling of the above exception, another exception occurred:
NOTE: Incompatible Exception Representation, displaying natively:

testtools.testresult.real._StringException: Traceback (most recent call last):
  File "/var/tmp/portage/dev-python/qiskit-aer-0.12.0/work/qiskit-aer-0.12.0/test/terra/noise/test_noise_model.py", line 252, in test_noise_model_from_lagos_v2
    result = AerSimulator().run(circ, noise_model=noise_model).result()
  File "/var/tmp/portage/dev-python/qiskit-aer-0.12.0/work/qiskit-aer-0.12.0-python3_9/install/usr/lib/python3.9/site-packages/qiskit_aer/jobs/utils.py", line 41, in _wrapper
    return func(self, *args, **kwargs)
  File "/var/tmp/portage/dev-python/qiskit-aer-0.12.0/work/qiskit-aer-0.12.0-python3_9/install/usr/lib/python3.9/site-packages/qiskit_aer/jobs/aerjob.py", line 106, in result
    return self._future.result(timeout=timeout)
  File "/usr/lib/python3.9/concurrent/futures/_base.py", line 446, in result
    return self.__get_result()
  File "/usr/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/usr/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/var/tmp/portage/dev-python/qiskit-aer-0.12.0/work/qiskit-aer-0.12.0-python3_9/install/usr/lib/python3.9/site-packages/qiskit_aer/backends/aerbackend.py", line 438, in _execute_circuits_job
    output = self._execute_circuits(aer_circuits, noise_model, config)
[...]

PR #2418, opened 2026-03-28
Add explicit json conversions

wshanks:
nlohmann_json 3.10.3 improved the detection of C++ iterables, so that more objects are correctly identified as iterable and passed through its iterable to array converter function. This change included detecting pybind11 py::handle objects pointing to Python dicts. The result of this change is that Python dicts are passed to functions accepting `json_t` objects the implicit conversion started treating them as iterables and producing JSON arrays of the dict keys rather than falling through to Aer's `std::to_json` converter and its `py::dict` case. So a noise model would be converted into a `["errors"]` array instead of an object with more nested data.

To address this issue, more explicit conversion of Python objects to JSON objects was inserted into the code. Where the noise model is passed in from Python in `execute`, an explicit `std:to_json` call was added. That alone was not enough because `std::to_json` itself was handling dicts without explicit conversion and so any nested dicts were falling into the same iterable-to-array implicit conversion problem. Explicit construction of the JSON object and conversion of the dict values with `std::json` were added to address the conversion issue recursively. Although no test failures flagged it as problematic, the `JSON::iterable_to_json_list` helper function was also updated to call `std::to_json` for completeness and to handle any untested cases where the noise model could contain a list of dicts. 

With these changes, Aer builds and passes all the tests with nlohmann_json 3.12.0.

Addresses the issues discussed in #1742

### DIFF
--- a/src/framework/pybind_json.hpp
+++ b/src/framework/pybind_json.hpp
@@ -213,7 +213,9 @@ json_t JSON::numpy_to_json(py::array_t<T, py::array::c_style> arr) {
 json_t JSON::iterable_to_json_list(const py::handle &obj) {
   json_t js = nl::json::array();
   for (py::handle value : obj) {
-    js.push_back(value);
+    json_t elem;
+    std::to_json(elem, value);
+    js.push_back(std::move(elem));
   }
   return js;
 }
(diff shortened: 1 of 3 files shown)


wshanks:
The force push was to add a more descriptive commit message.

src/framework/pybind_json.hpp
```diff
@@ -235,8 +237,15 @@ void std::to_json(json_t &js, const py::handle &obj) {
   } else if (py::isinstance<py::tuple>(obj) || py::isinstance<py::list>(obj)) {
     js = JSON::iterable_to_json_list(obj);
   } else if (py::isinstance<py::dict>(obj)) {
+
```
ihincks: cosmetic nit
```suggestion
```

ihincks (commented):
Thanks for taking the time to fix this issue, it makes sense to me.

I'm not sure whether or not this deserves a release note? I kind of suspect not. It's a sort of upgrade: you can now build against 3.10.3 and later if you want to. I'll leave it up to you.

src/framework/pybind_json.hpp
wshanks: Fixed in 182f2bd

ihincks (approved):
Thanks again.

wshanks:
> I'm not sure whether or not this deserves a release note? I kind of suspect not. It's a sort of upgrade: you can now build against 3.10.3 and later if you want to. I'll leave it up to you.

Since it is a long standing problem and has been brought up a few times in different issues by people trying to compile in different environments, it seemed worth adding a short note in case those people read the release notes, so I added one now.

PR #2420, opened 2026-04-02
fix: update example code

jwoehr:
### Summary

example code outdated

### Details and comments

backend syntax and basis gates


ihincks (approved):
Thanks for the updates, @jwoehr.

Issue #2415, opened 2026-03-24
Build fails on NVIDIA Blackwell (sm_120, CC 12.0) with CUDA 13.2 — patches required


dmeliksetian:
## Environment
- GPU: NVIDIA RTX PRO 4500 Blackwell (Compute Capability 12.0)
- Driver: 595.45.04
- CUDA: 13.2
- OS: Ubuntu (GCC 13.3.0)
- Qiskit Aer: 0.17.2
- Python: 3.12.3

## Problem
Building qiskit-aer from source fails with four separate issues when targeting
NVIDIA Blackwell (sm_120 / CC 12.0) with CUDA 13.2. All are caused by the source
predating the Blackwell architecture and CUDA 13.x.

### 1. `select_compute_arch.cmake` missing Blackwell
`cuda_select_nvcc_arch_flags` has a hardcoded arch list that stops at Ampere (8.6).
Fix: add version blocks for Hopper (11.8), Ada (12.0), and Blackwell (12.8) to the
known arch list and the named-arch elseif chain.

### 2. `CMakeLists.txt` — `cuda_select_nvcc_arch_flags` doesn't accept `12.0`
Even after patching the cmake file, the dotted version regex only matches single
digits (e.g. `8.6`), so `12.0` falls through to the unknown-arch error.
Fix: bypass `cuda_select_nvcc_arch_flags` entirely and hardcode:
```cmake
set(AER_CUDA_ARCH_FLAGS "-gencode arch=compute_120,code=sm_120")
set(AER_CUDA_ARCHITECTURES "120")
set(CMAKE_CUDA_ARCHITECTURES "120")
```

### 3. C++14 incompatible with CUDA 13.2 Thrust/CCCL
CUDA 13.2 bundles Thrust/CCCL which requires C++17. Aer defaults to C++14.
Fix: pass `-DCMAKE_CXX_STANDARD=17 -DCMAKE_CUDA_STANDARD=17` at build time,
or update `CMakeLists.txt` to set these when CUDA >= 12.

### 4. `thrust::unary_function` removed in Thrust 2.x
`src/simulators/statevector/chunk/thrust_kernels.hpp` inherits from
`thrust::unary_function` and `thrust::binary_function`, which were removed
in Thrust 2.x (bundled with CUDA 12+).
Fix: add a compatibility shim after `#include "misc/wrap_thrust.hpp"`:
```cpp
namespace thrust {
  template<typename Arg, typename Result>
  struct unary_function {
[...]

PR #2422, opened 2026-04-26
Fix nlohmann_json >= 3.11 incompatibility breaking noise simulation

### LINKED ISSUE #2421: Noise simulation fails with Invalid noise_params JSON: not an object when built against nlohmann_json >= 3.11
### Informations
 
- **Qiskit Aer version**: 0.17.2 (built from source)
- **Python version**: python 3.13.12
- **Operating system**: Ubuntu (nlohmann-json3-dev 3.11.3 from apt; GCC 13.3.0)

When qiskit-aer is built from source against a system-provided nlohmann_json >= 3.11, all noisy simulation fails with:
`ValueError: Invalid noise_params JSON: not an object.`
This affects all simulation methods (statevector, density_matrix, tensor_network) and all devices (CPU, GPU). The PyPI qiskit-aer wheel works because its build pipeline uses nlohmann_json 3.10.x via Conan; only source builds against modern system packages exhibit the bug.
CMake even prints a warning during configure:
-- Found nlohmann_json: ... (found version "3.11.3")
-- nlohmann_json version found: 3.11.3. WARNING: This version may not work!!!
-- Using system-provided nlohmann_json library

…but the build proceeds anyway, and noise simulation breaks at runtime.
This affects builds on Ubuntu 24+, Fedora 40+, and any current Linux distribution where nlohmann-json3-dev ≥ 3.11 is the default system version.

Build qiskit-aer from source on a system with nlohmann_json 3.11.x installed (e.g., apt install nlohmann-json3-dev on Ubuntu 24):
```
DISABLE_CONAN=ON python setup.py bdist_wheel
pip install dist/qiskit_aer-*.whl
```
Then run
```
from qiskit_aer.noise import NoiseModel, depolarizing_error

nm = NoiseModel()
nm.add_all_qubit_quantum_error(depolarizing_error(0.05, 1), ['h'])
qc = QuantumCircuit(2)
qc.h(0); qc.cx(0, 1); qc.measure_all()

sim = AerSimulator(noise_model=nm)
sim.run(transpile(qc, sim), shots=512).result().get_counts()
# ValueError: Invalid noise_params JSON: not an object.
```

The script should return a counts dictionary, e.g.:
`{'00': 250, '11': 262}`

### Suggested solutions
The root cause is in `src/framework/pybind_json.hpp`, which defines `to_json` and `from_json` overloads for pybind11::handle and pybind11::object inside the std:: namespace:
```
namespace std {
  void to_json(json_t &js, const py::handle &o);
  void from_json(const json_t &js, py::object &o);
}
```
Adding free functions to `std:`: is undefined behavior per the C++ standard — only specific template specializations are permitted. It worked in practice because nlohmann_json ≤ 3.10 used a permissive ADL lookup that happened to find these overloads. nlohmann_json 3.11 tightened ADL conformance, and the overloads are no longer found.
The fix is to move the conversion logic out of `std::` and into proper `nlohmann::adl_serializer<T>` template specializations — nlohmann's documented extension point for non-owned types, version-stable across all releases. Verified by:

1. Vendoring nlohmann_json 3.10.2 locally (overriding system 3.11.3) — bug disappears, confirming the version is the variable.
2. Refactoring to adl_serializer specializations against system 3.11.3 — bug also disappears, confirming the fix is the right shape.

A PR implementing the adl_serializer fix is in progress.

dmeliksetian:
### Summary
Fixes #2421. aer's pybind11 → nlohmann::json conversion in
`src/framework/pybind_json.hpp` defined `to_json`/`from_json` overloads
in the `std::` namespace. This worked under nlohmann_json ≤ 3.10's
permissive ADL but breaks under 3.11+, which is now the default
system-package version on Ubuntu 24, Fedora 40, etc. All noisy
simulation fails with `ValueError: Invalid noise_params JSON: not an object.`

This PR moves the conversion into `nlohmann::adl_serializer<T>`
specializations — nlohmann's documented, version-stable extension
point — so the build works correctly against all nlohmann_json
versions.

### Details and comments

**Root cause:** [keep the paragraph from the current body about ADL
tightening and the failure mode]

**Fix:** [keep the bullet list of changes]

**Verification:**
- Minimal noise repro (`depolarizing_error` on 2-qubit GHZ) now
  returns correct counts; previously raised the JSON error.
- Test suite on plain `main` + this fix, CPU-only build:
  - Noise test module: 97/97 pass
  - Full suite: 3038 pass, 132 skipped, 0 fail
- On a separate Blackwell + CUDA 13.2 build (per #2415): 216 of 221
  previously-failing noise tests now pass under parallel execution;
  the remaining are unrelated tensor_network GPU concurrency flakes
  that pass in serial mode.

### Checklist

- [x] I have added the tests to cover my changes (existing noise test
      module passes; no new tests needed since the fix restores
      previously-broken functionality).
- [x] I have updated the documentation accordingly (release note added
      under `releasenotes/notes/`).
- [x] I have read the CONTRIBUTING document.
-

### DIFF
--- a/releasenotes/notes/fix_nlohmann_json_3_11_compat-e2e1f6251eb7bcbc.yaml
+++ b/releasenotes/notes/fix_nlohmann_json_3_11_compat-e2e1f6251eb7bcbc.yaml
@@ -1,0 +1,10 @@
+---
+fixes:
+  - |
+    Fixed an incompatibility with nlohmann_json >= 3.11 that caused noisy
+    simulation to fail with ``ValueError: Invalid noise_params JSON: not an
+    object.`` when aer was built from source against a system-provided
+    nlohmann_json 3.11.x (the version shipped in current-generation Linux
+    distributions). The Python-to-JSON conversion now uses
+    ``nlohmann::adl_serializer`` specializations, which is the documented
+    extension point and is stable across nlohmann_json releases.
--- a/src/framework/pybind_basics.hpp
+++ b/src/framework/pybind_basics.hpp
@@ -108,7 +108,7 @@ py::object to_python(T &&obj) {
 template <>
 py::object to_python(json_t &&obj) {
   py::object pydata;
-  from_json(obj, pydata);
+  JSON::py_from_json(obj, pydata);
   return pydata;
 }
 
--- a/src/framework/pybind_json.hpp
+++ b/src/framework/pybind_json.hpp
@@ -55,24 +55,57 @@ using json_t = nlohmann::json;
 // Nlohman JSON <--> Python Conversion
 //------------------------------------------------------------------------------
 
-namespace std {
+namespace JSON {
 
 /**
  * Convert a python object to a json.
  * @param js a json_t object to contain converted type.
- * @param o is a python object to convert.
+ * @param obj is a python object to convert.
  */
-void to_json(json_t &js, const py::handle &o);
+void py_to_json(json_t &js, const py::handle &obj);
 
 /**
- * Create a python object from a json
- * @param js a json_t object
- * @param o is a reference to an existing (empty) python object
+ * Create a python object from a json.
+ * @param js a json_t object.
+ * @param o is a reference to a python object to populate.
  */
-void from_json(const json_t &js, py::object &o);
+void py_from_json(const json_t &js, py::object &o);
 
-} // end namespace std.
+} // end namespace JSON.
 
+namespace nlohmann {
+
+/**
+ * adl_serializer specialization for pybind11::handle.
+ *
+ * Provides version-stable Python -> JSON conversion through nlohmann's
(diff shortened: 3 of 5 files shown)


dmeliksetian:
Hi @hhorii and @chriseclectic, I've submitted this PR to fix the noise simulation failures on newer distros (Ubuntu 24+) caused by the nlohmann_json 3.11 incompatibility. The CI checks are currently 'Waiting for status to be reported'—could you please approve and run the actions so we can verify the fix? Thanks!

ihincks:
Thank you for your contribution and effort, @dmeliksetian, it is appreciated. #2418 was merged instead, which was a duplicate effort of this PR, so I'll close this one. Ultimately, we should move away from nlohmann altogether, it's the wrong choice to transfer these objects.

PR #2423, opened 2026-04-26
Fix cuda arch thrust compat

### LINKED ISSUE #2415: Build fails on NVIDIA Blackwell (sm_120, CC 12.0) with CUDA 13.2 — patches required

dmeliksetian:
Closes #2415.
### Summary
Adds support for building qiskit-aer with CUDA 12+ and NVIDIA GPUs of
compute capability 12.0+ (Blackwell architecture, e.g. RTX PRO
4500/5000/6000), neither of which is currently supported by the
released build system.

### Details and comments
Four atomic fixes corresponding to the four problems documented in #2415:

1. `cuda_select_nvcc_arch_flags` arch list updated to include Hopper (sm_90),
   Ada (sm_89), and Blackwell (sm_120). Regex fixed to match 3-digit
   architecture numbers.

2. cuQuantum/cuTENSOR library path detection updated for CUDA 12+ where
   the library directory layout changed.

3. `thrust::unary_function` shim replaced with inline
   `argument_type`/`result_type` typedefs. CUDA 12.6 ships Thrust 2.5
   which still has `thrust::unary_function` (deprecated, not removed),
   making the previous version-guarded shim cause ambiguous-symbol
   errors. The inline typedef approach works regardless of whether
   `thrust::unary_function` exists, is deprecated, or is fully removed.

4. Build improvements: pin pybind11 to <3.0, serialize GPU tests via
   `group_regex` in stestr, and add a defensive nullptr check in
   `circuit_executor.hpp`.

### Verification

Built successfully on:
- NVIDIA RTX PRO 4500 (Blackwell, sm_120)
- CUDA 13.2
- Python 3.13.12
- GCC 13.3.0
- Ubuntu

Smoke test (3-qubit GHZ on GPU statevector) returns correct counts.
GPU + cuQuantum + cuStateVec paths verified.

The full GPU test suite has remaining noise-related failures caused by
a separate nlohmann_json compatibility issue (filed as #2421, fix in
#2422); when that fix is also applied locally, 216/221 of those noise
failures pass under parallel execution and 5 pass in serial mode (the
remaining 30 are unrelated tensor_network GPU concurrency flakes).

### Checklist

- [x] Tests covered: existing GPU test suite confirms behavior.
- [x] Documentation: release note pending — will add if requested.
- [x] Read the CONTRIBUTING document.


dmeliksetian:
Hi @hhorii and @chriseclectic, I've submitted this PR to fix the noise simulation failures on newer distros (Ubuntu 24+) caused by the nlohmann_json 3.11 incompatibility. The CI checks are currently 'Waiting for status to be reported'—could you please approve and run the actions so we can verify the fix? Thanks!

PR #2435, opened 2026-05-22
Make mergify consistent with branch protection rule

wshanks:
Mergify will not run on a branch that is not up to date with the main branch if the branch protection rule requiring branches be up to date with the main branch is in place, unless mergify is configured to run merges serially, one at a time. Here the mergify configuration is modified to try to meet this condition.


ihincks (approved):
Seems likely to work, let's give it a try.

PR #2445, opened 2026-07-22
Fix controlled-gate parameter binding

### LINKED ISSUE #2391: Incorrect Results from SamplerV2 for parameterized circuits with CRX gates
### Informations

- **Qiskit Aer version**: 0.17.1
- **Python version**: 3.13.2
- **Operating system**: Mac OS

`SamplerV2` from `qiskit_aer.primitives` returns incorrect results for parameterized circuits with CRX gates and multiple parameter values.

```python
import json
from qiskit.circuit import Parameter
from qiskit_aer.primitives import SamplerV2

sampler = SamplerV2()

params = [ 0.1, 0.1]

qc = QuantumCircuit(2)
qc = h([0,1])
theta = Parameter("$\\theta$")
qc.crx(theta, 0, 1)
qc.measure_all()

result = sampler.run([(qc, params)], shots=1024).result()[0]
print(
    "error:",
    json.dumps([r.get_counts() for r in result_error.data.meas], indent=2),
)

```
I am always getting only "00" for the second parameter.

```
error: [
  {
    "00": 257,
    "10": 264,
    "01": 240,
    "11": 263
  },
  {
    "00": 1024
  }
]
```

Ideally for the minimal reproducer given, I would expect result for each parameter value to be identical, since parameter values are equal.   

### Suggested solutions

In the minimal example, all parameter values are the same. Thus, I would expect the output for each parameter value to be identical across all experiments, matching the distribution from the first run.

```python

qc_decomposed = qc.decompose()

result = sampler.run([(qc_decomposed, params)], shots=1024).result()[0]

```

### LINKED ISSUE #2403: [BUG] Wrong PUB Parameter Bind for Controlled Phase Gate in Primitives
### Informations

- **Python version**: 3.13.0
- **Operating system**: macOS Tahoe 26.1 (25B78)

Executing a `SamplerV2`/`EstimatorV2` from `qiskit_aer.primitives` with a quantum circuit containing a parametrized controlled phase gate, e.g. `estimator.run([qc, op, parameter_assignment])`, the result is wrong.

Below is a snippet that provides an example circuit. Alongside is an execution where the parameter is bound via `qc.assign_parameters`, where the result is expected.

```python
import numpy as np

from qiskit.circuit import Parameter
from qiskit_aer.primitives import EstimatorV2, SamplerV2

op = SparsePauliOp.from_list([("IZ", 1)])

p = Parameter("p")

qc = QuantumCircuit(2)
qc.h(0)
qc.cx(0, 1)
qc.cp(p, 0, 1)
qc.cx(0, 1)
qc.h(0)
# qubit 0 in state 1 if p = pi

estimator = EstimatorV2()
res1 = estimator.run([(qc, op, {p: np.pi})]).result()
res2 = estimator.run([(qc.assign_parameters({p: np.pi}), op)]).result()

print(res1[0].data.evs)
print(res2[0].data.evs)
```
Output:
```
1.0
-1.0
```
Correct output is -1.0

The expected behaviour is also provided above.

### Suggested solutions

Bug source not identified.
However, I have a suspicion that the controlled phase gates are entirely disregarded in the computation. The reason is that large quantum circuits containing only controlled phase gates as two-qubit operations are executed extremely quickly using the MPS simulator when the parameters are assigned in the pub.

sankalpsthakur:
### Summary

Fix PUB parameter binding for native parameterized controlled gates, including open-control gates and repeated controlled rotations.

### Details and comments

The parameterization scan relied on `ControlledGate.is_parameterized()`, which can miss public parameters proxied through the base gate. When open controls expand into leading and trailing `X` operations, the parameterization index must also point at the native controlled operation rather than the synthetic gate.

This change scans the operation's public parameters directly, records the native-operation offset produced during assembly, and detaches parameter lists before replacing symbolic values. Regressions cover closed and open CP/CU gates, repeated and distinct CRX instances, MCPhase, and multi-control offsets through both EstimatorV2 and SamplerV2.

Validation:

- `pytest -q test/terra/primitives/test_estimator_v2.py test/terra/primitives/test_sampler_v2.py test/terra/backends/test_parameterized_circuit.py test/terra/backends/test_runtime_parameterization.py`
- Ruff lint and format checks
- `git diff --check`

The macOS native test run used the Homebrew `libomp` runtime required by the local Aer build.

## Why this matters

Parameterized circuits are core to variational algorithms and runtime jobs. When controlled gates lose their bound parameters during Aer assembly, Estimator/Sampler results silently diverge from Qiskit semantics — the kind of bug that only shows up mid-experiment.

## Related

Fixes #2403
Fixes #2391

### AI/LLM disclosure

- OpenAI Codex was used to help generate or modify the implementation, tests, and PR description. I reviewed the resulting changes and ran the validation listed above.


sankalpsthakur:
Withdrawing this; no maintainer signal and we are cutting the queue.

PR #2446, opened 2026-07-22
Prevent empty multiplexer crashes

### LINKED ISSUE #2444: Error with StatePreparation.inverse()
### Informations

- **Python version**: 3.11.13
- **Operating system**: macOS 15.7.7

The Python process terminates abruptly during Aer execution, without reaching the final `print("end")` (see the script in the next section). 

Run the following script, for example with `uv run python repro.py`.

```python
import math

from qiskit.circuit.library import StatePreparation

backend = AerSimulator()
circuit = QuantumCircuit(1)
circuit.append(StatePreparation([0.5, math.sqrt(0.75)]).inverse(), [0])
circuit.measure_all()

transpiled = transpile(circuit, backend)
print([(inst.operation.name, len(inst.operation.params)) for inst in transpiled.data])

result = backend.run(transpiled, shots=16).result()
print(result.get_counts())
print("end")
```

The script prints only `[('multiplexer', 0), ('barrier', 0), ('measure', 0)]`. Then the process terminates before reaching `print("end")`.

It should work without abrupt termination.

### Suggested solutions
I have not really investigated the aer codebase, but this might be related to `StatePreparation.inverse()` producing a multiplexer instruction with empty `params`.

sankalpsthakur:
### Summary

Prevent native crashes when a definition-backed multiplexer reaches Aer with no matrix parameters.

### Details and comments

`StatePreparation(...).inverse()` can transpile to a `multiplexer` whose matrices live in its circuit definition rather than `operation.params`. Aer's normal assembly path forwarded the resulting empty matrix vector to native code, where `make_multiplexer` indexed the first element and terminated the Python process.

This change inlines definition-backed, empty-parameter multiplexers at the compiler boundary and refreshes the circuit operation types after rewriting. A native guard also rejects any malformed empty multiplexer that reaches assembly, turning the former out-of-bounds access into a Python `ValueError`.

Regression coverage verifies the original inverse-StatePreparation result, the recursive control-flow path, and the native rejection boundary.

Validation:

- Original issue reproducer completed twice instead of terminating
- 157 affected tests and 91 subtests passed; 24 tests skipped
- Full multiplexer, control-flow, and initialize suites passed
- The rebuilt native extension contains the new guard
- Black, Pylint, clang-format, Reno lint, and `git diff --check` passed

## Why this matters

An empty multiplexer reaching native code terminates the Python process outright — not a catchable Python exception. Any transpilation path that produces definition-backed multiplexers with no matrix params (e.g. inverse StatePreparation) can crash long-running simulation jobs without a stack trace.

## Related

Fixes #2444

### AI/LLM disclosure

- OpenAI Codex was used to help generate or modify the implementation, tests, and PR description. I reviewed the resulting changes and ran the validation listed above.


sankalpsthakur:
Withdrawing this; no maintainer signal and we are cutting the queue.

PR #2457, opened 2026-09-01
Support MidCircuitMeasure and measure_* instructions in Aer

### LINKED ISSUE #2456: Add mid-circuit measurement support in Qiskit Aer

Qiskit circuits containing measurements before the final instruction layer ("mid-circuit measurement", MCM) — used for classical feedback via `if_test`/`while_loop`/`switch_case`, and for repeated syndrome-style measurements — are only exercised indirectly today, through the control-flow test suite. There is no dedicated test module that validates MCM as a first-class, standalone feature: plain MCM with no feedback, MCM used for classical feedback, MCM followed by reset, deterministic MCM + correction, confirmation that ordinary final-only-measurement circuits still take the fast `measure_sampling` path, MCM under readout noise, and MCM through `SamplerV2`.

Separately, there is no way to attach noise to one specific measurement. `NoiseModel.add_quantum_error` targets an instruction by op name, and every `measure` in a circuit shares the same `"measure"` name, so a `QuantumError` added under that name applies before *every* measurement, including every mid-circuit one, rather than a chosen one. This blocks realistic noise modeling for circuits with several measurements at different points (e.g. MCM-heavy dynamic circuits, repeated syndrome measurements) where different measurements need independent error channels.

### Proposed solution

1. Add a dedicated `test_mid_circuit_measurement.py` suite covering plain MCM, classical feedback, MCM + reset, deterministic MCM + correction, the `can_sample` fast-path regression, MCM under readout noise, and `SamplerV2`, across `statevector`, `density_matrix`, and `matrix_product_state`.
2. Tighten the compiler's conditional-clbit scoping (a circuit-wide `is_conditional` bool → the precise set of clbits actually read by a downstream conditional) so register allocation for MCM circuits is exact rather than all-or-nothing.
3. Let a measurement be labelled via Qiskit's existing `Measure(label=...)` support, and surface that label to the Aer noise model as the op name `measure_<label>`, so a `QuantumError` can target one specific (including mid-circuit) measurement:

```python
from qiskit.circuit import Measure
from qiskit_aer.noise import NoiseModel, pauli_error

circuit.append(Measure(label="mcm0"), [0], [0])

noise_model = NoiseModel()
noise_model.add_quantum_error(pauli_error([("X", 1.0)]), "measure_mcm0", [0])
 ```

The error applies only immediately before the labelled measurement; every other (unlabeled) measurement keeps the current shared `"measure"` behaviour, including the fast-sampling path, unchanged.

### PR DESCRIPTION
**Pull request by Siddharthgolecha**
## Summary

Closes #2456.

This PR adds Qiskit Aer support for dedicated mid-circuit measurement instructions such as `measure_2`, used by:

```python
from qiskit_ibm_runtime.circuit import MidCircuitMeasure
```

These instructions are distinct from the standard Qiskit `"measure"` operation and may appear when using IBM Runtime mid-circuit measurement APIs or backend-specific measurement instructions.

Previously, circuits containing these operations could fail during transpilation or simulation with errors such as:

```text
qiskit.transpiler.exceptions.TranspilerError:
'HighLevelSynthesis is unable to synthesize "measure_2"'
```

This PR allows Aer to recognize and preserve `measure_*` instructions as measurement operations while keeping existing `"measure"` behavior unchanged.

### Support `MidCircuitMeasure` / `measure_*` instructions

Aer now treats instruction names matching:

```text
measure_*
```

as valid measurement operations.

This includes instructions such as:

```text
measure_2
```

used by `qiskit_ibm_runtime.circuit.MidCircuitMeasure`.

The measurement semantics remain the same from the simulator's perspective, but the original instruction name is preserved so that Aer can distinguish backend-specific measurement implementations.

Standard measurements remain unchanged:

```python
qc.measure(0, 0)
```

continue to use the existing `"measure"` instruction.

### Preserve `measure_*` instructions through Aer backend handling

`measure_*` instructions are now handled correctly in:

* Aer circuit assembly;
* basis-gate validation;
* backend `Target` construction;
* backend-property conversion;
* C++ circuit representation.

This prevents the transpiler from attempting to synthesize `measure_2` as an unsupported custom instruction.

### Support measurement-specific noise

Because the measurement instruction name is preserved, a noise model can now target a specific measurement implementation independently.

For example:

```python
noise_model.add_quantum_error(
    error,
    "measure_2",
    [0],
)
```

can apply an error specifically to `measure_2` without affecting ordinary `"measure"` instructions.

The same mechanism also supports labelled measurements represented internally as `measure_<label>`.

### Improve conditional-clbit handling

The compiler previously used a circuit-wide `is_conditional` flag to determine whether measurement results should also be written to Aer register storage.

This is replaced with `_conditional_clbits(circuit)`, which tracks the specific classical bits actually consumed by downstream conditional operations.

This makes register allocation more precise for circuits containing multiple measurements.

### Add dedicated mid-circuit measurement tests

Adds `test_mid_circuit_measurement.py` with coverage for:

* plain mid-circuit measurement;
* `MidCircuitMeasure` / `measure_*`;
* classical feedback via `if_test`;
* measurement followed by reset;
* deterministic measurement + correction;
* ordinary final-measurement `measure_sampling` regression;
* readout noise;
* measurement-specific quantum noise;
* `SamplerV2`;
* `statevector`;
* `density_matrix`;
* `matrix_product_state`.

### Additional compatibility fixes

* Serialize the Aer noise model to JSON before passing it to the C++ controller, avoiding Python → `nlohmann::json` conversion issues with pybind11 ≥ 3.
* Add the Apple Silicon Homebrew library path (`/opt/homebrew/lib`) to the CMake configuration.

## Example

```python
from qiskit_ibm_runtime.circuit import MidCircuitMeasure

qc = QuantumCircuit(1, 1)

qc.h(0)
qc.append(MidCircuitMeasure(), [0], [0])
qc.x(0)
qc.measure(0, 0)

simulator = AerSimulator()

# measure_2 should be accepted and simulated as a measurement operation.
```

## Backward compatibility

Existing standard measurements are unaffected.

Unlabelled Qiskit `Measure` instructions continue to use:

```text
measure
```

and retain the existing fast `measure_sampling` path where applicable.

## Test plan

* [x] `test_mid_circuit_measurement`
* [x] `test_control_flow`
* [x] `test_measure`
* [x] `test_noise`
* [x] `test_fusion`

The existing behavior for standard measurement and control-flow circuits remains unchanged.


Siddharthgolecha:
I have used Claude Code and IBM Bob

